PR #7, opened 2018-10-08
Relation tuple sets

natgavrilenko:


natgavrilenko:
No, it is not about efficiency. It is necessary for cloning conditional RMW events, such as atomic_add_unless. In Dartagnan, all events which execution depends on result of a read (conditional writes or conditional fences in Linux), "know" their read. And, if such events are cloned, they should know their corresponding cloned copy.

The problem is that one conditional read can have multiple dependent events. For example, a read can have a corresponding write and two conditional fences around the Read-Write pair. Thus, we cannot simply skip cloning of a read when it is called directly and do cloning at the moment of cloning its write, because there might be also fences which need a reference to the new read. 

To overcome the previous problem, we could also keep a reference to dependent events inside a read, so that a read clone can keep track of its already cloned fences. But this would create a circular dependency: a read has a reference to its fence and a fence has a reference to its read. And circular dependencies might cause issues with efficiency of garbage collection.

So, I used the following solution for cloning conditional RMWs. Whenever an event, which depends on a conditional read is cloned, it "asks" the read for a reference to its clone. If the read has been already cloned, then it returns a reference to the clone. Otherwise, it creates a new clone and memorizes it. In order to avoid returning reference to stale clones, we reset prepared clones when program.clone() is called (we clone events only if we clone a program).

There are also other ways to design cloning of conditional reads, but this one seems the cleanest to me.

natgavrilenko:
Let's keep DartagnanIdlTest. It might be useful later if we want to make changes IDL / Fixpoint encoding (and test them). But yes, we should probably move it dartagnan.test package. I want to write a comprehensive set of tests for Dartangnan, but I don't know when I will have enough time for this.

PR #17, opened 2019-03-24
Gui refactoring

hernanponcedeleon:
I took care of most of your comments + some more refactoring.
Also, I needed to modify the parsers. You recently change the parsers to deal only with files but I need them to deal also directly with text to avoid having to create dummy files with the content of the editors.
Please pay special attention to this to be sure I did not screw anything (the rest is all in ui plugin so it should not affect functionality).


natgavrilenko:
Height of archPane is set to zero, so it is invisible even when it must be shown

Issue #73, opened 2021-07-16
SMT portfolio


hernanponcedeleon:
Our current implementation uses Z3 API to generate the SMT formula.

I tried dumping the formulas to a *.smt2 file and used different SMT solvers to see how they perform:

<p align="center">
  <img width="510" alt="Screenshot 2021-07-16 at 09 46 15" src="https://user-images.githubusercontent.com/11671435/125911535-9f7d05fa-e5eb-48ba-86c8-cb0381d6a74a.png">
</p>

From the experiment it is obvious that Z3 might not be the best solver for us.

The idea is to support a portfolio of solvers. The easiest solution seems to be using a common API like [JavaSMT](https://github.com/sosy-lab/java-smt).

I implemented this (currently in branch `javastm`), but there still remain some things to be done:

- [x] there are currently some performance issues when closing contexts when formulas are big, see sosy-lab/java-smt#236
- [ ] JavaSMT depends on native binaries. This requires some changes in our `pom` files and how to package the application. I think all problems are solved now, but this requires some more testing, specially in different platforms
- [ ] because of the above, our jars do not have the `-jar-with-dependencies` postfix, we need to update documentation. Also about dependencies and options
- [ ] parse the configuration options from a file
- [ ] set up smt solver timeout
- [ ] JavaSMT doesn't seem to have support for `SREM` and `UREM`

hernanponcedeleon:
The performance issue was solved by setting option `solver.z3.usePhantomReferences` to `true` in the configuration.
However this is currently hardcoded in the code. It would be better to use configuration files for this.

hernanponcedeleon:
Merged.

Issue #79, opened 2021-08-06
Consolidate the use of configuration flags


hernanponcedeleon:
Currently the tool can be configure in many ways
- passing options
- setting flags in `GlobalSettings`
- setting flags in `Settings`

Now that we use `org.sosy_lab.common.configuration.Configuration` we can move all to a file which can be parsed.

hernanponcedeleon:
Fixed by #752

PR #81, opened 2021-08-09
Migration from z3 API to Javasmt

hernanponcedeleon:
This PR implements the changes for #73.


hernanponcedeleon:
From my point of view the code in `master` is incomplete. It is possible that e.g. `b1.getLastValueExpr(ctx)` returns a `BitVecExpr` and then the casting would throw an exception.
I think the reason we never run into this problem is that `getLastValueExpr` is used to encode the litmus assertions and we don't use bit-vectors for those (but we could).
Thus I consider the way we are handling this in the `javasmt` branch is better. The whole handling of bit-vectors in our code needs some re-work, but that is orthogonal to this PR.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -54,6 +59,19 @@ public static void main(String[] args) throws Exception {
...
+				if(options.getSettings().getSolverTimeout() > 0) {
+					// Converts timeout from secs to millisecs
+					Thread.sleep(1000 * options.getSettings().getSolverTimeout());
+					sdm.requestShutdown("Shutdown Request");
+					logger.warn("Shutdown Request");
+				}
+			} catch (InterruptedException e) {
+				// Verification ended, nothing to be done.
+			}});
+
```
ThomasHaas, line 74: The Thread creation should be moved more closely to the place of usage. Also, creating it only if a timeout is specified might be more efficient.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -85,45 +103,69 @@ public static void main(String[] args) throws Exception {
...
+                    config, 
+                    BasicLogManager.create(config), 
+                    sdm.getNotifier(), 
+                    options.getSMTSolver()); 
+			ProverEnvironment prover = ctx.newProverEnvironment(ProverOptions.GENERATE_MODELS);
+
+            Result result;
+    		switch(options.getAnalysis()) {
+				case RACES:
+					result = checkForRaces(ctx, task);
```
ThomasHaas, line 121: Missing ```break;```

dartagnan/src/main/java/com/dat3m/dartagnan/expression/BExpr.java
```diff
@@ -1,22 +1,26 @@
...
 public abstract class BExpr implements ExprInterface {
 
     @Override
-    public Expr toZ3Int(Event e, Context ctx) {
-        return ctx.mkITE(toZ3Bool(e, ctx), ctx.mkInt(1), ctx.mkInt(0));
+    public Formula toIntFormula(Event e, SolverContext ctx) {
+    	FormulaManager fmgr = ctx.getFormulaManager();
+		return fmgr.getBooleanFormulaManager().ifThenElse(toBoolFormula(e, ctx), 
+				fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ONE), 
+				fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ZERO));
```
ThomasHaas, line 19: refactor fmgr.getIntegerFormulaManager into a local variable

dartagnan/src/main/java/com/dat3m/dartagnan/expression/BNonDet.java
```diff
@@ -23,25 +27,30 @@ public IConst reduce() {
...
-        return ctx.mkITE(toZ3Bool(e, ctx), e1, e2);
+    public Formula toIntFormula(Event e, SolverContext ctx) {
+    	FormulaManager fmgr = ctx.getFormulaManager();
+		Formula e1 = precision > 0 ? 
+    			fmgr.getBitvectorFormulaManager().makeBitvector(precision, BigInteger.ONE) : 
+    			fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ONE);
+    	Formula e2 = precision > 0 ?
+    			fmgr.getBitvectorFormulaManager().makeBitvector(precision, BigInteger.ZERO) : 
+    			fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ZERO);
+        return fmgr.getBooleanFormulaManager().ifThenElse(toBoolFormula(e, ctx), e1, e2);
```
ThomasHaas, line 38: It would be clearer to do the "precision > 0" check only once and have two branches that initialize e1 and e2 to either Integers or BitVectors. This better shows that either both expressions will be Int or both will be BV.
```
Formula e1, e2;
if (precision > 0) {
    e1 = BV(1);
    e2 = BV(0);
} else {
    e1 = Int(1);
    e2 = Int(0);
}
return ifThenElse(toBoolFormula(e, ctx), e1, e2);
```

dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
```diff
@@ -41,20 +44,26 @@ public IConst reduce() {
...
+		return precision > 0 ? 
+				fmgr.getBitvectorFormulaManager().makeVariable(precision, name) : 
+				fmgr.getIntegerFormulaManager().makeVariable(name);
 	}
 
 	@Override
-	public BigInteger getIntValue(Event e, Model model, Context ctx) {
-		return new BigInteger(model.getConstInterp(toZ3Int(e, ctx)).toString());
+	public BigInteger getIntValue(Event e, Model model, SolverContext ctx) {
+		return new BigInteger(model.evaluate(toIntFormula(e, ctx)).toString());
```
ThomasHaas, line 66: model.evaluate(...) may return null if the variable is not present in the model. It might be better to perform a check against that instead of letting the code run into a NullPointerException. It might not be necessary as the NullPointerException is thrown at the place of error (instead of some later time) so it is easy to debug.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
```diff
@@ -136,21 +145,19 @@ public int getPrecision() {
...
-	        	enc = ctx.mkAnd(enc, ctx.mkBVSLE((BitVecExpr)toZ3Int(null,ctx), ctx.mkBV(max, precision)));					
-			}
+			boolean signed = !(type.equals(UINT) || type.equals(ULONG) || type.equals(USHORT) || type.equals(UCHAR));
+			enc = bmgr.and(enc, fmgr.getBitvectorFormulaManager().greaterOrEquals((BitvectorFormula) toIntFormula(null,ctx), fmgr.getBitvectorFormulaManager().makeBitvector(precision, min), signed));
+	        enc = bmgr.and(enc, fmgr.getBitvectorFormulaManager().lessOrEquals((BitvectorFormula) toIntFormula(null,ctx), fmgr.getBitvectorFormulaManager().makeBitvector(precision, max), signed));
 		} else {
-        	enc = ctx.mkAnd(enc, ctx.mkGe((IntExpr)toZ3Int(null,ctx), ctx.mkInt(min)));
-        	enc = ctx.mkAnd(enc, ctx.mkLe((IntExpr)toZ3Int(null,ctx), ctx.mkInt(max)));
+			enc = bmgr.and(enc, fmgr.getIntegerFormulaManager().greaterOrEquals((IntegerFormula)toIntFormula(null,ctx), fmgr.getIntegerFormulaManager().makeNumber(min)));
+			enc = bmgr.and(enc, fmgr.getIntegerFormulaManager().lessOrEquals((IntegerFormula)toIntFormula(null,ctx), fmgr.getIntegerFormulaManager().makeNumber(max)));
```
ThomasHaas, line 160: Refactor fmgr.getIntegerFormulaManager / fmgr.getBitVectorFormulaManager into local variables of their respective branches.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/COpBin.java
```diff
@@ -34,28 +36,73 @@ public String toString() {
...
             case UGTE:
-            	return e1.isBV() ? ctx.mkBVUGE((BitVecExpr)e1, (BitVecExpr)e2) : ctx.mkGe((ArithExpr)e1, (ArithExpr)e2);
+            	if(e1 instanceof IntegerFormula) {
+            		IntegerFormulaManager imgr = ctx.getFormulaManager().getIntegerFormulaManager();
+            		return imgr.greaterOrEquals((IntegerFormula)e1, (IntegerFormula)e2);
+            	}
+            	if(e1 instanceof BitvectorFormula) {
+            		BitvectorFormulaManager bvmgr = ctx.getFormulaManager().getBitvectorFormulaManager();
+            		return bvmgr.greaterOrEquals((BitvectorFormula)e1, (BitvectorFormula)e2, this.equals(GTE));
+            	}
```
ThomasHaas, line 105: The checks and the switch should be interchanged, i.e. first check the types of e1 and e2 and then perform a switch on the operation inside each of the branches (bool + bool, int + int, bv + bv). The code in each branch is more clearer because the respective FormulaManager can be stored in a local variable as well as the casted two expressions.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
```diff
@@ -61,36 +63,98 @@ public String toLinuxName(){
...
-            case SREM:
-            	return e1.isBV() ? ctx.mkBVSRem((BitVecExpr)e1, (BitVecExpr)e2) : ctx.mkBV2Int(ctx.mkBVSRem(ctx.mkInt2BV(32, (IntExpr)e1), ctx.mkInt2BV(32, (IntExpr)e2)), false);
-            case UREM:
-            	return e1.isBV() ? ctx.mkBVURem((BitVecExpr)e1, (BitVecExpr)e2) : ctx.mkBV2Int(ctx.mkBVURem(ctx.mkInt2BV(32, (IntExpr)e1), ctx.mkInt2BV(32, (IntExpr)e2)), false);
+            	bvmgr = ctx.getFormulaManager().getBitvectorFormulaManager();
+            	if(e1 instanceof IntegerFormula) {
+            		return bvmgr.toIntegerFormula(bvmgr.shiftRight(bvmgr.makeBitvector(32, (IntegerFormula)e1), bvmgr.makeBitvector(32, (IntegerFormula)e2), true), false); 
+            	} else {
+            		return bvmgr.shiftRight((BitvectorFormula)e1, (BitvectorFormula)e2, true); 
+            	}
```
ThomasHaas, line 157: The type check and the switch should be interchanged (see comment on COpBin)

dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
```diff
@@ -247,28 +254,35 @@ public BoolExpr encodeFinalRegisterValues(Context ctx){
...
                 }
-                enc = ctx.mkAnd(enc, ctx.mkImplies(lastModReg,
-                        ctx.mkEq(reg.getLastValueExpr(ctx), ((RegWriter)w1).getResultRegisterExpr())));
+                BooleanFormula same = reg.getLastValueExpr(ctx) instanceof BitvectorFormula ?
+                		fmgr.getBitvectorFormulaManager().equal(
+                				(BitvectorFormula)reg.getLastValueExpr(ctx), 
+                				(BitvectorFormula)((RegWriter)w1).getResultRegisterExpr()) :
+                		fmgr.getIntegerFormulaManager().equal(
+                				(IntegerFormula)reg.getLastValueExpr(ctx), 
+                				(IntegerFormula)((RegWriter)w1).getResultRegisterExpr());
```
ThomasHaas, line 270: It might be clever to add a Helper class that does support a generic equals method (that internally checks the types of the formulas and calls the respective managers). It seems this sort of code appears quite often.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Register.java
```diff
@@ -87,13 +95,16 @@ public Expr toZ3IntResult(Event e, Context ctx) {
 	}
 
 	@Override
-	public Expr getLastValueExpr(Context ctx){
-		return precision > 0 ? ctx.mkBVConst(getName() + "_" + threadId + "_final", precision) : ctx.mkIntConst(getName() + "_" + threadId + "_final");
+	public Formula getLastValueExpr(SolverContext ctx){
+		FormulaManager fmgr = ctx.getFormulaManager();
+		return precision > 0 ?
+				fmgr.getBitvectorFormulaManager().makeVariable(precision, getName() + "_" + threadId + "_final") :
+				fmgr.getIntegerFormulaManager().makeVariable(getName() + "_" + threadId + "_final");
```
ThomasHaas, line 102: Create the name only once just line in toIntFormula

dartagnan/src/main/java/com/dat3m/dartagnan/program/Register.java
```diff
@@ -71,14 +73,20 @@ public boolean equals(Object obj) {
...
 		String name = getName() + "(" + e.repr() + ")";
-		return precision > 0 ? ctx.mkBVConst(name, precision) : ctx.mkIntConst(name);
+		FormulaManager fmgr = ctx.getFormulaManager();
+		return precision > 0 ?
+				fmgr.getBitvectorFormulaManager().makeVariable(precision, name) :
+				fmgr.getIntegerFormulaManager().makeVariable(name);
 	}
 
-	public Expr toZ3IntResult(Event e, Context ctx) {
+	public Formula toZ3IntResult(Event e, SolverContext ctx) {
```
ThomasHaas, line 84: Rename to generic name.

dartagnan/src/main/java/com/dat3m/dartagnan/program/arch/aarch64/event/RMWStoreExclusiveStatus.java
```diff
@@ -48,13 +52,20 @@ public String toString() {
...
-		BoolExpr enc = ctx.mkAnd(
-                ctx.mkImplies(storeEvent.exec(), ctx.mkEq(regResultExpr, new IConst(BigInteger.ZERO, precision).toZ3Int(this, ctx))),
-                ctx.mkImplies(ctx.mkNot(storeEvent.exec()), ctx.mkEq(regResultExpr, new IConst(BigInteger.ONE, precision).toZ3Int(this, ctx)))
+		BooleanFormula enc = bmgr.and(
+				bmgr.implication(storeEvent.exec(), imgr.equal(
+						(IntegerFormula)regResultExpr, 
+						(IntegerFormula)new IConst(BigInteger.ZERO, precision).toIntFormula(this, ctx))),
+				bmgr.implication(bmgr.not(storeEvent.exec()), imgr.equal(
+						(IntegerFormula)regResultExpr, 
+						(IntegerFormula)new IConst(BigInteger.ONE, precision).toIntFormula(this, ctx)))
```
ThomasHaas, line 66: This code will break if precision > 0 since toIntFormula will return a BitVectorFormula that can't be cast to IntegerFormula. Again, a Helper method that can construct equality formulas might be good.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
```diff
@@ -47,8 +54,26 @@ protected CondJump(CondJump other) {
...
+			defaultCtx = SolverContextFactory.createSolverContext(
+			        config, 
+			        BasicLogManager.create(config), 
+			        ShutdownManager.create().getNotifier(), 
+			        Solvers.Z3);
+		} catch (Exception e) {
+			e.printStackTrace();
+		}
+    }
+
```
ThomasHaas, line 71: The default context should be created once globally. It should probably use the Solver specified by the User instead of Z3.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/rmw/cond/RMWReadCond.java
```diff
@@ -28,12 +31,14 @@
     }
 
     @Override
-    public void initialise(VerificationTask task, Context ctx) {
+    public void initialise(VerificationTask task, SolverContext ctx) {
         super.initialise(task, ctx);
-        z3Cond = ctx.mkEq(memValueExpr, cmp.toZ3Int(this, ctx));
+        z3Cond = ctx.getFormulaManager().getIntegerFormulaManager().equal(
+        		(IntegerFormula)memValueExpr, 
+        		(IntegerFormula)cmp.toIntFormula(this, ctx));
```
ThomasHaas, line 38: Is cmp.toIntFormula guaranteed to be an IntegerFormula (is it never a BitVectorFormula?)

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Address.java
```diff
@@ -42,22 +44,25 @@ public void setConstantValue(BigInteger value) {
...
+        return toIntFormula(ctx);
     }
 
-    public Expr getLastMemValueExpr(Context ctx){
-        return precision > 0 ? ctx.mkBVConst("last_val_at_memory_" + index, precision) : ctx.mkIntConst("last_val_at_memory_" + index);
+    public Formula getLastMemValueExpr(SolverContext ctx){
+        FormulaManager fmgr = ctx.getFormulaManager();
+		return precision > 0 ? 
+        		fmgr.getBitvectorFormulaManager().makeVariable(precision, "last_val_at_memory_" + index) : 
+        		fmgr.getIntegerFormulaManager().makeVariable("last_val_at_memory_" + index);
```
ThomasHaas, line 60: Create name only once.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
```diff
@@ -26,36 +35,68 @@ public Location getLocationForAddress(Address address){
...
                 e1 = e2;
             }
         }
         // Following SMACK, only address with constant values can have negative values.
         for(Address add : getAllAddresses()) {
         	if(!add.hasConstantValue()) {
-        		enc = ctx.mkAnd(enc, add.toZ3Int(ctx).isBV() ? ctx.mkGt(ctx.mkBV2Int((BitVecExpr) add.toZ3Int(ctx), false), ctx.mkInt(0)) : ctx.mkGt((IntExpr)add.toZ3Int(ctx), ctx.mkInt(0)));
+        		enc = bmgr.and(enc, add.toIntFormula(ctx) instanceof BitvectorFormula ?
+        				fmgr.getIntegerFormulaManager().greaterThan(fmgr.getBitvectorFormulaManager().toIntegerFormula((BitvectorFormula)add.toIntFormula(ctx), false), fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ZERO)) :
+        				fmgr.getIntegerFormulaManager().greaterThan((IntegerFormula)add.toIntFormula(ctx), fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ZERO)));
```
ThomasHaas, line 62: Use the new convertToIntegerFormula to avoid the case distinction.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
```diff
@@ -26,36 +35,68 @@ public Location getLocationForAddress(Address address){
...
         }
         
-        return ctx.mkAnd(enc, ctx.mkDistinct(getAllAddresses().stream().map(a -> a.toZ3Int(ctx).isBV() ? ctx.mkBV2Int((BitVecExpr) a.toZ3Int(ctx), false) : a.toZ3Int(ctx)).toArray(Expr[]::new)));
+        BooleanFormula distinct = getAllAddresses().size() > 1 ?
+        		fmgr.getIntegerFormulaManager().distinct(getAllAddresses().stream()
+                		.map(a -> convertToIntegerFormula(a.toIntFormula(ctx), ctx))
+                		.collect(Collectors.toList())) : 
+                bmgr.makeTrue();
+        
+        return bmgr.and(enc, distinct);
```
ThomasHaas, line 72: Refactor fmgr.getIntegerFormulaManager into local variable (it will always be used anyway)

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
```diff
@@ -26,36 +35,68 @@ public Location getLocationForAddress(Address address){
...
+                        Formula e1 = add.toIntFormula(ctx);
+                        fmgr.getBitvectorFormulaManager().toIntegerFormula((BitvectorFormula) e1, false);
+						return fmgr.getIntegerFormulaManager().equal((IntegerFormula) e1, fmgr.getIntegerFormulaManager().makeNumber(add.getValue().intValue()));})
+            		.toArray(BooleanFormula[]::new);
+    	} else {
+        	addrExprs = getAllAddresses().stream().filter(x -> !x.hasConstantValue())
+            		.map(add -> fmgr.getIntegerFormulaManager().equal((IntegerFormula) add.toIntFormula(ctx), fmgr.getIntegerFormulaManager().makeNumber(add.getValue().intValue())))
+            		.toArray(BooleanFormula[]::new);
+    	}
+        return fmgr.getBooleanFormulaManager().and(addrExprs);
```
ThomasHaas, line 99: - Refactor fmgr.getIntegerFormulaManager into local variable
- Use convertToIntegerFormula and get rid of case distinction (there is not much performance gain in checking for bv once vs. in every iteration).
- The bv check fails if there are no addresses (should rarely happen but may be the case for sequential programs)

dartagnan/src/main/java/com/dat3m/dartagnan/analysis/ScopeTypes.java
```diff
@@ -1,9 +1,9 @@
 package com.dat3m.dartagnan.analysis;
 
-public enum SolverTypes {
+public enum ScopeTypes {
```
ThomasHaas, line 3: Rename to SolvingMethod or something similar?

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
```diff
@@ -45,6 +92,48 @@ public void parse(String[] args) throws ParseException, RuntimeException {
...
+					break;
+				case "boolector":
+					smtsolver = BOOLECTOR;
+					break;
+				case "cvc4":
+					smtsolver = CVC4;
+					break;
+				case "yices2":
+					smtsolver = YICES2;
+					break;
```
ThomasHaas, line 126: adding a "z3" and "unspecified" case that both select z3 as solver might be more readable. A default case could throw the exception. This way you avoid all Ifs. Something like:
```
String solverOption = cmd.getOptionValue(SMTSOLVER_OPTION, "unspecified");
switch (solverOption) {
  // cases
  // ...
  case "z3":
  case "unspecified":
     smtsolver = Z3;
     break;
  case default: 
     // throw exception
}

File: dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java [hunk 21]
```diff
@@ -110,7 +110,7 @@ public Program getProgram() {
     public Model getZ3Model() {
         return model;
     }
-    public Context getZ3Context() {
+    public SolverContext getZ3Context() {
         return context;
     }
```
ThomasHaas, line 115: Rename to getSolverModel and getSolverContext

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -308,12 +308,11 @@ private void extractReadsFrom() {
...
             for (EventData read : addressedReads.getValue()) {
                 for (EventData write : addressWritesMap.get(address)) {
-                    BoolExpr rfExpr = rf.getSMTVar(write.getEvent(), read.getEvent(), context);
-                    Expr rfInterp = model.getConstInterp(rfExpr);
+                    BooleanFormula rfExpr = rf.getSMTVar(write.getEvent(), read.getEvent(), context);
                     // The null check is important: Currently there are cases where no rf-edge between
                     // init writes and loads get encoded (in case of arrays/structs). This is usually no problem,
                     // since in a well-initialized program, the init write should not be readable anyway.
-                    if (rfInterp != null && rfInterp.isTrue()) {
+                    if (model.evaluate(rfExpr) != null && model.evaluate(rfExpr)) {
```
ThomasHaas, line 315: Call model.evaluate only once (just like the old code did). Something like this:
```
rfEval = model.evaluate(rfExpr)
if (rfEval != null && rfEval) {
    //...
}
```.

File: dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java [hunk 23]
```diff
@@ -340,9 +339,8 @@ private void extractCoherences() {
             for (EventData w1 : addressedWrites.getValue()) {
                 coherenceMap.put(w1, new HashSet<>());
                 for (EventData w2 : addressWritesMap.get(address)) {
-                    BoolExpr coExpr = co.getSMTVar(w1.getEvent(), w2.getEvent(), context);
-                    Expr coInterp = model.getConstInterp(coExpr);
-                    if (coInterp != null && coInterp.isTrue()) {
+                	BooleanFormula coExpr = co.getSMTVar(w1.getEvent(), w2.getEvent(), context);
+                    if (model.evaluate(coExpr) != null && model.evaluate(coExpr)) {
```
ThomasHaas, line 343: Model.evaluate only once (like above)

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -152,33 +155,35 @@ public BoolExpr encodeBase(Context ctx) {
...
 
     // Initalizes everything just like encodeBase but also encodes all
     // relations that are needed for the axioms (but does NOT encode the axioms themselves yet)
-    public BoolExpr encode(Context ctx) {
-        BoolExpr enc = encodeBase(ctx);
+    public BooleanFormula encode(SolverContext ctx) {
+    	BooleanFormula enc = encodeBase(ctx);
         for (Axiom ax : axioms) {
-            enc = ctx.mkAnd(enc, ax.getRelation().encode(ctx));
+            enc = ctx.getFormulaManager().getBooleanFormulaManager().and(enc, ax.getRelation().encode(ctx));
```
ThomasHaas, line 172: Move ctx.getFormulaManager().getBooleanFormulaManager out of the loop (just like in the method "consistent")

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
```diff
@@ -122,62 +126,67 @@ public TupleSet getMaxTupleSet(){
...
-                        BoolExpr isExecPair = ctx.mkAnd(isPair, store.exec());
-                        enc = ctx.mkAnd(enc, ctx.mkEq(isPair, pairingCond(thread, load, store, ctx)));
+                    	BooleanFormula isPair = exclPair(load, store, ctx);
+                    	BooleanFormula isExecPair = bmgr.and(isPair, store.exec());
+                        enc = bmgr.and(enc, bmgr.equivalence(isPair, pairingCond(thread, load, store, ctx)));
 
                         // If load and store have the same address
-                        BoolExpr sameAddress = ctx.mkEq(((MemEvent)load).getMemAddressExpr(), (((MemEvent)store).getMemAddressExpr()));
-                        unpredictable = ctx.mkOr(unpredictable, ctx.mkAnd(isExecPair, ctx.mkNot(sameAddress)));
+                        BooleanFormula sameAddress = imgr.equal((IntegerFormula)((MemEvent)load).getMemAddressExpr(), (IntegerFormula)(((MemEvent)store).getMemAddressExpr()));
```
ThomasHaas, line 154: Might cause problems if we use bitvectors for addresses (the address encoding does check for BitVectors AFAIK)

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/local/BasicRegRelation.java
```diff
@@ -74,25 +83,38 @@ BoolExpr doEncodeApprox(Collection<Event> regReaders, Context ctx) {
...
 
                 if(writers.isEmpty() || writers.get(0).getCId() >= regReader.getCId()){
-                    enc = ctx.mkAnd(enc, ctx.mkEq(register.toZ3Int(regReader, ctx), new IConst(BigInteger.ZERO, register.getPrecision()).toZ3Int(ctx)));
+                	BooleanFormula equal = register.toIntFormula(regReader, ctx) instanceof BitvectorFormula ?
+                			fmgr.getBitvectorFormulaManager().equal(
+                					(BitvectorFormula)register.toIntFormula(regReader, ctx), 
+                					(BitvectorFormula)new IConst(BigInteger.ZERO, register.getPrecision()).toIntFormula(ctx)) :
+                			fmgr.getIntegerFormulaManager().equal(
+                					(IntegerFormula)register.toIntFormula(regReader, ctx), 
+                					(IntegerFormula)new IConst(BigInteger.ZERO, register.getPrecision()).toIntFormula(ctx));
```
ThomasHaas, line 92: Generic Helper method for equality might help again.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/local/BasicRegRelation.java
```diff
@@ -74,25 +83,38 @@ BoolExpr doEncodeApprox(Collection<Event> regReaders, Context ctx) {
...
-                        )));
+                        enc = bmgr.and(enc, bmgr.equivalence(edge, clause));
+                        
+                        BooleanFormula equal = ((RegWriter)regWriter).getResultRegisterExpr() instanceof BitvectorFormula ?
+                        		fmgr.getBitvectorFormulaManager().equal(
+                                		(BitvectorFormula)((RegWriter) regWriter).getResultRegisterExpr(),
+                                		(BitvectorFormula)register.toIntFormula(regReader, ctx)) :
+                                fmgr.getIntegerFormulaManager().equal(
+                                		(IntegerFormula)((RegWriter) regWriter).getResultRegisterExpr(),
+                                		(IntegerFormula)register.toIntFormula(regReader, ctx));
```
ThomasHaas, line 115: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
```diff
@@ -86,56 +96,73 @@ protected BoolExpr encodeApprox(Context ctx) {
         ));
 
         for(Event e : eventsInit) {
-            enc = ctx.mkAnd(enc, ctx.mkEq(intVar("co", e, ctx), ctx.mkInt(0)));
+            enc = bmgr.and(enc, fmgr.getIntegerFormulaManager().equal(intVar("co", e, ctx), fmgr.getIntegerFormulaManager().makeNumber(BigInteger.ZERO)));
```
ThomasHaas, line 99: Refactor fmgr.getIntegerFormulaManager into local variable for the whole method

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelRf.java
```diff
@@ -88,86 +90,100 @@ public TupleSet getMaxTupleSet(){
...
 
         boolean canAccNonInitMem = task.getSettings().getFlag(Settings.FLAG_CAN_ACCESS_UNINITIALIZED_MEMORY);
         boolean useSeqEncoding = task.getSettings().getFlag(Settings.FLAG_USE_SEQ_ENCODING_REL_RF);
 
         for(Tuple tuple : maxTupleSet){
             MemEvent w = (MemEvent) tuple.getFirst();
             MemEvent r = (MemEvent) tuple.getSecond();
-            BoolExpr edge = this.getSMTVar(tuple, ctx);
+            BooleanFormula edge = this.getSMTVar(tuple, ctx);
+            IntegerFormula a1 = w.getMemAddressExpr() instanceof BitvectorFormula ?
```
ThomasHaas, line 108: There are so many copies of this code. A "convertToIntegerFormula" method in some Helper class would be better. Basically the new method you used in Memory but refactored into a static Helper class.
Also, the check for Addresses being BitVectors is not done cosistently (a lot of code just assumes Integers)

dartagnan/src/test/java/com/dat3m/dartagnan/AbstractDartagnanTest.java
```diff
@@ -64,19 +71,29 @@
         this.settings = settings;
     }
 
-    @Test
+    private void initSolverContext() throws Exception {
```
ThomasHaas, line 74: This should return the Context instead of assigning to some member variable since the context is used only locally anyway.

dartagnan/src/test/java/com/dat3m/dartagnan/AbstractDartagnanTest.java
```diff
@@ -64,19 +71,29 @@
...
             Program program = new ProgramParser().parse(new File(path));
-            if (program.getAss() != null) {
-                Context ctx = new Context();
-                Solver solver = ctx.mkSolver(ctx.mkTactic(Settings.TACTIC));
-                VerificationTask task = new VerificationTask(program, wmm, target, settings);
-                assertEquals(expected, runAnalysis(solver, ctx, task));
-                ctx.close();
-            }
+            VerificationTask task = new VerificationTask(program, wmm, target, settings);
+            initSolverContext();
```
ThomasHaas, line 90: If this method returns the Context, a try-with-resource scope can be used to guarantee cleanup on case of exceptions.

dartagnan/src/test/java/com/dat3m/dartagnan/AbstractSvCompTest.java
```diff
@@ -33,75 +39,74 @@
...
+    private SolverContext ctx;
+    
     public AbstractSvCompTest(String path, Wmm wmm, Settings settings) {
         this.path = path;
         this.wmm = wmm;
         this.settings = settings;
     }
 
-    @Test(timeout = TIMEOUT)
+    private void initSolverContext() throws Exception {
```
ThomasHaas, line 50: See comment in AbstractDartagnanTest

dartagnan/src/test/java/com/dat3m/dartagnan/DartagnanArrayValidTest.java
```diff
@@ -42,21 +48,34 @@
...
     private final Settings settings;
+    private SolverContext ctx;
 
     public DartagnanArrayValidTest(String path, Wmm wmm, Settings settings) {
         this.path = path;
         this.wmm = wmm;
         this.settings = settings;
     }
 
+    private void initSolverContext() throws Exception {
```
ThomasHaas, line 59: See comment in AbstractDartagnanTest.

dartagnan/src/test/java/com/dat3m/dartagnan/DartagnanBranchTest.java
```diff
@@ -83,14 +90,26 @@ public DartagnanBranchTest(String path, Result expected, Wmm wmm, Settings setti
         this.settings = settings;
     }
 
+    private void initSolverContext() throws Exception {
```
ThomasHaas, line 93: See comment in AbstractDartagnanTest

dartagnan/src/test/java/com/dat3m/dartagnan/SafeCTest.java
```diff
@@ -104,13 +113,25 @@ public SafeCTest(String path, Wmm wmm, Arch target, Settings settings, Result ex
         this.expected = expected;
     }
     
+    private void initSolverContext() throws Exception {
```
ThomasHaas, line 116: See comment in AbstractDartagnanTest

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -36,22 +43,53 @@ public String getVerdict(){
...
-            	case ASSUME:
-            		result = runAnalysisAssumeSolver(solver, ctx, task);
-            		break;
-            	case TWOSOLVERS:
-                    result = runAnalysis(solver, ctx, task);
-                    break;
+            Result result = Result.UNKNOWN;
+
+            ShutdownManager sdm = ShutdownManager.create();
+        	Thread t = new Thread(() -> {
```
ThomasHaas, line 49: Only create Thread if timeout is set.

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -36,22 +43,53 @@ public String getVerdict(){
...
+    			} catch (InterruptedException e) {
+    				// Verification ended, nothing to be done.
+    			}});
+
+            try {
+            	t.start();
+                Configuration config = Configuration.builder()
+                		.setOption("solver.z3.usePhantomReferences", "true")
+                		.build();
+				SolverContext ctx = SolverContextFactory.createSolverContext(
```
ThomasHaas, line 65: Use ctx with try-with-resource statement to do proper cleanup. Especially when a timeout interrupts the solving, we end up leaking a context

ThomasHaas (changes requested):
I have added comments for almost all issues I found.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
hernanponcedeleon: As discussed, the performance overhead is almost null so no need to create it only if a timeout is passed. I'm not sure what you mean by `closely to the place of usage` ... the timeout is supposed to be used for the whole running of the process so I place it early in the running process

ThomasHaas: I'm talking about the fact that the created Thread is started over 40 lines later in the code, so its creation can be moved to the point of first usage (around line 107).

hernanponcedeleon: Actually, my comment above contradicts the the decision I made when I put the `start` at that point. I'm moving all those together at the point where the start currently is

dartagnan/src/main/java/com/dat3m/dartagnan/expression/BExpr.java
hernanponcedeleon: About to be pushed

dartagnan/src/main/java/com/dat3m/dartagnan/expression/BNonDet.java
hernanponcedeleon: About to be pushed

dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
hernanponcedeleon: About to be pushed

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
hernanponcedeleon: I agree we can use a global context (maybe at the program level an initialise it when the program is initialise).
I think the used solver is irrelevant. This context is only used to check if a `CondJump` is a goto, i.e. its condition is always true.

ThomasHaas: Initializing at the program level is fine I suppose. The used solver is irrelevant as long as all solvers are available. But if a user decides to e.g. only install PRINCESS on the system, then using Z3 will break. It is somewhat mitigated by the fact that Z3 is available for all operating systems but I still can imagine use-cases where Z3 might not be available for some reason or another.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
hernanponcedeleon: What should be the behaviour if `rfEval == null`? In the old code the body iff the if would just not be executed. If this is the intended behaviour we can just do `if (rfEval)` right?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
hernanponcedeleon: The problem that we have right now it that the information about which solver the user selected is not available in this class (or `Program.class`) and making that information available would require passing information which is irrelevant for anything else between classes.
Once we have a better way of handling configuration options as #79, this could be done much easier.

ThomasHaas: Can't this information be part of the VerificationTask? Maybe in the Settings?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
ThomasHaas: As the code comment suggests, there are some cases where some rf edge is not encoded and in that case we generally assume that that read-from relation simply didn't happen (i.e. rf(e1, e2) not encoded == rf(e1, e2) is false). This can happen due to active set reasoning. ``` if (rfEval) ``` doesn't work because it implicitly unboxes, causing a NullPointerException if the edge was not encoded (rfEval is a boxed boolean)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
hernanponcedeleon: It might, but I don't see the point in adding stuff to the Settings if we will (hopefully soon) get rid of this class

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
hernanponcedeleon: ok ... `if(null)` would throw and exception instead of consider that the condition is not satisfied.

hernanponcedeleon:
A general comment on an issue that repeated several times in the review: the handling/conversion of bit-vector and integers.
I agree with all you points and partially implemented some of your suggestions as having a `Helper` to encapsulate `equals`.
However, I consider some changes (e.g. unsafe cast to `IntegerFormula` when converting addresses) more related to #82 than this PR. While the current status is not optimal, the proposed code does not deteriorates the quality of the code that is already there and thus I propose to handle those problems as part of #82 so this can be merge ASAP.

ThomasHaas:
I agree that the inconsistent usage of conversions is orthogonal to this PR. I just mentioned it because this flaw of the master branch was partially addressed by your code changes (sometimes you added the missing type checks). Either way, we need to address this problem together with #82 in a different PR.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
hernanponcedeleon: I moved the `defaultCtx` to `Program` and "initialise" the field when the program is initialised.
The good thing about this is that when the program is initialised, the context makes use of the correct solver.
The problem is that there are many parts of the code (e.g. preprocessing, equivalence) that make use of `CondJump.isGoto()` (which requires the `defaultCtx`) and are called before `Program.initialise()`. I avoid a `NullExpception` problem by checking if the context was already initialised (which in most cases was not) and if not, I initialise it. Unfortunately here we run into the same problem that we cannot pass the correct solver so we use Z3 by default.
Your proposal of using the `VerificationTask` would have the same problem (unless we initialise the `defaultCtx` in one of the first passes of the program, i.e. before preprocessing and commuting equivalence classes).

hernanponcedeleon: For now I'll revert this change since it results in the following problem with Z3
```
ASSERTION VIOLATION
File: ../src/ast/shared_occs.cpp
Line: 79
UNEXPECTED CODE WAS REACHED.
Z3 4.8.10.0
Please file an issue with this message and more detail about how you encountered it at https://github.com/Z3Prover/z3/issues/new
```
I suspect it is related to Z3Prover/z3#5107, but until JavaSMT does not release new bindings for 4.8.11.0 I cannot be sure.

ThomasHaas: We can keep Z3 for now, but if you decide to do initialization this way, I would rather use a private static field and a public static getter. The getter can initialize everything if the value is still null (so it will automatically get initialized on first usage). But generally I would probably move the default context to a different place since Program is not the only class that could use a default context. If you have this new Helper class, maybe you can create a static method there that has the functionality that is needed by CondJump.isGoto(). Then the helper class would own the default context. But I suppose we can fix this problem outside of this PR.

hernanponcedeleon:
Your changes seem ok to me. Two things that are still missing before merging
- [ ] Update README to add solver options
- [ ] Figure it out if everything is ok in terms of licenses

ThomasHaas (approved):
All issues have been addressed. If there are no more problems we can merge this into master.

hernanponcedeleon:
@ThomasHaas can you please check if the new README makes sense or we need to add some more description? Once we are done with that I can merge.

ThomasHaas:
The only things that are missing from the README are the fact that we also support .i files and that we can choose between 3 different solving methods (incremental, assume, two solvers). The latter is probably quite irrelevant for the user, but once we add Refinement as another option (in some other PR), we will want to give a choice between solving methods so we might already do it now.

hernanponcedeleon:
`*.I` are just preprocessed `*.c` files so no need to mention them in the options. We agreed on avoiding mentioning the method option for the time being (in might confuse the user). We will re-visit this decision when we merge the refinement method.

Issue #83, opened 2021-08-16
Atomicity block in Witness


hernanponcedeleon:
In the way we encode atomicity, it is possible there exists a context switch inside an atomic block as long as this does not interfere with the verification results, i.e. the order of instructions the solver returns can be re-arrange by interchanging non-conflicting instructions in such a way that we obtain an execution where instruction of the atomic block are one next to the other.

SVCOMP requires no context switch during an atomic block, thus we need to re-arrange the order we obtain from the solver when we create the witness.

hernanponcedeleon:
Merged

PR #86, opened 2021-08-18
Witness reorder

hernanponcedeleon:
This PR tackles #83.
The solves has the freedom to given an order (based on `acyclic(hb)`) which can perform a context switch during an atomic block. This is not allowed by SVCOMP.
When building the witness, we perform a re-ordering to keep all elements of an atomic block one next to the other.


ThomasHaas (approved):
If I didn't accidentally added any bugs, the changes can be merged.

PR #87, opened 2021-08-19
60

hernanponcedeleon:
We create use a new RMW store that cannot fail and use it in the compilation of AtomicFetchOp


ThomasHaas:
The code is not right. If you set ```execVar = true```, you will force the event to be executed in every execution.
Instead you should do ```execVar = cfVar```, which will give the intended effect.
Also I'm not too happy with the extra class. A simple extra flag ("NoFail" or "Strong") would be sufficient. Furthermore, the current implementation will still create the unneeded execVar in the solver context due to the call to ```super.initialiize(...)```. This would be avoided with a flag (I have a working implementation on the haas branch for reference).

hernanponcedeleon:
I basically replace the content of the corresponding classes from your branch so this should be exactly what you described, but please check

ThomasHaas:
The change looks correct. But I think you need to make changes to AtomicXchg as well (it still uses weak behaviour). And then we might want to actually parse the C-Code correctly to also handle AtomicCmpXchg as well (that is the only operation which has strong and weak versions, but we always compile to a weak CAS right now).

ThomasHaas:
Do you want to handle AtomicCmpXchg in this PR as well?

hernanponcedeleon:
Yes, I'll try to add the changes to this PR as well.
If I understood it correctly, we need to use the two way of compiling we had at different points:
- weak: the one using the status register
- strong: the one with fake dependency

I also need to add the `Etype.STRONG` when parsing and use it to decide the compilation scheme

ThomasHaas:
The easiest change is to simply create the RMWStore with the appropriate strongness without changing anything else. This will already be correct. A little more involved is to optimize the ```strong``` case just like you did for FetchOp and Xchg.
As you said, you will need to adapt the parsing. But as far as I can tell, it's just a matter of calling ```String.contains("strong")``` on the CAS operation and depending on the result, you can tag the generated CAS with ```EType.STRONG```.

hernanponcedeleon:
Optimising for strong would result in convoluted code due to the cases and I don't think this would improve performance in a measurable way, so I opted not do to it.

There is still the problem that we are not parsing the second `mo` tag, but this was already the case for our previous version which could still fail (if I understood correctly, the second `mo` is for when the comparison fails), thus independent of this code.

ThomasHaas:
We did not implement the parsing of the second ```mo``` mainly because it was not clear how the compilation works (no paper addresses it). As far as I remember, most compilers I tested using https://godbolt.org/ did generate the same code irrespective of the second ```mo```. Furthermore, the store in the fail case is most likely targeting a local address, making the seocnd ```mo```irrelevant.

ThomasHaas (approved):
Everything looks fine to me now.

PR #88, opened 2021-08-19
Removed duplicated code and clean up of code

hernanponcedeleon:
Method and MethodTyples classes were duplicated so I removed the later.
I also clean up some code, and made more consistent the use of options both from the UI and the console.


dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
```diff
@@ -80,11 +80,8 @@ public void parse(String[] args) throws ParseException, RuntimeException {
         programFilePath = cmd.getOptionValue("input");
         targetModelFilePath = cmd.getOptionValue("cat");
 
-        if(cmd.hasOption("target")) {
-            target = Arch.get(cmd.getOptionValue("target"));
-        }
-
-        method = MethodTypes.fromString(cmd.getOptionValue(METHOD_OPTION, INCREMENTAL.toString()));
+        target = Arch.get(cmd.getOptionValue("target", Arch.NONE.toShortString()));
```
ThomasHaas, line 83: Can you double check that this is correct? I remember that I looked through the code before and didn't change this cause there was a slight difference between ```target == NONE``` and ```target == null```.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Arch.java
```diff
@@ -1,7 +1,7 @@
 package com.dat3m.dartagnan.wmm.utils;
 
 public enum Arch {
-    NONE, ARM, ARM8, POWER, TSO;
+    ARM, ARM8, NONE, POWER, TSO;
```
ThomasHaas, line 4: Did you reorder the options alphabetically? I find NONE as a first option a lot clearer cause it has a special meaning and is the default value in most cases.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
hernanponcedeleon: None of the way we have to call the verification was using a `null` target
- `Dartagnan` had this code
```
        if(target == null) {
            System.out.println("Compilation target cannot be inferred");
            System.exit(0);
            return;
        }
```
- In the UI there was always a target as default
- For the UnitTest we were always setting this arch.

We now make it explicit that the default `Arch` is `NONE`

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Arch.java
hernanponcedeleon: Since the `Selector` ordered them using the enumeration order, I did this.
However I now make that the `Selector` class order whatever collection of elements we use.

ThomasHaas: Hmm, the ```Selector``` ordered them according to enumeration order? If so, why don't we use NONE first? It sounds to me like you want an alphabetical order in the ```Selector``` but isn't the order as we have it in the enum better? NONE first sounds like the best option, especially if we default to it.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
ThomasHaas: Ok, great

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Arch.java
hernanponcedeleon: When the selector is expanded, the options are shown in the order coming from the array that is used to create the selector. This order coincides with the enum order. I wanted to display the order alphabetically because I think it is nicer to the user, in particular when there are quite a few options.

Unless specified (which we do), the selector selects by default the first option.
I modified the selector created to do the ordering here (this implies that now also the solvers get ordered). I also implement all enum classes we use as options in a consistent way and added a `getDefault` method do now it is clear in each class which is the default option (before this info was done during the parsing or UI panels creation).

ThomasHaas: I understand that ```Selector``` uses the order of the array. But I'm wondering why the option NONE should be alphabetically ordered with all the other architectures. Our initial enum layout was NONE and then the Architectures in alphabetical order. I would say this layout is ideal, cause it puts the special NONE value first (which also happens to be the default) and then orders the remaining one's alphabetically. I would use just that order for the user as well, cause having the common value NONE all over the place (it might be at very different places for different options) feels wrong to me. NONE should probably the first item for all ```Selectors```, even for the Alias analysis.

ThomasHaas:
The new commit contains .smt test files. Are those intended to be there?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Arch.java
hernanponcedeleon: I don't think relying in the order that enums are defined is good, this I added a method that returns the intended order. I also create a helper class which handles the order for `Solvers`. As you suggested, for options having a `NONE`, I put this first in the order and the rest based on alphabetical order of they `toString()` representation.

hernanponcedeleon:
> The new commit contains .smt test files. Are those intended to be there?

No, sorry ... I just removed them

ThomasHaas (approved):
Everything looks fine to me. You can merge it.

Issue #89, opened 2021-08-20
Replacing explicit event constructors by static factory methods


ThomasHaas:
To reduce coupling and give a central place to construct events for a program, we should get rid of calling constructors directly and instead use static factory methods.
Let me demonstrate this by an example.

Old:
```
Local local = new Local(...);
RMWStoreExclusive rmwStore = new RMWStoreExclusive(...);
Fence iSync = new Fence("ISync");
```

New:
```
Local local = Events.newLocal();
RMWStoreExclusive rmwStore = Events.AARCH64.newRMWStoreExclusive();
Fence iSync = Events.POWER.newISyncFence();
```

As the example suggests, we can even use inner classes to differentiate between events of different archictures.
We could even use something like ```Events.DAT3M.```for all our internal events and ```Èvents.[ARCH].``` for all architecture specific events that need further compilation to our internal representation.

The benefits are as follows:
- There is less coupling between needed events and actually used implementation.
- All available events are accessible in a single place, which makes it easier to get an overview without searching through the extensive directory structure we have right now.
- We can easily add/remove Event classes without changing the calling code. E.g. we might choose to implement strong/weak RMW events using two separate classes but the calling code can use ```Events.newRMWStoreExclusive(strongFlag)``` which returns the appropriate implementation. Similarly, we might also add two functions ```Events.newWeakRMWStoreExclusive()``` and ```Events.newStrongRMWStoreExlusive()``` independent of whether we actually use two implementing classes or just one class with a strongness flag.

PR #92, opened 2021-08-23
Eventfactory

ThomasHaas:
This PR addresses #89. The major changes are as follows:
- A new static factory class ```Events``` is used to create events.
- Linux specific events are moved to the Linux directory (rmw.cond was only used for CLitmus tests and nowhere else)
- RMW exclusive events where moved from AArch64 to the rmw directory (we use those events not only for Arm but also for Power!)
- ```Events``` also contains new convenience functions for creating fake ctrl dependencies, unconditional jumps (gotos) and jumps with negated conditions.

The structure of the new ```Events``` class mirrors the directory structure of our ```events``` folder (with some changes). It has various inner classes to achieve this structure.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/Events.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/Events.java
@@ -1,0 +1,355 @@
+package com.dat3m.dartagnan.program;
+
+import com.dat3m.dartagnan.expression.*;
+import com.dat3m.dartagnan.expression.op.BOpUn;
+import com.dat3m.dartagnan.expression.op.COpBin;
+import com.dat3m.dartagnan.expression.op.IOpBin;
+import com.dat3m.dartagnan.program.arch.aarch64.event.StoreExclusive;
+import com.dat3m.dartagnan.program.arch.linux.event.*;
+import com.dat3m.dartagnan.program.arch.linux.event.cond.*;
+import com.dat3m.dartagnan.program.arch.tso.event.Xchg;
+import com.dat3m.dartagnan.program.atomic.event.*;
+import com.dat3m.dartagnan.program.event.*;
+import com.dat3m.dartagnan.program.event.pthread.*;
+import com.dat3m.dartagnan.program.event.rmw.*;
+import com.dat3m.dartagnan.program.memory.Address;
+import com.dat3m.dartagnan.program.svcomp.event.BeginAtomic;
+import com.dat3m.dartagnan.program.svcomp.event.EndAtomic;
+
+import java.math.BigInteger;
+
+public class Events {
+
+    // Static class
+    private Events() {}
+
+
+    public static Local newLocal(Register register, ExprInterface expr, int cLine) {
+        return new Local(register, expr, cLine);
+    }
+
+    public static Local newLocal(Register register, ExprInterface expr) {
+        return newLocal(register, expr, -1);
+    }
+
+    public static CondJump newJump(BExpr cond, Label target) {
+        return new CondJump(cond, target);
+    }
+
+    public static CondJump newJumpUnless(ExprInterface cond, Label target) { return newJump(new BExprUn(BOpUn.NOT, cond), target); }
+
+    public static IfAsJump newIfJump(BExpr expr, Label label, Label end) {
+        return new IfAsJump(expr, label, end);
+    }
+
+    public static IfAsJump newIfJumpUnless(ExprInterface expr, Label label, Label end) {
+        return newIfJump(new BExprUn(BOpUn.NOT, expr), label, end);
+    }
+
+    public static CondJump newGoto(Label target) {
+        return newJump(BConst.TRUE, target);
+    }
+
+    public static CondJump newFakeCtrlDep(Register reg, Label target) { return newJump(new Atom(reg, COpBin.EQ, reg), target); }
+
+    public static Fence newFence(String name) {
+        return new Fence(name);
+    }
+
+    public static FenceOpt newFenceOpt(String name, String opt) {
+        return new FenceOpt(name, opt);
+    }
+
+    public static Init newInit(IExpr address, IConst value) {
+        return new Init(address, value);
+    }
+
+    public static Label newLabel(String name) {
+        return new Label(name);
+    }
+
+    public static Load newLoad(Register register, IExpr address, String mo, int cLine) {
+        return new Load(register, address, mo, cLine);
+    }
+
+    public static Load newLoad(Register register, IExpr address, String mo) {
+        return newLoad(register, address, mo, -1);
+    }
+
+    public static Store newStore(IExpr address, ExprInterface value, String mo, int cLine) {
+        return new Store(address, value, mo, cLine);
+    }
+
+    public static Store newStore(IExpr address, ExprInterface value, String mo) {
+        return newStore(address, value, mo, -1);
+    }
+
+    public static Cmp newCompare(IExpr left, IExpr right) {
+        return new Cmp(left, right);
+    }
+
+    public static Skip newSkip() {
+        return new Skip();
+    }
+
+    public static FunCall newFunctionCall(String funName, int cLine) {
+        return new FunCall(funName, cLine);
+    }
+
+    public static FunRet newFunctionReturn(String funName, int cLine) {
+        return new FunRet(funName, cLine);
+    }
+
+    public static RMWLoad newRMWLoad(Register reg, IExpr address, String mo) {
+        return new RMWLoad(reg, address, mo);
+    }
+
+    public static RMWStore newRMWStore(RMWLoad loadEvent, IExpr address, ExprInterface value, String mo) {
+        return new RMWStore(loadEvent, address, value, mo);
+    }
+
+    public static RMWLoadExclusive newRMWLoadExclusive(Register reg, IExpr address, String mo) {
+        return new RMWLoadExclusive(reg, address, mo);
+    }
+
+    public static RMWStoreExclusive newRMWStoreExclusive(IExpr address, ExprInterface value, String mo) {
+        return new RMWStoreExclusive(address, value, mo);
+    }
+
+    public static RMWStoreExclusive newRMWStoreExclusive(IExpr address, ExprInterface value, String mo, boolean isStrong) {
+        return new RMWStoreExclusive(address, value, mo, isStrong);
+    }
+
+    public static RMWStoreExclusiveStatus newRMWStoreExclusiveStatus(Register register, RMWStoreExclusive storeEvent) {
+        return new RMWStoreExclusiveStatus(register, storeEvent);
+    }
+
+
+    public static class Pthread {
+        private Pthread() {}
+
+        public static Create newCreate(Register pthread_t, String routine, Address address, int cLine) {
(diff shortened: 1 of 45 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/Events.java
```diff
@@ -0,0 +1,355 @@
...
+import com.dat3m.dartagnan.program.event.*;
+import com.dat3m.dartagnan.program.event.pthread.*;
+import com.dat3m.dartagnan.program.event.rmw.*;
+import com.dat3m.dartagnan.program.memory.Address;
+import com.dat3m.dartagnan.program.svcomp.event.BeginAtomic;
+import com.dat3m.dartagnan.program.svcomp.event.EndAtomic;
+
+import java.math.BigInteger;
+
+public class Events {
```
hernanponcedeleon, line 21: I think EventFactory would be a better name. It is more conventional and less confusing with the `Event` class

dartagnan/src/main/java/com/dat3m/dartagnan/program/Events.java
```diff
@@ -0,0 +1,355 @@
...
+
+    public static RMWStore newRMWStore(RMWLoad loadEvent, IExpr address, ExprInterface value, String mo) {
+        return new RMWStore(loadEvent, address, value, mo);
+    }
+
+    public static RMWLoadExclusive newRMWLoadExclusive(Register reg, IExpr address, String mo) {
+        return new RMWLoadExclusive(reg, address, mo);
+    }
+
+    public static RMWStoreExclusive newRMWStoreExclusive(IExpr address, ExprInterface value, String mo) {
```
hernanponcedeleon, line 115: I think it would be better to keep a unique constructor for the `RMWStoreExclusive` event and make this method to call the one below with false as the last parameter.

ThomasHaas: That was just a mistake. Will fix this.

hernanponcedeleon:
There are a couple of missing uses of the factory. See comments in the code

dartagnan/src/main/java/com/dat3m/dartagnan/program/Events.java
ThomasHaas: I thought about it. But I think it gets unecessarily long and quite redundant if we use ```Factory``` on every layer, e.g.
```EventFactory.PthreadFactory``` is redundant. It makes sense if we use singleton classes but not for purely static classes.
Also the top-level class needs to have the same name as the file AFAIK, so ```ProgramFactory``` as a file name is not possible if the top level class is called ```EventFactory```. Either way, ```ProgramFactory``` is IMO not a good name cause the factory doesn't produce programs but only events.

In short, I can rename the file and the top-level class to ```EventFactory``` but I would keep the inner classes as is.

**Comment by ThomasHaas** [comment]
What do you think about the naming of the factory methods themselves? For the initial change I stuck with a naming very close to the original constructor calls:
- Every method name starts with ```new```
- Most method names include the exact name of the implementing class. I only did a few changes like replacing ```Xchg```with ```Exchange```.

But I'm open to changing names.

Also, I would like to clean up the code construction some more. There are some constructs that are repeatedly found, like e.g. fake dependencies. It might be interesting to add a class ```EventFactory.Compund``` or ```EventFactory.Complex``` that has some helper methods to create common sets of events (e.g. Label + CondJump as fake dep). 

I would also like to get rid of overly specific events. For example, ```RMWStoreExclusiveStatus``` is tied to ```RMWStoreExclusive``` but its implementation is completely generic in the event whose status should be tracked. We could simply change it to e.g. ```ExecutionStatus``` which tracks whether some arbitrary event was executed. We would only use it for ```RMWStoresExclusive``` for now, but we might want to use it for other events in the future.
Similarly, we could get rid of ```FenceOpt``` which is just ```Fence``` with a special tag. We could just let the factory method do the tagging that ```FenceOpt``` does. And if in any case we need a separate class for these Fences, we can easily change the implementation of the factory method. Also ```RMWLoad``` and ```RMWLoadExclusive``` are nothing but regular ```Loads``` with an extra tag (btw. why is ```RMWLoadExclusive``` not tagged as ```RMW```?).
What do you think about simplifying the structure and reducing the number of classes we have?

**Comment by hernanponcedeleon** [comment]
> What do you think about the naming of the factory methods themselves? For the initial change I stuck with a naming very close to the original constructor calls:

The naming seems good to me, I think they are quite descriptive.

> Also, I would like to clean up the code construction some more. There are some constructs that are repeatedly found, like e.g. fake dependencies. It might be interesting to add a class `EventFactory.Compund` or `EventFactory.Complex` that has some helper methods to create common sets of events (e.g. Label + CondJump as fake dep).

If I understand correctly, you refer to situations where a couple of events are generally created al together. We could have some factory methods that create a couple of events at the same time and return a list instead of a single one (if this is what you refer to)

> I would also like to get rid of overly specific events. For example, `RMWStoreExclusiveStatus` is tied to `RMWStoreExclusive` but its implementation is completely generic in the event whose status should be tracked. We could simply change it to e.g. `ExecutionStatus` which tracks whether some arbitrary event was executed. We would only use it for `RMWStoresExclusive` for now, but we might want to use it for other events in the future.

If the implementation does not become more complex, I'm ok with generalising already. However if the implementation gets more complex and we don't yet have a a use case, I don't see the point.

> Similarly, we could get rid of `FenceOpt` which is just `Fence` with a special tag. We could just let the factory method do the tagging that `FenceOpt` does. And if in any case we need a separate class for these Fences, we can easily change the implementation of the factory method. Also `RMWLoad` and `RMWLoadExclusive` are nothing but regular `Loads` with an extra tag (btw. why is `RMWLoadExclusive` not tagged as `RMW`?).

For what I can tell, the tag was not added because we were never filtering `EType.RWM` intersection `Type.Load`. Adding the tag should not break anything.

> What do you think about simplifying the structure and reducing the number of classes we have?

The less classes we have, the better. Now with the factory it is as easy as creating a new factory method, calling the constructor of `Load` and just adding the corresponding tag.

**Comment by ThomasHaas** [comment]
I added another (experimental) commit that removes some events like RMWLoad, RMWLoadExclusive and FenceOpt.

**Comment by hernanponcedeleon** [comment]
With the new factory, I think we should keep only one constructor per event class and expose the different ways to create the same class of events only from the factory. For example, in the factory we have two methods
```
public static Load newLoad(Register register, IExpr address, String mo, int cLine) {
        return new Load(register, address, mo, cLine);
    }

    public static Load newLoad(Register register, IExpr address, String mo) {
        return newLoad(register, address, mo, -1);
    }
```
and in `Load.class` we have
```
public Load(Register register, IExpr address, String mo, int cLine) {
        super(address, mo, cLine);
        this.resultRegister = register;
        addFilters(EType.ANY, EType.VISIBLE, EType.MEMORY, EType.READ, EType.REG_WRITER);
    }

    public Load(Register register, IExpr address, String mo) {
    	this(register, address, mo, -1);
    }
```
The second constructor is useless. The same happens in other classes.

Also, `AtomicStore` and `AtomicLoad` are pretty much identical to `Store` and `Load` except that they have a compilation. What do you think of moving the compilation to `Store` and `Load` respectively and get id of those two classes too?

ThomasHaas:
How do you intend to move e.g. ```AtomicStore``` to ```Store``` if it compiles down to a Store + various other operations? If we did this, then the compilation would be somewhat recursive (a ```Store``` gets compiled to multiple events, including another ```Store```). I think it is best to seperate between primitive events (no further compilation required) and non-primitive events (need to be compiled). Another problem with reducing the amount of classes we use is the fact that we cannot nicely adapt the string representation as necessary. Even if two events are basically identical, they still might have different ```toString``` representations or use different (architecture specific) tagging for good reasons. This limits the amount of reduction we can do, so I'm not sure if we can actually get rid of many classes. 

About the constructors, I will change that. I would also like to make the constructors not directly accessible but unfortunately that's not possible.

hernanponcedeleon:
It might be a good idea of creating `primitive` and `non-primitive` interfaces to make this clear in the code.
Method `compileRecursive` would only exists for classes implementing the first one.
We have other kind of information that could be used to e.g. differentiate if a `Store` should be compiled down or not.
For example, we have two levels of `Mo` orders: `com.dat3m.dartagnan.program.atomic.utils.Mo` (C-level event which needs to be compiled) and `com.dat3m.dartagnan.program.arch.aarch64.utils.Mo` (architecture level event which does not need to be compiled).

We could use things like the `Mo` or the tag too decided the `toString()` ... thing would make the method a little longer, but we will end up will less classes.

About the constructors: can't you mark them as protected and at least limit their visibility?

ThomasHaas:
I think it is better to have separate classes than a single class with a huge implementation for all sorts of Mo tags and architectures. Cause you will have to change this implementation whenever you add some new Mo or new Architecture.
If it is just about ```toString``` we could also provide the primitive events with a ```setRepresentation``` method that can get called from the factory to change the representation as needed. 
I think events have mainly 3 distinguishing components:
- their syntactic representation (```toString```)
- their tagging
- their semantic encoding

If events only differ in the first two, we could solve this in the factory methods. Otherwise, we need separate classes. But I'm not too convinced that this will be clean. For example, ```RMWLoad``` is just a load with an extra tag. On the other hand, ```RMWStore``` is not simply a store with some extra tag and hence needs an extra class. Now we have a discrepancy when we remove the ```RMWLoad``` class (which I did in the most recent commit). Especially because ```RMWStore``` takes an ```RMWLoad``` as parameter. To solve this, I changed ```RMWStore``` to take any ```Load``` but now you can combine such a ```RMWStore``` with a regular (non-rmw) load which is strange.

Coming up with a proper and clean solution is something we need to discuss further, so I think I will revert the last change that got rid of some events and only keep the changes that are strictly better. We should probably handle this in a different discussion/PR when we also consider clearing up the whole compilation process and possibly introduce compileable and non-compilable (primitive) events.

hernanponcedeleon:
I agree we can further discuss this somewhere else.
Just a comment about the example you gave on RMWLoad/Store. I actually liked it because it also removed some castings. The problem about using arbitrary Loads with RMWStores can be solved via a check/assertion like `load.is(Type.RMW)`

ThomasHaas:
I totally got rid of the ```cLine``` parameter in constructors. I think this parameter is unfitting for construction, since it is simply not needed to obtain a working event. I think such additional/optional parameters should be set by the factory method and do not need to be specified on construction. I'm thinking of giving a similar treatment to ```mo``` for memory events.

In general, I would try to reduce the constructors down to only the essential parameters and add any further/optional parameters via setters. These setters can be used by the Factory to imitate complex constructions.
Maybe we could even create a sort of EventBuilder that allows fluent construction like this:
```Store store = EventBuilder.store([essential parameters]).withCLine(cLine).withMo(EType.Relaxed).withFilters(EType.RMW).create()```
I'm not sure if this is doable in Java without writing tons of boilerplate code though.

hernanponcedeleon:
I like the idea of using a builder (and overall your current approach of start using common design pattern when possible).
It is true that we might end up having some boilerplate code (`builder.withCline()` will probably be pretty much the same as the `setCLine()` for the event), however using a builder also allows to make the constructors private (if you put both the constructor and the builder in the same class).

PR #100, opened 2021-08-28
Removed the buggy alias flag

hernanponcedeleon:
I removed the buggy alias flag (otherwise the Linux tests timeout and the build fails) and I tried to add a work around for the problem we had with the alias analysis.
The problem was that `processRegs` was handling local events such as `r1 <- r2` and `r <- mem` but not `r1 <- r2 + X` and thus the set of address it computed was an under approximation. For such cases I just added the whole `maxAddressSet`, i.e. we fully over approximate.


ThomasHaas (approved):
Since the tests all work fine and the changes are trivial, there is no room for me to reject the PR :). You can merge it.

Issue #110, opened 2021-09-02
Problem with alias analysis


hernanponcedeleon:
The alias analysis is (once again) buggy. The problem is here https://github.com/hernanponcedeleon/Dat3M/blob/16df56e24c2e91b0e44daf384f3007d5a4fabf82/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/alias/AliasAnalysis.java#L58-L68

This code is wrong and it does not represent the rule stated by the comment. Based on the rule, the event should be a `Local`, but the code iterates over `MemEvent`. This means that the analysis can say that `r` points-to `mem0` while we should return the whole `maxAddressSet` (unless we have further precise informatoin) for code like this 
```
r = load(mem0)
```
This is the cause of the problem we are having with `ticketlock` in `refinement`

The rule in the comment is actually covered by this part of the code
https://github.com/hernanponcedeleon/Dat3M/blob/16df56e24c2e91b0e44daf384f3007d5a4fabf82/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/alias/AliasAnalysis.java#L173-L176

hernanponcedeleon:
Solved in 44859cb48e024820ab0cf4cf3cccb928d757a205

PR #111, opened 2021-09-02
Remove wrong case for alias analysis

hernanponcedeleon:
See #110


ThomasHaas (approved):
I don't know if this fixes (all) the bugs, but since the tests still go through you can merge it.

Issue #108, opened 2021-09-01
isGoto()


hernanponcedeleon:
We need to find a way of implementing `CondJump.isGoto()` without relying on a solver.
One possible way is to use `ExpInterface.reduce()`, but we might need to improve this to reduce a `ExpInterface` instead of a `IConst`.

PR #112, opened 2021-09-02
Computing isGoto()

hernanponcedeleon:
This PR targets #108.

I improved the typing of the reduce method (I can return `BConst` or `IConst` now) and I use this to implement `isGoto()` instead of having to make a call to the solver.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
@@ -54,8 +54,8 @@ public class BConst extends BExpr implements ExprInterface {
 	}
 
 	@Override
-	public IConst reduce() {
-		return value ? IConst.ONE : IConst.ZERO;
+	public ExprInterface reduce() {
+		return new BConst(value);
 	}
 	
 	public boolean getValue() {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
@@ -54,7 +54,7 @@ public class BConst extends BExpr implements ExprInterface {
 	}
 
 	@Override
-	public ExprInterface reduce() {
+	public BConst reduce() {
 		return this;
 	}
 	
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
@@ -55,7 +55,7 @@ public class BConst extends BExpr implements ExprInterface {
 
 	@Override
 	public ExprInterface reduce() {
-		return new BConst(value);
+		return this;
 	}
 	
 	public boolean getValue() {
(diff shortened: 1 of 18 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/expression/BConst.java
```diff
@@ -54,8 +54,8 @@ public boolean getBoolValue(Event e, Model model, SolverContext ctx){
 	}
 
 	@Override
-	public IConst reduce() {
-		return value ? IConst.ONE : IConst.ZERO;
+	public ExprInterface reduce() {
+		return new BConst(value);
```
ThomasHaas, line 58: Just use ```return this```.
We should avoid creating new BConst all the time. Ideally, we would simply use a static factory method that only ever returns the same two instances (TRUE and FALSE).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/StdProcedures.java
```diff
@@ -45,19 +42,6 @@
...
-			visitor.programBuilder.addChild(visitor.threadCount, EventFactory.newStore(address, value, "NA"));
-			return;
-		}
-//		if(name.startsWith("READ_ONCE")) {
-//			System.out.println(ctx.getText());
-//			IExpr address = (IExpr)ctx.call_params().exprs().expr(0).accept(visitor);
-//			ExprInterface value = (ExprInterface)ctx.call_params().exprs().expr(1).accept(visitor);
-//			visitor.programBuilder.addChild(visitor.threadCount, new Store(address, value, "NA"));
-//			return;
-//		}
```
ThomasHaas, line 60: What do have these changes to do with the PR?

ThomasHaas (changes requested):
I don't really understand the idea of the reduce method. Wouldn't it make more sense to "simplify" the expressions, even if it is only partially doable? Right now, the method throws an exception if a full reduction to a constant is not possible, but I would simply return the original expression if no simplifciation is possible (or a partial simplficiation was done). What advantage do we have in throwing exceptions in such cases?

Also, you should override the reduce method in ```IExpr``` and ```BExpr``` to return ```IConst``` and ```BConst``` instead of ```ExprInterface``` (you can specialize the return types!). That way, you don't have to do the casting in a lot of cases.

hernanponcedeleon:
The initial use case of `reduce()` was to be used in cases where the reduction was guaranteed to end up in a constant, e.g. to reduce things like `mem0 + (0*8) + (1*0)` to `mem0 + 1`.
The simplification would probably not have much impact on the solver, but it makes reading the printed program much easier.
In particular this was used to for the initial values of locations that must always be a constant (but in `_SMACK_init()` might not).

The new use case (in `isGoto()`) is the first one where an exception can be thrown.

One disadvantage with returning partial reductions is that I could not override the return type of e.g. `IExpr` to `IConst` and thus I would still need to do most of the casts.

One alternative I see is to have two methods, `reduce()` returning `XConst` (according to the type) and `simplify` returning `XExp` ... then I can keep using `reduce` for the initial use case and use `simplify` for `isGoto()` where I would check if the simplification returns something `instanceof BConst`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/StdProcedures.java
hernanponcedeleon: Nothing, that code was related to the thesis of one bachelor student but which we will not integrate and thus we can get rid of this. If just made it here because only here I noticed.

ThomasHaas:
We can keep ```reduce``` if you say that in basically all use cases it is safe. We can add the ```simplify``` method in some other PR. So if you specalize the return types for ```IExpr.reduce``` and ```BExpr.reduce```, I will approve the PR.

ThomasHaas (approved):
It looks fine now.

Issue #85, opened 2021-08-18
Assume method not working for MathSAT5


hernanponcedeleon:
MathSAT5 only allows assumptions as negated literals.
We need to change the way we encode the assumption to satisfy the requirement that assumptions are negated literals.

PR #119, opened 2021-09-08
Solver with assumption using just literals

hernanponcedeleon:
See #85.

I tested in the Linux container and now the litmus tests using `assume` method and `mathsat5` work.


ThomasHaas:
I don't like the fact that this simple change to the assume solver spreads over the ```Utils``` and ```VerificationTask``` class.
I would simply integrate this directly into the assume solver method like this:
```
        BooleanFormulaManager bmgr = ctx.getFormulaManager().getBooleanFormulaManager();
        BooleanFormula assumptionLiteral = bmgr.makeVariable("DAT3M_assertion_assumption");
        BooleanFormula assumedAssertion = bmgr.implication(assumptionLiteral, task.encodeAssertions(ctx));
        prover.addConstraint(assumedAssertion);
        
        prover.isUnsatWithAssumptions(singletonList(assumptionLiteral));
```

This way, you only change one class, it is more readable because you don't need to check the implementation across multiple classes, it is more obvious that you actually assume a single literal and it is clear what this assumption is exactly assuming. I think you end up writing even less code and you don't pollute external classes with code only used by the assume solver.

ThomasHaas (approved):
Looks fine to me. You can merge when the tests are finished.

PR #126, opened 2021-09-27
Update JavaSMT version and add yices2 as a solver

hernanponcedeleon:


ThomasHaas:
So Yices2 should now be useable out-of-the-box in Linux systems, right? I cannot test this PR on my Mac but since it it is a tiny change, I would simply approve this PR if you are fine with that.

hernanponcedeleon:
I already tested it in Linux and "it works" (I did not do heavy testing, I just checked that the verification starts and finishes with the correnc SMT solver)

ThomasHaas (approved):

PR #129, opened 2021-09-29
Compilation schemes and new tests

hernanponcedeleon:
This PR cons tis on mainly 3 changes:

- Added an encoding for `SREM` and `UREM` operations. I only did it for Integer encoding for two reasons: We currently only use this for concurrency benchmarks (thus no use of bitvectors) and I was able to do some simplifications in the definition because since we use Integers, there are no overflows.
- Did a couple of modifications to the compilation schemes of several `Atomic` events: moved a `isync` barrier to the correct place in `CAS`, added a couple of ctrl-dependencies for `AtomicLoad` (to match tables in papers) and used `lwsync` instead of `isync` in `AtomicFecthOp` following a real compiler (from godbolt.org) rather than a paper.
- Added some new lock benchmarks


ThomasHaas (approved):
Everything looks fine to me.

PR #130, opened 2021-09-30
CSV

hernanponcedeleon:
In this PR the solving times from the unit tests are saved into CSV files.
These files can be used to automatically and easily generate graphs to compare the different solvers.

I considered a couple of options not to always generate the CSV files in case that it degrades performance, but the overhead is almost nothing so I decided to always generate such files.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -13,6 +13,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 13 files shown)


ThomasHaas (approved):
I cleaned up the code for the path generation.
It might not be as copy&paste-friendly as before, but the previous way to get the class name was kinda ugly. 
If you agree with my changes, we can merge this.

PR #139, opened 2021-10-10
New model-checking-based Solver

ThomasHaas:
This PR aims to replace our current search-based `SaturationSolver` with a new model-checking based version that we tested on the `experimental` branch. The new method is a lot easier and seems to be faster in almost all cases. Furthermore, it has no incompleteness issues and it can handle litmus tests as well. 

Summary:
- Replaced the search-based refinement procedure with a model-checking-based refinement procedure (we still call it `SaturationSolver` but the name might be misleading now). Everything related to the previous search algorithm has been removed (with minor exceptions).
- Added new custom data structures `EdgeList` and `EdgeSet`, however, the latter is not used yet.
- Added generation of graphs for visualization of violations via Graphviz. Currently it is very rudimentary and hardcoded for a specific use-case but we can generalize the code in a separate PR if needed.
- Added new Refinement baseline Wmm options (ATOMIC_RMW)


hernanponcedeleon:
This PR has been superseded by #156

Issue #137, opened 2021-10-09
Abstraction/Usage of Alias Analysis


ThomasHaas:
Our current alias analysis is used in a very specific manner:
- It is created by the `Wmm` on initialization
- It computes address sets and attaches them to `MemEvents`

This usage pattern forces the alias analysis to be based on address sets. This is unnecessarily restrictive and forbids most advanced types of analysis.
Instead, we should treat the alias analysis just like we treat our `BranchEquivalence`. It gets instantiated once on/before initialization and is then accessible via the `VerificationTask` object. The alias analysis object should provide simple methods to check aliasing information: `mayAlias(e1, e2)`, `mustAlias(e1, e2)` and possibly more if needed.
During the encoding phase, all relevant objects have access to the `task` object, so we can easily use e.g. `task.getAliasAnalysis().mayAlias(e1, e2)` instead of relying on the fact that the `MemEvents` `e1` and `e2` themselves contain any aliasing information. This gives us the freedom to replace our current analysis by any other analysis in a modular manner.

hernanponcedeleon:
This issue is solved in `config` and can be closed once that PR is merged, right @ThomasHaas ?

ThomasHaas:
Yes, this is solved

PR #146, opened 2021-10-24
[Draft] Restructuring + Configurability

ThomasHaas:
This PR aims to solve issue #79. As a byproduct, the structure of program processing has been changed drastically. Further modularization of analysis, processing and encoding procedures has to be done.

Key changes are:
- `GlobalSettings` is now a singleton instance and its options can be loaded via a `Configuration`.
- A lot of the settings from `GlobalSettings` has been moved to the individual classes that define and need them. This was not always possible as the core structure of the code does not support this fully. In particular, static methods cannot have configurations injected which affects parsing as well as all verification methods.
- To allow more injection opportunities, the whole processing of `Program` (from dead code elimination over unrolling and compiling up to symmetry reduction) has been modularized into separate classes that are configurable.
- A lot of precondition checking has been replaced by calls to google's `Preconditions` class.
- Probably some more stuff.

Current problems / incomplete reworks:
- Loading of `Configurations` from a file or the command line has not been added yet, all code will use the default configs which essentially coincide with what we had set in `GlobalSettings` before.
- There is no standard config file right now. We would probably add a folder `config/` that contains at least a single `.properties` file with configurable options, but we might also want to modularize this into different config files.
- To allow better config injection, we need to encapsulate (static) parsing code and verification code into instantiable classes.
- Generally, algorithmic code and data objects need to get separated more. In particular, `Program` and `Wmm` should be pure data objects and e.g. the encoding generation should be moved elsewhere. This allows configurability of these algorithmic components.
- Temporarily, the compilation `target` and `settings` are still existent and the `VerificationTask` will merge these options into a provided config. Ideally, these options shall be directly integrated into the config. But to achieve this, more rework is needed.
- The naming and description of configuration options is very tentative.
- For now, all configurable options of any class also have regular accessors to read/modify the options programatically. In most cases this is not needed. We need to discuss what needs we actually have in this regard.


xeren:
This update targets the removal of `target` and `settings` in `VerificationTask`.  It removes the class `Settings` entirely, while adding its three main parameters to `BaseOptions` and `UiOptions` where further work is required.
- As a result of moving `Settings#alias`, `AliasAnalysis` was annotated with `Options` and its call was moved from `Wmm` to `VerificationTask`.  This addresses Issue #137 but does not resolve it yet.

hernanponcedeleon:
The branch has conflicts (`GlobalSettings.java`). @xeren can you please resolve them?
Also, it seems that all your commits are **unverified**. Do you have your GitHub account details either when you push from the console or if you use an IDE?

The latest changes this not trigger the CI. I'm not sure if it is because your commits are unverified, due to the conflict (I assume it is the latter), or because the PR is marked as a DRAFT

xeren:
I resolved the conflict by ignoring the change to the default for `GlobalSettings.FIXED_MEMORY_ENCODING` (now `MemoryEncoder.shouldUseFixedMemoryEncoding`).  This may have to be adjusted later.

hernanponcedeleon:
Can we now remove this PR or is there any code that might be useful as a guide for finishing the configuration migration?

ThomasHaas:
We can most likely remove this.

Issue #152, opened 2021-11-21
Atomicity encoding for atomic blocks is buggy


ThomasHaas:
@hernanponcedeleon and me discussed this problem before.
I'm opening this issue so we do not forget. I think one of the weaver benchmarks (test-easy8?) showed this behaviour.

xeren:
This might be caused by the relaxed memory order, if another consistency model than SC is used.  Locks should be at least Acquire-Release.  Can you describe the test that was used?  I never encountered any 'weaver' benchmark.

ThomasHaas:
We know why this is caused. It is simply that our atomicity encoding for atomic blocks (SVCOMP) is incorrect.
I don't have the exact reason in mind right now, but IIRC it had to do with cases where an ordering was not directly caused by a `com` relation between 2 atomic blocks but indirectly over a third thread (via com;po;com).

We have observed this problem in only one benchmark so far, so it is a rather rare problem.
If we replace the atomic blocks by locks, it works as expected.

xeren:
Would that benchmark work on this model?
```
let hb = (po \ rmw) | com
let atomic = rmw | rmw^-1
acyclic hb | (hb ; atomic) | (atomic ; hb)
empty rmw^-1 & com
```

ThomasHaas:
You can check out the benchmark here: [Link](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/weaver/test-easy8.wvr.c)

The benchmark is really short and can be understood by hand. The program is safe as `c=1` will always hold after both threads join (and hence the program aborts before reaching the error). If `thread2` runs last, it is obvious as it does `c=1`. If `thread1` runs last, then it will fail its assume since `thread2` will have set `b=1`.

xeren:
The above model had flaws, but the program was safe under it.  The PR features a modification that should fit the existing tests and `easy8`.

hernanponcedeleon:
The cat files makes sense to me and the CI did not complain. I'll make a run in the server using the whole set of svcomp benchmarks and if all pass, I'll merge

PR #153, opened 2021-11-26
altered svcomp.cat

xeren:
`rmw` is integrated into `hb`, effectively implying edges with all events in an atomic block, if at least one participates in communication.  As a result, the additional emptiness axiom becomes obsolete.
Should fix issue #152.


hernanponcedeleon:
The cat file seems correct, but it affects performance quite badly

Screenshot 2021-11-29 at 09 17 45

I discussed with @ThomasHaas and we think this cat file would also be correct

```
SVCOMP

(* Define co (and fr) *)
include "cos.cat"

(* All communication relations *)
let com = (rf | fr | co)

(* Sequential consistency *)
let hb = po | com | (rmw^-1;(com & ext))
acyclic hb as sc
```

I'm running the experiments on this version to see if the performance is better.

Alternative we could also add `| ((com & ext);rmw^-1)` to `hb`.

hernanponcedeleon:
Here are the results about performance

Screenshot 2021-11-30 at 09 44 10

2021-11-27: Rene's version
2021-11-28: SVCOMP 22 (buggy) version
2021-11-29: Version I commented above

I checked and all 80 benchmarks that SVCOMP can solve and the current cat file cannot are from the WMM folder.
I randomly run a couple of them and the end up being solved in ~3 minutes. Thus, we can expect that in the SVCOMP setting (i.e., 15 min instead of 2 min I used for this test), we should solve the same amount of benchmarks.

Issue #154, opened 2021-11-28
RelInverse does not respect min sets


ThomasHaas:
`RelInverse` does not properly respect the min sets and always does the same encoding which gives wrong results in certain cases.

PR #155, opened 2021-11-28
Fixed RelInverse to respect min sets (issue #154)

ThomasHaas:
This address issue #154


hernanponcedeleon:
This PR targets `master`, it should be `development`

PR #162, opened 2021-12-04
[Draft] Improved unit tests

ThomasHaas:
This PR aims to simplify the process of setting up unit tests by using the JUnit rule system.


ThomasHaas:
I have only added the new features in a test class callled `CLocksAlternative`. @hernanponcedeleon can you check if the changes are fine? If so, I will change all the unit tests to the new format.

hernanponcedeleon:
Overall it looks great (I just did a small style modification).
We might eventually want to do some modifications to generate one file per WMM as we did in one of the PLDI branches, but I think that should be easy.
This PR still improves the state of what we have in development so it is good.
@ThomasHaas if you implement the missing test classes, I'll merge.

ThomasHaas:
I created a new `AbstractCTest` class for C-based tests. But unlike the other abstract test classes, I did not put the actual test code there. I think that having all test methods in the base class can be problematic if one wants to run different test methods on different benchmark sets, e.g., `Refinement` on `CLocksLarge`/`SvCompLarge` and `Incremental` on `CLocks`/`SvCompConcurrency`.
For now, I kept the test methods in the Litmus test base class and the Svcomp base class, because on those it is very reasonable that we run a single technique over the whole set.
But for C benchmarks, it is more likely that we want to be able to pick the verification technique more precisely.
Moving the test methods into/from the base class is a non-issue though as they reduce to a single line of copyable code, so we can easily adapt to whatever we need.

hernanponcedeleon:
This branch now has conflicts due to merging `164` into `development`. @ThomasHaas  can you solve them please?

PR #166, opened 2021-12-07
Improved symmetry breaking

ThomasHaas:
This PR extends the functionatlity of symmetry breaking slightly. `GlobalSettings` has some new options to configure symmetry breaking.

Overall, I didn't see a performance improvement in breaking on e.g. `co` instead of `rf` or in changing the order of breaking variables (based on "synchronisation degree"). But this needs more testing.


hernanponcedeleon:
Effectively, the only change is the way of sorting the row, right? 
The rest of the changes seem to be just to have better control of the configuration, i.e. to be able to use a relation `!= rf` for the symmetry breaking.

ThomasHaas:
Exactly. It is just the sorting that is different.

PR #176, opened 2021-12-09
174

hernanponcedeleon:
This PR gives the user more control over `clang` (both directly when checking that the code compiles and when the flags are passed to `smack`).

Notice that we could do the same for the compilation in the `svcomp` module. However for the competition the flags are pretty much fixed (all tasks are preprocessed anyway), thus I chose not to change those.

The CI will trivially because this change only affects the behavior when `*.c` files are passed as input (the CI uses the `*.bpl` ones), so please test that it works from the console (you need to set `CFLAGS`, see the updated `README`).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
@@ -21,8 +21,8 @@ public class Compilation {
 		
     	ArrayList<String> cmd = new ArrayList<String>();
     	cmd.addAll(asList("smack", "-q", "-t", "--no-memory-splitting"));
-        cmd.add("--clang-options=-I" + System.getenv().get("DAT3M_HOME") + "/include/ " + clangFlags);
-    	cmd.addAll(asList("-bpl", System.getenv().get("DAT3M_HOME") + "/output/" + name + ".bpl"));
+        cmd.add("--clang-options=-I" + System.getenv("DAT3M_HOME") + "/include/ " + clangFlags);
+    	cmd.addAll(asList("-bpl", System.getenv("DAT3M_HOME") + "/output/" + name + ".bpl"));
     	cmd.add(file.getAbsolutePath());
     	
     	ProcessBuilder processBuilder = new ProcessBuilder(cmd); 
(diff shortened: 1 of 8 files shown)


ThomasHaas:
If we have a required flag, wouldn't it make sense to enable that one automatically and then add `CFLAGS`?
I think all necessary flags should be set by default.

hernanponcedeleon:
I was worried that doing so could overwrite the flags, e.g., `clang -I path1 -I path2`.
Apparently this is not a problem, so I'll fix the needed flags and put in the `README` that `CFLAGS` can be used to pass flags to the compiler.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
```diff
@@ -17,10 +17,11 @@
 	
 	public static void compileWithSmack(File file) throws Exception {
 		String name = file.getName().substring(0, file.getName().lastIndexOf('.'));
-
+		String clangFlags = System.getenv().get("CFLAGS") != null ? System.getenv().get("CFLAGS") : "";
```
ThomasHaas, line 20: I would use `System.getenv().getOrDefault(""))` instead.
Also, it seems that there is an overload of `.getenv()` that directly takes the key in case you do not do additional checks (e.g. for `DAT3M_HOME`)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
```diff
@@ -17,10 +17,11 @@
...
+		
     	ArrayList<String> cmd = new ArrayList<String>();
     	cmd.addAll(asList("smack", "-q", "-t", "--no-memory-splitting"));
-        cmd.add("--clang-options=-DCUSTOM_VERIFIER_ASSERT -fno-vectorize -fno-slp-vectorize -I" + System.getenv().get("DAT3M_HOME") + "/include/");    		    		
+        cmd.add("--clang-options=-I" + System.getenv().get("DAT3M_HOME") + "/include/ " + clangFlags);
```
ThomasHaas, line 24: Here you can use the short version `System.getenv("DAT3M_HOME)`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
```diff
@@ -43,8 +44,10 @@ public static void compileWithSmack(File file) throws Exception {
 	}	
 
 	public static void compileWithClang(File file) throws Exception {
+		String clangFlags = System.getenv().get("CFLAGS") != null ? System.getenv().get("CFLAGS") : "";
+		
     	ArrayList<String> cmd = new ArrayList<String>();
-    	cmd.addAll(asList("clang", "-S", "-o"));
+    	cmd.addAll(asList("clang", "-S", clangFlags, "-o"));
     	cmd.add(System.getenv().get("DAT3M_HOME") + "/output/test.s");
```
ThomasHaas, line 51: Same two changes as suggested above.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
```diff
@@ -55,7 +58,6 @@ public static void compileWithClang(File file) throws Exception {
     		String errorString = CharStreams.toString(new InputStreamReader(proc.getErrorStream(), Charsets.UTF_8));
 			throw new Exception(errorString);
     	}
-    	// TODO(HP): Can this be removed?
     	File testFile = new File(System.getenv().get("DAT3M_HOME") + "/output/test.s");
```
ThomasHaas, line 61: Short version of `getenv`

ThomasHaas (changes requested):

hernanponcedeleon:
I did not know there was a shorter version. I applied it consistently to the whole code base.

ThomasHaas (approved):
Looks good to me

PR #186, opened 2021-12-19
Coverage

hernanponcedeleon:
The purpose of this PR is to increase our unit test coverage.
In particular, by adding test that detect different exceptions due to errors in the parsing / validation of the programs.


dartagnan/src/test/java/com/dat3m/dartagnan/exceptions/RuntimeExceptionsTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void noModelBNonDet() throws Exception {
+	    Configuration config = Configuration.builder()
+	            .setOption("solver.z3.usePhantomReferences", "true")
+	            .build();
+	    try (SolverContext ctx = SolverContextFactory.createSolverContext(
+	            config,
+	            BasicLogManager.create(config),
+	            ShutdownManager.create().getNotifier(),
+	            org.sosy_lab.java_smt.SolverContextFactory.Solvers.Z3);
+	         ProverEnvironment prover = ctx.newProverEnvironment(ProverOptions.GENERATE_MODELS))
```
ThomasHaas, line 68: I think we have a `TestHelper` class that provides this functionality.

dartagnan/src/test/java/com/dat3m/dartagnan/exceptions/RuntimeExceptionsTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void noModelINonDet() throws Exception {
+	    Configuration config = Configuration.builder()
+	            .setOption("solver.z3.usePhantomReferences", "true")
+	            .build();
+	    try (SolverContext ctx = SolverContextFactory.createSolverContext(
+	            config,
+	            BasicLogManager.create(config),
+	            ShutdownManager.create().getNotifier(),
+	            org.sosy_lab.java_smt.SolverContextFactory.Solvers.Z3);
+	         ProverEnvironment prover = ctx.newProverEnvironment(ProverOptions.GENERATE_MODELS))
```
ThomasHaas, line 85: same as above

ThomasHaas (commented):
I like this :). I think this will be very helpful in the future when we finally get around to do changes to the frontend.

dartagnan/src/main/java/com/dat3m/dartagnan/asserts/AssertBasic.java
```diff
@@ -34,7 +36,12 @@ public AbstractAssert removeLocAssertions(boolean replaceByTrue) {
 
     @Override
     public BooleanFormula encode(SolverContext ctx) {
-        return op.encode(e1.getLastValueExpr(ctx), e2.getLastValueExpr(ctx), ctx);
+    	Formula f1 = e1 instanceof LastValueInterface ? ((LastValueInterface)e1).getLastValueExpr(ctx) : null;
+    	Formula f2 = e2 instanceof LastValueInterface ? ((LastValueInterface)e2).getLastValueExpr(ctx) : null;
+    	if(f1 == null || f2 == null) {
+    		throw new RuntimeException("AssertBasic can only relate LastValueInterface objects: " + this);
+    	}
+        return op.encode(f1, f2, ctx);
```
ThomasHaas, line 44: I think you could do the check for `LastValueInterface` directly in the constructor.
Also, you can use `Preconditions.checkState` (if you check the state of the current object like in this case) or `Preconditions.checkArgument` if you check the arguments provided (if you move the code to the Constructor).

dartagnan/src/test/java/com/dat3m/dartagnan/exceptions/RuntimeExceptionsTest.java
```diff
@@ -0,0 +1,75 @@
...
+import com.dat3m.dartagnan.parsers.program.utils.ProgramBuilder;
+import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.event.CondJump;
+import com.dat3m.dartagnan.program.event.Skip;
+import org.junit.Test;
+import org.sosy_lab.java_smt.api.ProverEnvironment;
+import org.sosy_lab.java_smt.api.SolverContext;
+import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
+
+public class RuntimeExceptionsTest {
```
ThomasHaas, line 21: In this whole class I would double check the type of exception you expect. We have the tendency to throw a lot of `RuntimeException` rather than a more specific type of error. I think we should be more concrete about errors and add a lot more custom exception classes for this.

dartagnan/src/test/java/com/dat3m/dartagnan/witness/BuildWitnessTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class BuildWitnessTest {
+
+    @Test
+    public void BuildWriteEncode() throws Exception {
+    	Program p = new ProgramParser().parse(new File(ResourceHelper.TEST_RESOURCE_PATH + "witness/lazy01-O0.bpl"));
+    	Wmm wmm = new ParserCat().parse(new File(ResourceHelper.CAT_RESOURCE_PATH + "cat/svcomp.cat"));
+    	VerificationTask task = new VerificationTask(p, wmm, Arch.NONE, new Settings(0, 0));
+    	
+    	SolverContext ctx = TestHelper.createContext();
```
ThomasHaas, line 32: You should wrap this into a try-with-resources block (probably irrelevant for tests but still).

dartagnan/src/test/java/com/dat3m/dartagnan/exceptions/RuntimeExceptionsTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test(expected = RuntimeException.class)
+    public void diffPrecisionIfExpr() throws Exception {
+    	IfExpr ifE = new IfExpr(BConst.TRUE, new Register("a", 0, 32), new Register("b", 0, 64));
+    	ifE.getPrecision();
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void noModelBNonDet() throws Exception {
+	    try (SolverContext ctx = com.dat3m.dartagnan.utils.TestHelper.createContext();
```
ThomasHaas, line 50: Add static import for the `TestHelper.createContext` method

dartagnan/src/test/java/com/dat3m/dartagnan/exceptions/RuntimeExceptionsTest.java
```diff
@@ -0,0 +1,75 @@
...
+	    		ProverEnvironment prover = ctx.newProverEnvironment(ProverOptions.GENERATE_MODELS))
+	    {
+	    	BNonDet nonDet = new BNonDet(32);
+	    	nonDet.getBoolValue(null, prover.getModel(), ctx);
+	    }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void noModelINonDet() throws Exception {
+	    try (SolverContext ctx = com.dat3m.dartagnan.utils.TestHelper.createContext();
```
ThomasHaas, line 60: Same as above.

dartagnan/src/test/resources/exceptions/IlegalJump.litmus
```diff
@@ -0,0 +1,22 @@
+PPC DETOUR0285
```
ThomasHaas, line 1: The filename should be "IllegalJump.litmus" (just a typo!)

dartagnan/src/test/resources/exceptions/SeveralAssignements.bpl
```diff
@@ -0,0 +1,12 @@
+// generated by SMACK version 2.4.1 for corral
```
ThomasHaas, line 1: Typo in the filename "Assign_e_ments"

dartagnan/src/test/java/com/dat3m/dartagnan/exceptions/RuntimeExceptionsTest.java
hernanponcedeleon: I implemented two new types of exceptions for the cases in this test: `TypeException` for problems related to precision (i.e., type) and `BuildException` for problems related to building or validating the program.
But I agree that we might want to implement similar changes to the overall code.
I will keep this in mind while I add new unit tests to this PR.

ThomasHaas: I would make the exception names more informative and precise: `TypeMismatchException` (maybe even `ExprTypeMismatchException`) and `InvalidProgramException` (or `MalformedProgramException)`.
The name `BuildException` only makes sense because we have a class `ProgramBuilder` but if you renamed that class, the exception name would become odd. This indicates that the exception name is not informative about the type of error that occurred but only about where the error occurred, which I think is wrong design.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/exception/UninitialisedVariableException.java
```diff
@@ -0,0 +1,8 @@
+package com.dat3m.dartagnan.parsers.program.exception;
+
+public class UninitialisedVariableException extends RuntimeException {
```
ThomasHaas, line 3: You might wanna extend from `MalformedProgramException`. I would say an uninitialized variable is also a case of wrongly formed program.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/exception/UnrollException.java
```diff
@@ -0,0 +1,8 @@
+package com.dat3m.dartagnan.parsers.program.exception;
+
+public class UnrollException extends IllegalStateException {
```
ThomasHaas, line 3: I think this might be superfluous. I haven't checked all usages of it though but I think it might be too specific again (`IllegalStateException` might be sufficient on its own).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -156,7 +158,7 @@ public Location getOrErrorLocation(String name){
         if(locations.containsKey(name)){
             return locations.get(name);
         }
-        throw new ParsingException("Location " + name + " has not been initialised");
+        throw new UninitialisedVariableException("Location " + name + " has not been initialised");
```
ThomasHaas, line 161: Seeing this usage, I think `UninitialisedVariableException` should not have inherited `MalformedProgramException` unless these methods are **only** used during program construction.
Even then, `IllegalStateException` seems to be the most fitting exception.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -183,7 +185,7 @@ public Register getOrErrorRegister(int thread, String name){
                 return register;
             }
         }
-        throw new ParsingException("Register " + thread + ":" + name + " is not initialised");
+        throw new UninitialisedVariableException("Register " + thread + ":" + name + " is not initialised");
```
ThomasHaas, line 188: `IllegalStateException`

dartagnan/src/main/java/com/dat3m/dartagnan/program/Thread.java
```diff
@@ -78,7 +79,7 @@ public Register getRegister(String name){
 
     public Register addRegister(String name, int precision){
         if(registers.containsKey(name)){
-            throw new RuntimeException("Register " + id + ":" + name + " already exists");
+            throw new AlreadyExistentVariableException("Register " + id + ":" + name + " already exists");
```
ThomasHaas, line 82: I would probably go for either `IllegalStateException` or a custom exception that is a little more general, something like `DuplicateException` or `AlreadyPresentException`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/arch/linux/event/cond/FenceCond.java
```diff
@@ -48,6 +49,6 @@ protected BooleanFormula encodeExec(SolverContext ctx){
 
     @Override
     protected RecursiveAction unrollRecursive(int bound, Event predecessor, int depth) {
-        throw new RuntimeException("FenceCond cannot be unrolled: event must be generated during compilation");
+        throw new UnrollException("FenceCond cannot be unrolled: event must be generated during compilation");
```
ThomasHaas, line 52: `UnsupportedOperationException` should do the trick. Possibly a custom `ProgramProcessingError`.

dartagnan/src/test/java/com/dat3m/dartagnan/DataRaceTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    protected Provider<Settings> getSettingsProvider() {
+        return Provider.fromSupplier(() -> new Settings(bound, 0));
+    }
+
+    protected Provider<Wmm> getWmmProvider() {
+        return GlobalSettings.ATOMIC_AS_LOCK ?
+                Providers.createWmmFromName(() -> "svcomp-locks") :
+                Providers.createWmmFromName(() -> "svcomp");
+    }
```
ThomasHaas, line 62: These methods can get "inlined" unless you intend to create subclasses of this test class to override them (that is the reason by these methods exist in the abstract base classes). But even without inheritence in mind, it is not bad to have these methods. So it is up to you if you want to keep them.

dartagnan/src/test/java/com/dat3m/dartagnan/DataRaceTest.java
```diff
@@ -0,0 +1,131 @@
...
+                continue;
+            }
+            return br.readLine().contains("false") ? FAIL : PASS;
+
+        } catch (Exception e) {
+            System.out.println(e.getMessage());
+            System.exit(0);
+        }
+        return null;
+    }
```
ThomasHaas, line 113: It might be better to move this method (and make it static) to `ResourceHelper`. At least I can imagine that all these `readExpected` methods should be reusable for various use-cases.

ThomasHaas (changes requested):
I have added a bunch of feedback but I did not yet review all the new test classes completely.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/LastValueInterface.java
```diff
@@ -0,0 +1,10 @@
+package com.dat3m.dartagnan.expression;
+
+import org.sosy_lab.java_smt.api.Formula;
+import org.sosy_lab.java_smt.api.SolverContext;
+
+public interface LastValueInterface extends ExprInterface {
```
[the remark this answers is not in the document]
hernanponcedeleon: It will not work but not due to this interface, but because of the parser. The grammar only accepts variables (`Location` or `Register`) and constants. Thus this interface is in sync with the parser.
`herd7` does not accept either the test you proposed so I don't really see the point in having support for this.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/Redusable.java
```diff
@@ -0,0 +1,7 @@
+package com.dat3m.dartagnan.expression;
+
+public interface Redusable extends ExprInterface {
+
+	IConst reduce();
+
```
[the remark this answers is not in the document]
hernanponcedeleon: As part of this PR I clean up a little bit the code of `reduce()`. It used to be implemented for all expressions, but only ever used for `IExpr`.
I think we already discussed in the past about the difference between `reduce()` and `simplify()`. Reduce is just used to allocate memory, parse axioms that define the value of constants create the value of `Init` events.
There are no exceptions now: the method is only called for classes that implement the `Reducible` interface.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/exception/ExprTypeMismatchException.java
```diff
@@ -0,0 +1,8 @@
+package com.dat3m.dartagnan.parsers.program.exception;
+
+public class ExprTypeMismatchException extends RuntimeException {
```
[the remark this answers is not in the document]
hernanponcedeleon: I think this error type goes beyond encoding. I consider `ExprTypeMismatchException` as a `MalformedProgramExcpetion `

dartagnan/src/main/java/com/dat3m/dartagnan/program/Thread.java
hernanponcedeleon: If this call gets in, it would result in a malformed program, thus I'll go with `MalformedProgramException`

dartagnan/src/main/java/com/dat3m/dartagnan/expression/LastValueInterface.java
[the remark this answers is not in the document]
ThomasHaas: In general it seems like an artificial restriction though, but it seems unlikely that we ever need it. So I'm fine with this.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/Redusable.java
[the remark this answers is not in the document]
ThomasHaas: It is fine for now. But if we ever get around to do proper simplification, `reduce` seems to become obsolete.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/exception/ExprTypeMismatchException.java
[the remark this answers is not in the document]
ThomasHaas: I agree that it may be seen as a `MalformedProgramException` if the types of the program are not cast properly. But I'm wondering when this type check happens? If this exception is only thrown during program construction, then it can be a `MalformedProgramException`, if it is only ever thrown during encoding, it should be a `EncodingException`, if it can be thrown in both cases, it can stay as it is (i.e. inherit from `RuntimeException`).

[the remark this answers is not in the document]
hernanponcedeleon: This is now checked using `Preconditions.check()` in the constructor of `IExprBin`

ThomasHaas (changes requested):
I think the whole `boogie`-folder is related to parsing so it should be moved to the `parsing`-folder. Also, the `ParserErrorListener` should throw a `ParsingException`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/Reducible.java
```diff
@@ -0,0 +1,7 @@
+package com.dat3m.dartagnan.expression;
+
+public interface Reducible extends ExprInterface {
```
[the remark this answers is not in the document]
hernanponcedeleon: With the current code, it does not ... the question is: do we want to keep them separated and say `IExpr` implements both `ExprInterface` and `Reducible` or just say that `IExpr` implements `Reducible` (and by extension `ExprInterface`).

While from a code perspective I'm inclined to keep them separated (i.e. removed the `extends ExprInterface`), I don't see a use for `Reducible` other than `IExp` thus using the `extends` seems natural. But I'm fine with any of them. 

What do you think?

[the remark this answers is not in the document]
ThomasHaas: I think we can keep it as it is for now. It will get reworked properly once we redesign all the expression code.

dartagnan/src/test/java/com/dat3m/dartagnan/witness/BuildWitnessTest.java
```diff
@@ -0,0 +1,65 @@
...
+    	    sOptions[5] = "1";
+    	    sOptions[6] = "-create_witness";
+    	    sOptions[7] = ResourceHelper.TEST_RESOURCE_PATH + "witness/lazy01-O0.bpl";
+    	    DartagnanOptions option = new DartagnanOptions();
+    	    option.parse(sOptions);
+    	    
+    		WitnessGraph graph = new WitnessBuilder(p, ctx, prover, res).buildGraph(option);
+    		// Write to file
+    		graph.write();
+    		assert(new File(System.getenv("DAT3M_HOME") + "/output/witness.graphml").exists());
```
ThomasHaas, line 56: I think the file never gets deleted so this test might be meaningless. Maybe we can leave this part out for now.

ThomasHaas (changes requested):
Apart from a last minor issue, I think this code looks good.

dartagnan/src/test/java/com/dat3m/dartagnan/witness/BuildWitnessTest.java
hernanponcedeleon: Keep in mind that each instance of the CI runs in a different "container" (not sure if this is the correct term) meaning that the file from the previous run of the CI does not exists and thus the test still checks that the file was correctly created in this run.

ThomasHaas: Ok, for the CI it makes sense then. It won't work locally though. Maybe add a comment that this works for the CI only?

hernanponcedeleon: I can also delete the file create by this tests ... this will not guarantee 100% that the test is not trivial locally because you can still have a witness file even if the test failed because e.g. you used the svcomp runner. Another way to reduce the chances of this happening is to use the program name for the witness file instead of just `witness.graphml` (I need to double check this would not cause problems for svcomp but I don't think so).

ThomasHaas: Different names is probably the best idea. But even if you just delete the file at the end (which you probably should either way), the test would only be trivial for the first run since it will delete whatever file is there, making sure that the second run is an actual test (unless you create a new file inbetween each two tests, but that seems unlikely).

hernanponcedeleon: The new version even checks that the file does not exist before writing the witness. I also modified the bpl filename to avoid the situation where the user has a witness file due to using the svcomp runner with the same task we use for the unit test.

ThomasHaas (approved):
Looks good to me.

Issue #145, opened 2021-10-20
Logger on/off in the CI


hernanponcedeleon:
We want to keep the log on in the development branch but we want to have it off when we do a PR (otherwise the console in the CI is impossible to read). There might be a way to keep it on while automatically disabling it when running the tests, see [this](https://stackoverflow.com/questions/13935715/disabling-log4j-logs-during-maven-test-phase).

PR #191, opened 2021-12-21
Keep log from UI and console, but silent it when running tests

hernanponcedeleon:
The log is `off` when running tests (both from the console or IDE), but it is `on` when running the tool from the UI (this might be useful for "debugging") or the console (this is specially useful for SVCOMP).


hernanponcedeleon:
Closes #145

PR #193, opened 2021-12-30
Consistent use of Preconditions.checkX

hernanponcedeleon:
Now most of the checks are performed using `Preconditions.checkX`.
This will probably also increase the coverage (at least if branches).


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -103,12 +108,16 @@ public static void main(String[] args) throws Exception {
                     options.getSMTSolver());
                  ProverEnvironment prover = ctx.newProverEnvironment(ProverOptions.GENERATE_MODELS))
             {
-                Result result;
+                Result result = Result.UNKNOWN;
+            	Preconditions.checkArgument(Arrays.asList(Analysis.values()).contains(options.getAnalysis()),
```
ThomasHaas, line 112: This is just a complex way to check for `NULL` (a value of enum type is always contained in `Enum.values` unless it is null).

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -103,12 +108,16 @@ public static void main(String[] args) throws Exception {
...
-                Result result;
+                Result result = Result.UNKNOWN;
+            	Preconditions.checkArgument(Arrays.asList(Analysis.values()).contains(options.getAnalysis()), 
+            			"Unrecognized analysis: " + options.getAnalysis());
                 switch (options.getAnalysis()) {
                     case RACES:
                         result = checkForRaces(ctx, task);
                         break;
                     case REACHABILITY:
+                    	Preconditions.checkArgument(Arrays.asList(Method.values()).contains(options.getMethod()),
```
ThomasHaas, line 119: same as above

dartagnan/src/main/java/com/dat3m/dartagnan/asserts/AssertNot.java
```diff
@@ -25,10 +26,8 @@ AbstractAssert getChild(){
 
     @Override
     public BooleanFormula encode(SolverContext ctx) {
-        if(child != null){
-            return ctx.getFormulaManager().getBooleanFormulaManager().not(child.encode(ctx));
-        }
-        throw new RuntimeException("Empty assertion clause in " + this.getClass().getName());
+    	Preconditions.checkNotNull(child, "Empty assertion clause in " + this.getClass().getName());
```
ThomasHaas, line 29: Don't use `checkNotNull` here. It should be done in the constructor of `AssertNot`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/BNonDet.java
```diff
@@ -42,10 +43,8 @@ public BooleanFormula toBoolFormula(Event e, SolverContext ctx) {
 	@Override
 	public boolean getBoolValue(Event e, Model model, SolverContext ctx) {
 		Boolean value = model.evaluate(toBoolFormula(e, ctx));
-		if(value != null) {
-			return value;
-		}
-		throw new RuntimeException("No value in the model for " + this);
+		Preconditions.checkNotNull(value, "No value in the model for " + this);
```
ThomasHaas, line 46: `InvalidArgumentException` is better here (although not ideal). So use `Preconditions.checkArgument(value != null, ...)` instead.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
```diff
@@ -54,10 +55,8 @@ public Formula toIntFormula(Event e, SolverContext ctx) {
 	@Override
 	public BigInteger getIntValue(Event e, Model model, SolverContext ctx) {
 		Object value = model.evaluate(toIntFormula(e, ctx));
-		if(value != null) {
-			return new BigInteger(value.toString());			
-		}
-        throw new RuntimeException("No value in the model for " + this);
+		Preconditions.checkNotNull(value, "No value in the model for " + this);
```
ThomasHaas, line 58: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
```diff
@@ -26,12 +26,8 @@
...
 
     public CondJump(BExpr expr, Label label){
-        if(label == null){
-            throw new IllegalArgumentException("CondJump event requires non null label event");
-        }
-        if(expr == null){
-            throw new IllegalArgumentException("CondJump event requires non null expression");
-        }
+    	Preconditions.checkArgument(label != null, "CondJump event requires non null label event");
+    	Preconditions.checkArgument(expr != null, "CondJump event requires non null expression");
```
ThomasHaas, line 30: `Preconditions.checkNotNull`

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/CondJump.java
```diff
@@ -168,9 +164,7 @@ public CondJump getCopy(){
 
     @Override
     protected RecursiveFunction<Integer> compileRecursive(Arch target, int nextId, Event predecessor, int depth) {
-        if(successor == null){
-            throw new MalformedProgramException("Malformed CondJump event");
-        }
+    	Preconditions.checkNotNull(successor, "Malformed CondJump event");
```
ThomasHaas, line 167: Either use `MalformedProgramException` or use `Preconditions.checkState`.
`Preconditions.checkNotNull` and `Preconditions.checkArgument` are both intended to validate input parameters only.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/MemEvent.java
```diff
@@ -42,24 +43,18 @@ public void initialise(VerificationTask task, SolverContext ctx) {
     }
 
     public Formula getMemAddressExpr(){
-        if(memAddressExpr != null){
-            return memAddressExpr;
-        }
-        throw new IllegalStateException("Attempt to access not initialised address expression in " + this);
+    	Preconditions.checkNotNull(memAddressExpr);
```
ThomasHaas, line 46: `Preconditions.checkState` (as the method has no input)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/MemEvent.java
```diff
@@ -42,24 +43,18 @@ public void initialise(VerificationTask task, SolverContext ctx) {
...
+    	return memAddressExpr;
     }
 
     public Formula getMemValueExpr(){
-        if(memValueExpr != null){
-            return memValueExpr;
-        }
-        throw new IllegalStateException("Attempt to access not initialised value expression in " + this);
+    	Preconditions.checkNotNull(memValueExpr);
```
ThomasHaas, line 51: `Preconditions.checkState` (as the method has no input)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/MemEvent.java
```diff
@@ -42,24 +43,18 @@ public void initialise(VerificationTask task, SolverContext ctx) {
...
+    	return memValueExpr;
     }
 
     public ImmutableSet<Address> getMaxAddressSet(){
-        if(maxAddressSet != null){
-            return maxAddressSet;
-        }
-        throw new IllegalStateException("Location set has not been initialised for memory event " + this);
+    	Preconditions.checkNotNull(maxAddressSet);
```
ThomasHaas, line 56: `Preconditions.checkState` (as the method has no input)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/pthread/Create.java
```diff
@@ -53,6 +55,8 @@ public Create getCopy(){
 
     @Override
     protected RecursiveFunction<Integer> compileRecursive(Arch target, int nextId, Event predecessor, int depth) {
+    	Preconditions.checkArgument(Arrays.asList(Arch.values()).contains(target), "Compilation to " + target + " is not supported for " + this);
```
ThomasHaas, line 58: Again, this is just a complex null check. Use `Preconditions.checkNotNull` or `Precondtions.checkArgument(target != null)`.
I think I would prefer the latter in this case.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/pthread/End.java
```diff
@@ -44,6 +46,8 @@ public End getCopy(){
 
     @Override
     protected RecursiveFunction<Integer> compileRecursive(Arch target, int nextId, Event predecessor, int depth) {
+    	Preconditions.checkArgument(Arrays.asList(Arch.values()).contains(target), "Compilation to " + target + " is not supported for " + this);
```
ThomasHaas, line 49: same as above

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/pthread/Join.java
```diff
@@ -71,6 +73,8 @@ public Join getCopy(){
 
     @Override
     protected RecursiveFunction<Integer> compileRecursive(Arch target, int nextId, Event predecessor, int depth) {
+    	Preconditions.checkArgument(Arrays.asList(Arch.values()).contains(target), "Compilation to " + target + " is not supported for " + this);
```
ThomasHaas, line 76: same

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/pthread/Start.java
```diff
@@ -67,12 +69,13 @@ public Start getCopy(){
 
     @Override
     protected RecursiveFunction<Integer> compileRecursive(Arch target, int nextId, Event predecessor, int depth) {
+    	Preconditions.checkArgument(Arrays.asList(Arch.values()).contains(target), "Compilation to " + target + " is not supported for " + this);
```
ThomasHaas, line 72: same

dartagnan/src/main/java/com/dat3m/dartagnan/utils/equivalence/BranchEquivalence.java
```diff
@@ -106,9 +107,7 @@ public Class getEquivalenceClass(Event x) {
 
     public BranchEquivalence(Program program) {
         this.program = program;
-        if (!program.isCompiled())
-            throw new IllegalArgumentException("The program needs to be compiled first.");
-
+        Preconditions.checkState(program.isCompiled(), "The program needs to be compiled first.");
```
ThomasHaas, line 110: I would go with `Preconditions.checkArgument` as this is input validation.

ThomasHaas (changes requested):

ThomasHaas:
I'm not sure that replacing everything with `Preconditions` is the correct way.
Preconditions should only be used to check that the caller does not mess up anything:
- `checkNotNull` = Caller passed null value.
- `checkArgument` = Caller passed unacceptable values.
- `checkState` = Caller called method in wrong state / at wrong time.

Whenever some exception is unrelated to the caller, then none of the `Preconditions` are appropriate.
In these contexts, one either throws a custom exception or uses `Verify.verify` from `guava`.
For something like invariants (which should never be violated), one uses `assert`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/pthread/Create.java
```diff
@@ -53,6 +54,8 @@ public Create getCopy(){
 
     @Override
     protected RecursiveFunction<Integer> compileRecursive(Arch target, int nextId, Event predecessor, int depth) {
+    	Preconditions.checkArgument(target != null, "Target cannot be null");
```
ThomasHaas, line 57: Since you changed the error message, a `checkNotNull` is more appropriate. When I suggested `checkArgument` it was with respect to the old message.
I think the cleanest way is to always do `checkNotNull` and if some event can only be compiled to specific targets, there should additionally be a `checkArgument`.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
```diff
@@ -109,26 +110,32 @@ public void parse(String[] args) throws ParseException, RuntimeException {
     }
 
     public Method getMethod(){
+    	Preconditions.checkNotNull(method, "Method cannot be null");
```
ThomasHaas, line 113: These checks are not appropriate in a getter. Do these checks during construction if nothing is allowed to be null.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
hernanponcedeleon: Those fields are not set during the constructor but actually when `parse()` is called.
This check was actually to be sure that the getter is called after `parse()` so I change the precondition to `check state()` to reflect this.

dartagnan/src/main/java/com/dat3m/dartagnan/program/arch/tso/event/Xchg.java
```diff
@@ -65,23 +66,21 @@ public Xchg getCopy(){
...
-
-            List<Event> events = eventSequence(
-                    load,
-                    store,
-                    updateReg
-            );
-            return compileSequenceRecursive(target, nextId, predecessor, events, depth + 1);
-        }
-        throw new UnsupportedOperationException("Compilation of xchg is not implemented for " + target);
+    	Preconditions.checkArgument(target == Arch.TSO, this + " can only be compiled to " + Arch.TSO);
```
ThomasHaas, line 69: Same as above.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
ThomasHaas: There are two generals rules we should try to satisfy at all times:
- An object always has valid state (simple accesses to query its state via getters should not cause problems). If getters throw exceptions, a caller cannot even inspect the objects state without causing problems. That is a big no-no.
- Validation should happen as early as possibly (ideally during construction, if possible)

In this case, the `XOptions` classes are always used by first calling the constructor and then immediately calling its parse method. Before calling `parse`, the object is basically in an invalid state, therefore`parse` should be part of the construction. Simply make the constructor and the `parse` method protected and then offer a static factory method `fromArgs` to construct a fully valid object. Then you do not need to do any checks inside the getters and the callers code simplifies.

svcomp/src/main/java/com/dat3m/svcomp/options/SVCOMPOptions.java
```diff
@@ -75,14 +73,27 @@ public SVCOMPOptions(){
...
+        try {
+            options.parse(args);
         }
-
+        catch (Exception e){
+            if(e instanceof UnsupportedOperationException){
+                System.out.println(e.getMessage());
+            }
+            new HelpFormatter().printHelp("SVCOMP Runner", options);
+            System.exit(1);
```
ThomasHaas, line 88: I don't like that this method terminates the whole program on failure. I would rather let the caller handle the failure and terminate if necessary. I think forced termination is something that should happen as close to the entry point of the program as possible.

I think it should be like this: 
- If the class is completely public, you should make as few assumptions about the calling environment as possible. In this case, printing help messages and terminating is something that not every caller would expect. For example, it is reasonable to expect a caller to parse something and on failure to fall back onto e.g. standard options. This is something that the current design forbids.
- If you expect a certain calling context, then the visibility of this class or at least its constructing methods should be as limited as possible. This means you should either have them package private w.r.t. to the caller or even make this class an inner class of the caller if there is only one such caller (e.g. if only `Dartagnan` uses `DartagnanOptions` and only `SvcompRunner` uses `SvcompIOptions`). Then you can incorporate all this special behaviour on failure as you like.
If this scoping is not possible for some reason, then we should at least document the expected callers via comments.

I think you did some of this parsing in `ApplicationTests` as well and I believe you would not want to have printed messages or forced termination in those cases. I think `System.exit()` tries to terminate the whole JVM, so you wouldn't be able to add unit tests for wrong parsing (it would terminate the whole test suit I believe).

ThomasHaas (changes requested):

svcomp/src/main/java/com/dat3m/svcomp/options/SVCOMPOptions.java
hernanponcedeleon: I think the best would be letting `parse()` throw an exception and catch it in the `Dartagnan` main class. Then the catch can print the help.

ThomasHaas: The first option is the cleanest one as it has the least assumptions.

hernanponcedeleon: Printing the help is still done in the `parse()` for two reasons:
- `HelpFormatter().printHelp()` needs access to the `XOption` object
- I think getting the options help is always useful when there was a parsing problem.

Overall, the handling of exceptions was a mess from the application level. I tried to clean it up. Overall, now none of the `main` methods throws exceptions anymore and always ends "graciously" (maybe with exit code 1). Different exceptions are now catch and a corresponding msg is displayed. When the problem comes from the verification, the msg is logged and either `TIMEOUT` or `ERROR` is printed; if the problem was parsing an options then the helper is printed; when the problem was due to parsing the program or the CAT model (missing file, parsing problem, compilation problem), the reason is printed.

ThomasHaas (commented):
This looks much better now. I haven't gone through the full code yet to check if everything was replaced properly, but at least the current changes are fine. I'll approve once I checked the whole code base.

ThomasHaas (approved):
I think this PR looks good enough to merge.

PR #195, opened 2022-01-04
Config

hernanponcedeleon:
This PR incorporates some of the changes present in #146 (it was easier to re-implement some of the changes rather than solving the conflicts in that branch). The main changes are the following
- Use of `ProgramProcessor` (unrolling, compilation, simplifier, etc)
- Use of `Encoder` (not yet used for `Wmm` since this requires some more work due to the presence of the `baselinemodel` in `RefinementVerificationTask`)
- Use of configurations (I tested this from the console and works correctly):
  - The name of all options are in a single class `OptionNames` to keep track of all the options rather than having this split over the whole code (the options are split, the options names are not). 
  - Not all `GlobalSettings` have yet been replaced by `Options`
  - Some option `--help` is still missing. I tried using `collectOptions` as it was in #146 but this is not working and it requires further investigation.
- Simplification and removal of obsolete code (mostly related to `ExprInterface`)
- `Base` class was split into 3 different classes (one for each method)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -0,0 +1,73 @@
...
+        try {
+            DeadCodeElimination.fromConfig(config).run(program);
+            BranchReordering.fromConfig(config).run(program);
+            Simplifier.fromConfig(config).run(program);
+            LoopUnrolling.fromConfig(config).run(program);
+            if(constantPropagation) {
+              ConstantPropagation.fromConfig(config).run(program);            	
+            }
+            Compilation.fromConfig(config).run(program);
+            if(atomicBlocksAsLocks) {
```
hernanponcedeleon, line 60: I think I'm moved this at some point. If the order of a pass is important, we should add some precondition to it (as we do for e.g. compilation or constant propagation)

dartagnan/src/main/java/com/dat3m/dartagnan/verification/RefinementTask.java
```diff
@@ -64,45 +101,39 @@ public BooleanFormula encodeBaselineWmmConsistency(SolverContext ctx) {
...
-                task.getSettings()
-        );
+        return new RefinementTaskBuilder()
+                .withWitness(task.getWitness())
+                .withConfig(task.getConfig())
+                .build(task.getProgram(),task.getMemoryModel());
     }
 
-    private static Wmm createDefaultWmm() {
+    //TODO: This code is outdated and was replaced in the CAAT branch.
```
hernanponcedeleon, line 115: Can we remove then the method?

ThomasHaas: No, it needs to get updated with the code that sits in development.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
ThomasHaas: The Precondition checks a processing pass should only be related to the processing step itself, that is, it should only check for properties that have to hold for the pass to work. If there is some other reason that forces us to order processing steps (i.e. the steps are interchangable in principle but the combined result is not what we want) then these checks should be inside the `ProcessingManager` either via `assert` or `Verify.verify(...)`.

As an example: `ConstantPropagation` assumes no loops, so it can have a Precondition that checks that the program is unrolled. But whether the program is compiled or not does not matter much, it should work in both cases. However, we have reasons of why we do not want to do it after compilation, so the `ProcessingManager` should have an assertion that checks that the program is not compiled before calling the `ConstantPropagation` code.
In the case of `AtomicBlocksAsLocks`, the code comments say that it can be used anytime in principal, so it should not have any Precondition, yet we still might want to run it before some other passes.

ThomasHaas: Of course, if you rely on the fact that no exclusive events are present, then it should be a precondition on the pass.
But `AtomicBlocksAsLocks` does not have this assumption, so we shouldn't add any preconditions to it.
And in fact, it might be correct to execute at any time we want. The comment I added was just because we used to do this as a first step back then and that has changed. It doesn't mean its wrong though.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -0,0 +1,75 @@
...
+                Simplifier.fromConfig(config),
+                LoopUnrolling.fromConfig(config)
+        ));
+        if(constantPropagation) {
+            programProcessors.add(ConstantPropagation.fromConfig(config));
+        }
+        programProcessors.add(Compilation.fromConfig(config));
+        if(atomicBlocksAsLocks) {
+            // TODO: Do we really want to execute this after compilation?
+            programProcessors.add(AtomicAsLock.fromConfig(config));
```
xeren: I propose moving `AtomicAsLock` before `Compilation`, as it generates events that should get individual `cId`s.

xeren:
`sosy_lab.common.configuration.OptionCollector` relies on source code content to fetch default values, since Java does not provide any reflection on this.  To find source file at runtime, it assumes one of those project structures: class files are in `build/classes/main` or `bin`, source files are in `src/main/java` or `src`, or as a fallback, class and source files share directories.  This project has a default maven structure and stores class files in `target/classes` and source files in `src/main/java`.  This results in all options declared as having no default value, and a warning, even in execution outside of a jar.
We could do one of the following:
- implement an own `OptionCollector`
- restructure the project to fit into their pattern
- file an issue at `sosy-lab`
- ignore

hernanponcedeleon:
> `sosy_lab.common.configuration.OptionCollector` relies on source code content to fetch default values, since Java does not provide any reflection on this. To find source file at runtime, it assumes one of those project structures: class files are in `build/classes/main` or `bin`, source files are in `src/main/java` or `src`, or as a fallback, class and source files share directories. This project has a default maven structure and stores class files in `target/classes` and source files in `src/main/java`. This results in all options declared as having no default value, and a warning, even in execution outside of a jar. We could do one of the following:
> 
> * implement an own `OptionCollector`
> * restructure the project to fit into their pattern
> * file an issue at `sosy-lab`
> * ignore

Can we just implement our own `OptionCollector` which extends theirs and just overrides `getSourcePath`?

ThomasHaas:
The help messages from the latest commit do not seem to work.
I get a lot of options unspecific to Dartagnan and then it fails with `Exception in thread "main" java.lang.ClassFormatError: JVMCFRE149 Unknown constant tag 19 in class file; class=META-INF/versions/9/module-info, offset=54`

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionInfo.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.stream.Stream;
+
+/**
+ * Collects all {@link Option}s of a program.
+ * <p>
+ * Mimics {@link org.sosy_lab.common.configuration.OptionCollector OptionCollector} with reduced functionality.
+ */
+public final class OptionInfo {
+
+    private static final Pattern IGNORED_CLASSES = Pattern.compile("^org\\.sosy_lab\\.common\\..*Test(\\$.*)?$");
```
hernanponcedeleon, line 24: This is very nice, but somehow it "polutes" the output with the options coming from javaSMT. While it is true that we support those options and they (indirectly) affect our behavior, I think in general we won't use them and I would avoid showing then to the user. Can we use `IGNORED_CLASSES` to filter all options coming from them and not just test clases?

ThomasHaas: We could also introduce a custom annotation on top of the usual `@Options`, and then we only look for classes that have this custom annotation attached. That is probably a lot easier than playing around with patterns. We could even have two help options then, one printing only our custom options and one that also prints all JavaSMT options.

xeren: Alternatively, we filter our own classes and only show the options that we defined ourselves.

ThomasHaas: We should definitely improve the description output provided by --help. One big problem is that the value ranges are not clear at all, especially for all our enums.

hernanponcedeleon: I just pushed a change that shows enum options, but many descriptions can be improved. For all options it would be nice to show the default value (I did not find a nice way of doing this yet). I would also order the options alphabetically.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
ThomasHaas: From the code comments, it seems that this pass is actually non-functional right now. I haven't tested it, but we might be better off removing this for now especially because I think that the parsing code still performs this transformation.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
```diff
@@ -19,40 +26,75 @@
...
     private final Relation rel;
 
-    public SymmetryBreaking(VerificationTask task) {
-        this.task = task;
-        this.symm = task.getThreadSymmetry();
+    @Option(name = BREAK_SYMMETRY_ON_RELATION,
+            description = "The relation on which symmetry breaking should happen." +
+            "Empty, if symmetry shall not be encoded.",
+            secure = true)
+    private String symmBreakRelName = "rf";
```
hernanponcedeleon, line 45: Use `RelatinNameRepository`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -0,0 +1,109 @@
...
+        return new WmmEncoder(memoryModel, task);
+    }
+
+    @Override
+    public void initializeEncoding(SolverContext ctx) {
+        for(String relName : Wmm.BASE_RELATIONS) {
+            memoryModel.getRelationRepository().getRelation(relName);
+        }
+
+        /*for (Axiom ax : memoryModel.getAxioms()) {
```
hernanponcedeleon, line 39: Can we remove this?

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionInfo.java
ThomasHaas: As @xeren said, the default values must be extracted from the source code and can't be obtained via simple class information. So we would need to incorporate the default values directly into the description text (which is a little unfortunate).

dartagnan/src/main/java/com/dat3m/dartagnan/program/arch/linux/event/cond/RMWReadCond.java
```diff
@@ -32,8 +29,8 @@
     }
```
hernanponcedeleon, line 29: shouldn't this class override `cfImpliesExec()`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/Relation.java
```diff
@@ -175,7 +191,7 @@ protected BooleanFormula getExecPair(Event e1, Event e2, SolverContext ctx) {
             e1 = e2;
             e2 = temp;
         }
-        BranchEquivalence eq = task.getBranchEquivalence();
+        BranchEquivalence eq = analysisContext.get(BranchEquivalence.class);
```
hernanponcedeleon, line 194: why not `requires()`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/Relation.java
```diff
@@ -189,7 +205,7 @@ protected final BooleanFormula getExecPair(Tuple t, SolverContext ctx) {
     }
 
     protected void removeMutuallyExclusiveTuples(Set<Tuple> tupleSet) {
-        BranchEquivalence eq = task.getBranchEquivalence();
+        BranchEquivalence eq = analysisContext.get(BranchEquivalence.class);
```
hernanponcedeleon, line 208: why not `requires()`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/local/BasicRegRelation.java
```diff
@@ -30,7 +30,7 @@
     void mkTupleSets(Collection<Event> regReaders) {
         maxTupleSet = new TupleSet();
         minTupleSet = new TupleSet();
-        BranchEquivalence eq = task.getBranchEquivalence();
+        BranchEquivalence eq = analysisContext.get(BranchEquivalence.class);
```
hernanponcedeleon, line 33: why not `requires()`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
ThomasHaas: Will fix.

dartagnan/src/main/java/com/dat3m/dartagnan/program/arch/linux/event/cond/RMWReadCond.java
ThomasHaas: You fell into the same trap as I did. Those events are not conditional, their condition is only checked by the associated `RMWStoreCond`. The design sucks.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/Relation.java
ThomasHaas: I did not do it because `RelationAnalysis` did this already. And you can think of those methods as being part of `RelationAnalysis` in some extended sense. It's unfortunate that a lot of "external" code is mixed in the relation classes right now but I think that will change once we reworked the analysis.
But for now, I can also change it to `requires` (I guess we can replace all `get` by `requires`, it will cause no harm).

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionInfo.java
hernanponcedeleon: I think there are ways of [programmatically obtain the value of class fields](https://stackoverflow.com/questions/13400075/reflection-generic-get-field-value), however I think it would require to make the fields public which I'm not sure is a good design. 

I think manually incorporating this info is a no-go: it would be too easy to change the value and forget updating it in the description. Since the information is anyway logged, I would prefer not to give any such information in the help than risking giving a wrong one.

ThomasHaas: You don't need to make the fields public. The problem is that you need an instance of a class to read the value of its field. But you won't be able to obtain such an instance. If the fields were static it would work without an instance.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
hernanponcedeleon: Since the option is set off and the code is completely in a separate class, I'm fine with keeping the code and fixing it in a different PR.

ThomasHaas: My suggestion is to remove it only from the `ProcessingManager` and not completely. But I guess we have no users that would accidentally enable this option so it doesn't really matter I guess. Then let's keep it and add a TODO/FIXME

hernanponcedeleon: `AtomicBlocksAsLocks` already has both

ThomasHaas: You are right.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/RefinementTask.java
ThomasHaas: You fixed this only partially. In the development version we could have enabled all of the various baseline options simultanously, but in the new version you have to decide for a single one (it is not possible to have uniproc + oota).
A completely different alternative would be to allow a `.cat` file to be specified as baseline wmm, but then we need to provide some `.cat` files that can be used for refinement.
Either way, we can keep it as it is for now and try to extend this option later. We run refinement without any baseline model most of the time anyways.

hernanponcedeleon: Now you can assign any of these 3 options. The general solution using `.cat` could be nice, but until we have some empirical evidence on how useful this is (even just with the 3 options we have) I would not go in that direction.

ThomasHaas (approved):
Looks good to me. I think we can use this a new bases for further developement even if there are still many unfinished aspects like actually parsing configuration files, getting rid of `GlobalSettings` completely and some more.

hernanponcedeleon:
I'll merge once the CI passes then. We can solve the unfinished aspects in separated PRs.

Issue #200, opened 2022-01-11
Understanding C-Litmus conditional events


ThomasHaas:
I took a look into the special events we generate for C-Litmus (Linux) tests as they do not compile to our internal LISA-like core language. These special events consist of a few types of `ReadCond` events that, unlike their name, always perform a read but they also have a conditional defined on top of the read value (i.e., it depends on the `resultRegister`).
Then there are `RMWStoreCond` and `FenceCond` that both are actually conditional on the value read by `ReadCond`.

Let's look at `RMWStoreCond` first. This one performs a store operation if the conditional associated with its `RMWReadCond` is satisfied. This is effectively nothing but a conditional in between the `Read` and the `Store`:
`RMWReadCond`;`RMWStoreCond` <=> `RMWRead`;`CondJump(Label)`;`RMWStore`;`Label`
And indeed, this event is only generated in the compilation of atomic operations like CAS or add_unless.

**Question**: What is the reason that we do not insert branching there instead of using these special events? Is it because of added ctrl dependencies? Under Linux, those control dependencies do not propagate past the merging point so it would at most add a single ctrl edge between the read and the store.
And if those do make problems, can we add special branching instructions that do not cause ctrl-edges (or simply a tag for existing branching instructions). In a discussion about supporting `IMM`, we also found the need for such "invisible" branching to properly model CAS operations on the language level.

Now let's look at `FenceCond`. Its working is basically the same as that of `RMWStoreCond` in that it executes if and only if the associated read satisfies some condition. Again, a `FenceCond` that is placed after a `ReadCond` may as well be modelled by regular branching operations I think.
However, we frequently generate this strange code during compilation:
```
FenceCond(condRead)
condRead
condStore
local
FenceCond(condRead)
```
where there is a conditional fence **before** the load it depends on. That means this barrier only executes if the subsequent read satisfies some condition. This obviously cannot be modelled by adding simple branching instructions.

**Question**: How does it make sense to place a fence before a read but only execute it if that read satisfies something? Does Linux really have such kind of barriers? Is it possible to make this Fence unconditional or move it past the read so that is can be modelled via branching?

xeren:
Don't forget the `ctrl` edge between the `RMWReadCond` event and its connected `RMWWriteCond`, as well as the dependency to the compared value.  It might be required to work with `Assume` and `NonDet`, instead, since those constructions break dependency chains.  I once tried the naive way and broke several tests.

ThomasHaas:
[...]

PR #201, opened 2022-01-11
Package restructuring

ThomasHaas:
This PR has not semantics changes at all. It only changes the package-structure of the project:
- The `filter`-package got moved to `program` (it has more to do with program/events than it has to do with wmms).
- Various packages like `atomic` and `svcomp` were merged into the `event`-package.
- The `event`-package is subdivided into 3 packages `core`, `arch` and `lang` with the intention that `core` contains all events that make up our internal language (~LISA), `arch` contains all hardware-specific events (currently `tso` and `aarch64`) and `lang` contains all language specific events (`svcomp`, `catomics`, `pthread` and also `linux`).
- The non-`core` events should all be events that need to be compiled into our core language. Unfortunately, `linux` contains some non-compilable events that are not modelled in `core`. We need to address issue #200 and either get rid of those events or make them part of `core`.


ThomasHaas:
We have various `utils` packages with single classes in `arch` and `lang`. What do you guys think of moving these classes into the parent package and removing all these `utils` to reduce the nesting of package?

xeren:
I support the flattening of the package structure, except `com.dat3m.dartagnan.utils`.  It allows more usage of package-private accessible members.

ThomasHaas:
I wouldn't want to flatten the top-level `utils` package. I'm talking more about all these small `utils`-packages that contain a single `EType` or `Mo` class and possibly also the one that contains `EventCache` (renamed from `ThreadCache)`. 

Also, we should probably streamline the naming scheme of `Mo` and `EType`. I think `Mo` is essentially a subset of `EType`, concerned with the tags of memory operations. I would also rename `EType` to `ETag` or just `Tag`, because the word "type" suggests a distinction between something like Load, Store and Fence but we use it for all kinds of attached information, e.g. visibility, mo and strongness.

hernanponcedeleon:
I fully agree in flattening the `.utils` in event packages (also the one containing `EventCache`).

I think originally there was more of a distinction between these classes: while `Mo` was (and still is) about memory ordering, we did so many modifications to `EType` that I'm not even sure we have a "unified use" for them. Since at the end of all of them are tags (and all their uses match the use of tags in cat), I propose to create a tag package with maybe some further sub-packages like `mo`, `arch`, etc.

PR #202, opened 2022-01-11
Interface Event

xeren:
Restructures `com.dat3m.dartagnan.program.event.Event` as an interface.  This enables the separation of parsed, unrolled and compiled events, etc. in a stepwise manner, due to diamond-shaped inheritance.  `RegWriter` and `RegReaderData` may now extend `Event`.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/AbstractEvent.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/AbstractEvent.java
@@ -1,0 +1,12 @@
+package com.dat3m.dartagnan.program.event;
+
+public abstract class AbstractEvent extends Event {
+
+    protected AbstractEvent() {
+        super();
+    }
+
+    protected AbstractEvent(AbstractEvent other) {
+        super(other);
+    }
+}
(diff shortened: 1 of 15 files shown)


ThomasHaas:
I'm not sure I see the benefit yet. Can you elaborate on a particular use-case where this has any impact? Do you have any case in mind where we would not inherit from `AbstractEvent`? 
In particular, the fact that the `Event` interface describes a lot of state-like information such as all the `ids` and filters but also all the utility functionality such as `wasExecuted` makes it a very bad abstraction in my opinion.

xeren:
I think we agree that certain Program transformations are mandatory in this project.  Some subclasses of `Event` are not to be unrolled, or compiled, or encoded.  Interfacing enables us to transform this runtime exceptions into respective type errors.

ThomasHaas:
I don't think adding interfaces just for the sake of type checks is a good idea. And I don't think you will be able to enforce any compile time checks with this unless you introduce classes like `ParsedProgram`, `UnrolledProgram` and `CompiledProgram` that each contain appropriate versions of Thread-Classes that in turn expose their list of events in the appropriate type.
You could go with generics to simplify this, but the resulting type hierarchy sounds like a pain in the ass nonetheless.

Runtime checks are a lot easier to enforce by just adding an `annotation` like e.g `@NotCompilable` or simply adding an appropriate getter to check the property. As long as these checks happen early in the process, a runtime error is as valuable as a compile time error. And in this case, we have plenty of checks in place to immediately find any problematic event.

hernanponcedeleon:
Can this PR be closed or is there any intent to pursue this @xeren?

xeren:
Yes, I'll close it.

PR #203, opened 2022-01-12
Compilation

hernanponcedeleon:
The PR introduces the visitor pattern for events as an interface `EventVisitor` and adds an implementation `CompilationVisitor` performing the compilation logic. This means that the whole compilation logic is in this single class.
Currently, we have a single visitor for all architectures, but we could split this further and have a visitor for each single architecture.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/CompilationVisitor.java
```diff
@@ -0,0 +1,860 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class CompilationVisitor implements EventVisitor<List<Event>> {
+
+	Arch target;
+	
+	public CompilationVisitor(Arch target) {
+		this.target = target;
+	}
```
ThomasHaas, line 40: You can make the `target` a `private final` field. Also, check for `null` only once on construction and get rid of all the unnecessary `Precondition.notNull` checks in the visit methods.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/CompilationVisitor.java
```diff
@@ -0,0 +1,860 @@
...
+                        casFail,
+                            storeExpected,
+                        casEnd
+                );
+                break;
+            }
+            case POWER:
+            case ARM8: {
+                String loadMo = Tag.ARMv8.extractLoadMoFromCMo(mo);
+                String storeMo = Tag.ARMv8.extractStoreMoFromCMo(mo);
```
ThomasHaas, line 513: It is kind of ugly that we use ARMv8 tags like this even for Power. But I think this is an orthogonal issue to this PR.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/EventVisitor.java
```diff
@@ -0,0 +1,78 @@
...
+import com.dat3m.dartagnan.program.event.lang.pthread.*;
+import com.dat3m.dartagnan.program.event.lang.svcomp.*;
+
+public interface EventVisitor<T> {
+    
+	// Basic events
+	default T visit(Assume e) { return visit((Event)e); };
+	default T visit(Cmp e) { return visit((Skip)e); };
+	default T visit(CondJump e) { return visit((Event)e); };
+	default T visit(Event e) { return visit((Event)e); };
```
ThomasHaas, line 19: Endless recursion incoming :).

I'm thinking that it might be better to give the different `visit` methods also different names.
While it is somewhat easier to not do so, it is always easy to accidentally call the wrong methods.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Compilation.java
```diff
@@ -80,7 +81,7 @@ private int compileThread(Thread thread, int nextId) {
         pred.setCId(nextId++);
 
         while (toBeCompiled != null) {
-            List<Event> compiledEvents = toBeCompiled.compile(target);
+            List<Event> compiledEvents = toBeCompiled.accept(new CompilationVisitor(target));
```
ThomasHaas, line 84: The `CompilationVisitor` has immutable state, so just create it once and reuse it for all calls.

ThomasHaas (changes requested):
I'm not a big fan of having the `CompilationVisitor` in a different package than `Compilation` because we end up splitting the same functionality over two packages.
I would suggest to make the `CompilationVisitor` an inner class of `Compilation` and rename it to just `Visitor`. You could even make it a private inner class, because it is unlikely that we ever need to compile single events (if the need arises, we can make it public again). This way, the fact that `Compilation` uses a `Visitor` to perform compilation becomes an implementation detail.
You can see a similar approach in the `Reasoner` class for `caat` which uses a private inner visitor class.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/CompilationVisitor.java
hernanponcedeleon: Using such tags should not have any effect with `power.cat`. I think this ugly code is a co-product of having a single match statement in the switch for `ARM8` and `POWER`. Actually, we ended up converting things to power fences based on the extracted `loadMo` and `storeMo`, so we can achieve the same reasoning directly about the `Tag.C11`.

But as you said, this is orthogonal to this PR and I think we can treat in a different one where we implement one visitor per architecture.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/EventVisitor.java
hernanponcedeleon: Instead of null, I throw an exception. Each concrete class implementing a visitor should specify the behaviour for `visitEvent()` since this is the default method call for any method that is not override.

I also added the event class name to each method to avoid problems (this is also what the automatically generated visitor for antlr do).

ThomasHaas: I don't think visitors have to specify a behavior for the default case. Despite the interface, visitors do not have to have a return value and in fact, you probably wouldn't use return values for quite a few cases like e.g. implementing a ExprSimplifierVisitor which only visits events like `MemEvent` and `Local`.
And for these types of visitors, visiting any type of event they do not care about is basically a NOOP.
If the interface forces an exception there, you would need to overwrite the default behaviour with empty methods which is kinda ugly. Also, if you want this method to be overwritten in the first place, then you could simply avoid making it default.

hernanponcedeleon: By "visitors have to specify a behavior for the default case" I meant "for visitEvent()". All default implementations in the interface end up pointing to that method. Therefore, if a concrete visitor just overrides a couple of methods (which will probably be the case for most use-cases), for the remaining ones `visitorEvent()` will be called and thus that one must have a proper implementation. But the implementation is use-case specific (e.g., for compilation is just returns a singleton list with the input). You proposed to use `null` in this case, but that is dangerous because unless the method using the return of the visitor checks for not null, this would be unnoticed. That is why I think throwing an exception (only) for `visitEvent()` in the interface to force the programmer to implement **that** method for each visitor is the way to go.

ThomasHaas: No, if you want the implementor to implement `visitEvent` you simply make it non-default. Then it is impossible to forget it.
But you argue about return types and my argument is that not all visitors have return types. Those that do will have to overwrite `visitEvent, those that do not will not want to do so. And I think there are many cases where the visitor does not use return types, possibly more than there are that use return types.

hernanponcedeleon: I think that setting the default implementation of `visitEvent()` to `null` is just too dangerous. I would remove the default implementation of it, thus forcing the developer to provide an implementation, and if in that particular case a NOOP is needed, `null` can be provided as an implementation.

ThomasHaas: Removing the default implementation is ok, it is better than the default exception. Actually, throwing an exception as default behaviour in any interface or class is always stupid because the methods can be made non-default/abstract instead (I think we have this case in some abstract RMW base classes).

ThomasHaas (approved):
I haven't checked all details, but it looks good overall.
I think `visitX`/`accept` should probably have an extra parameter to pass data around, but we can add that when we have the need.

Issue #138, opened 2021-10-10
LISA parser


hernanponcedeleon:
[Here](https://github.com/paulmckrcu/litmus) there is a large corpus of litmus test for the LKMM.
Half of them are probably the ones we have in our test suite, however the are half we cannot currently handle because we do not have a parser for LISA.
We should implement this to have a larger set of LKMM tests.

ThomasHaas:
We have to double-check that these tests are actually valid for our current LKMM. The repo has various different definitions of the memory model (5 versions I think) that all differ from our current model I believe. So "wrong results" in some tests might be caused by differing memory models.

hernanponcedeleon:
We compare our results with those generated by herd, so as long as we run both tools with the same models, this should not be a problem.

hernanponcedeleon:
[This](https://github.com/herd/herdtools7/blob/b86aec8db64f8812e19468893deb1cdf5bbcfb83/litmus/LISACompile.ml#L92) needs to be used as a resource to understand the semantics of the LISA instructions.

hernanponcedeleon:
I pushed a branch with the implementation of the parser. The parser is yet not complete, but currently we can parse and verify (w.r.t our usual LKMM cat file) ~4000 new tests.

The problem is that I did not manage yet to make herd work with these tests (apparently herd requires some configuration file to be able to understand the semantics of the instructions) so I could not create the expected results to test if the parser is correct.

hernanponcedeleon:
I manage to make herd work for around half of the tests (2387). From those, two parsing problems remains. There is also one wrong result which I think it is related to control dependency yes. This is the program
```
LISA WWC+o-branch-o+o-branch-o
(* Allowed: Control dependencies not transitive.  Also allowed by PowerPC. *)
{
x = 0;
y = 0;
}
P0               | P1               | P2          ;
r[once] r1 x     | r[once] r2 y     | w[once] x 2 ;
mov r3 (eq r1 0) | mov r4 (eq r2 0) |             ;
b[] r3 END1      | b[] r4 END2      |             ;
w[once] y 1      | w[once] x 1      |             ;
END1:            | END2:            |             ;
exists(0:r1=2 /\ 1:r2=1 /\ x = 2)
```
Interestingly, we also report a wrong result (`PASS`) for `PPC`.

ThomasHaas:
[...]

PR #210, opened 2022-01-16
138

hernanponcedeleon:
This PR partially solves #138. It creates a LISA parser and compare the results with those benchmarks that I manage to far to run with herd.


ThomasHaas:
So are there still parsing issues for some benchmarks, or is the parsing fine but cannot verify half of the remaining tests against herd7?

hernanponcedeleon:
From the whole 4819 tests, we still have parsing problems in 113. However all of them fit in the following categories (and probably have some semantics that we currently do not handle)
- `lderef` access: "'lderef (*lockless_dereference*)" according to [this file](https://github.com/paulmckrcu/litmus/blob/master/models/pem.2016.01.20a.bell)
- `ghostacq` access: I was not able to find any mention about this kind of access in the bell files.
- `rb_dep` memory order for barriers: "'rb_dep (*smp_read_barrier_depends*)" according to [this file](https://github.com/paulmckrcu/litmus/blob/master/models/pem.2016.01.20a.bell)
- Weird use of `*` in locations definition like here
```
{
x0 = x1;
x1 = 0;
x2 = 42;
0:r5=x2;
}
 P0              | P1                 ;
 w[assign] x0 r5 | f[rcu_read_lock]   ;
 f[sync]         | r[deref] r1 x0     ;
 w[once] x1 7    | r[once] r2 r1      ;
                 | f[rcu_read_unlock] ;
locations [1:r1*;]
exists (1:r2=7)
```
Somehow it seems to be a pointer (even if the `*` is after the variable) related to the use of `deref` access.

ThomasHaas:
Let me see if I understand this. The tests are written in LISA and have various associated WMMs (from the repo you linked).
But we want to verify them against our own `lkmm.cat`, so the parser actually translates LISA -> Linux. In other words, we do not have a pure LISA parser but a LISA-to-Linux compiler is built in as well(*).

Now the problem is that you do not know how to translate those 3 tags `lderef`, `ghostacq` and `rb_dep` to Linux tags?
Cause in pure LISA, you would treat them just like any other tag. None of them have special semantics.

The weird declaration `locations [1:r1*;]` seems to just mark the register `1:r1` to contain address-values.
Maybe LISA needs to distinguish between registers containing values and registers containing addresses? I think our code does not do this distinction in the first place, so those declarations seem to be ignorable.

(*) It would be cleaner to have a pure LISA parser and then a processing-pass (LISA2LKMM) that translates all the tags.

hernanponcedeleon:
I think the LISA parser should parse the tag as it is. Then it is the WMM the one giving semantics to the (base) event based on the tag.

Something that will cause problems in our current implementation and we should decide how to handle are tags differing in the use of caps, e.g., `Acquire` vs `acquire`. Our current `linux-kernel.cat` uses `Acquire`, but then the matching `*.bell` and `*.def` files use `acquire`, thus I assume herd treat them as the same. 

If we treat them as the same (probably the best is to convert everything to upper case during parsing of program and cat files and use only upper case tags), the implementation of the LISA visitor becomes trivial because we just create a new `MemEvent` with the corresponding tag (which is nothing more than the parsed string). However, since we don't use `*.bell` files (where tags are defined as an enumeration type for herd), we still need to "hardcode" all allowed tags in the grammar. However extending this (if we add new LISA litmus) is trivial, because we just need to extend the grammar without modifying the visitor.

hernanponcedeleon:
> I think we should ignore case. But you will still run into cases where some memory model uses `acquire` while some other uses `acq`. And I expect that the only way to go about this is by defining compilation schemes for tags. Essentially, a compiler that translates a `.bell` file to another.

You are right, in fact this already happens with tests in this suite which use `sync` instead of `Sync-rcu`. 

Since this will cause more problems as soon as we want to run LISA tests against e.g. ARMv8, what about using JSON files to write mappings externally? Then we can create a program pass that
- has a `Map<String, String> lisa2arch` which gets populated by parsing the JSON file
- it traverses the program after compilation and if an event hast tag `t`, it also adds (there is no need to remove the original tags) `lisa2arch(t)` to the event

Then if the user wants to use litmus tests written in LISA, it only needs to provide a JSON file that maps every tag used in the `*.litmus` file the the corresponding tag in the `.cat` file.

ThomasHaas:
> Since this will cause more problems as soon as we want to run LISA tests against e.g. ARMv8, what about using JSON files to write mappings externally? Then we can create a program pass that
> 
>     * has a `Map<String, String> lisa2arch` which gets populated by parsing the JSON file
> 
>     * it traverses the program after compilation and if an event hast tag `t`, it also adds (there is no need to remove the original tags) `lisa2arch(t)` to the event
> 
> 
> Then if the user wants to use litmus tests written in LISA, it only needs to provide a JSON file that maps every tag used in the `*.litmus` file the the corresponding tag in the `.cat` file.

That is possible. But I feel like you won't be able to map tags uniformly like this. I think you can have e.g. the same tag for writes and fences, but the compilation may translate them differently due to one being for writes and another being for fences.
Similarly, RMW operations with a tag will probably pass that tag to the generated load/store pair after compilation (or can you specify separate tags for the generated load/store pair in LISA?).
But it seems likely that e.g. "RMW[acq]" should generate "r[acq];w[]" for many memory models, i.e., the plain LISA compiler will generate `r[acq];w[acq]` but the compilation to some hardware might remove the `acq` tag from stores.
Also, you should remove the old tag, because it may conflict with another tag in the `.cat`.

hernanponcedeleon:
> That is possible. But I feel like you won't be able to map tags uniformly like this. I think you can have e.g. the same tag for writes and fences, but the compilation may translate them differently due to one being for writes and another being for fences. 

AFAIK, LISA is almost our core language, the only exception is `RWM` (which is compiled to `R[];(L[])*;W[]` in base). While I think it is unprovable that a tag would be converted to something different for a `MemEvent` and for a `Fence`, we can simply have more information in the JSON file, i.e., explicitly state for which type of event the transformation is. Something like this
```
"S": {
    "acquire": "N/A",
    "release": "Release",
  },
"L": {
    "acquire": "Acquire",
    "release": "N/A",
  },
"F": {
    "acquire": "Acquire",
    "release": "Release",
  },
```

>Similarly, RMW operations with a tag will probably pass that tag to the generated load/store pair after compilation (or can you specify separate tags for the generated load/store pair in LISA?). But it seems likely that e.g. "RMW[acq]" should generate "r[acq];w[]" for many memory models, i.e., the plain LISA compiler will generate `r[acq];w[acq]` but the compilation to some hardware might remove the `acq` tag from stores. 

I don't think we should mix a pass where tags are converted with the compilation pass. The "tag transformation" pass should convert `RMW[acq]` to `RMW[Acquire]` and the compilation pass will convert (this we are already doing) `RMW[Acquire]` to `R[Acquire];W[]`.

> Also, you should remove the old tag, because it may conflict with another tag in the `.cat`.

If it conflicts, it means that the tag is used in the target WMM and thus the mapping should be the identity, or am I missing something?

ThomasHaas:
> AFAIK, LISA is almost our core language, the only exception is `RWM` (which is compiled to `R[];(L[])*;W[]` in base). While I think it is unprovable that a tag would be converted to something different for a `MemEvent` and for a `Fence`, we can simply have more information in the JSON file, i.e., explicitly state for which type of event the transformation is. [...]

I think this is realistic. Consider e.g. C11 which has release fences as well as release writes. When compiling to e.g. ARMv8, the store will keep its release tag, while the fence will get the `DMB.ISH` tag. And this compilation is purely on the tags itself.

> I don't think we should mix a pass where tags are converted with the compilation pass. The "tag transformation" pass should convert `RMW[acq]` to `RMW[Acquire]` and the compilation pass will convert (this we are already doing) `RMW[Acquire]` to `R[Acquire];W[]`.

Ok, but then you perform the tag transformation before compilation. Then you also need to distinguish between `rmw` as well as `w`, `r` and `f`. 

> > Also, you should remove the old tag, because it may conflict with another tag in the `.cat`.

> If it conflicts, it means that the tag is used in the target WMM and thus the mapping should be the identity, or am I missing something?

No, assume that the source file uses `REL` as release tag for both stores and fences. Now assume the target WMM also uses `REL` for stores but `FREL` for fences. Now a compilation should map `w[REL] -> w[REL]` but `f[REL] -> f[FREL]`.
If you retain the `REL`-tag for `f[REL]`, then any usage of `[REL]` in the `cat`-file also applies to the fences which may be unintended as those should not be tagged by `REL` (the `.bell` file would forbid this).

hernanponcedeleon:
> I think this is realistic. Consider e.g. C11 which has release fences as well as release writes. When compiling to e.g. ARMv8, the store will keep its release tag, while the fence will get the `DMB.ISH` tag. And this compilation is purely on the tags itself.
> 

There is no compilation involved in litmus test. LISA is supposed to be flexible to allow you write programs and run them on different WMM, but for the example you mention, I think you are supposed to write directly
`W[Rel]` and `F[DMB.ISH]`.

> Ok, but then you perform the tag transformation before compilation. 

Yes.

> Then you also need to distinguish between `rmw` as well as `w`, `r` and `f`.

Yes. We need to support any event that can be tagged, i.e. All visible ones I think.

> No, assume that the source file uses `REL` as release tag for both stores and fences. Now assume the target WMM also uses `REL` for stores but `FREL` for fences. Now a compilation should map `w[REL] -> w[REL]` but `f[REL] -> f[FREL]`. If you retain the `REL`-tag for `f[REL]`, then any usage of `[REL]` in the `cat`-file also applies to the fences which may be unintended as those should not be tagged by `REL` (the `.bell` file would forbid this).

Again, my understanding is that you are supposed to use `F[FREL]` directly.

Before this PR I never really used LISA in Herd so I'm not fully sure that my understanding of how tags should be used in those are correct. I think that our need to have this tag mappings is due to the lack of support for the whole machinery herd has with `.bell` and `.def` files. However, if we are going with the idea of supporting tag mappings, I agree that we need to handle both cases you mentioned above. 

One of the side effects of diverting from herd in the use of tag mappings is that we can "reuse" LISA tests for several architectures, e.g. we can write `F[sync]` and have `sync -> mfence` and `sync -> DMB.ISH` for TSO and ARM respectively.

ThomasHaas:
> > I think this is realistic. Consider e.g. C11 which has release fences as well as release writes. When compiling to e.g. ARMv8, the store will keep its release tag, while the fence will get the `DMB.ISH` tag. And this compilation is purely on the tags itself.
> 
> There is no compilation involved in litmus test. LISA is supposed to be flexible to allow you write programs and run them on different WMM, but for the example you mention, I think you are supposed to write directly `W[Rel]` and `F[DMB.ISH]`.

I'm a little confused. Don't we have a case right now that the tags provided in the LISA programs do not match the tags needed for the WMM we want run it on (LKMM)? So either you change the WMM to work with these tags, or you change the LISA code to provide different tags. Both approaches are some form of compilation.
And my impression was that we have tests for LISA code designed for WMM X, but we want to run those tests under WMM Y  (LKMM). We do so by changing the tagging of the LISA code appropriately.
And now I don't find it hard to believe that you can have LISA code that is written using C11 tags but you want to test the ARMv8 model on it so you provide a simple tag-mapping (which is enough for programs without RMWs I think). In principal, you could run a tranformer on the LISA code that changes tagging before we parse it into Dartagnan (similar to how Boogiesan runs on .bpl files).

> Before this PR I never really used LISA in Herd so I'm not fully sure that my understanding of how tags should be used in those are correct. I think that our need to have this tag mappings is due to the lack of support for the whole machinery herd has with `.bell` and `.def` files. However, if we are going with the idea of supporting tag mappings, I agree that we need to handle both cases you mentioned above.
> 
> One of the side effects of diverting from herd in the use of tag mappings is that we can "reuse" LISA tests for several architectures, e.g. we can write `F[sync]` and have `sync -> mfence` and `sync -> DMB.ISH` for TSO and ARM respectively.

This mapping is exactly what we do right now to verify against LKMM.
I interpret such a tag-mapping as a light-weight compiler where events are compiled to same events but with different tags. This may or may not be sufficient depending on how much the target architectures/wmms differ. But at least for language-level wmms, you would expect the code to remain identical except possibly for tags, and I think C11, RC11, IMM and Linux all belong into this category.
In general, if you would allow a more general mapping function that maps single tagged events to a sequence of tagged events, then you have a full compiler. But I think expressing these full compilation schemes in some simple file format is not that easy.

hernanponcedeleon:
> I'm a little confused. Don't we have a case right now that the tags provided in the LISA programs do not match the tags needed for the WMM we want run it on (LKMM)? So either you change the WMM to work with these tags, or you change the LISA code to provide different tags. Both approaches are some form of compilation. 

I initially mentioned the idea of having tags mapping due to two mismatches I found. Notice this is related to LISA tests that were intended to be run against the `LKMM`; see (*).
- litmus tests use e.g., `acquire` while the `.cat` file uses `Acquire`
- [Some litmus](https://github.com/hernanponcedeleon/Dat3M/blob/138/litmus/LISA/LISA1Rd1G.litmus) use `F[sync]` while the `.cat` file uses `Sync-rcu`

> And my impression was that we have tests for LISA code designed for WMM X, but we want to run those tests under WMM Y (LKMM). 

(*) I'm not sure about this. I tried to find LISA examples (not only on herd's repo, but also on other repos) and the only place where I found LISA litmus is the repo linked in #138. Specifically, herd repo only uses the corresponding ISA for hardware WMM, e.g. [here](https://github.com/herd/herdtools7/blob/master/catalogue/aarch64/tests/CAS%2Bdata1.litmus) or [here](https://github.com/herd/herdtools7/blob/master/catalogue/x86/tests/2%2B2W.litmus). For `LKMM`, test are only written in `C` in herd's repo, like [here](https://github.com/herd/herdtools7/blob/master/catalogue/linux/tests/MP%2Brelacq.litmus). Paul McKenney even mentioned that [the official `LKMM` only accepts C-language litmus tests](https://github.com/paulmckrcu/litmus/issues/10). 

[This](https://github.com/herd/herdtools7/blob/ec247deebcaeb2505ef9e887c48910bb8dd109ec/CHANGES.txt#L22) and [this](https://github.com/herd/herdtools7/blob/ec247deebcaeb2505ef9e887c48910bb8dd109ec/litmus/LISACompile.ml#L92) suggest that LISA tests are not intended to be run against `LKMM`. I'm not entirely sure what LISA tests are for.

> We do so by changing the tagging of the LISA code appropriately. 

Tag (used in litmus tests) and sets or unary predicates (used in cat files) are not the same; see (**).

> This mapping is exactly what we do right now to verify against LKMM. 

If you mean the original code of this PR, I agree. Otherwise I don't understand what you mean. In the current code in `development` I don't think we do any tag transformation for `LKMM`.

> I interpret such a tag-mapping as a light-weight compiler where events are compiled to same events but with different tags. This may or may not be sufficient depending on how much the target architectures/wmms differ. But at least for language-level wmms, you would expect the code to remain identical except possibly for tags, and I think C11, RC11, IMM and Linux all belong into this category. 

I find the word "compilation" too overloaded. But yes, this was kind of my idea of the tag mapping, to convert between (possible only language-level) wmms tags but keeping the events as they are.

> In general, if you would allow a more general mapping function that maps single tagged events to a sequence of tagged events, then you have a full compiler. But I think expressing these full compilation schemes in some simple file format is not that easy.

If we stick with the idea of "compilation" involving both `lang -> arch -> core`, then we could probably have the whole compilation mapping as a JSON file and allow any kind of compiler mapping (here we would do both event conversion and tag conversion).

(**) I tried to dig into herd's code to understand the whole tag business. I think we currently make no distinction between tags and sets (unary predicates). Tags are the ones used in the litmus files, while sets are the ones used in cat. Herd provides a way to [convert tags to sets](https://github.com/herd/herdtools7/blob/b86aec8db64f8812e19468893deb1cdf5bbcfb83/herd/libdir/c11.bell) and this might be what we are missing and generating the mismatches I mentioned above. Herd also [differentiate memory orders from other tags](https://github.com/herd/herdtools7/blob/master/lib/memOrderOrAnnot.mli) and [this file](https://github.com/herd/herdtools7/blob/ec247deebcaeb2505ef9e887c48910bb8dd109ec/herd/CAction.ml#L321) kind of explicitly show the mismatch I mentioned.

hernanponcedeleon:
As mentioned in #138, this PR finally just adds a simple LISA parser.
I'm not fully happy with adding 3 events during the parsing of `rmw`. The alternative would be to create a new `RMW` event and then create the corresponding 3 events when compiling. But with our current distinction of core, arch, lang ... I was not sure where to make this new event fit.

ThomasHaas:
I guess one can argue about whether `LISA` is a language or if it is a fictional architecture. But since we intend to not compile `LISA` to any architecture like you would do for languages, I think it makes sense to see it as an architecture.
So it would fit into `arch/lisa` I think.

ThomasHaas (approved):
I think this is fine and can be merged.
I haven't checked all the details though.

PR #212, opened 2022-02-14
Alias method as OptionInterface

hernanponcedeleon:
This PR adds a new enum type (similar to Arch, Method, etc) to select between the different alias analysis.

Interestingly, in cases where I expected both analyses to behave (i.e. size of the maxTupleSets) differently like  mutex_musl, they do not.


ThomasHaas (approved):
LGTM

PR #213, opened 2022-02-14
Constant propagation

hernanponcedeleon:
This PR fixes the problems in the constant propagation which was already part of `development` but set off due to the problems. It also adds a dead store elimination pass (called DeadAssignementElimination to avoid confusion). Finally it fixes some problems with event tagging.

The analysis can be improved in at least two ways:
- we don't have a proper join for events having more than one predecessor (i.e. certain labels). Basically the join
 right now just returns TOP. Despite this, the analysis is able to perform several propagations.
- it could be possible to perform further simplifications, i.e.
```
r1 = load(y)
r2 = r1;
store(x, r2)
```
is not currently converted to
```
r1 = load(y)
store(x, r1)
```
However this is not exactly constant propagation. I'll handle this situation in a different PR


ThomasHaas:
I think adding joins (actually meets, no?) is not particularly difficult. To perform constant propagation (or more general propagation that avoids intermediate registers), you can linearly traverse the thread storing a single `Map<Register, IExpr>`.
Additionally, you can store a `Map<Label, Map<Register, IExpr>>` that stores all the inflows of each label.
Whenever you encounter a jump, you can make a copy of your current map and join/meet it to the value that is stored for the target label.
Whenever you encounter a `Label`, you look up the value associated to that label and join/meet it with whatever value you currently have, unless the previous statement was a goto cause that one does not produce direct inflow (i.e. you have no "current value" and you just take whatever is the label map).

Also, by not relying on the listener sets of jumps, you get accurate information. The listener sets usually contain outdated values (maybe we should have a pass that fixes this?).

dartagnan/src/main/java/com/dat3m/dartagnan/asserts/AssertBasic.java
```diff
@@ -35,4 +38,16 @@ private String valueToString(LastValueInterface value){
...
+	@Override
+	public List<Register> getRegs() {
+		List<Register> regs = new ArrayList<>();
+		if(e1 instanceof Register) {
+			regs.add((Register) e1);
+		}
+		if(e2 instanceof Register) {
+			regs.add((Register) e2);
+		}
+		return regs;
```
ThomasHaas, line 51: Isn't it possible that `e1` and `e2` are complex expressions? Don't you need to traverse them for the registers via "e.getRegs"?.

dartagnan/src/main/java/com/dat3m/dartagnan/asserts/AssertNot.java
```diff
@@ -27,4 +31,11 @@ public BooleanFormula encode(SolverContext ctx) {
     public String toString() {
         return "!" + child.toString();
     }
+
+    @Override
+	public List<Register> getRegs() {
+		List<Register> regs = new ArrayList<>();
+		regs.addAll(child.getRegs());
+		return regs;
```
ThomasHaas, line 39: you could also just return `child.getRegs` or if you really need a copy, you can do "return new ArrayList(child.getRegs())".

dartagnan/src/main/java/com/dat3m/dartagnan/asserts/AssertTrue.java
```diff
@@ -15,4 +20,9 @@ public BooleanFormula encode(SolverContext ctx) {
     public String toString(){
         return "true";
     }
+
+    @Override
+	public List<Register> getRegs() {
+		return new ArrayList<>();
```
ThomasHaas, line 26: Do you need mutability? Otherwise, you should return `Collections.emptyList()`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ConstantPropagation.java
```diff
@@ -163,21 +171,19 @@ else if(current instanceof Load) {
...
-//        				String.format("Expression %s got no value after constant propagation analysis", oldValue));
-//        		
-//        		e = EventFactory.newLocal(reg, newValue);
+        		Register reg = ((Local)current).getResultRegister();
+        		
+        		IExpr oldValue = (IExpr) ((Local)current).getExpr();
+        		IExpr newValue = evaluate(oldValue, currentThreadId);
+        		newValue = newValue instanceof ITop ? oldValue : newValue; 
+        		Preconditions.checkState(newValue != null, 
+        				String.format("Expression %s got no value after constant propagation analysis", oldValue));
```
ThomasHaas, line 184: For checking of invariants, use `Verify.verify` instead. Preconditions are meant to check for mistakes made by the caller, not by the algorithm itself.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DeadAssignementElimination.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.sosy_lab.common.configuration.Configuration;
+import org.sosy_lab.common.configuration.InvalidConfigurationException;
+
+// This is just Dead Store Elimination, but the use of the term "Store" can be confusing in our setting 
+public class DeadAssignementElimination implements ProgramProcessor {
```
ThomasHaas, line 19: Typo in "Assign(e)ment".

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DeadAssignementElimination.java
```diff
@@ -0,0 +1,87 @@
...
+        logger.info("#Events after DSE: " + program.getEvents().size());
+    }
+
+    private void eliminateDeadStores(Thread thread) {
+    	
+    	List<Event> removed = new ArrayList<>();
+    	
+    	// We traverse the events in inverse order and mark those that can be removed
+    	ArrayList<Event> sorted = new ArrayList<Event>(thread.getEvents());
+    	sorted.sort((Event e1, Event e2) -> e2.compareTo(e1));
```
ThomasHaas, line 54: You can use `Comparator.reverseOrder()`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DeadAssignementElimination.java
```diff
@@ -0,0 +1,87 @@
...
+    					.filter(s -> s instanceof MemEvent && !removed.contains(s))
+    					.noneMatch(s -> ((MemEvent)s).getAddress().getRegs().contains(rw.getResultRegister()));
+    			// We use register to pass the parameters to new threads
+    			boolean notUsedByOtherThread = program.getEvents().stream()
+    					.filter(o -> !o.getThread().equals(e.getThread()) && o instanceof RegReaderData && !removed.contains(o))
+    					.noneMatch(s -> ((RegReaderData)s).getDataRegs().contains(rw.getResultRegister()));
+    			boolean notUsedByAssertion = !program.getAss().getRegs().contains(rw.getResultRegister());
+				if(notUsedByRegReader && notUsedByMemEvent && notUsedByOtherThread && notUsedByAssertion) {
+    				removed.add(e);
+    			}
```
ThomasHaas, line 72: Wouldn't it be easier to store a list of used registers?
Initially it is empty(*), but whenever you encounter a `MemEvent` that uses a register, you mark it as "used".
Whenever you encounter a local assignment, you check if it assigns to a used register and if not, you delete it. Otherwise, you add its dependencies to the used registers as well.

(*) To incorporate assertions and other threads, you can initially fill it with registers related to those.

ThomasHaas (changes requested):
I don't like the usage of the `propagationMap` in `ConstantPropagation`. Since propagation happens thread-locally, there is no need to maintain such a map (unless you want to inspect that map externally for some reason).
You can have a single map locally created inside the function for each thread and pass the map to `evaluate` (instead of a threadid for lookup).

dartagnan/src/main/java/com/dat3m/dartagnan/asserts/AssertBasic.java
hernanponcedeleon: No, `e1` and `e2` are `LastValueInterface` which is only extended by `IConst`, `Location`, `MemoryObject` and `Register`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DeadAssignementElimination.java
hernanponcedeleon: You also need to consider the case of `Jump`, since those also read registers. From an implementation point of view, you also need to split the case of `Load` and `Store` (and not just tread them as `MemEvent`) since for the former you need to add to "used" `getResultRegister` and for the later `getDataRegs`. I tried all this and the code becomes larger, thus I don't this what you propose is easier.

hernanponcedeleon:
All changes implemented

hernanponcedeleon:
The failure in the CI was due to the fact that the alias analysis calls `IOpBin.combine()` and we throw and exception if the operator is >>> (due to lack of support in `BigInteger`). I fixed the problem by just returning `null` from the `Collector` visitor if this is the operator. My understanding is that is the way of handling imprecision. @xeren can you please confirm?

xeren:
> The failure in the CI was due to the fact that the alias analysis calls `IOpBin.combine()` and we throw and exception if the operator is >>> (due to lack of support in `BigInteger`). I fixed the problem by just returning `null` from the `Collector` visitor if this is the operator. My understanding is that is the way of handling imprecision. @xeren can you please confirm?

Yes, with this, the expression is marked as imprecise.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DeadAssignementElimination.java
ThomasHaas: I have pushed a new version that does what I proposed. I think it is a lot clear and the code size is pretty much identical (if it works correctly)

ThomasHaas (approved):
LGTM

PR #215, opened 2022-02-28
Improving Constant Propagation code

hernanponcedeleon:
This PR improves the code of constant propagation in the following ways
- constant propagation now also reduces `BExpr`s (the previous code only worked with `IExpr`s)
- evaluated expressions are further simplified using `ExprSimplifier` (hopefully it removed some burden from the SMT-solver)
- rather than completely creating new events (using the event factory), things like `address` and `memValue` are not marked as final anymore and a new `EventVisitor` takes care of updating them using the computed map.

Now that more propagations are performed, I noticed a problem with some benchmarks in `SVCOMP-cf` having overflows, thus I removed them from our unit tests.

Now on `ClockTests`, constant propagation constantly gives a ~12% performance improvement.


hernanponcedeleon:
The latest changes look good to me. If @ThomasHaas approves I'll merge.

ThomasHaas (approved):
LGTM.

The only thing which might be better to do is to compute `getDataRegs` on demand instead of on construction/whenever the used expressions change. This way the data regs always match with the actual expression without any need to maintain explicit state.
But I think we can change this in a separate PR.

PR #217, opened 2022-03-03
Replace duplicated bound option

hernanponcedeleon:
This PR removed the duplicated bound option (once in the LoopUnrolling pass and once in the svcomp witness builder) by storing the bound as program metadata after unrolling


ThomasHaas (approved):
LGTM

PR #220, opened 2022-03-07
Centralized precision

hernanponcedeleon:
This PR centralises the precision for variables for which a precisions is not given (i.e. any variable if the smack code was generated using `--integer-encoding != bit-vector`) and addresses seen if the boogie code was generated with `--integer-encoding bit-vector`.


dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpUn.java
```diff
@@ -96,9 +98,9 @@ public Formula encode(Formula e, SolverContext ctx) {
     		case MINUS:
     				return bvmgr.subtract(bvmgr.makeBitvector(32, BigInteger.ZERO), bv);
     		case BV2UINT:
-    				return bvmgr.toIntegerFormula(bv, false);
+    				return ARCH_PRECISION > -1 ? bv : bvmgr.toIntegerFormula(bv, false);
```
ThomasHaas, line 101: I'm wondering if this code should perform an extension of the provided `bv` instead.
E.g. if you have a `bv8` and convert it to `int` on a 32 bit architecture, it will convert it to ´bv32´ instead.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -53,7 +54,7 @@ public static Load newLoad(Register register, IExpr address, String mo, int cLin
     }
 
     public static Load newLoad(Register register, IExpr address, String mo) {
-        return newLoad(register, address, mo, -1);
+        return newLoad(register, address, mo, ARCH_PRECISION);
```
ThomasHaas, line 57: The `-1` here is a Cline argument, not a precision argument.
This is true for all the methods you changed in `EventFactory`.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpUn.java
hernanponcedeleon: At this point in the code, we don't have information about the precision of the expression (if so, we could just extend `ARCH_PRECISION - prec` bits), thus the best we can do is to convert the expression to an int, and then just create a `bv` with `ARCH_PRECISION`.

ThomasHaas: I think you do have that information via `bvmgr.getLength(bv)`.
So you can extend by `ARCH_PRECISION - bvmgr.getLength(bv)`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpUn.java
```diff
@@ -94,36 +96,37 @@ public Formula encode(Formula e, SolverContext ctx) {
 			BitvectorFormula bv = (BitvectorFormula)e;
 	    	switch(this) {
     		case MINUS:
-    				return bvmgr.subtract(bvmgr.makeBitvector(32, BigInteger.ZERO), bv);
+    			return bvmgr.subtract(bvmgr.makeBitvector(32, BigInteger.ZERO), bv);
```
ThomasHaas, line 99: There is a `bvmgr.negate` function that does what you want (not to be confused with `bvmgr.not` which is the logical ~ operator). It will also keep the precision of the original bv-expression.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
```diff
@@ -183,9 +184,11 @@ protected BooleanFormula encodeApprox(SolverContext ctx) {
...
                 lastCo = bmgr.and(lastCo, bmgr.not(relation));
 
-                IntegerFormula a1 = convertToIntegerFormula(w1.getMemAddressExpr(), ctx);
-                IntegerFormula a2 = convertToIntegerFormula(w2.getMemAddressExpr(), ctx);
-                BooleanFormula sameAddress = imgr.equal(a1, a2);
+                Formula a1 = w1.getMemAddressExpr();
+                Formula a2 = w2.getMemAddressExpr();
+                BooleanFormula sameAddress = ARCH_PRECISION > -1 ? 
+                		bvmgr.equal((BitvectorFormula)a1, (BitvectorFormula)a2) : 
+                		imgr.equal((IntegerFormula)a1, (IntegerFormula)a2);
```
ThomasHaas, line 191: Isn't this what `Utils.generalEquals` does?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
```diff
@@ -215,12 +218,18 @@ protected BooleanFormula encodeApprox(SolverContext ctx) {
...
+                    Formula v1 = w1.getMemValueExpr();
+                    Formula v2 = init.getBase().getLastMemValueExpr(ctx,init.getOffset());
+                    // The boogie file might have a different type (Ints vs BVs) that the imposed by ARCH_PRECISION
+                    // In such cases we perform the transformation 
+                    BooleanFormula sameAddress = ARCH_PRECISION > -1 ? 
+                    		bvmgr.equal(convertToBitvectorFormula(a1, ctx), convertToBitvectorFormula(a2, ctx)) : 
+                        	imgr.equal(convertToIntegerFormula(a1, ctx), convertToIntegerFormula(a2, ctx));
+                    BooleanFormula sameValue = ARCH_PRECISION > -1 ? 
+                    		bvmgr.equal(convertToBitvectorFormula(v1, ctx), convertToBitvectorFormula(v2, ctx)) : 
+                        	imgr.equal(convertToIntegerFormula(v1, ctx), convertToIntegerFormula(v2, ctx));
```
ThomasHaas, line 232: Again, `generalEquals` should do the trick

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelRf.java
```diff
@@ -179,13 +180,18 @@ protected BooleanFormula encodeApprox(SolverContext ctx) {
...
+            // The boogie file might have a different type (Ints vs BVs) that the imposed by ARCH_PRECISION
+            // In such cases we perform the transformation 
+            BooleanFormula sameAddress = ARCH_PRECISION > -1 ? 
+            		bvmgr.equal(convertToBitvectorFormula(a1, ctx), convertToBitvectorFormula(a2, ctx)) : 
+            		imgr.equal(convertToIntegerFormula(a1, ctx), convertToIntegerFormula(a2, ctx));
+            Formula v1 = w.getMemValueExpr();
+            Formula v2 = r.getMemValueExpr();
+            BooleanFormula sameValue = ARCH_PRECISION > -1 ? 
+            		bvmgr.equal(convertToBitvectorFormula(v1, ctx), convertToBitvectorFormula(v2, ctx)) : 
+                	imgr.equal(convertToIntegerFormula(v1, ctx), convertToIntegerFormula(v2, ctx));
```
ThomasHaas, line 194: Same as above

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpUn.java
hernanponcedeleon: Ohhh I was looking to the methods from `BitvectorFormula` instead of the manager. Thanks for the pointer

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
hernanponcedeleon: No precisely. This code might potentially be unsafe because the lack of conversions (e.g. `convertToBitvectorFormula`). See the detail comment below.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
hernanponcedeleon: if `ARCH_PRECISION == -1`, addresses will have this precision and thus when encoded they will be `IntegerFormula`s. If the boogie file was generated by telling smack to use bit-vectors, it is possible that the accessed address of a `MemEvent` has `precision > 0` (e.g. in the presence of pointers) and thus when encoded they will be `BitvectorFormuala`s. Thus the conversions are required.

I just discovered this because the svcomp-loop benchmarks (which were generated by tell smack to use bv) were failing. It seems there were no problems related to `Init` events and then I just missed the change there (but the change should also be applied there).

ThomasHaas: I'm not sure I can follow. As far as I see you do have conversion operators when comparing to `Init` events (for the sake of defining `lastCo`) but you do not have those conversions when dealing with normal `Writes` (see the above comment on that). But according to what you say, exactly there you should run into problems, no?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
ThomasHaas: The code snippet has no conversion operators so it will break if there is a type mismatch just like `generalEquals` would.
In fact, it is more likely to break in my eyes.

hernanponcedeleon: Exactly, since this code does not apply the conversions, it can run into problems (even if our current tests do not expose this).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
hernanponcedeleon: Sorry, it was the other way round. In summary, the same code (using conversions) shall be used everywhere. I think the best is to do the globalFal checking + conversions in the definition of general equal and use this method everywhere

ThomasHaas (approved):
LGTM

PR #221, opened 2022-03-14
Javasmt

hernanponcedeleon:
Update javaSMT to version 3.12.0 and make use of the new statistics feature.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
@@ -96,6 +96,13 @@ public class RefinementSolver {
 
         logger.info("Refinement procedure started.");
         while (!prover.isUnsat()) {
+        	if(iterationCount == 0) {
+        		String smtStatistics = "\n ===== SMT Statistics (after first iteration) ===== \n";
+        		for(String key : prover.getStatistics().keySet()) {
+        			smtStatistics += String.format("\t%s -> %s\n", key, prover.getStatistics().get(key));
+        		}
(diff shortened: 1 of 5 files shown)


hernanponcedeleon:
The failure in the CI was due to a problem with javaSMT (i.e. external to us) which does not accept the same key for statistics twice. Z3 seems to be generating those for variable elimination (maybe due to variables of different types). I'll try to open an issue in their repo, but I need to provide a way of reproducing the problem which is generally not easy (I need to see if I can create an smt2 file and the problem persists if javasmt generates the formula from the file).

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -148,6 +155,14 @@ public static Result run(SolverContext ctx, ProverEnvironment prover, Refinement
...
                         " ===== Final Iteration: {} =====\n" +
                         "Native Solving/Proof time(ms): {}", iterationCount, curTime - lastTime);
+		
+        if(logger.isDebugEnabled()) {        	
+            String smtStatistics = "\n ===== SMT Statistics (after final iteration) ===== \n";
+    		for(String key : prover.getStatistics().keySet()) {
+    			smtStatistics += String.format("\t%s -> %s\n", key, prover.getStatistics().get(key));
+    		}
+    		logger.debug(smtStatistics);
+        }
```
ThomasHaas, line 165: I think this should be at the very end of the verification (after the bound check) so it is aligned with the way the statistics of the other solving methods are produced.

ThomasHaas (changes requested):

ThomasHaas (approved):
LGTM

PR #222, opened 2022-03-19
Benchmarking

hernanponcedeleon:
This PR brings some small improvements which allows easy benchmarking. 

Changes include
- storing the verification result in the csv file
- having a simple C file per lock benchmark with macros that allows to generated relaxed variations
- new unit tests (one for each architecture, only safe version of benchmarks) that generates several cvs files which make it easy to run script for automatic graph generation.
- automatic and manual detection of spinloops (which result in one of the tests to return PASS instead of UNKNOWN)


hernanponcedeleon:
Now we generate csv for all tools that we normally compare against so the PR is truly ready for review.

ThomasHaas (approved):
LGTM

PR #223, opened 2022-04-01
Mod vs rem

hernanponcedeleon:
This PR adds encoding implementation to `srem` and `urem` in the case of BVs. As clarified by the java-SMT authors, what their API calls `modulo` is actually the reminder. You can compute `mod` for that one. I used this from [here](https://llvm.org/docs/LangRef.html)
> This instruction returns the remainder of a division (where the result is either zero or has the same sign as the dividend, op1), not the modulo operator (where the result is either zero or has the same sign as the divisor, op2) of a value.

to implement `mod`.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
@@ -127,7 +127,16 @@ public enum IOpBin {
             	case UDIV:
             		return bvmgr.divide(bv1, bv2, this.equals(DIV));
             	case MOD:
+            		BooleanFormulaManager bmgr = ctx.getFormulaManager().getBooleanFormulaManager();
+            		BitvectorFormula rem = bvmgr.modulo(bv1, bv2, true);
+            		// Check if rem and bv2 have the same sign
+            		BooleanFormula cond = bvmgr.greaterThan(bvmgr.multiply(rem, bv2), bvmgr.makeBitvector(32, 1), true);
+            		// If they have the same sign, return the reminder, otherwise invert it
+            		bmgr.ifThenElse(cond, rem, bvmgr.negate(rem));
+            	case SREM:
             		return bvmgr.modulo(bv1, bv2, true);
+            	case UREM:
+            		return bvmgr.modulo(bv1, bv2, false);
             	case AND:
             		return bvmgr.and(bv1, bv2);
             	case OR:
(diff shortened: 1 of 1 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
```diff
@@ -127,7 +127,16 @@ public Formula encode(Formula e1, Formula e2, SolverContext ctx){
             	case UDIV:
             		return bvmgr.divide(bv1, bv2, this.equals(DIV));
             	case MOD:
+            		BooleanFormulaManager bmgr = ctx.getFormulaManager().getBooleanFormulaManager();
+            		BitvectorFormula rem = bvmgr.modulo(bv1, bv2, true);
+            		// Check if rem and bv2 have the same sign
+            		BooleanFormula cond = bvmgr.greaterThan(bvmgr.multiply(rem, bv2), bvmgr.makeBitvector(32, 1), true);
```
ThomasHaas, line 133: This always creates bv32, right? The number of bits should match the operands.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
```diff
@@ -127,7 +127,16 @@ public Formula encode(Formula e1, Formula e2, SolverContext ctx){
...
+            		// If they have the same sign, return the reminder, otherwise invert it
+            		bmgr.ifThenElse(cond, rem, bvmgr.negate(rem));
```
ThomasHaas, line 135: return is missing. This code falls through into the next switch case.

ThomasHaas (changes requested):

ThomasHaas:
I'm fine with the changes, but I think performing multiplication to check the signs of bit vectors is overkill, isn't it?
Is it even correct with overflows?
I think the easier and most correct version is to extract the top-most bit of both bv's and compare them for equality. That should also be the most trivial encoding for the solver.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
```diff
@@ -130,8 +130,10 @@ public Formula encode(Formula e1, Formula e2, SolverContext ctx){
             		BooleanFormulaManager bmgr = ctx.getFormulaManager().getBooleanFormulaManager();
             		BitvectorFormula rem = bvmgr.modulo(bv1, bv2, true);
             		// Check if rem and bv2 have the same sign
-            		BitvectorFormula srem = bvmgr.extract(rem, bvmgr.getLength(rem), bvmgr.getLength(rem), true);
-            		BitvectorFormula sbv2 = bvmgr.extract(rem, bvmgr.getLength(bv2), bvmgr.getLength(bv2), true);
+            		int rem_length = bvmgr.getLength(rem);
+            		int bv2_length = bvmgr.getLength(bv2);
+            		BitvectorFormula srem = bvmgr.extract(rem, rem_length-1, rem_length-1, true);
+            		BitvectorFormula sbv2 = bvmgr.extract(rem, bv2_length-1, bv2_length-1, true);
```
ThomasHaas, line 136: You extract twice from `rem`. The second should be `bv2`.

ThomasHaas (changes requested):

ThomasHaas (approved):
LGTM.

PR #224, opened 2022-04-02
Svcomp compilation

hernanponcedeleon:
This PR removes the duplicated `C -> boogie` compilation logic in the svcomp subproject and allows to easily pass different options to smack using an env variable.


ThomasHaas (approved):
I guess the configurability via config files (which we still not have) is not really needed for Smack/Clang. It makes sense to configure them using environment variables.

Issue #225, opened 2022-04-03
[BUG] Constant propagation


hernanponcedeleon:
The following benchmark shows there is a problem with constant propagation
```
❯ java -jar dartagnan/target/dartagnan-3.0.0.jar cat/svcomp.cat ../sv-benchmarks/c/pthread-divine/ring_1w1r-2.c --method=assume --bound=2
FAIL
❯ java -jar dartagnan/target/dartagnan-3.0.0.jar cat/svcomp.cat ../sv-benchmarks/c/pthread-divine/ring_1w1r-2.c --method=assume --bound=2 --program.processing.constantPropagation=false
UNKNOWN
```
but I still could not identify where exactly is the problem.

ThomasHaas:
Might be related to #329

hernanponcedeleon:
This issue is not reproducible in the current version of `development`
```
> java -jar dartagnan/target/dartagnan-3.1.1.jar cat/svcomp.cat ../benchmarks/ring_1w1r-2.c --method=assume --bound=2 --program.processing.constantPropagation=true
[26.11.2022] 11:29:33 [INFO] GitInfo.CreateGitInfo - Git branch: development
[26.11.2022] 11:29:33 [INFO] GitInfo.CreateGitInfo - Git commit ID: de28cf515802a9d4870019e711c8ab0b3f42bffe
UNKNOWN
```
```
> java -jar dartagnan/target/dartagnan-3.1.1.jar cat/svcomp.cat ../benchmarks/ring_1w1r-2.c --method=assume --bound=2 --program.processing.constantPropagation=false
[26.11.2022] 11:30:11 [INFO] GitInfo.CreateGitInfo - Git branch: development
[26.11.2022] 11:30:11 [INFO] GitInfo.CreateGitInfo - Git commit ID: de28cf515802a9d4870019e711c8ab0b3f42bffe
UNKNOWN
```

ThomasHaas:
So we can close this issue?

PR #226, opened 2022-04-03
225

hernanponcedeleon:
The bug described in #225 was due to calling `map.clear()` when resetting the map. The problem is that if we then merge the cleared map `m1` with another `m2`, we basically get `m2` and loose the fact that other values (the ones from the path we are clearing) are possible. Instead of clearing the map, we just set all values to TOP.


ThomasHaas:
The whole intention was to only get `m2` because there is no other possible path. I think this "fix" will only hide the bug by making the constant propagation super inaccurate.

ThomasHaas:
Since your loop is non-terminating there is indeed only a single path in the program that reaches the end of the original program and on that one we have `r0=0`. After processing of Dartagnan there are 2 paths for the `END_OF_T` label but those two will get merged resulting in `TOP` at that label (not that it realls matters there).

hernanponcedeleon:
Ok, I wanted to keep the example as simple as possible but you are right that this would not really be a problem if the loop is non terminating. However if you consider the similar scenario where
- the loop is terminating
- the program is such that you know the loop will be executed at least once
- you add an assertion `r0 == 1` at the end
The assertion should then be satisfied but the current analysis will still say that the value is zero

ThomasHaas:
I don't think the analysis will say that the value is `0`. Because if the loop is terminating, it has conditional jumps that leave the loop and all those jumps will accumulate their result into after the loop.

The idea of when to merge is quite simple. In our IR, there are exactly 2 different ways to reach some label: Either by a jump or by a direct predecessor. If the predecessor is a `goto`, however, that case can be excluded (that's why the clear happens) and we are left with only indirect predecessors via jumps.

Just to clarify the intention of why the map should get cleared. If we have
```
goto b1;
X
...
b1:
```
Then the code line `X` is either a `Label` or it is unreachable code. In the former case, all paths that reach `X` have to use a jump that was already processed by constant propagation. Whenever such a jump is processed, the values it propagates are merged into the value associated with `X`. In other words, the value associated with `X` is already the correct one as soon as we reach it (no merging with anything required).
In the case that the code is unreachable, we would have an empty map that accumulates dead assignments until the first non-dead label is reached in which case we merge. Here we get imprecise results at best because we merge some unreachable value with reachable values, potentially adding a few `TOP` results.

hernanponcedeleon:
This needs re-work so closing it for now.

PR #232, opened 2022-04-21
Format as program metadata

hernanponcedeleon:
This PR adds the format (litmus or bpl) as program metadata and provides a reliable and stable way of checking the format. Currently we are checking if the program has a "global" assertion to decide that the format is litmus, but this does not work anymore once we update inline assertions.


ThomasHaas (approved):
LGTM. But I'm wondering if the enum should be called `Format` because it suggests that the program is stored/represented differently. I guess that is true for the assertions of the program, so I'm not fully against it.
Otherwise, I would probably have called it `Source` or `SourceLang(uage)`.

PR #235, opened 2022-04-27
Compare expected with result in external tools

hernanponcedeleon:
The external tool unit tests were executing the corresponding tool, but never actually comparing the obtained result with the generated one. This PR adds this functionality.


ThomasHaas:
I would make `getResult` take the output parameter as input and avoid the `output` field as an instance member altogether.

ThomasHaas (approved):
LGTM

PR #236, opened 2022-04-28
Build on windows

xeren:
Adds the settings profile 'windows'.  While linux builds require the .so files of Z3 to be present and mac-os needs .dylib files, my machine successfully passed the test suites with just the .dll files.  This should improve the platform-independent development.

Tested on:
Edition: Windows 10 Home (64bit)
Version: 21H2
Processor: Intel Core i5 3450 @3.1GHz
Maven Version: 3.8.1

This PR also moves the coverage plugin setting into the purely module-specific section of the POM.  I was not able to find any discussion on why each profile had its own coverage settings, but suspect it may have some function.

### DIFF
--- a/dartagnan/pom.xml
+++ b/dartagnan/pom.xml
@@ -311,6 +311,78 @@
       </build>
     </profile>
 
+    <profile>
+      <activation>
+        <os>
+          <family>windows</family>
+        </os>
+      </activation>
+      <id>windows</id>
+      <dependencies>
+        <!-- Z3 dependencies (Windows) -->
+        <dependency>
+          <groupId>org.sosy-lab</groupId>
+          <artifactId>javasmt-solver-z3</artifactId>
+          <version>4.8.10</version>
+          <type>dll</type>
+          <classifier>libz3</classifier>
+        </dependency>
+        <dependency>
+          <groupId>org.sosy-lab</groupId>
+          <artifactId>javasmt-solver-z3</artifactId>
+          <version>4.8.10</version>
+          <type>dll</type>
+          <classifier>libz3java</classifier>
+        </dependency>
+      </dependencies>
+
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-dependency-plugin</artifactId>
+            <version>3.1.1</version>
+            <executions>
+              <execution>
+                <id>copy</id>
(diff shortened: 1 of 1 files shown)


dartagnan/pom.xml
```diff
@@ -417,6 +311,78 @@
...
+                <phase>validate</phase>
+                <goals>
+                  <goal>copy-dependencies</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <outputDirectory>${project.dependency.path}</outputDirectory>
+              <artifactItems>
+                <!-- Z3 native libraries (MacOS) -->
```
hernanponcedeleon, line 364: MacOS -> Windows

hernanponcedeleon (commented):
Why only z3 is set up for windows? [JavaSMT has also support for other solvers](https://github.com/sosy-lab/java-smt)

xeren:
To this date, JavaSMT supports four solvers on Windows:
- <a href="http://mathsat.fbk.eu/">MathSAT5</a> requires three DLLs to be available.  This should now be added in this PR.  However, it fails at 9 tests, `seqlock-6.bpl`, `chase-lev-5` and `chase-lev-6` in three methods each, because of missing support for a modulo operation.
- <a href="http://www.philipp.ruemmer.org/princess.shtml">Princess</a> is written in Scala, and executed entirely inside a JVM, thus should not require additional resources, but sosy-lab has still to deploy a maven artifact for this.
- <a href="https://ultimate.informatik.uni-freiburg.de/smtinterpol/">SMTInterpol</a> is written in Java, and also requires no additional configuration.  Tests still running, exceeding quadruple the time of Z3.
- <a href="https://github.com/Z3Prover/z3">Z3</a> requires two DLLs, as configured in this PR.  This is checked by github.

I switched the solver for testing by temporarily replacing all 12 occurrences of `org.sosy_lab.java_smt.SolverContextFactory.Solvers.Z3` in this project.

hernanponcedeleon:
Yes, in my experience Z3 is the best solver for our whole test suite, this is the reason is the default solver and why we probably should always use it in the CI (besides missing features in mathSAT, boolector does not support interfer theory, princess tends to consume too much memory and cvc4 segfaults in many cases).

I did not test this PR because I don't have a windows machine, but since you said you test it on your side and the configuration seems ok, I'll ho ahead an merge.

PR #237, opened 2022-05-01
Do not simplify atoms using signed comparisons

hernanponcedeleon:
The simplifications done by constant propagation are unsound (when using `ARCH_PRECISION != -1`, i.e. bit-vectors) for `Atom`s if the comparison is signed. 

This was causing problems with `linuxrwlock` if `ARCH_PRECISION == 64`.

This PR fixes this issue.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/COpBin.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/COpBin.java
@@ -58,6 +58,18 @@ public enum COpBin {
         }
     }
 
+    public boolean isSigned() {
+        switch(this){
+            case EQ: case NEQ:
+            case GTE: case LTE: case GT: case LT:
+            	return true;
+            case UGTE: case ULTE: case UGT: case ULT:
+            	return false;
+            default:
+                throw new UnsupportedOperationException(this + " cannot be inverted");
+        }
+    }
+
     public BooleanFormula encode(Formula e1, Formula e2, SolverContext ctx) {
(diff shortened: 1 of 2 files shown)


ThomasHaas:
Can you elaborate what exactly causes this problem? Disabling all optimization with signed comparison seems to be overkill.
I don't think that EQ and NEQ care about signedness at all, but you disable optimizations with them as well. If that is really necessary, then there is more likely a bug in the Simplifier code.
Also, if this only happens with BV, why not check the precision of the expression (or just `ARCH_PRECISION`) and still optimize integers?

I guess the problem is related to either of the following:
- `BigInteger` doesn't have a precision and hence we might accidentally treat negative values as positive. This can happen due to either having overflow (should be UB?) or due to having bit-operations that set the most significat bit (e.g., via `or`).
In `linuxwrlock` there shouldn't be such a problem though.
- If the problem is only under `ARCH_PRECISION=64` but not the true`ARCH_PRECISION=32` (the benchmark uses 32-bit ints), then some operations are likely performed with the wrong precision. The LLVM code will contain `.bv32` operations, even if we use `ARCH_PRECISION=64` and we have to be sure to consider that precision when simplifiying.

Maybe we should create our own value class that has two different implementations `Integer` and `BitVector`. The former is just a wrapper around `BigInteger`, the latter could be a `BigInteger` + precision (or `long` if we only care about 64 bits max), making sure that all the bv-specific behaviour is correctly modelled.

hernanponcedeleon:
This needs re-work, so closing it for now.

PR #238, opened 2022-05-04
Use isync instead of lwsync in compilation to power

hernanponcedeleon:


ThomasHaas (approved):
LGTM. One of the easiest reviews so far :P

PR #241, opened 2022-05-09
Use BigInteger.reminder() instead of BigInteger.mod() in

hernanponcedeleon:
`Chase-lev` under `RC11` used to failed sporadically. The reason was probable that we were using the wrong `BigInteger` method when combining values. I tested that benchmarks several (>20) times now and it does not fail. Without the changes of this PR it failed 1 out of ~5 times I tried. Thus it seems the problem is actually fixed.

@ThomasHaas the other day you managed to add a break point somewhere and detected that we had a rf-edge for two events not having the same address. I did not find the way of doing this. Can you please check if this does not happen anymore with the proposed changes?


ThomasHaas (approved):
I can confirm that we get consistent behaviour now; the semantics of our expression matches the semantics of the encoded formula.

PR #243, opened 2022-05-11
Imm

hernanponcedeleon:
This PR adds support for the intermediate memory model IMM described [here](https://plv.mpi-sws.org/imm/paper.pdf).

I added all litmus tests described in the paper. Example 3.10 is a special case. The paper shows a problem in the compilation from the promising semantics to ARMv8. The promising semantics forbids this behaviour and thus IMM should also. To handle this, IMM allows to types of RMW events, a weak and a strong version (page 69:3 describes this). We don't have such strong RMW events and thus the test uses the weak version which makes the behaviour observable (as expected according to IMM). Notice that this only means we lack support for the instruction, not that we are unsound. I manually added the tag (which otherwise we don't use) for this example and indeed we report that the program is safe. Since we don't really have a litmus parser for IMM and real C code will not generate this strong RMW event, I don't really see the point of adding such kind of events.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -73,6 +73,14 @@ private C11() {}
...
         public static final String MO_SC                = "SC";
 
+        public static String extractStoreMo(String cMo) {
+            return cMo.equals(C11.MO_ACQUIRE_RELEASE) ? C11.MO_RELEASE : cMo;
+        }
+
+        public static String extractLoadMo(String cMo) {
+            return cMo.equals(C11.MO_ACQUIRE_RELEASE) ? C11.MO_ACQUIRE : cMo ;
+        }
+
```
ThomasHaas, line 83: I think you need to handle the case of extracting `storeMo` from `acq` and vice versa.
I mean that `extractStoreMo(MO_ACQUIRE)` and `extractLoadMo(MO_RELEASE)` should probably result in `MO_RELAXED`.
Otherwise, the RMW compilation schemes end up producing acquire writes and release loads which don't make sense I think.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/local/RelCASDep.java
```diff
@@ -0,0 +1,30 @@
...
+import com.dat3m.dartagnan.wmm.utils.Tuple;
+import com.dat3m.dartagnan.wmm.utils.TupleSet;
+
+import static com.dat3m.dartagnan.wmm.relation.RelationNameRepository.CASDEP;;
+
+public class RelCASDep extends StaticRelation {
+
+    public RelCASDep(){
+        term = CASDEP;
+        forceDoEncode = true;
```
ThomasHaas, line 16: Why the `forceDoEncode`?

ThomasHaas (changes requested):

ThomasHaas (approved):
LGTM

PR #249, opened 2022-05-14
coLast variables

hernanponcedeleon:
coLast variables are also needed for liveness (not only for litmus tests) thus we encode them in both cases.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
@@ -49,10 +49,10 @@ public class VerificationTask {
     // Data objects
     private final Program program;
     private final Wmm memoryModel;
+    private final EnumSet<Property> property;
     private final WitnessGraph witness;
     private final Configuration config;
     private final Context analysisContext;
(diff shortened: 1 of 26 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
```diff
@@ -44,6 +48,16 @@
...
+    @Option(
+            name = OptionNames.PROPERTY,
+            description = "The property to check for: reachability (default), liveness, races.",
+            toUppercase=true)
+    private EnumSet<Property> property = EnumSet.of(Property.getDefault());
+
+	public EnumSet<Property> getProperty() { return property; }
+
+    // =====================================================================
+
```
ThomasHaas, line 60: `BaseOptions` already has Property defined. So just add a `setProperty` method to the `VerificationTaskBuilder` and let the constructor of `VerificationTask` take the property as an input parameter. This avoids the duplication of having two identical property options in two different objects.

ThomasHaas (changes requested):

ThomasHaas (approved):
LGTM

PR #251, opened 2022-05-18
No-optimisable tag

hernanponcedeleon:
This PR adds a new tag. Events having this tag should be neither removed (e.g. `DeadAssignmenetElimination`) nor simplified (e.g. `ConstanPropagation`). The following events have this tag
- any event in a litmus test (litmus tests are used to model a specific behaviour. If the effect of an optimisation would be desirable, the litmus test would already have this in its code),
- bound events,
- spinloops, and
- fake control-flow dependencies.

As as side effect, we can now apply more aggressive simplifications in the expression simplifier. Constant propagation takes care of not calling the expression simplifier if the event should not be optimised.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
@@ -23,6 +23,8 @@ public final class Tag {
     public static final String ASSERTION    = "ASS";
     public static final String BOUND   		= "BOUND";
     public static final String SPINLOOP   	= "SPINLOOP";
+    // Some events should not be optimized (e.g. fake dependencies) or delete (e.g. bounds)
+    public static final String NOOPT   		= "NOOOPT";
 
     // =============================================================================================
     // =========================================== ARMv8 ===========================================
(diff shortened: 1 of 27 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -23,6 +23,8 @@ private Tag() { }
     public static final String ASSERTION    = "ASS";
     public static final String BOUND   		= "BOUND";
     public static final String SPINLOOP   	= "SPINLOOP";
+    // Some events should not be optimized (e.g. fake dependencies) or deleted (e.g. bounds)
+    public static final String NOOPT   		= "NOOOPT";
```
ThomasHaas, line 27: A lot of Os :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -141,6 +141,7 @@ public List<Event> visitAtomicFetchOp(AtomicFetchOp e) {
         Store store = newRMWStoreExclusive(address, dummyReg, Tag.ARMv8.extractStoreMoFromCMo(mo), true);
         Label label = newLabel("FakeDep");
         Event fakeCtrlDep = newFakeCtrlDep(resultRegister, label);
+        fakeCtrlDep.addFilters(Tag.NOOPT);
```
ThomasHaas, line 144: Why not integrate the addition of the tag directly into the function `newFakeCtrlDep`?

ThomasHaas (changes requested):

ThomasHaas (approved):
LGTM

PR #252, opened 2022-05-20
Implement complement in filter and relation visitor

hernanponcedeleon:
This PR implements the complement of expressions (needed to update our LKMM) both in filter and relation visitor.

I tested this code in the feature branch where I'm updating the LKMM and it works as expected.


ThomasHaas (approved):
LGTM

PR #255, opened 2022-05-22
Unrolling

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/annotations/FunCall.java
```diff
@@ -1,10 +1,10 @@
-package com.dat3m.dartagnan.program.event.core;
+package com.dat3m.dartagnan.program.event.core.annotations;
 
 import com.dat3m.dartagnan.program.event.Tag;
 import com.dat3m.dartagnan.program.event.visitors.EventVisitor;
 
 
-public class FunCall extends Event {
+public class FunCall extends CodeAnnotation {
```
hernanponcedeleon, line 7: no need to add ANY as tags, it already has it from CodeAnnotation

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
```diff
@@ -79,74 +85,134 @@ public void run(Program program) {
...
+        }
+    }
 
-                if (bound == 1 || jump.hasFilter(Tag.SPINLOOP)) {
-                    Label target = (Label) jump.getThread().getExit();
-                    newPred = EventFactory.newGoto(target);
-                    newPred.addFilters(Tag.BOUND, Tag.NOOPT);
-                    if(jump.hasFilter(Tag.SPINLOOP)) {
-                    	newPred.addFilters(Tag.SPINLOOP, Tag.NOOPT);
+    // =================== new test code ========================
```
hernanponcedeleon, line 115: Remove comment if the code is stable

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
```diff
@@ -79,74 +85,134 @@ public void run(Program program) {
...
-                    // Needed for the current implementation of copy in If events
-                    //TODO: Is this needed anymore since we got rid of If events?
-                    if (bound != 1) {
-                        newPred = cur.getCopy();
-                    }
-                    predecessor.setSuccessor(newPred);
+                Map<Event, Event> copyCtx = new HashMap<>();
+                List<Event> copies = copyPath(loopBegin, loopBackJump, copyCtx);
+                ((Label)copyCtx.get(loopBegin)).setName(loopBegin.getName() + "_" + iterCounter);
+                //TODO: The following was testing code and needs to be updated for real usage.
```
hernanponcedeleon, line 167: what about this?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -57,7 +57,7 @@ public List<Event> visitJoin(Join e) {
         return eventSequence(
                 load,
                 AArch64.DMB.newISHBarrier(),
-                newJumpUnless(new Atom(resultRegister, EQ, IValue.ZERO), e.getLabel())
+                newJumpUnless(new Atom(resultRegister, EQ, IValue.ZERO), (Label) e.getThread().getExit())
```
hernanponcedeleon, line 60: shall we make `getExit()` directly return a label instead of always cast?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/CondJump.java
```diff
@@ -29,10 +30,9 @@ public CondJump(BExpr expr, Label label){
 
     protected CondJump(CondJump other) {
 		super(other);
-		this.label = other.label4Copy;
+		this.label = other.label;;
 		this.expr = other.expr;
-		Event notifier = label != null ? label : other.label;
-		notifier.addListener(this);
+		this.label.addListener(this);
```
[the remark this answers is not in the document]
ThomasHaas: I think quite a bit of code relies on the listener set of a `Label` containing all its associated jumps.
But I think it is reasonable to add this set only for labels and name it appropriately (e.g. `jumpSet`).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
ThomasHaas: It's a matter of whether we want special annotation events or not. This code is also the reason why I didn't create a PR for the branch yet. We need to discuss if we want to have annotation events and what their purpose should be. But we can always add them later if needed, so I'm fine with just getting rid of that code for now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
ThomasHaas: I think that is reasonable. But do all threads have a label at the end? I'm thinking of litmus tests and init threads.

hernanponcedeleon: Yes, see `ProgramBuilder:L52`

ThomasHaas: Ok, but during construction at least, it is not true that `thread.getExit` is a label. `Thread.append` does not maintain that invariant either. I guess it would make sense to create the exit event on thread construction and have the `append` method insert events before the exit event (basically, the exit event always gets moved to the end).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/CondJump.java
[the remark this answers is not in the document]
ThomasHaas: i removed the listeners. Only `Label` has a set `jumpSet` now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
ThomasHaas: I removed the annotation code.

PR #258, opened 2022-05-24
Use Linux.LOCK_READ in EventData.isLock()

hernanponcedeleon:
The last PR added new events to the RMW relation, but those changes were not reflected in the CAAT solver resulting in wrong results for tests using lock/unlock.
This PR solves. the issue.


ThomasHaas (approved):
LGTM

PR #259, opened 2022-05-25
Target

hernanponcedeleon:
Events coming from pthread related calls (e.g. `pthread_create` or `pthread_join`) need different memory orderings for e.g. `C11` or `LKMM`. This PR adds `C11` and `LKMM` as "compilation" targets (and removes the `NONE` one which used to be used for those two cases).


ThomasHaas:
Do you still add fences around the `pthread` events? I couldn't find that code anywhere (but neither could I find it in development?). If so, can can you move them to the compilation code now? If you don't add those fences, then I'm surprised that the release/acquire tagging is enough for synchronization.

hernanponcedeleon:
That code is still in a feature branch, and as you said, I'll move those fences to the visitor once I merge this PR into that branch. The code in the LKMM visitor in this PR does not have any real effect right now because all the code that allows to verify C code in the LKMM in currently only in that branch.

ThomasHaas:
Ok, that makes sense.

ThomasHaas (approved):
LGTM

PR #264, opened 2022-05-31
Lkmm2arch

hernanponcedeleon:
This PR implements a compilation mapping from the Linux kernel API to Power following the inline assembly from the LK repo (links are provided as comments in the code). It also adds an abstract compilation test. Right now this is only used as the base for LKMM2Power but once we have the other compilation mapping (LKMM2TSO and LKMM2ARM8) implemented or C11 litmus tests (for which we already have all the compilation mappings), all those will also extend this abstract test,

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
@@ -436,6 +436,38 @@ class VisitorPower extends VisitorBase implements EventVisitor<List<Event>> {
         );
 	}
 	
+	//	https://github.com/torvalds/linux/blob/master/arch/powerpc/include/asm/barrier.h
+	@Override
+	public List<Event> visitLoad(Load e) {
+		Register resultRegister = e.getResultRegister();
+		IExpr address = e.getAddress();
+		String mo = e.getMo();
+		
+        Load load = newLoad(resultRegister, address, "_rx");
+        Fence optionalMemoryBarrier = mo.equals(Tag.Linux.MO_ACQUIRE) ? Power.newSyncBarrier() : null;
+        
+        return eventSequence(
+                load,
+                optionalMemoryBarrier
+        );
+	}
+
+	//	https://github.com/torvalds/linux/blob/master/arch/powerpc/include/asm/barrier.h
+	@Override
+	public List<Event> visitStore(Store e) {
+		ExprInterface value = e.getMemValue();
+		IExpr address = e.getAddress();
+		String mo = e.getMo();
+
+        Store store = newStore(address, value, "_rx");
+        Fence optionalMemoryBarrier = mo.equals(Tag.Linux.MO_RELEASE) ? Power.newSyncBarrier() : null;
+        
+        return eventSequence(
+                optionalMemoryBarrier,
+                store
+        );
+	}
+	
 	//		https://github.com/torvalds/linux/blob/master/arch/powerpc/include/asm/barrier.h
 	@Override
 	public List<Event> visitFence(Fence e) {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
@@ -413,40 +413,6 @@ class VisitorPower extends VisitorBase implements EventVisitor<List<Event>> {
 	// all three methods have the same implementation.
 	
 	@Override
-	public List<Event> visitRMWOpReturn(RMWOpReturn e) {
-		Register resultRegister = e.getResultRegister();
-		IOpBin op = e.getOp();
-		IExpr value = (IExpr) e.getMemValue();
-		IExpr address = e.getAddress();
-		String mo = e.getMo();
-		
-        Register dummyReg = e.getThread().newRegister(resultRegister.getPrecision());
-        Local localOp = newLocal(dummyReg, new IExprBin(resultRegister, op, value));
-
-        // Power does not have mo tags, thus we use null
-        Load load = newRMWLoadExclusive(resultRegister, address, null);
-        Store store = newRMWStoreExclusive(address, dummyReg, null, true);
-        Label label = newLabel("FakeDep");
-        Event fakeCtrlDep = newFakeCtrlDep(resultRegister, label);
-
-        Fence optionalMemoryBarrierBefore = mo.equals(Tag.Linux.MO_MB) ? Power.newSyncBarrier()
-                : mo.equals(Tag.Linux.MO_RELEASE) ? Power.newLwSyncBarrier() : null;
-        Fence optionalMemoryBarrierAfter = mo.equals(Tag.Linux.MO_MB) ? Power.newSyncBarrier()
-        		: mo.equals(Tag.Linux.MO_ACQUIRE)? Power.newISyncBarrier() : null;
-
-        
-        return eventSequence(
-                load,
-                localOp,
-                optionalMemoryBarrierBefore,
-                store,
-                fakeCtrlDep,
-                label,
-                optionalMemoryBarrierAfter
-        );
-	};
-
-	@Override
 	public List<Event> visitRMWOp(RMWOp e) {
 		Register resultRegister = e.getResultRegister();
 		IOpBin op = e.getOp();
@@ -470,9 +436,43 @@ class VisitorPower extends VisitorBase implements EventVisitor<List<Event>> {
 
         
         return eventSequence(
+                optionalMemoryBarrierBefore,
                 load,
                 localOp,
+                store,
+                fakeCtrlDep,
+                label,
+                optionalMemoryBarrierAfter
+        );
+	};
+
+	@Override
+	public List<Event> visitRMWOpReturn(RMWOpReturn e) {
+		Register resultRegister = e.getResultRegister();
+		IOpBin op = e.getOp();
+		IExpr value = (IExpr) e.getMemValue();
+		IExpr address = e.getAddress();
+		String mo = e.getMo();
+		
+        Register dummyReg = e.getThread().newRegister(resultRegister.getPrecision());
+        Local localOp = newLocal(dummyReg, new IExprBin(resultRegister, op, value));
+
+        // Power does not have mo tags, thus we use null
+        Load load = newRMWLoadExclusive(resultRegister, address, null);
+        Store store = newRMWStoreExclusive(address, dummyReg, null, true);
+        Label label = newLabel("FakeDep");
+        Event fakeCtrlDep = newFakeCtrlDep(resultRegister, label);
+
+        Fence optionalMemoryBarrierBefore = mo.equals(Tag.Linux.MO_MB) ? Power.newSyncBarrier()
+                : mo.equals(Tag.Linux.MO_RELEASE) ? Power.newLwSyncBarrier() : null;
+        Fence optionalMemoryBarrierAfter = mo.equals(Tag.Linux.MO_MB) ? Power.newSyncBarrier()
+        		: mo.equals(Tag.Linux.MO_ACQUIRE)? Power.newISyncBarrier() : null;
+
+        
+        return eventSequence(
                 optionalMemoryBarrierBefore,
+                load,
+                localOp,
                 store,
                 fakeCtrlDep,
                 label,
@@ -504,9 +504,9 @@ class VisitorPower extends VisitorBase implements EventVisitor<List<Event>> {
 
         
         return eventSequence(
+                optionalMemoryBarrierBefore,
                 load,
                 localOp,
-                optionalMemoryBarrierBefore,
                 store,
                 fakeCtrlDep,
                 label,
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
@@ -10,6 +10,8 @@ import com.dat3m.dartagnan.program.event.core.*;
 import com.dat3m.dartagnan.program.event.lang.catomic.*;
 import com.dat3m.dartagnan.program.event.lang.linux.RMWCmpXchg;
 import com.dat3m.dartagnan.program.event.lang.linux.RMWFetchOp;
+import com.dat3m.dartagnan.program.event.lang.linux.RMWOp;
+import com.dat3m.dartagnan.program.event.lang.linux.RMWOpReturn;
 import com.dat3m.dartagnan.program.event.lang.linux.RMWXchg;
 import com.dat3m.dartagnan.program.event.lang.pthread.Create;
 import com.dat3m.dartagnan.program.event.lang.pthread.End;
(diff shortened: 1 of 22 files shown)


dartagnan/src/test/java/com/dat3m/dartagnan/compilation/AbstractCompilationTest.java
```diff
@@ -0,0 +1,147 @@
...
+        .withSolverTimeout(timeoutProvider.get())
+        .build(program1Provider.get(), wmm1Provider.get(), propertyProvider.get()));
+    protected final Provider<VerificationTask> task2Provider = Provider.fromSupplier(() ->
+    	VerificationTask.builder()
+    	.withConfig(Configuration.builder()
+    		.setOption(INITIALIZE_REGISTERS,String.valueOf(DO_INITIALIZE_REGISTERS))
+    		.build())
+    	.withTarget(targetProvider.get())
+    	.withSolverTimeout(timeoutProvider.get())
+    	.build(program2Provider.get(), wmm2Provider.get(), propertyProvider.get()));
```
ThomasHaas, line 106: It might be better to have an explicit method that creates the Provider instead of 2 lambdas.
Possibly something like
```
protected Provider<VerificationTask> createTaskProvider(Provider<Program> progProvider, Provider<Wmm> wmmProvider, Provider<Arch> targetProvider) {  ...  }
```
You could then call this method twice, using the appropriate other providers.

ThomasHaas (commented):
I haven't checked in detail if the compilation scheme is correct, but I guess we have the tests for that :).

dartagnan/src/test/java/com/dat3m/dartagnan/compilation/AbstractCompilationTest.java
```diff
@@ -0,0 +1,147 @@
...
+    // =================== Modifiable behavior ====================
+
+    protected abstract Provider<Arch> getSourceProvider();
+    protected abstract Provider<Arch> getTargetProvider();
+
+    protected Provider<Integer> getTimeoutProvider() {
+        return Provider.fromSupplier(() -> 0);
+    }
+
+    protected long getTimeout() { return 10000; }
```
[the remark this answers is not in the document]
hernanponcedeleon: I'm having some trouble to understand the code "wrapping" all our unit test infrastructure. AFAICS, the same problem happens in all our unit tests. I think `getTimeout` was introduced about the time of bb294cd6b1dcbfb06c5375f41ba9dc2215a4a0ed to showdown the test if the timeout expired. However I would expect that to be used with `ShutdownRule` which is never used in our code. I'm not really sure why we allow to configure `getTimeoutProvider()` and do not just make it call `getTimeout()` so we don't have a single point of configuration.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -285,4 +296,336 @@ public List<Event> visitDat3mCAS(Dat3mCAS e) {
...
+		
+        // Power does not have mo tags, thus we use null
+        Load load = newLoad(resultRegister, address, null);
+        Fence optionalMemoryBarrier = mo.equals(Tag.Linux.MO_ACQUIRE) ? Power.newLwSyncBarrier() : null;
+        
+        return eventSequence(
+                load,
+                optionalMemoryBarrier
+        );
+	}
```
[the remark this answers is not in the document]
hernanponcedeleon: Before the code of this PR, `VisitorPower` would rely on the default implementation from `EventVisitor` for `Load`s, which is just `visitEvent()` which in turn just returns `Collections.singletonList(load);`. If the code does not contain LKMM events, then both implementation would match. However I agree this might  be dangerous if we are not cautious. The most reasonable solution I see is to introduce a new event `LKMMLoad` (similar to how we have `AtomicLoad`) to be sure that the code from above only applies to LKMM events.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -285,4 +296,336 @@ public List<Event> visitDat3mCAS(Dat3mCAS e) {
...
+	public List<Event> visitFence(Fence e) {
+		String mo = e.getName();
+        Fence optionalMemoryBarrier = mo.equals(Tag.Linux.MO_MB) 
+        		|| mo.equals(Tag.Linux.MO_WMB) || mo.equals(Tag.Linux.MO_RMB)
+        		|| mo.equals(Tag.Linux.BEFORE_ATOMIC) || mo.equals(Tag.Linux.AFTER_ATOMIC) ?
+        		Power.newSyncBarrier() : null;
+        
+        return eventSequence(
+                optionalMemoryBarrier
+        );
```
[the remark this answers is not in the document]
ThomasHaas: No, C11 fences are `AtomicFences` and no plain fences. They have a special visitor method that gets called.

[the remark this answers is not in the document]
hernanponcedeleon: You are right. Then I think the most reasonable (to tackle this and the comment above) would be to add special `LKMMLoad`, `LKMMStore` and `LKMMFence` events to exactly mimic what we are doing with C11.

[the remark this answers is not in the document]
ThomasHaas: That is possible, but then you have to make sure that those also get used in the C-Litmus parser if you want to support compilation of C-Litmus to PPC.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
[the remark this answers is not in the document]
ThomasHaas: I don't think your code necessarily reduces to equivalent code if there are no LKMM events.
The difference is that the default implementation of `EventVisitor` will keep the load as is, while your implementation will create a new load. This new load may end up being different than the visited one. 
E.g., visiting an exclusive load will return a plain load. Also existing tags gets "deleted".

That being said, it is probably best to have `LKMMLoad` and friends as special events.

hernanponcedeleon:
> I haven't checked in detail if the compilation scheme is correct, but I guess we have the tests for that :).

While I'm quite confident about its correctness now, this is not bullet proof. There was a point where all tests were passing and I still had a bug in the semantics of cmpxchg which I found just because it introduced a deadlock in qspinlock.

dartagnan/src/test/java/com/dat3m/dartagnan/compilation/AbstractCompilationTest.java
```diff
@@ -82,47 +75,33 @@ protected Provider<Integer> getTimeoutProvider() {
...
-    protected final Provider<VerificationTask> task2Provider = Provider.fromSupplier(() ->
-    	VerificationTask.builder()
-    	.withConfig(Configuration.builder()
-    		.setOption(INITIALIZE_REGISTERS,String.valueOf(DO_INITIALIZE_REGISTERS))
-    		.build())
-    	.withTarget(targetProvider.get())
-    	.withSolverTimeout(timeoutProvider.get())
-    	.build(program2Provider.get(), wmm2Provider.get(), propertyProvider.get()));
+    protected final Provider<VerificationTask> task1Provider = createTask(program1Provider, wmm1Provider, propertyProvider, sourceProvider, () -> 1);    
+    protected final Provider<VerificationTask> task2Provider = createTask(program2Provider, wmm2Provider, propertyProvider, targetProvider, () -> 1);
```
ThomasHaas, line 84: Don't you need to set the option `INITIALIZE_REGISTERS` anymore? What was the purpose of this before?
If you keep it like this, I would use the qualified name `Providers.createTask` instead of using a static import.

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
```diff
@@ -60,11 +63,13 @@ public static Provider<Program> createProgramFromFile(Supplier<File> fileSupplie
...
+                                                        Supplier<Arch> targetSupplier, Supplier<Integer> boundSupplier, boolean initiRegs) {
     	return Provider.fromSupplier(() -> VerificationTask.builder().
+    	        withConfig(Configuration.builder()
+    	                .setOption(INITIALIZE_REGISTERS, String.valueOf(initiRegs))
+    	                .build()).
     			withTarget(targetSupplier.get()).
     			withBound(boundSupplier.get()).
-    			withSolverTimeout(timeoutSupplier.get()).
     			build(programSupplier.get(), wmmSupplier.get(), propertySupplier.get()));
     }
```
ThomasHaas, line 74: I would rather have this method take a `Provider<Configuration>` as input and then create the appropriate provider in the unit tests. It's a lot more general.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -285,4 +299,336 @@ public List<Event> visitDat3mCAS(Dat3mCAS e) {
...
+		IOpBin op = e.getOp();
+		IExpr value = (IExpr) e.getMemValue();
+		IExpr address = e.getAddress();
+		String mo = e.getMo();
+		
+        Register dummyReg = e.getThread().newRegister(resultRegister.getPrecision());
+        Local localOp = newLocal(dummyReg, new IExprBin(resultRegister, op, value));
+
+        // Power does not have mo tags, thus we use null
+        Load load = newRMWLoadExclusive(resultRegister, address, null);
```
ThomasHaas, line 410: `RMWOp` is the non-returning counterpart of `RMWOpReturn`, right? So technically, it should not have a `resultRegister` to begin with since it returns nothing (I guess `RMWOp` has a result register because it inherits it?).
I don't mind using the `resultRegister` as a dummy here, since no-one is gonna read it, but we should have at least a comment saying that this is supposed to not return anything and that `resultRegister` is just used as a dummy.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -285,4 +299,336 @@ public List<Event> visitDat3mCAS(Dat3mCAS e) {
...
+        Register dummy = resultRegister;
+		if(resultRegister == value){
+            dummy = e.getThread().newRegister(resultRegister.getPrecision());
+        }
+
+		Load load = newRMWLoadExclusive(dummy, address, null);
+        Local optionalUpdateReg = dummy != resultRegister ? newLocal(resultRegister, dummy) : null;
+        Store store = newRMWStoreExclusive(address, new IExprBin(dummy, e.getOp(), value), null, true);
+        Label label = newLabel("FakeDep");
+        Event fakeCtrlDep = newFakeCtrlDep(dummy, label);
```
ThomasHaas, line 482: I get that there can be issues when the written `value` expression coincides with the `resultRegister`.
Can a similiar issue happen when `address` coincides with `resultRegister` (does that even make sense?).
And why do the other `RMW` events do not take the possibility of `resultRegister == value` into account?

Lastly, I noticed that the naming `e.getMemValue()` is very misleading as it does not correspond to the value the operation will eventually write to memory! It might be better to rename this to just `e.getValue()` or `e.getOperandValue()`, but this is an issue orthogonal to this PR.

hernanponcedeleon: This is the same pattern that is being used in `visitorLKMM`. My understanding is that the other methods do not need to perform the check because they do not fetch the old value.

Not comparing with `address` can indeed generate problems (good catch!), and this is already a bug in `visitorLKMM`. The following program fails under LKMM when it should clearly be safe.
```
C C-atomic-fetch-simple-01

{
  atomic_t x[2] = {0,0};
}

P0(atomic_t *x) {
  int r0; int r1;
  r0 = 1;
  r0 = atomic_fetch_add(2,x+r0);
}

exists (0:r0 != 0 \/ x[1] != 2)
```
I also think it is not enough to compare `resultRegister == value`. We should probably use `value.getRegs().contains(resultRegister)` which is the one I used to solve the problem with `address` from above.

Yes, the `e.getMemValue()` is an unfortunately artifact of inheriting from MemEvent.

ThomasHaas: I think the other events do use `value` in some way after performing the load into `resultRegister`. That means the load can affect the `value`, which should never be intended.
Maybe we shouldn't bother with the check and always create dummy registers to just avoid any of this special reasoning? The only disadvantage should be that we end up with a lot of dummy registers. But then I'm wondering if we could even reuse the same dummy register (per thread) over and over again (this might negatively affect static analyses though).

Also, I noticed that `RMWAbstract` has its own member `value` but it overrides `getMemValue` to return it. I think it would be reasonable to just add a new getter and setter instead of reusing `getMemValue` there.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -404,10 +404,9 @@ public List<Event> visitRMWOp(RMWOp e) {
...
         Register dummyReg = e.getThread().newRegister(resultRegister.getPrecision());
-        Local localOp = newLocal(dummyReg, new IExprBin(resultRegister, op, value));
-
         // Power does not have mo tags, thus we use null
-        Load load = newRMWLoadExclusive(resultRegister, address, null);
+        Load load = newRMWLoadExclusive(dummyReg, address, null);
+        Local localOp = newLocal(dummyReg, new IExprBin(dummyReg, op, value));
         Store store = newRMWStoreExclusive(address, dummyReg, null, true);
         Label label = newLabel("FakeDep");
         Event fakeCtrlDep = newFakeCtrlDep(resultRegister, label);
```
ThomasHaas, line 412: The fake dependency should use the dummy reg.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -438,10 +437,9 @@ public List<Event> visitRMWOpReturn(RMWOpReturn e) {
 		String mo = e.getMo();
 		
         Register dummyReg = e.getThread().newRegister(resultRegister.getPrecision());
-        Local localOp = newLocal(resultRegister, new IExprBin(dummyReg, op, value));
-
         // Power does not have mo tags, thus we use null
         Load load = newRMWLoadExclusive(dummyReg, address, null);
+        Local localOp = newLocal(resultRegister, new IExprBin(dummyReg, op, value));
         Store store = newRMWStoreExclusive(address, resultRegister, null, true);
```
ThomasHaas, line 443: Here you might have an issue if `address` depends on `resultRegister`. You should probably update the dummy with the localOp, then perform the store and then set `resultRegister=dummy`.

ThomasHaas (approved):
I think it looks fine now.

hernanponcedeleon:
> I think it looks fine now.

I'll think twice next time I said "adding a new compilation mapping is trivial" 🤪

ThomasHaas:
> > I think it looks fine now.
> 
> I'll think twice next time I said "adding a new compilation mapping is trivial" 🤪

At least now you know about all the pitfalls. It will be easier next time, maybe even trivial 😄 .

PR #261, opened 2022-05-28
Witness

hernanponcedeleon:
This PR adds the option to generate a violation witness as a dot file.

PR #265, opened 2022-06-02
Do not merge events with the same Cline

hernanponcedeleon:
#261 took some code from the `cna-verification` where I added an option to decide if merging or not events in the witness graph. Since we agreed that merging events was not that useful at the end (and merging them cause problems like missing po edges), mi idea was to completely remove the merging. However I just remove the option but kept the old merging code. This PR completely remove the merging of events in the graph generation.


ThomasHaas (approved):
There is not much room to complain about this PR :P. Looks good to me.

PR #267, opened 2022-06-10
Lkmm2 arm

hernanponcedeleon:
This PR adds the compilation mappings from LKMM to ARMv8.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
@@ -318,7 +318,6 @@ class VisitorArm8 extends VisitorBase implements EventVisitor<List<Event>> {
 
 	// Following
 	// 		https://elixir.bootlin.com/linux/v5.18/source/arch/arm64/include/asm/atomic_ll_sc.h#L259
-	// Instead of using a comparison, it uses a XOR
 	@Override
 	public List<Event> visitRMWCmpXchg(RMWCmpXchg e) {
 		Register resultRegister = e.getResultRegister();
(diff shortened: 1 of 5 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -223,4 +239,291 @@ public List<Event> visitDat3mCAS(Dat3mCAS e) {
...
+	@Override
+	public List<Event> visitRMWCmpXchg(RMWCmpXchg e) {
+		Register resultRegister = e.getResultRegister();
+		IExpr address = e.getAddress();
+		ExprInterface value = e.getMemValue();
+		String mo = e.getMo();
+
+		Register dummy = e.getThread().newRegister(e.getResultRegister().getPrecision());
+        Label casEnd = newLabel("CAS_end");
+        CondJump branchOnCasCmpResult = newJump(new Atom(new IExprBin(dummy, IOpBin.XOR, (IExpr) e.getCmp()), NEQ, IValue.ZERO), casEnd);
```
ThomasHaas, line 331: Are you sure that we want to mimic the real compilation here? I think the `XOR` can harm performance substantially without affecting the semantics at all.

hernanponcedeleon: I think when there is inline assembly involved, we should be as close as possible to the real implementation. If performance is really an issue, I agree we should use a semantic equivalent, but cheaper, model (at the end of the day, we have several "modelling assumptions"). However I have been testing this quite a lot with real code from the kernel and performance is not an issue at all.

ThomasHaas: Just to make sure: Did you test this with integer encoding or bit vector encoding?
When using BV-encoding, there is no issue at all cause a XOR is cheap. If not, then this causes a conversion from IA to BV, which can always be dangerous.
In some SVCOMP benchmarks, just having a boolean negation implement via `XOR 1 `already caused serious issues in the past to the point. So I would be surprised if something like this doesn't happen here as well.

hernanponcedeleon: Fair question ... when I mentioned that I tested this with real code, I was always using the BV-encoding. Since we don't have any real code from the kernel not suing BVs (i.e. where only the compilation would introduce bit-precise reasoning), I did the following: I changed all other encodings of `IOpBin` operators to just use random integer operations (I know this is completely unsound, I just wanted to measure the performance impact of adding a single bit-precise operation) and test the compilation on qspinlock. You are right, the overhead is considerable (~10X) thus I will change the compilation to use `==` instead or `xor` and just keep a comment about the rationale of the difference between the real compilation and the encoding.

ThomasHaas (approved):
LGTM

PR #268, opened 2022-06-15
Missing parenthesis in IValue.hashCode()

hernanponcedeleon:
The missing parenthesis allow `IValue.hashCode()` to return zero which can (and does) result in a division by zero in `IExpBin.hashCode()`


ThomasHaas (approved):
LGTM.

PR #269, opened 2022-06-15
LKMM races

hernanponcedeleon:
This PR adds support to detect data races in the LKMM in all methods except refinement.
I marked the PR as a DRAFT so we can discuss how we want to handle this from the refinement side.


ThomasHaas:
If you add the `flag` syntax, why not use it as a general way to specify additional properties in general?
I mean, you could add negated versions of axioms in general (as consistency axioms).
And whenever you add `flag` in front of any axiom, then this axiom is considered a "property" to be checked for.

ThomasHaas:
I thought about this a little more, and I suggest to do this:
- Bring back the `negated` flag for `Axiom`s. 
- Possibly add an `inconsistent` method as counterpart of `consistent`. Its default implementation can just return the negation of ´consistent´. I think this will fail for encoding cyclicity (i.e, not acyclic), but we can throw a `NotSupportedException` for now.
- Add a `isProperty` flag to mark axioms as properties rather than consistency axioms.
- When encoding consistency, encode only axioms that are *not* flagged as ´isProperty`.
- When encoding properties, encode all axioms that are flagged as `isProperty`. Here it is important to conjoin all properties with `OR` instead of `AND`, because we want to find a violation of any property.
- Extend the syntax of our CAT with either `flag axiom` or `property axiom`. The former follows herd7 and specifies _bad behavior_ while the latter would specify _good behaviour_ (and hence would get negated in the encoding). The second possibility would more closely resemble the way assertions work (they indicate what should hold, i.e. good behavior).

As a side note: In the future we will have to be careful to treat property axioms differently than consistency axioms in static analyses. It is, e.g., incorrect to propagate knowledge from properties downwards since unlike consistency axioms, they don't have to hold! Unless we only check a single property during the verification, in which case we can exploit property-based information. This may come in handy when introducing multi-threading (running an instance per property to check for).

hernanponcedeleon:
I agree with most of what you wrote. However I do not see a use case for the `inconsistent` method and `property axiom`. Isn't `property axiom` the same as `flag ~axiom`?

At the theory level we have 6 axioms (`empty`, `irreflexive`, `acyclic`, `not-empty`, `reflexive`, `cyclic`). Implementation-wise we will have 3 with a corresponding `negated` field. We already have an implementation for `consistent` for the first 3; the following 2 is just the negation of the counter-part; for `cyclic.consistent()` we can use the encoding for detecting cycles we used to have for porthos.

On top of that, for some of them we will set a `flag` field. Not-flagged axioms are used as consistency filters by calling `consistent()` in the `WmmEncoder`. Flagged axioms are used to encode a property by calling `consistent()` from the `PropertyEncoder`.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/RefinementTask.java
```diff
@@ -123,7 +123,7 @@ private Wmm createDefaultWmm() {
 	        repo.addRelation(porfco);
 	        Relation porfcofr = new RelUnion(porfco, fr);
 	        repo.addRelation(porfcofr);
-	        baseline.addAxiom(new Acyclic(porfcofr));
+	        baseline.addAxiom(new Acyclic(porfcofr, false, false));
```
ThomasHaas, line 126: It might be better to keep the simple constructor as well as the new complex constructor. I don't have a strong opinion about this though.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Axiom.java
```diff
@@ -47,13 +51,21 @@ public Relation getRelation() {
         return rel;
     }
 
+    public boolean isFlagged() {
+        return flag;
+    }
+
```
ThomasHaas, line 57: I think the name `flagged` is non-descriptive. We should find a better name to indicate the meaning of this.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/Relation.java
```diff
@@ -183,6 +183,12 @@ public final BooleanFormula getSMTVar(Event e1, Event e2, SolverContext ctx) {
         return getSMTVar(new Tuple(e1, e2), ctx);
     }
 
+    public BooleanFormula getSMTCycleVar(Tuple edge, SolverContext ctx) {
+        return !getMaxTupleSet().contains(edge) ?
+        		ctx.getFormulaManager().getBooleanFormulaManager().makeFalse() :
+                edge(getName() + "-cycle", edge.getFirst(), edge.getSecond(), ctx);
+    }
+
```
ThomasHaas, line 191: Isn't this just `getSMTVar` with a slightly different name?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
```diff
@@ -126,20 +126,37 @@ public BooleanFormula consistent(SolverContext ctx) {
...
+            for(Event e : rel.getEncodeTupleSet().stream().map(t -> t.getFirst()).collect(Collectors.toSet())){
+            	BooleanFormula in = bmgr.makeFalse();
+            	for(Tuple pre : rel.getEncodeTupleSet().getBySecond(e)) {
+            		in = bmgr.or(in, rel.getSMTCycleVar(pre, ctx));
+            	}
+            	BooleanFormula out = bmgr.makeFalse();
+            	for(Tuple post : rel.getEncodeTupleSet().getByFirst(e)) {
+            		out = bmgr.or(out, rel.getSMTCycleVar(post, ctx));
+            	}
+            	enc = bmgr.and(enc, bmgr.implication(Utils.cycleVar(rel.getName(), e, ctx), bmgr.and(in , out)));
```
ThomasHaas, line 143: The meaning of `rel.getSMTCycleVar` is never encoded (i.e. its derivation rules). 
Shouldn't you just use `rel.getSMTVar` here?

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Axiom.java
hernanponcedeleon: I think we should use `flag` to be consistent with herd. However I do agree the name is not very descriptive so I added some documentation

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/Relation.java
hernanponcedeleon: Yes, but the name is not a parameter, it is extracted from the `Relation` object.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
hernanponcedeleon: The rest of the cycle encoding was in the next page of the porthos paper and I missed it XD.

> Shouldn't you just use rel.getSMTVar here?

No, that one should refer to the cycle edges. The new parts refer to the original relation edge.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Axiom.java
ThomasHaas: I don't think the internal name we give to the variable has to match the syntax. We can keep `flag` as syntax in the`.cat` file and still use some more descriptive name internally. But I'm also fine with the comment for now.

hernanponcedeleon:
I removed the [DRAFT] tag because the code should in theory work (*) with refinement. Data-races (or any other flagged axiom) is now encoded as any other property so the solver should work (potentially having to pay the price of cutting most of the WMM).

(*) The problem is that right now refinement does not allow to cut recursive relations. However this is independent of this PR and it should be tackled independently.

I added a new property options (CAT_OPTIONS) that allows to easily switch-off flagged axioms (this might considerable improve performance of e.g. LKMM). Initially I thought this could be use to avoid the current problem with refinement, but cutted relations are computed independently of flagged relations not being encoded.

ThomasHaas:
I can take a look into the cutting and/or avoiding flagged relations for CAAT.

hernanponcedeleon:
@ThomasHaas I'm ok with your latest changes, so if you are ok with the rest, I'll merge once the CI passes.

ThomasHaas (approved):
You can go ahead and merge

PR #270, opened 2022-06-16
RefinementSolver

ThomasHaas:
- Now `RefinementSolver` tracks time spent in generating refinement clauses.
- Improved time spent in generating refinement clauses by avoiding duplicate clauses.
  This does not affect solving times, because the solver already detected duplicate clauses it seems (albeit more expensively).
- Minor change in reusing refinement clauses for the bound check.


hernanponcedeleon:
LGTM

PR #271, opened 2022-06-20
C11

hernanponcedeleon:
This PR makes several changes.
- it divides the `dartagnan-expected` file into several ones. This allows to run the same tests (e.g. those in `/litmus/C11`) under different memory models (e.g. `RC11`, `IMM`, `C11 (*)`)
- it extends the `LitmusC` parser which before was used only for LKMM tests and can now handle C11-atomics
- (*) adds `c11.cat`. This file uses fixed points in the  RHS of differences and thus we cannot trust the results (because we currently compute arbitrary fixed-points instead of least ones). However the results on litmus tests match those with herd and this allows to also test different compiler schemes correctness (see below), thus I don't see a problem in keeping the cat file and the C11-litmus unit test, but we should not claim we can handle C11
- adds an option to specify if Power compilation should be sync-leading or sync-trailing. This shows that one compilation is broken from C11 to Power
- our previous compilation scheme was following the RC11 paper. This was not the real compilation for C11 (this PR fixes this). The compilation was needed to avoid OOTA in RC11, thus we add an option to chose that scheme.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -37,7 +37,14 @@
 
 class VisitorArm8 extends VisitorBase implements EventVisitor<List<Event>> {
 
-	protected VisitorArm8() {}
+	// If the source WMM does not allow OOTA behaviors (e.g. RC11)
+	// we need to strength the compilation following the paper
+	// "Outlawing Ghosts: Avoiding Out-of-Thin-Air Results"
+	private final boolean nooota;
```
ThomasHaas, line 43: Again, this is very misleading. "rc11Compilation" or "strongCompilation" are better fits for this.

dartagnan/src/test/java/com/dat3m/dartagnan/benchmarking/DartagnanRC1.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+import static com.dat3m.dartagnan.utils.ResourceHelper.TEST_RESOURCE_PATH;
+import static com.dat3m.dartagnan.utils.Result.*;
+import static com.dat3m.dartagnan.configuration.Arch.*;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class DartagnanRC1 extends AbstractCTest {
```
ThomasHaas, line 26: Typo (RC11).

dartagnan/src/test/java/com/dat3m/dartagnan/compilation/C112Aarch64Test.java
```diff
@@ -0,0 +1,38 @@
...
+import com.dat3m.dartagnan.utils.rules.Providers;
+import com.dat3m.dartagnan.wmm.Wmm;
+import com.dat3m.dartagnan.configuration.Arch;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+
+@RunWith(Parameterized.class)
+public class C112Aarch64Test extends AbstractCompilationTest {
```
ThomasHaas, line 13: `C11ToAarch64Test` might be a better name (`C112` looks odd).

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionNames.java
```diff
@@ -12,8 +12,8 @@ public class OptionNames {
 	public static final String VALIDATE = "validate";
 	
 	// Compilation Options
-	public static final String NOOOTA = "compilation.nooota";
-	public static final String SYNCSCHEME = "compilation.syncScheme";
+	public static final String USE_RC11_TO_ARCH_SCHEME = "compilation.rc11ToArch";
+	public static final String USE_C_TO_POWER_SCHEME = "compilation.cToPower";
```
ThomasHaas, line 16: I think here the `USE` prefix is not right, since its not a boolean value. I would leave it at `C_TO_POWER_SCHEME`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AtomicityPropagation.java
```diff
@@ -73,6 +71,7 @@ public Event visitMemEvent(MemEvent e) {
 			// Accesses to an atomic object without a concrete mo are considered SC
     		if(atomics.contains(e.getAddress()) && e.canRace()) {
     			e.addFilters(Tag.C11.ATOMIC, Tag.C11.MO_SC);
+    			e.removeFilters(Tag.C11.NONATOMIC);
     		}
```
ThomasHaas, line 75: I don't think this is enough. We have this special `mo` member in `MemEvent` which should get changed. Otherwise `e.canRace` will continue to return `true`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -30,24 +30,24 @@
...
 	// The paper does not defines the mappings for RMW but we derive them
 	// using the same pattern as for Load/Store	
-	private final SyncScheme scheme;
+	private final PowerScheme cToPowerScheme;
 	// If the source WMM does not allow OOTA behaviors (e.g. RC11)
 	// we need to strength the compilation following the papers
 	// "Repairing Sequential Consistency in C/C++11"
 	// "Outlawing Ghosts: Avoiding Out-of-Thin-Air Results"
-	private final boolean nooota; 
+	private final boolean useRC11Scheme;
```
ThomasHaas, line 46: I think the comment can get updated. LKMM, for example, seems to successfully use dependencies on the language level to avoid OOTA.
What is true is that *non-dependency* tracking models may need to use the strong version of `noOOTA`.
So maybe say that `RC11` is not dependency-tracking and hence uses a strong definition of `noOOTA`. This requires stronger compilation schemes to hardware.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemEvent.java
```diff
@@ -68,6 +68,11 @@ public String getMo(){
         return mo;
     }
 
+    public void setMo(String mo){
+        this.mo = mo;
+        addFilters(mo);
+    }
```
ThomasHaas, line 74: I think it is safer to remove the old `mo` from the `filters` before setting the new `mo`.
This way, calling code doesn't have to remove any filters manually (e.g. `AtomicityPropagation`).
Also, since we allow `NULL` values, you should check that the new value is not null before adding it to the filters.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemEvent.java
hernanponcedeleon: The question is: do we want to allow calling `e.setMo(null)` (meaning we would guard `addFilters` by the `!= null` check) or not (and thus use `Preconditions.checkNotNull`)?

ThomasHaas: The problem that I see is that the default value of `mo` is `null` which seems to be correctly treated in our code.
So it suggests that `null` is valid. In principle, I'm all for removing `null`, but would you then put e.g. `NONATOMIC` as the default value? Or would you keep the default value `null`, but won't allow the value to be reset to `null` once it was changed?

ThomasHaas: I'm fine with that

ThomasHaas: You might still want to do an `null`-check on the old mo-value before removing it from the filters.
I think right now it may not be a problem, but some set implementations may throw exceptions when trying to remove `null`.

ThomasHaas (approved):
LGTM.

PR #272, opened 2022-06-25
Removed unnecessary value constraints on non-executed events

ThomasHaas:
Fixed special case that caused some value constraints to be active even for non-executed events.


hernanponcedeleon:
LGTM

PR #273, opened 2022-06-29
Make the output of litmus test be similar to the one of herd7

hernanponcedeleon:
Herd7 reports `Ok` or `No` only based on the outcome of the (reachability) condition, but not if there are other violations like data races. Thus if a program does not violate safety but it has a data race, the output should be something like
```
Flag data-race
Condition exists X
No
```
This PR implements this.

### DIFF
--- a/ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
+++ b/ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
@@ -18,6 +18,7 @@ import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.common.configuration.Configuration;
 import org.sosy_lab.common.log.BasicLogManager;
 import org.sosy_lab.java_smt.SolverContextFactory;
+import org.sosy_lab.java_smt.api.Model;
 import org.sosy_lab.java_smt.api.ProverEnvironment;
 import org.sosy_lab.java_smt.api.SolverContext;
 import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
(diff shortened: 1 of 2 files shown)


ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -101,7 +105,15 @@ private void run(){
                     }
                     // Verification ended, we can interrupt the timeout Thread
                     t.interrupt();
+            		for(Axiom ax : wmm.getAxioms()) {
+                		if(ax.isFlagged() && TRUE.equals(prover.getModel().evaluate(CAT.getSMTVariable(ax, ctx)))) {
+                			verdict += "Flag " + (ax.getName() != null ? ax.getName() : ax.getRelation().getName());
+                		}                			
+            		}
```
ThomasHaas, line 112: This should be in the `buildVerdict` method. See the comment there.

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -101,7 +105,15 @@ private void run(){
...
                     buildVerdict(result);
+            		if(TRUE.equals(prover.getModel().evaluate(REACHABILITY.getSMTVariable(ctx)))) {
+            			safetyViolationFound = true;
+            		}
```
ThomasHaas, line 116: It seems that you store the `safetyViolationFound` before building the verdict, but reference it inside `buildVerdict`.
So it will always have its default value when building the verdict, no?

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -113,9 +125,9 @@ private void run(){
 
     private void buildVerdict(Result result){
         StringBuilder sb = new StringBuilder();
-        sb.append("Condition ").append(program.getAss().toStringWithType()).append("\n");
-        sb.append(program.getFormat().equals(LITMUS) ? result == FAIL ? "Ok" : "No" : result).append("\n");
-        verdict = sb.toString();
+        sb.append("\n").append("Condition ").append(program.getAss().toStringWithType()).append("\n");
+        sb.append(program.getFormat().equals(LITMUS) ? safetyViolationFound ? "Ok" : "No" : result).append("\n");
+        verdict += sb.toString();
```
ThomasHaas, line 130: This method is the only one relying on `safetyViolationFound`, so it should be a parameter rather than a member variable. Also, why don't we just pass the `prover` object (or just the `model`) and let this method compute the whole verdict, as its name suggests?
It could check the flagged axioms, the safety violation and even liveness from the model and produce whatever it needs to produce.

ThomasHaas (changes requested):

ThomasHaas (approved):
LGTM

PR #274, opened 2022-07-01
Output folder and as an env variable

hernanponcedeleon:
This PR allows to set the output folder (where log, bpl, csv, etc files are created) using environment variable `DAT3M_OUTPUT` rather than fixing it to be `$DAT3M_HOME/output`.


ThomasHaas (approved):
LGTM

PR #276, opened 2022-07-07
Liveness fix

hernanponcedeleon:
This PR solves two problems with the liveness encoding based on two wrong assumptions the code was making
- the old `markAnnotatedSpinLoops` assumed a `LoopEnd` is followed by a buckjump
- the old `encodeLiveness` assumed each spinloop has a single exit

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/FindSpinLoops.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/FindSpinLoops.java
@@ -10,7 +10,7 @@ import com.dat3m.dartagnan.program.event.core.Label;
 import com.dat3m.dartagnan.program.event.core.MemEvent;
 import com.dat3m.dartagnan.program.event.core.utils.RegReaderData;
 import com.dat3m.dartagnan.program.event.core.utils.RegWriter;
-import com.dat3m.dartagnan.program.event.lang.svcomp.LoopEnd;
+import com.dat3m.dartagnan.program.event.lang.svcomp.LoopStart;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/FindSpinLoops.java
```diff
@@ -63,23 +63,31 @@ public void run(Program program) {
...
-        	    // for (int tmp = (__VERIFIER_loop_begin(), 0); __VERIFIER_spin_start(),  \
-        	    //     tmp = cond, __VERIFIER_spin_end(!tmp), tmp;)
-        		Event spinloop = curr.getSuccessors().stream().filter(e -> e instanceof CondJump && ((CondJump)e).isGoto()).findFirst().get();
-        		spinloop.addFilters(Tag.SPINLOOP, Tag.NOOPT);
-                ((CondJump)spinloop).getLabel().addFilters(Tag.SPINLOOP, Tag.NOOPT);
-        		spinloops++;
-        	}
+            // Find start of spinloop.
+            if (curr.getSuccessor() instanceof LoopStart) {
+                Label label = (Label) curr;
```
ThomasHaas, line 75: This only works if the IR has the form
```
loopLabel:
#LoopStart
```
This might be the case when using the `await_while` construct but I think it is a bit risky. If the user explicitly places a `__VERIFIER_spin_start` not directly after the label, the code will trigger a `ClassCastException`.
I think we should at least check if `curr` is a `Label`, and if it is not, ignore the `LoopStart` and log a warning that the `LoopStart` event is misplaced.

Also, wouldn't it be nicer to have a single `__VERIFIER_spin_loop` that is placed somewhere before the start of the spinloop (like `__VERIFIER_loop_bound`)? Cause right now, I think it is impossible to specify a spinloop without using `await_while`, which might also be a good thing...

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/FindSpinLoops.java
hernanponcedeleon: I agree about the possibly unsafe cast. I'll fix that.

> Also, wouldn't it be nicer to have a single `__VERIFIER_spin_loop` that is placed somewhere before the start of the spinloop (like `__VERIFIER_loop_bound`)? Cause right now, I think it is impossible to specify a spinloop without using `await_while`, which might also be a good thing...

The purpose of adding the three `__VERIFIER_X` methods related to spinloops was to support this annotation in a way which is consistent with GenMC and thus all benchmarks from the DRC would work out of the box. Since right now we don't have other use cases (where the general idea you propose would work better) and changing this would not work with e.g. the CNA code, I think we should keep using this one for the moment.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/FindSpinLoops.java
```diff
@@ -72,6 +72,10 @@ private void markAnnotatedSpinLoops(Thread t){
         while (curr != null) {
             // Find start of spinloop.
             if (curr.getSuccessor() instanceof LoopStart) {
+            	if(!(curr instanceof Label)) {
+            		logger.warn("LoopStart does not match expected use (ot should be followed by a Label)");
```
ThomasHaas, line 76: "It should be preceeded by a label".

PR #278, opened 2022-07-12
Benchmarking

hernanponcedeleon:
This PR contains the following changes
- new unit tests for evaluating `cutting`
- new scripts for generating the csv files for each plot and the script to create figures from such files
- new unit tests for litmus in the benchmarking folder. This introduces some redundancy, but it allows us to decouple the configurations for the tests that run in the CI than those we want for benchmarking (similar to how we currently do for the C benchmarks). This is needed because e.g. in the CI we want to run only the `two solvers` instead of `refinement` to be able to test data races
- some small refactoring


ThomasHaas (approved):
Looks good to me.
I haven't checked all the details, but since the changes are very isolated (a single new benchmarks package), there shouldn't be any problems.

PR #279, opened 2022-07-14
Benchmarking

hernanponcedeleon:
This PR splits the benchmarking tests further to avoid constant problems with maven (due to trying to run tests in parallel).


hernanponcedeleon:
I agree that is a good idea if we decide to stick to the idea of using JUnit test for benchmarking (right now I am not getting any more of the common problems, so for the time being seems safe to keep using this).

PR #283, opened 2022-07-19
Fix jump condition after VERIFIER_assert

hernanponcedeleon:
We add jumps after assertions to terminate an execution as soon as a violation is found. However, the condition of the jump after a `__VERIFIER_assert` was of the following form
```
assert_X <- exp
if(assert_X != exp) goto END
```
making the jump not to be taken ever.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -3,29 +3,30 @@
...
+    public static final String LABEL        	= "LB";
+    public static final String CMP          	= "C";
+    public static final String IFI          	= "IFI";	// Internal jump in Ifs to goto end 
+    public static final String JUMP    			= "J";
+    public static final String VISIBLE      	= "V";
+    public static final String REG_WRITER   	= "rW";
+    public static final String REG_READER   	= "rR";
+    public static final String ASSERTION    	= "ASS";
+    public static final String BOUND   			= "BOUND";
+    public static final String EARLYTERMINATION	= "BOUND";
```
ThomasHaas, line 25: You didn't change the tag.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -200,7 +201,8 @@ class SpinLoop {
         BooleanFormula atLeastOneStuck = bmgr.makeFalse();
         for (Thread t : program.getThreads()) {
             BooleanFormula isStuck = isStuckMap.getOrDefault(t, bmgr.makeFalse());
-            BooleanFormula isDone = t.getCache().getEvents(FilterBasic.get(Tag.BOUND)).stream()
+            BooleanFormula isDone = t.getCache().getEvents(FilterUnion
+            		.get(FilterBasic.get(Tag.BOUND), FilterBasic.get(Tag.EARLYTERMINATION))).stream()
```
ThomasHaas, line 205: I think you can remove the check for `Tag.BOUND` cause you added the new tag to all bound events.

ThomasHaas (commented):
There are 3 different points that generate assertions?
Is there a way to refactor this? At least the two occurrences in `VisitorBoogie` can be refactored into a single method I believe.
Also, `Stdprocedures.__assert` is yet another method that generates assertions, but unlike the other 3 cases, it does not add a terminating jump. Is that an oversight?

PR #284, opened 2022-07-20
Ask for model only when result is FAIL

hernanponcedeleon:
`ReachabilityResult.buildVerdict()` calls `solver.evaluate()` even when there is no model (i.e. `result != FAIL`) which causes an exception.

### DIFF
--- a/ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
+++ b/ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
@@ -126,7 +126,9 @@ public class ReachabilityResult {
 		// TODO We might want to output different messages once we allow to check LIVENESS from the UI
 		sb.append("Condition ").append(program.getAss().toStringWithType()).append("\n");
 		sb.append(program.getFormat().equals(LITMUS) ? (model != null && TRUE.equals(model.evaluate(REACHABILITY.getSMTVariable(ctx)))) ? "Ok" : "No" : result).append("\n");
-		model.close();
+		if(model != null) {
+			model.close();			
+		}
         verdict = sb.toString();
     }
 
--- a/ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
+++ b/ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
@@ -18,10 +18,10 @@ import org.sosy_lab.common.ShutdownManager;
 import org.sosy_lab.common.configuration.Configuration;
 import org.sosy_lab.common.log.BasicLogManager;
 import org.sosy_lab.java_smt.SolverContextFactory;
-import org.sosy_lab.java_smt.api.Model;
 import org.sosy_lab.java_smt.api.ProverEnvironment;
 import org.sosy_lab.java_smt.api.SolverContext;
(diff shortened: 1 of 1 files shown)


ThomasHaas (commented):
Shouldn't the call to `prover.getModel` already throw an exception?
I think you might want to pass the prover to `buildVerdict` and only extract the model if the result was `FAIL`.

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -114,18 +114,18 @@ private void run(){
...
-        		if(ax.isFlagged() && TRUE.equals(m.evaluate(CAT.getSMTVariable(ax, ctx)))) {
+        		if(ax.isFlagged() && TRUE.equals(prover.getModel().evaluate(CAT.getSMTVariable(ax, ctx)))) {
         			sb.append("Flag " + (ax.getName() != null ? ax.getName() : ax.getRelation().getName())).append("\n");
         		}
     		}
         }
 		// TODO We might want to output different messages once we allow to check LIVENESS from the UI
 		sb.append("Condition ").append(program.getAss().toStringWithType()).append("\n");
-		sb.append(program.getFormat().equals(LITMUS) ? (result.equals(FAIL) && TRUE.equals(m.evaluate(REACHABILITY.getSMTVariable(ctx)))) ? "Ok" : "No" : result).append("\n");
+		sb.append(program.getFormat().equals(LITMUS) ? (result.equals(FAIL) && TRUE.equals(prover.getModel().evaluate(REACHABILITY.getSMTVariable(ctx)))) ? "Ok" : "No" : result).append("\n");
```
ThomasHaas, line 128: I would make sure to call `prover.getModel` only once and store it in a variable (or null, if there is model).
It is not clear if the underlying solver reconstructs the model on each call.

Also, models should get closed properly (`model.close`).

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -116,16 +117,16 @@ private void run(){
...
+    	for(Axiom ax : wmm.getAxioms()) {
+        	if(ax.isFlagged() && model != null && TRUE.equals(model.evaluate(CAT.getSMTVariable(ax, ctx)))) {
+        		sb.append("Flag " + (ax.getName() != null ? ax.getName() : ax.getRelation().getName())).append("\n");
+        	}
+    	}
 		// TODO We might want to output different messages once we allow to check LIVENESS from the UI
 		sb.append("Condition ").append(program.getAss().toStringWithType()).append("\n");
-		sb.append(program.getFormat().equals(LITMUS) ? (result.equals(FAIL) && TRUE.equals(prover.getModel().evaluate(REACHABILITY.getSMTVariable(ctx)))) ? "Ok" : "No" : result).append("\n");
+		sb.append(program.getFormat().equals(LITMUS) ? (model != null && TRUE.equals(model.evaluate(REACHABILITY.getSMTVariable(ctx)))) ? "Ok" : "No" : result).append("\n");
+		model.close();
```
ThomasHaas, line 129: You need a null-check here.

ThomasHaas (approved):
LGTM

PR #286, opened 2022-07-22
Static read modify write

xeren:
Attempts to further populate 'conditional-must' tuples in `rmw` and `rcu-rscs`.


ThomasHaas:
Your fix is backwards :).
There shouldn't be non-`MemEvents` in the tuple sets, and in particular, no jump events (they are not visible).
It seems that the jumps created by `pthread_mutex_lock` events are tagged by `LOCK` but the code in `RelRMW` assumes that `LOCK`-tagged events are loads only (filtering additionally by `Tag.READ` should do the trick).
This bug seems to be there for a long time, but it didn't cause problems these extra edges where never part of the `encodeTupleSet` and hence got skipped in the encoding.
However, your changed code tries to encode the full `maxTupleSet` instead of the `encodeTupleSet`, triggering an exceptions on the errouneously added rmw-edges.
That being said, for all non-exclusive rmw accesses encoding just the `encodeTupleSet` is sufficient, however, for exclusive events we need (or will need in the future) to encode `rmw` edges even if they are not important for consistency (in particular, they are not in the encodeSet).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
```diff
@@ -23,58 +23,46 @@ public RelCrit(){
...
-                            }
-                            for(Event otherUnlock : thread.getCache().getEvents(FilterBasic.get(Tag.Linux.RCU_UNLOCK))){
-                                if(lock.getCId() < otherUnlock.getCId() && otherUnlock.getCId() < unlock.getCId()){
-                                    relation = bmgr.and(relation, bmgr.not(this.getSMTVar(lock, otherUnlock, ctx)));
-                                }
-                            }
-                            enc = bmgr.and(enc, bmgr.equivalence(this.getSMTVar(tuple, ctx), relation));
-                        }
-                    }
+        for(Tuple tuple : maxTupleSet) {
```
ThomasHaas, line 49: Should be `encodeTupleSet`. But you still need to go over the `maxTupleSet` to check for interferring locks/unlocks.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
```diff
@@ -136,56 +102,53 @@ public TupleSet getMaxTupleSet(){
...
-
-                        // Can be executed if addresses mismatch, but behaviour is "constrained unpredictable"
-                        // The implementation does not include all possible unpredictable cases: in case of address
-                        // mismatch, addresses of read and write are unknown, i.e. read and write can use any address
-                        storeExec = bmgr.or(storeExec, isPair);
-                    }
+        BooleanFormula enc = bmgr.makeTrue();
+        BooleanFormula unpredictable = bmgr.makeFalse();
+        Map<Event, BooleanFormula> map = new HashMap<>();
+        for(Tuple tuple : maxTupleSet) {
```
ThomasHaas, line 108: Should be `encodeTupleSet` for all non-exclusive events. This is particularly important for Svcomp, because atomic blocks create huge `rmw`-relations of which not a single one needs to get encoded.
For exclusive events, I think you always need to encode the pairing conditions (not sure though).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
xeren: We would still need to extends `encodeTupleSet` over alternatives with intermediates.  Those need to be encoded, as well.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
xeren: I slightly changed the semantics of the `excl` variables for this purpose:  It now accounts for the `MATCHADDRESS` tag of the store and is encoded for all non-min exclusive pairs, while the `rmw` relationship is defined on top of it.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
```diff
@@ -136,58 +101,65 @@ public TupleSet getMaxTupleSet(){
...
+        BooleanFormula unpredictable = bmgr.makeFalse();
+        for(Event store : task.getProgram().getCache().getEvents(
+                FilterIntersection.get(FilterBasic.get(Tag.WRITE), FilterBasic.get(Tag.EXCL)))) {
+            checkState(store instanceof MemEvent, "non-memory event participating in '" + getName() + "'");
+            BooleanFormula storeExec = bmgr.makeFalse();
+            for(Tuple t : maxTupleSet.getBySecond(store)) {
+                MemEvent load = (MemEvent) t.getFirst();
+                if(minTupleSet.contains(t)) {
+                    storeExec = bmgr.or(storeExec, load.exec());
+                    continue;
```
ThomasHaas, line 114: The pairing conditions still need to hold in this case. In particular, under RISC-V and Power the store must fail if there is an address mismatch (that is why there is the MATCHADDRESS Tag which is added to RISC-V and Power SC events, but not to ARMv8 one's).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
ThomasHaas: I'm not sure what you mean by "alternatives with intermediates".
I think `RelCrit` behaves nicely in that it has no intrinsic meaning to the anarchic semantics and is only relevant for the memory model. So if a tuple is not in the `encodeSet` then no-one cares about the value of that edge in an execution.
This is unlike the case of exclusive events in `RelRMW`, because pairing conditions need to be satisfied due to the anarchic semantics (under e.g. RISC-V an unpaired or wrongly paired SC has to fail).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
xeren: This means that `minTupleSet` is currently greater than it should be, even for ARMv8, whose encoding still entails equal addresses.  Seems like an unprecedented case, is there no test on ARMv8 with reserve/exclusive instructions with different addresses that requires a consistent execution with even the store executed?
Shouldn't the entire `rmw` be a subset of `loc`?

ThomasHaas: I don't think we have litmus tests for ARMv8 that have LL/SC pairs accessing different addresses. We do have such tests for RISC-V though (and some of those break in this PR).
I think in all cases, a rmw-edge should only be established when the addresses match. In the case of Power/Risc-V this is justified because the store has to fail if addresses mismatch. In the case of ARMv8, mismatching addresses give rise to a kind of undefined behavior which we do not fully model (who the hell does that write mismatching LL/SC pairs to begin with?!).

So I would say a LL/SC pair is in the `minTupleSet` iff
- there is never an intervening operation
- In the case of ARMv8, both events are must-alias.

Note that the address condition is irrelevant in the case of RISC-V/Power, because if the addresses mismatch, the store has to fail (i.e. not get executed). So in every execution that contains both events, they have to go to the same address and hence form a rmw-pair.
But care has to be taken that this failure condition gets encoded under all circumstances, because it is part of the anarchic semantics. It's a little unfortunate, that `RelRMW` encodes both the rmw-edges as seen from the memory model as well as the anarchic semantics of LL/SC operations.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
ThomasHaas: I just looked at the original code and now I get what you mean. Since critical sections can get nested, we use an encoding that works a little different to the usual pairing encodings. In particular, to decide the pairing of an outer pair of lock/unlock, we need to decide the pairing of the inner nested lock/unlocks. 
But then I think your `addMatchingTupleSet` method is incorrect for this case. It seems to me that it will fail in the case of
```
rcu_lock; // 1
rcu_lock; // 2
rcu_unlock; // 3
rcu_unlock; // 4
```
I suspect that the pair (1, 4) will not be found since there are intervening operations.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/stat/StaticRelation.java
```diff
@@ -47,4 +55,35 @@ protected BooleanFormula encodeApprox(SolverContext ctx) {
...
+    protected void addMatchingTupleSet(List<Event> events, String open, String close, Consumer<Tuple> outMust) {
+        ExecutionAnalysis exec = analysisContext.get(ExecutionAnalysis.class);
+        // assume order by cId
+        // assume cId describes a topological sorting over the control flow
+        for(int end = 1; end < events.size(); end++) {
+            Event store = events.get(end);
+            if(!store.is(close)) {
+                continue;
+            }
+            int start = range(end - 1, -1).filter(i -> exec.isImplied(store, events.get(i))).findFirst().orElse(0);
```
ThomasHaas, line 68: This doesn't work and will always default to `0`. If you want to have a stream that iterates backwards, you need to use `IntStream.iterate(end - 1, i -> i >= 0, i -> i - 1)`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/stat/StaticRelation.java
```diff
@@ -47,4 +55,35 @@ protected BooleanFormula encodeApprox(SolverContext ctx) {
         }
         return enc;
     }
+
+    protected void addMatchingTupleSet(List<Event> events, String open, String close, Consumer<Tuple> outMust) {
```
ThomasHaas, line 59: The naming of the method is bad. I would call it `findMatchingPairs` and rename the parameters to something like `tagA` and `tagB` (or even use FIlters as input).
Also, you can call the consumer `mustMatch` and additionally add a `mayMatch` consumer.
The `mayMatch` consumer can be used to populate `maxTupleSet`.
I think this makes this method a lot more generic and reusable**.
That being said, this method is not usable for `RelCrit`, so it has a single valid user right now. You might as well put it into `RelRMW` as a helper method.

** If you want to go all the way, you could even add an option to change between "nested matching" (used by `RelCrit`) and "linear/direct matching" (used by LL/SC). In this case, it might even be worth to create a class that encapsulates these features.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/stat/StaticRelation.java
xeren: Thanks, that's correct.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
xeren: You were right: `rcu-rscs` must not behave like `rmw`.  The algorithm should now result in {(1,4),(2,3)}.

There are still imprecisions;  In the following program, the new iteration yields the additional edge (1,6) in `maxTupleSet`.
```
rcu_lock; // 1
if x
    rcu_unlock; // 2
else
    rcu_unlock; // 3
if y
    rcu_lock; // 4
else
    rcu_lock; // 5
rcu_unlock; // 6
```
Assume that (1,6) was the only tuple added to `encodeTupleSet`.  All others have to be encoded, as well, since the encoding is
```
(= rcu-rscs(1,6)
  (and
    exec(1)
    exec(6)
    (not rcu-rscs(1,2))
    (not rcu-rscs(1,3))
    (not rcu-rscs(4,6))
    (not rcu-rscs(5,6))
  )
)
```

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/stat/StaticRelation.java
xeren: I decided to inline the method again, since both usages were indeed incompatible.  Feel free to outline it again, if the need arises.

ThomasHaas: I agree with the inlining. If there is only one user, there is not much advantage in having a separate method (the only one is that it possibly has a descriptive name).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
ThomasHaas: Yes, it depends on how smart you do the static matching and the encoding.
You can establish that that the 4 cases you have listed there are actually valid `minTuples`.
Now you either encode `rcu-rscs(1,2) <=> exec(1) /\ exec(2)` (which enlarges your `encodeSet`) and then use `rcu-rscs(1,2)` or you directly inline `exec(1) /\ exec(2)` consistently.
I would go for whatever is easier to implement.
The difference in the encodings is marginal and won't be noticable in runtime at all.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
```diff
@@ -23,59 +30,95 @@ public RelCrit(){
...
-            removeMutuallyExclusiveTuples(maxTupleSet);
         }
         return maxTupleSet;
     }
 
-    // TODO: Not the most efficient implementation
-    // Let's see if we need to keep a reference to a thread in events for anything else, and then optimize this method
+    @Override
+    public void addEncodeTupleSet(TupleSet tuples) {
+        LinkedList<Tuple> queue = new LinkedList<>(intersection(tuples, maxTupleSet));
```
ThomasHaas, line 77: You can usse `Deque<Tuple>` with `ArrayDeque<Tuple>` as implementation. Its more natural and likely faster (not that performance matters).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
```diff
@@ -23,59 +30,95 @@ public RelCrit(){
...
+                }
+            }
+            for (Tuple t : maxTupleSet.getByFirst(lock)) {
+                if (isOrdered(lock, t.getSecond(), unlock)) {
+                    queue.add(t);
+                }
+            }
+        }
+    }
+
```
ThomasHaas, line 97: Just a little remark on this.
The `encodeSet` has a single purpose: Make sure that *at least* those relationships are encoded because someone needs them.
Now this makes all of the following approaches valid:
- Whenever someone enlarges the `encodeSet`, add whatever you additionally need to the `encodeSet` (as you do right now).
- On encoding time, enlarge the `encodeSet` as you need (i.e. do it once as a preprocessing/postprocessing step). It is sound to enlarge the `encodeSet` arbitrarily.
- On encoding time, just encode more than is forced by the `encodeSet`. It is sound to encode more than necessary (it would be nice to reflect this in the `encodeSet` though, so I would prefer the second point over this one).

I don't know which of these solutions fits best here, but I just want to mention them.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
ThomasHaas: The issue of allowing execution of the SC under POWER/RISC-V even if its address mismatches with the LL is still there.
In other words, an SC that is not part of any rmw-edge should not get executed at all (under POWER and RISC-V).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
xeren: Inlining would have to be recursive.  I implemented the artificial enlarging.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
xeren: You were right.  As long as `RelRMW` encodes the anarchic semantics of store-exclusives, that conditional is obsolete.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
xeren: Done, thanks!

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
xeren: For a relation without dependencies, it should not matter when this closure is computed.  I rather stick to the first approach, as to keep modifications to `encodeTupleSet` in one phase.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelRMW.java
```diff
@@ -136,58 +137,61 @@ public TupleSet getMaxTupleSet(){
...
+            MemEvent store = (MemEvent) tuple.getSecond();
+            enc = bmgr.and(enc, bmgr.equivalence(
+                    getSMTVar(tuple, ctx),
+                    minTupleSet.contains(tuple) ?
+                            getExecPair(tuple, ctx) :
+                            // Relation between exclusive load and store
+                            bmgr.and(
+                                    store.exec(),
+                                    exclPair(load, store, ctx),
+                                    generalEqual(load.getMemAddressExpr(), store.getMemAddressExpr(), ctx))));
```
ThomasHaas, line 177: I think you can check if the store has `Tag.MATCHADDRESS` and if so, drop the condition on the addresses, since `exclPair` already implies it. It's a micro optimization though.

ThomasHaas (approved):
LGTM

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/RelCrit.java
```diff
@@ -23,59 +30,95 @@ public RelCrit(){
...
         if(minTupleSet == null){
-            minTupleSet = new TupleSet();
-            // Todo
+            getMaxTupleSet();
         }
         return minTupleSet;
     }
 
     @Override
     public TupleSet getMaxTupleSet(){
```
hernanponcedeleon, line 39: Since the method populates both min and max sets, it make more sense to extract the populating in a new method which is then call by both `getMinTupleSet()` and `getMaxTupleSet()` whenever one of the sets is `null`.

ThomasHaas: I agree about this. I will fix this

PR #287, opened 2022-07-22
Task analysis context

xeren:
Move the reference and ownership over `Context` from `VerificationTask` to the model checking implementations.


hernanponcedeleon:
The PR should target the development branch

ThomasHaas:
You need to change the target branch to `development`.

I like the changes of this PR a lot. In particular the fact that the encoders are now external from the task.
If you want to take it one step further, it would also be nice to move the program preprocessing and the program/wmm analyses outside the task object.

xeren:
> The PR should target the development branch

Thanks for pointing this out.  I keep disregarding that option.  It is indeed intended for `development`.

> it would also be nice to move the program preprocessing and the program/wmm analyses outside the task object.

Coming right up.

ThomasHaas:
> > it would also be nice to move the program preprocessing and the program/wmm analyses outside the task object.
> 
> Coming right up. How about a common base class for the solvers, called `SolverBase` or `ModelChecker`?

I agree with a common base class for the solving methods (and to finally get rid of the static methods).
You can place the analyses and processing code into that base class for now, but don't try to maximize code re-usage via that base class. While the solver/model checking algorithms we have share quite a bit of logic, I think it is best to let them really have full control over the whole process of processing, analyzing and finally solving the verification task.
They are (almost) "entrypoints" so to speak, and we have to be very careful about reusing common code for them.

ThomasHaas:
Do you still want to make the actual solving methods non-static?
Eventually we will do this, but we can also do so in a separate PR and keep this one as it is for now.

xeren:
That change I am willing to add to this PR, but it is not finished yet.

ThomasHaas:
I like the changes. I will take a look later and if there is nothing more questionable, we can merge.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -14,19 +15,35 @@
...
+        return a;
     }
 
-    private void run(VerificationTask task, Context context) {
+    private void run(Wmm memoryModel, VerificationTask task, Context context) {
         // Init data context so that each relation is able to compute its may/must sets.
-        Wmm memoryModel = task.getMemoryModel();
         for (Axiom ax : memoryModel.getAxioms()) {
             ax.getRelation().updateRecursiveGroupId(ax.getRelation().getRecursiveGroupId());
         }
```
ThomasHaas, line 49: I think this code is dangerous, cause there is a mismatch between the provided memory model and the memory model inside the task.
That being said, I don't think any relation ever accesses `task.getMemoryModel` so they won't see any mismatch.
I understand that you did this to handle `Refinement`, but now that `VerificationTask` is a simple data container, I think it is reasonable to have the `RefinementSolver` construct a new `VerificationTask` object with the correct memory model and pass that one to the relation analysis (we can extend the `Builder` to allow creating copies of a preexisting task with minor changes).
I also think that this makes intuitive sense: The task that gets encoded by `RefinementSolver` is fundamentally a different one to the one that it will eventually solve.

ThomasHaas:
@hernanponcedeleon I think this branch is ready to get merged. There are more things to do, like getting rid of some more static methods and making more use of the fact that we can instantiate solvers now, but I think we can do this in a different PR.

ThomasHaas (approved):

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
```diff
@@ -4,34 +4,60 @@
...
 import static java.util.Collections.singletonList;
 
-public class AssumeSolver {
+public class AssumeSolver extends ModelChecker {
 
     private static final Logger logger = LogManager.getLogger(AssumeSolver.class);
 
+    private final SolverContext ctx;
+    private final ProverEnvironment prover;
+    private final VerificationTask task;
```
hernanponcedeleon, line 28: All solvers have these. Can we simply move them to `ModelChecker` as `protected`? The class `TwoSolvers` have two `ProverEnviroments`, but I guess it can use the one in `ModelChecker` as one and just add a second one in that class.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -41,6 +41,8 @@ public List<Axiom> getAxioms() {
         return axioms;
     }
 
+    public Set<Relation> getRelations() { return relationRepository.getRelations(); }
```
hernanponcedeleon, line 44: `Wmm.getRelationRepository().getRelation()` is also used all over the place. Shall we also provide a convenient method `Wmm.getRelation()` (I know @ThomasHaas wants to completely rewrite how we handle Wmms so this could be part of that intended change)

ThomasHaas: I agree with this change (I was contemplating about doing it already), but I think we can defer it to the rework of `Wmm`.
That rework definitely intends to get rid of this ugly `RelationRepository` business.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
ThomasHaas: I wouldn't do this right now. I think the whole idea of explicitly providing the `ProverEnvironment` and `SolverContext` as input is a little outdated.
We only do this to be able to access the solver model to generate some output (e.g. check the type of violation).
However, now that we have actual `ModelChecker` instances, it is reasonable that these manage their own `ProverEnviroment/SolverContext` and just make a `Model` accessible, similar to how the SMT solver works: On `FAIL`, we can call `modelChecker.getCounterexample` (or `modelChecker.getModel`). Ideally, we generalize the `ExecutionModel` class to extract complete models (including violated assertions/properties as well as local events) so that we can return those instead of SMT-solver models.
That is what I meant by saying
> making more use of the fact that we can instantiate solvers now
All this requires quite a bit of rework though.

For the time being, I think it is fine to have those variables in each class separately. We will need to think about what exactly we want to achieve with the `ModelChecker` base class (surely not just sharing 3 variables!), and once we decided, we can refactor if necessary.

PR #289, opened 2022-07-24
Allow relation aliases

hernanponcedeleon:
This PR adds support for alias between relations which allows to handle cat files with things like `let r1 = co`.


ThomasHaas:
The current implementation does not allow to figure out aliases of a relation, which can cause some problems when copying relations from one `Wmm` to another (the aliases won't be copied). This would happen when cutting a `Wmm`, but I don' think it will cause problems right now.
Also, base relations strangely have no name (only a `term`) but they can have an arbitrary amount of aliases. This is somewhat counterintuitive.
I will fix this in a future PR.

ThomasHaas (approved):
LGTM

PR #290, opened 2022-07-26
Remove AMO events

hernanponcedeleon:
As discussed with @ThomasHaas, many of our analyses rely on the assumption that a `MemEvent` is either a `Load` or a `Store`. Adding AMO events break this assumption. While it caused no problems for the litmus tests, this is dangerous and thus we decided to revert that change (i.e. remove AMO events for the time being).


ThomasHaas (approved):
LGTM

PR #291, opened 2022-07-26
Allow ExecutionStatus to track dependency

hernanponcedeleon:
This PR explicitly marks which `ExecutionStatus` events should track dependencies (and in the case of RISCV). If new models allow to track dependency from stores, the logic of relations does not need to be changed anymore. Instead the correct type of `ExecutionStatusWithTrackingDependency` needs to be created when creating events.


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/local/BasicRegRelation.java
```diff
@@ -51,6 +52,9 @@ private void mkTupleSets() {
                 Dependency.State r = dep.of(regReader, register);
                 for(Event regWriter : r.may) {
                     maxTupleSet.add(new Tuple(regWriter, regReader));
+                    if(regWriter instanceof ExecutionStatus && ((ExecutionStatus)regWriter).getTrackDep()) {
+                    	maxTupleSet.add(new Tuple(((ExecutionStatus)regWriter).getStatusEvent(), regWriter));
+                    }
```
ThomasHaas, line 57: You can also add the tuple to `minTupleSet`.

hernanponcedeleon:
@ThomasHaas your changes LGTM, thus once the CI passes I'll merge.

PR #294, opened 2022-07-27
Improved value propagation in dependency encoding

ThomasHaas:
This PR improves the encoding that matches `RegWriter` results with `RegReaders` reading those results.
The encoding we generate is usually of the form `guard => regValue = regWriterResult`,  where `guard` is a boolean guard variable that makes sure that the `regWriter` is the most recent one and that both events got executed.
However, in the special case that there is a unique `regWriter` from which the `regValue` must get obtained (must-dependency + execution of `regReader` implies execution of `regWriter`), we can get rid of the guard altogether.
Now the generated formula is an unconditional equality `regValue = regWriterResult`, which allows the solver to eliminate one of the two variables in the preprocessing.

Some quick tests indicate large improvements on SafeStack.


hernanponcedeleon:
This works quite well, specially for safe-stack

ARM8-old
ARM8

PR #295, opened 2022-07-28
Multireads + no final reg values for non-litmus test + alias analysis extension

ThomasHaas:
This PR adds two encoding optimizations:
- Avoid encoding of final reg values for non-litmus tests.
- (Toggable option, dsiabled by default in this PR) Avoid encoding at-most-one constraints for reads. The uniproc axiom of memory models already disallows more than one read from happening. In the case of Refinement, the solver will decide on the first rf-edge it finds, even if multiple exist.

Edit: It also adds a simple alias analysis extension that established must-alias based on equality of address expressions.

I have no idea if those optimizations have measurable impact, so we should run some tests first.
While the first optimization is a no-brainer, the second one is not and needs testing.


dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -38,9 +43,32 @@ final class Config {
                 description = "General type of analysis that approximates the 'loc' relationship between memory events.")
         private Alias method = Alias.FIELD_SENSITIVE;
 
+        private final boolean addEqualityAnalysis = true;
```
hernanponcedeleon, line 46: Why not adding this as a configuration option?

ThomasHaas: I didn't add it (yet) because I was not sure if this option is so obvious that it should simply be always enabled.
I would rather avoid cluttering the configuration settings with options that are no-brainers.
On the other hand, the read-relaxation should become configurable.

hernanponcedeleon: I agree about not cluttering the configurations ... but if it will always be enabled, we don't need this filed at all

ThomasHaas: Yes, I can remove it completely. For debugging/testing purposes, it was very useful though.

PR #296, opened 2022-07-29
Lkmm to riscv

hernanponcedeleon:
This PR adds support to compile LKMM code into RISCV.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
@@ -306,8 +306,7 @@ class VisitorArm8 extends VisitorBase implements EventVisitor<List<Event>> {
 				optionalMemoryBarrier = AArch64.DMB.newISHBarrier();
 				break;
 			default:
-				optionalMemoryBarrier = null;
-				break;
+				throw new UnsupportedOperationException("Compilation of fence " + e.getName() + " is not supported");
 		}
 
 		return eventSequence(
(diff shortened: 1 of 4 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -306,8 +306,7 @@ public List<Event> visitLKMMFence(LKMMFence e) {
 				optionalMemoryBarrier = AArch64.DMB.newISHBarrier();
 				break;
 			default:
-				optionalMemoryBarrier = null;
-				break;
+				throw new UnsupportedOperationException("Compilation of fence " + e.getName() + " is not supported");
```
ThomasHaas, line 309: Is there no LKMM fence that does not get compiled to NOOP? I would have imagined that there exist some very weak barriers that are removed on compilation.

hernanponcedeleon: If that is the case, we should explicitly handle that case, but defaulting to compile some barriers as NOOPT is dangerous

ThomasHaas: I totally agree. I was just wondering if we did implicitly handle some weak barriers in the default case, which we end up not doing anymore. But it seems that the switch is exhaustive and there are indeed no barriers that get removed upon compilation (I guess when compiling to TSO, some barrier would definitely go).

ThomasHaas (approved):
LGTM

PR #297, opened 2022-07-29
Improved co encoding

ThomasHaas:
Fixed minor bug in RelCo.applyLocalConsistencyMinSet.
Refactored, simplified and hopefully improved co-encoding by avoiding to encode some transitively implied clock constraints.

I don't know if this improves the solving times in any way. It reduces the co-encoding a little bit though.

Edit: Added simple trick to reduce number of theory literals produced in co-encoding. The idea is this: 
if there are w1, w2 such that both coherence orders are possible, then we usually encode one direction via `c(w1) < c(w2)` and the other via `c(w2) < c(w1)`, which results in two different theory literals. However, due to the distinctness constraint on clock variables, we can soundly change the second version to `c(w2) <= c(w1)` instead. Now notice that `c(w2) <= c(w1) IFF not (c(w1) < c(w2))`, that is, we can reuse the first clock constraint but negated.
This reusage of the same theory atom might help the solver.

Edit 2: I reverted the last change as it negatively impacted `SafeStack` (by a lot). I cannot really tell why that happens, but maybe the fact that literals appear in both polarities can disallow some reasoning/optimization.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
@@ -8,6 +8,7 @@ import com.dat3m.dartagnan.program.event.core.Init;
 import com.dat3m.dartagnan.program.event.core.MemEvent;
 import com.dat3m.dartagnan.program.filter.FilterBasic;
 import com.dat3m.dartagnan.program.filter.FilterMinus;
+import com.dat3m.dartagnan.utils.dependable.DependencyGraph;
 import com.dat3m.dartagnan.wmm.analysis.WmmAnalysis;
 import com.dat3m.dartagnan.wmm.relation.Relation;
 import com.dat3m.dartagnan.wmm.utils.Tuple;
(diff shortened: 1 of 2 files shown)


hernanponcedeleon:
It is not yet clear if there is any performance improvement. On LFDS, there seem to be a ~7% improvement, but since that set contains unsafe benchmarks, some oscillation can be expected and this "improvement" might just be noise.
Since anyway the size of the encoding is reduced, I'll merge.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
```diff
@@ -220,7 +217,47 @@ protected BooleanFormula encodeApprox(SolverContext ctx) {
         }
         return enc;
     }
-    
+
+    /*
+        Returns a set of co-edges (w1, w2) (subset of maxTupleSet) whose clock-constraints
+        do not need to get encoded explicitly.
+        The reason is that whenever we have co(w1,w2) then there exists an intermediary
```
hernanponcedeleon, line 224: "there exists" should be "there might exist", right? The `co` edge might actually be a direct one.

ThomasHaas: No, the explanation is exactly for those co-edges that do not need to get encoded (and for those there exists such an intermediary). I can reformulate the sentence a little bit to make this more clear.

ThomasHaas:
Before you merge, I want to test another thing.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/base/memory/RelCo.java
hernanponcedeleon: Now I get it. The `co(W1,W2)` in the second sentence refers to those defined in the first sentence and not any pair.

ThomasHaas:
I added another optimization of distinctness constraints.
In `development`, we add a single distinctness constraint over all clock variables of stores.
However, this ends up relating the clock constraints of stores that may never access the same address.
We can partition the stores into aliasing classes, such that stores in different classes will never alias.
It suffices to add a distinctness constraint per class.

hernanponcedeleon:
LGTM

PR #299, opened 2022-08-04
Alternative/simpler Co-encoding

ThomasHaas:
This PR replaces the co-encoding by a new one, which is cleaner and avoids the distinctness constraints.
I don't know if it affects performance in any way though.
There is also some minor refactoring in `RelCo`.


hernanponcedeleon:
The code is no only cleaner, but it also indeed improves performance (~15%) in Locks and LFDS benchmarks.

PR #300, opened 2022-08-12
SAT-based acyclicity/totality encoding

ThomasHaas:
This PR adds the option "encoding.wmm.idl2sat" to enforce a SAT-based encoding for totality of `co` and the acyclicity axiom. This is needed to e.g. make use of Boolector which cannot handle integer constraints.
Together with bv-based data flow encoding and bit-blasting, we can get pure SAT encodings this way.

I also removed the co antisymmetry option as it seems pointless now.


dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -497,6 +522,6 @@ private void extractCoherences() {
             }
 
             coherenceMap.put(addr, Collections.unmodifiableList(sortedWrites));
-        }
+        }*/
```
hernanponcedeleon, line 525: Can we get rid of this?

ThomasHaas: Good catch, I totally forgot about this change.
I'm actually not sure about that one. 
The uncommented version works for any encoding of `co`, whereas the commented version assumes clock variables (IDL-encoding).
Now, the clock-variable version is surely cheaper to extract, so we should probably use it for the IDL-encoding.
That being said, model extraction is rarely an issue (except for MathSAT5).
Maybe I will make the `useSATEncoding`-variable of `RelCo` accessible, and then the `ExecutionModel` can decide on which one to use.

hernanponcedeleon: sounds good

PR #301, opened 2022-08-17
Fix RefinementSolver + new idl2Sat option

ThomasHaas:
This PR fixes an issue with the new `encoding.wmm.idl2sat` option and Refinement.
I also added better error message printing.


dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -126,6 +126,11 @@ private Result run() throws InterruptedException, SolverException, InvalidConfig
         symmEncoder.initializeEncoding(ctx);
         baselineEncoder.initializeEncoding(ctx);
 
+        //FIXME: This is an ugly fix to inject the configuration into the RelCo-instance that does NOT get encoded.
+        // This is needed because the ExecutionModel will see the settings of the non-encoded instance
+        // which may not align with the encoded one.
+        memoryModel.getRelationRepository().getRelation(CO).initializeEncoding(ctx);
+
```
hernanponcedeleon, line 133: Can we move the option annotation to the `RelationRepositoy` and pass the `useSATEncoding` as a flag during the constructor?

ThomasHaas: We usually don't get to configure the `Wmm` before its relations get constructed (I think the Parsing code doesn't know about configurations).
I could probably split the injection of configurations from the initialization of the encoding (IIRC, `initializeEncoding` on `Relations` doesn't do anything at all except injecting configurations). And then there could be a `Wmm.configure(Configuration)` which passes a configuration to all its relations.

ThomasHaas: To make this more clean, I reworked a few things:
- `Relation`, `Axiom` and `Wmm` have now configure methods.
- All `ModelCheckers` will call `Wmm.configureAll` to inject configurations.
- `Wmm` will on construction add all base relations to the relation repo (it was strange that `RelationAnalysis` was the one who made sure that all base relations are present in the `Wmm`).

PR #302, opened 2022-08-17
Positive version of RISC-V

ThomasHaas:
This PR adds an alternative and fully positive formulation of the RISC-V memory model: `riscv-pos.cat`.
With this, cutting is needed no more.

The idea is that the negative formulation of `r2` ends up basically expressing what is known as `rdw` (read-different-write?) in Power. `rdw = po-loc & fre;rfe`, however, is completely positive so replacing `r2` by `rdw` yields a positive model.

While this PR does not add unit tests for this model, I ran the tests locally (all litmus tests, as well as all compilation tests) without encountering any wrong results.


hernanponcedeleon:
I tried the litmus tests with the new model and indeed there are no problem. I propose the following:
- lets rename the original `riscv.cat` to `cut-riscv.cat` and add a note in the header that this is the original version, which is not semi-positive and thus requires cutting
- lets rename the new version as `riscv.cat` and add a note in the header that this is an alternative and positive version. With this changes, the CI will take care that the versions are "equivalent" for our tests cases

ThomasHaas:
If you just want to make sure that the CI runs the correct version, you can simply change the `Providers.WmmFromArchitecture` implementation to return `riscv-pos.cat`. In fact, this makes sure that all our unit tests use the desired model.
I think we should not call the original one `cut-riscv` but simply `riscv-orig.cat` (or just keep it at `riscv.cat`). Having "cut" in the name is odd, because it does not say anything about the model itself but something about how one of our algorithms (CAAT) treats it.

PR #304, opened 2022-08-22
Wmm cleanup

ThomasHaas:
This PR cleans up code surrounding `Wmm` and `Relation`.

- `Relations` do not remember if they got encoded or not anymore. This behaviour to skip encoding on subsequent calls was dangerous.
- `WmmEncoder` makes sure that each relation gets encoded only once without relying on maintaining various flags and doing recursive encodings.
- As a result of the previous changes, `Relation.doForceEncode` and `Relation.isEncoded` are gone. The function `Relation.doEncode` is gone as well. `Relation.encode` has got a trivial implementation now (we may even remove it altogether as well).
- Also, `Relation.recursiveGroupId` as well as `Recursion.isRecursive` are gone now. Neither of them had any uses.
- `UnaryRelation` and `RelMinus` are not allowed in recursions. These relations used to check for this by examining their now gone `recursiveGroupId`. This functionality has been moved to `WmmAnalysis` which will now throw appropriate exceptions (including a new `MalformedMemoryModelException`) if the memory model contains unsupported recursions.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
@@ -1,12 +1,15 @@
 package com.dat3m.dartagnan.encoding;
 
+import com.dat3m.dartagnan.utils.dependable.DependencyGraph;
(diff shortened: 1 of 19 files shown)


hernanponcedeleon:
LGTM

PR #305, opened 2022-08-23
Cf opt

hernanponcedeleon:
Due to our branch analysis, the CF encoding has unnecessary redundancies like
This PR makes the following simplifications to the CF encoding
- We can completely skip adding constraints `cf(E0) == cf(E0)` and `cf(E0) => cf(E0)`
- We simply constraints `T == cf(E0)` and `T => cf(E0)` to `cf(E0)`
- We simplify `enc /\ T` to `enc`

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
@@ -154,7 +154,7 @@ public class ProgramEncoder implements Encoder {
             }
 
             if(!e.cf().equals(cfCond)) {
-            	enc = bmgr.and(enc, cfCond.equals(bmgr.makeTrue()) ? e.cf() : cfEncoder.apply(e.cf(), cfCond));
+            	enc = bmgr.and(enc, e.cf().equals(bmgr.makeTrue()) ? cfCond : cfEncoder.apply(e.cf(), cfCond));
             }
             if(!e.encodeExec(ctx).equals(bmgr.makeTrue())) {            	
                 enc = bmgr.and(enc, e.encodeExec(ctx));
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
@@ -153,7 +153,12 @@ public class ProgramEncoder implements Encoder {
                 }
             }
 
-            enc = bmgr.and(enc, cfEncoder.apply(e.cf(), cfCond), e.encodeExec(ctx));
+            if(!e.cf().equals(cfCond)) {
+            	enc = bmgr.and(enc, cfCond.equals(bmgr.makeTrue()) ? e.cf() : cfEncoder.apply(e.cf(), cfCond));
+            }
+            if(!e.encodeExec(ctx).equals(bmgr.makeTrue())) {            	
+                enc = bmgr.and(enc, e.encodeExec(ctx));
+            }
             pred = e;
         }
         return enc;


ThomasHaas:
You should check if `FormulaManager.simplify` does these types of simplifications.
From my testing so far, it indeed does perform all those simplifications if the underlying solver is `Z3`. I have not yet tested what `Yices2` and `MathSAT5` do, but you can bet that those trivial simplifications are done.
So I would generally avoid adding such special case reasoning because all you get is more complicated code without any practical benefit. In fact, if you look into the current implementation of `RelCo.encodeSAT/RelCo.encodeIDL`, you will even see that I rely on solver-sided simplifications to keep the code as simple as possible.

hernanponcedeleon:
I don't think we should always rely on the SMT solver (which we do not have control over) to make such simplifications, at least not in cases where doing the simplification at the java level is trivial.

It is true that `z3`, `yices2` and `mathsat5` do these simplification, but neither `princess` no `cvc4` do them and thus this consumes unnecessary memory. This is what `ctx.getFormulaManager().simplify(enc)` returns
- princess:
```
(((cf(E0) & cf(E0)) & cf(E0)) & cf(E0))
(((cf(E0) & cf(E0)) & cf(E0)) & cf(E0))
(cf(E0) & cf(E0))
(cf(E0) & cf(E0))
```
- cvc4
```
(and (and (and (= |cf(E0)| true) (= |cf(E0)| |cf(E0)|)) (= |cf(E0)| |cf(E0)|)) (= |cf(E0)| |cf(E0)|))
(and (and (and (= |cf(E0)| true) (= |cf(E0)| |cf(E0)|)) (= |cf(E0)| |cf(E0)|)) (= |cf(E0)| |cf(E0)|))
(and (= |cf(E0)| true) (= |cf(E0)| |cf(E0)|))
(and (= |cf(E0)| true) (= |cf(E0)| |cf(E0)|))
```

ThomasHaas:
The fact that `CVC4` does not show simplifications is because `JavaSMT` doesn't forward any calls to `CVC4` (it simply returns the original formula). In general, `FormulaManager.simplify` is a very crude approximation of what the solver will eventually do, even for `Z3`, `Yices2`, and `MathSAT5`.

I think we should rely on the trivial simplfications, because there is simply no way that any solver will retain statically satisfied clauses, multiple identical literals per clause, or "true/false" literals.
Any solver that does not do these simplifications is simply not worth using. Princess might be a prime example for this as it performs incredibly bad on our formulas. In fact, Princess was made for an entirely different purpose (quantified presburger arithmetic), so I wouldn't even consider it as a valid alternative that we should optimize for.

The point is, none of these optimizations/simplications use any domain-knowledge or context-based information. They don't even modify the formula in an equisatisfiable way but literally keep it logically identical on the boolean level.
This sort of trivial reasoning is exactly what any SAT-engine can and will do with local rewriting.
I'm all for using high-level domain knowledge or context-sensitive information that the solver is unlikely to predict, but I wouldn't want to bloat the code with impactless code that just ends up making it harder to modify and read (also, the extra code is even branching, increasing code nesting and complexity.)
If you apply the same reasoning to all our code base, you can easily get dozens of lines of effectively dead code.

Lastly, if you really care about these formula simplifications, you could write a `FormulaTransformer` that traverses the whole formula and removes these redundancies everywhere. I actually implemented this approach to do arithmetic simplifications from `ax=ay` to `x=y` as this is not done by `FormulaManager.simplify` using `Z3` (because simplify won't go into theory literals!). As it turns out (and was expected), this also had zero impact on solving because the solver was doing it already on its own.

hernanponcedeleon:
I don't have counter arguments or some performance result showing that these changes have a practical impact, thus I am closing this PR.

PR #308, opened 2022-09-07
Fix pom file in ui plugin to run from console

hernanponcedeleon:
Our README states that we can open the UI by running
```
java -jar ui/target/ui-3.1.0.jar
```
but this command was failing. This PR fixes this.

**How to test:** simply run the command above from the console
**Expected result:** the UI should open


ThomasHaas (approved):
LGTM

PR #310, opened 2022-09-12
Include cos.cat in riscv.cat

hernanponcedeleon:
File `riscv.cat` uses `fre` which we can interpret (i.e. we always include its definition), but herd does not.
Thus we need to include `cos.cat` to be able to run the herd tests.


hernanponcedeleon:
I agree. The import would be useful (specially to avoid problems like this where dartagnan can handle a model, but herd cannot) and it should be easy to implement

PR #312, opened 2022-09-25
Set Cline to all parsed memory events

hernanponcedeleon:
All parsed memory events should have a Cline (only "modeling" events should use `-1`), otherwise this can easily "break" some po-edges in the graphviz witnesses.


ThomasHaas (approved):
Not sure what the `allocationRegs` were used for, but all deletions are welcomed.
LGTM.

PR #313, opened 2022-09-26
Fencerel takes filters as input

hernanponcedeleon:
CAT allows `fencerel` to use any set (i.e. filters implementation-wise) and not just base ones.
This PR generalizes `RelFencerel` to accommodate this.

Now we can use the original `let barrier = ...` from the LKMM.


ThomasHaas (approved):
LGTM

PR #314, opened 2022-09-28
Fix symmetry breaking issue

ThomasHaas:
- Fixed(?) a rare issue with symmetry breaking that caused unsound result.
- Changed refinement to do symmetry breaking based on the original memory model.


hernanponcedeleon (approved):
LGTM

PR #315, opened 2022-09-28
Wmm Restructuring

xeren:
This PR is a massive restructuring of the package `com.dat3m.dartagnan.wmm`.  It contains several interface changes.  Be careful!
### Summary
The following changes 

- ~~`PropertyEncoder` is now responsible of 'LastCoConstraints', since it manages both current reasons of its existence.~~
- ~~`RelationRepository` is merged into `Wmm`.~~
- ~~`R^*` is replaced with `(R|id)^+`.~~
- ~~`id` is replaced with `[V]`.~~
- ~~`Constraint` unifies both `Axiom` and `Definition` (formerly `Relation`)~~
- ~~The unused constant `Relation.PostFixApprox` becomes the option `encoding.wmm.postFixApproximate`.~~
- Adds (yet unused) capability of n-ary unions and intersections.
- `Relation.encodeTupleSet` is moved into an analysis-like manager.
- `Relation.maxTupleSet` etc. are moved into `RelationAnalysis`.
- Every relation in the model becomes redefinable.
- The `Relation` class hierarchy is flattened such that all current implementations directly extend the common base.
- That common base is now called `Definition`, which was extracted from `Relation`.  Those two classes now refer to each other.
- `RecursiveRelation` and `RecursiveGroup` become obsolete and are removed.

Edit: The items 1 to 5 are already merged into `development`.  Point 6 was discarded.


ThomasHaas:
I'm not really a fan of the default naming for relations. I think I would prefer unnamed relations instead. They would only be identified by their definition.
For example, the output of `wmm.toString` is quite unreadable as compared to before due to the default naming.
Also, we want to distinguish between named and unnamed relations when flattening sequences of binary unions into n-ary unions. While explicitly checking for the default naming works, it feels awkward to do so.

I did try to implement a pass to do this flattening, but the current framework misses redefinitions as well as the capability to delete relations (and axioms).
I think we should try to get the functionality of `Wmm` to a level where the implementation of flattening is possible and easy, to make sure that the rework is actually worthwhile.

ThomasHaas:
Another issue I see is that the active set (encodeSet) computation is now completely hidden inside the `WmmEncoder`/`EncodeSets`. 
How would one go about computing partial encode sets for lazy encodings?
My only guess would be to create alternative "lazy" versions of `WmmEncoder` and `EncodeSets` to get the desired behaviour.

xeren:
> How would one go about computing partial encode sets for lazy encodings?

It should be no problem to add methods like below to `WmmEncoder`.  Keeping track of `alreadyEncoded` would be the responsibility of `RefinementSolver`.
```
Map<Relation,Set<Tuple>> getEncodeSetClosure(
        Map<Relation,Set<Tuple>> reasons,
        Map<Relation,Set<Tuple>> alreadyEncoded);

void encodeTuples(Map<Relation,Set<Tuple>> tuples);
```

ThomasHaas:
Those methods would be helpful. I agree that `alreadyEncoded` needs to be tracked elsewhere (e.g., in the `RefinementSolver`).

I would probably name the second method like this:
```
BooleanFormula encodeRelations(Map<Relation, Set<Tuple>> tuplesToBeEncoded);
```
Then the current (default) implementation can call this with the appropriate map that contains everything computed by the active set analysis (btw. we should eventually rename encode sets to active sets).

For the first one, I'm not sure why you called the first parameter `reasons`. I assume that map expresses what the user wants to encode, i.e., the parameters would be `toBeEncoded` and `alreadyEncoded` instead. The default implementation would pass as first argument a map filled with the encode sets provided by the axioms (+ those needed for rf and co) and an empty map for `alreadyEncoded`.
Then the full eager encoding would be achieved by
```
Map<Relation,Set<Tuple>> toBeEncoded = ... // add axioms etc.
Map<Relation,Set<Tuple>> alreadyEncoded = Map.of();
BooleanFormula enc = encodeTuples/Relations(getEncodeSetClosure(toBeEncoded, alreadyEncoded));
```
right?
I would like this design quite a bit more than the current one. Those methods give nice flexibility to the caller and are less state-dependent.

ThomasHaas:
Relations that are defined via `let r = ...` are unnamed now.
The parser adds an alias but the relation itself remains unnamed.

xeren:
This exposes a bug in my default implementation of `Definition#substitute(Relation,Relation)`.  I will have to add implementations to all definitions.  Since the method makes no sense to be public at all, I will implement the visitor `DefinitionSubstitutor`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -66,20 +66,27 @@ public Relation getRelation(String name) {
...
-    public void addAlias(String name, Relation relation) {
-        addName(name, relation);
+    public Relation addName(String name, Relation relation) {
+        if (relation.named) {
+            addAlias(name, relation);
+            return relation;
+        }
+        Relation r = newRelation(name);
+        substitute(relation, r);
+        deleteRelation(relation);
```
ThomasHaas, line 76: This code is really dangerous. By doing substitutions like this, references held by external code become outdated.
Certainly nothing that should be expected when changing the name of a relation.
I would probably do something like this:
- Relations have no name by default, likely by having a `List<String> names` of names, initially empty. The first entry of this list is the main name, other entries are aliases.
- The Wmm holds a `Set<Relation> relations` containing all relations as well as a `Map<String, Relation> aliasMap` for the names/aliases.
- When calling `addName/addAlias`, the map is checked for duplicates. If none, the map is updated and the name is appended to `rel.names`. The first added alias is automatically the main name of the relation.
- Maybe there is a `setName` function to set the main name (either inserting the name at the front, or moving an already existing alias to the front). We have no use-case for this yet though.
- I also don't think we need to be able to do lookups of relations by term, so we can get rid of those and only store true names/aliases. Search by term can be done by iterating over the set of relations, if necessary.

I think you won't get any cleaner solution than the above. It supports exactly what is needed (arbitrary number of names per relation) without any hidden traps and moderate state management.

xeren: I was glad that I found a way to describe a relation's string representation that is both final (and therefore robust against restructuring) and human-readable.  Since the ability to rename was not required and the names of relations are known to `VisitorBase` before the creation of the relation, there should be no reason not to give it a name setter, yet.
I agree with you, concerning the substitution feature.  The only current usage is unnecessary.  I will remove it.
The term mapping implements a feature.  Duplicate expressions like `co | fr` in `tso.cat` are recognized and unified into one relation.  I did not plan to remove this, but I could do it, as cases like that will be rare and easily fixable in the input.  In the above case, both duplicates are removed during simplification.
As for the handling of aliases, after all those fixes were applied, I saw no obstacle to add that, as well.

ThomasHaas: >I was glad that I found a way to describe a relation's string representation that is both final (and therefore robust against restructuring) and human-readable. Since the ability to rename was not required and the names of relations are known to VisitorBase before their creation, there should be no reason to let them keep a name setter.

Technically, relations can be unnamed (which is also useful). By forcing default names onto those relations, you end up having to do special case reasoning to distinguish between named and unnamed relations, which feels awkward.
While this allows you to have a final string representation (using the dummy names), I don't see why this would be beneficial (it's also less human-readable IMO). After all, our whole Wmm structure is not immutable and is also not intended to be.
The new implementation of `Relation` also does not check equality based on naming, or computes its hashcode from it.
This gives it a lot less reason to have a final name on construction, particularly a non-null one.
Lastly, I think there are cases where the name of a relation is unknown during its first construction:
```
let com = rf | co | rf^-1;co // used fr, but without its name
let fr = rf^-1;co // Only here, the relation is named for the first time
```
I guess a processing step after the parsing could also eliminate these duplicates, but it might be easiest to do this already during the parsing (like it is currently done, IIRC). With fixed initial names, you would need to reconstruct the whole Wmm (which is not too bad in terms of performance, but might feel too clunky in practice).

>The term mapping implements a feature. Duplicate expressions like co | fr in tso.cat are recognized and unified into one relation. I did not plan to remove this, but I could do it, as cases like that will be rare and easily fixable in the input. In the above case, both duplicates are removed during simplification.

Ok, I see that in this case, having fixed names and fixed terms is a useful feature. We wouldn't want to have structurally equivalent duplicates like multiple `(co | fr)` in our Wmm. But I don't think this feature necessitates a term mapping in the Wmm. The parser itself could likely maintain a `Term -> Relation` map, or query the Wmm repeatedly for terms it is about to construct.

Overall, I think there are only two reasonable approaches:
- Make the whole Wmm, its relations, and its axioms all immutable. In this case, the Wmm is seen as a big syntactical expression and you need a Transformer/Rewriter to do any changes. This has nice properties but might feel overly clunky in practice (recursiveness + immutability is also a big problem).
- Make the Wmm, its relations and its axiomsm mutable. This is what general graph structures usually are. This makes (simple) transformations easy to implement but requires careful state management to avoid inconsistent data. The state management should therefore be kept to a minimum. It allows for more freedom but also for more ways to break the state. In practice, however, I believe this to be a minor issue because we follow general principles like
-- not relying on cached state during transformation/processing phases
-- not keeping state from the transformation/processing phases
-- not mutating the data during the analysis phase

Our old implementation did follow the first approach of immutability (not full immutablity though).
I felt that working with this was a pain when it came to doing any sorts of transformations of even analyzing it properly (dummy recursive relations that were not returned by `.getRelations`???).
So our new approach should remedy this issue, even if it introduces some more ways to break data consistency.

ThomasHaas: I forgot to add: the default names `_x` are just numbers in disguise, no? Maybe you are better off giving all Relations an `id` rather than a name by default.

xeren: If there was any other use for that `id` other than identifying the instance and being transformed into a string.  But this way relation instances are unique anyway; there is no need to define equivalence classes by id.  Therefore I do the string transformation once instead of each time a name is fetched for a edge or clock variable.

I think I should replace the `_` prefix with something like `#`, to avoid name collision with input-defined relations.

ThomasHaas: I think the underscore is totally fine. But I don't mind using any other special symbol instead.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
```diff
@@ -198,9 +222,19 @@ private RelationGraph getOrCreateGraphFromRelation(Relation rel) {
...
 
+    private RelationGraph createGraphFromRelation(Relation rel) {
         RelationGraph graph;
-        Class<?> relClass = rel.getClass();
+        Class<?> relClass = rel.getDefinition().getClass();
+        List<Relation> dependencies = rel.getDependencies();
+        RelationGraph[] graphs = new RelationGraph[dependencies.size()];
+        for (int i = 0; i < graphs.length; i++) {
+            graphs[i] = getOrCreateGraphFromRelation(dependencies.get(i));
+        }
```
ThomasHaas, line 237: The early construction of the dependencies does not interact well with cutting and the check for special relations, because now dependencies get needlessly constructed. Fortunately, this causes only computation overhead but does not affect the result of CAAT in any way.

xeren: I moved dependency transformation after the special and cut-relations tests.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: Sounds good. 
The thing about naming a relation the same as a term is one of those "issues" that are practically non-existant. I wouldn't bother with that at all. Writing defensive code against that makes everything more complicated for virtually zero benefit.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/ReadModifyWrites.java
```diff
@@ -0,0 +1,20 @@
+package com.dat3m.dartagnan.wmm.definition;
+
+import com.dat3m.dartagnan.wmm.Definition;
+import com.dat3m.dartagnan.wmm.Relation;
+
+/*
+    NOTE: Changes to the semantics of this class may need to be reflected
+    in RMWGraph for Refinement!
+ */
```
hernanponcedeleon, line 9: This class has not semantics information anymore. The comment can go

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -64,51 +125,737 @@ The former is disallowed by assumption, so we have co(w3, w2) and hence c(w3) <
...
+                            must.add(tuple);
+                        }
+                    }
+                }
+            }
+            return new Knowledge(may, must);
+        }
+        @Override
+        public Knowledge visitMemoryOrder(Relation rel) {
+            logger.info("Computing maxTupleSet for memory order");
```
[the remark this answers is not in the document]
ThomasHaas: I think logging shouldn't really be done inside those methods at all.
It would be fine to have the Relation Analysis finish, and then iterate through all interesting relations (rf, co, rmw and possibly the axiom relations) and output their sizes in a central place.
Logging within those methods may make sense if they compute many intermediate results and we want to observe those, but then I would likely do this with trace-level logging.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
```diff
@@ -0,0 +1,358 @@
...
+        Relation r = wmm.getRelation(n);
+        if (r != null) {
+            namespace.put(n, r);
+            addCurrentName(r);
+            return r;
+        }
+        FilterBasic f = FilterBasic.get(n);
+        namespace.put(n, f);
+        addCurrentName(f);
+        return f;
```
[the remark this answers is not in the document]
xeren: Correct.  It should also be equivalent to the following:
```
Object o = namespace.computeIfAbsent(n,
    k -> Objects.requireNonNullElseGet(
        wmm.getRelation(k),
        () -> FilterBasic.get(k)));
addCurrentName(o);
return o;
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
```diff
@@ -0,0 +1,358 @@
...
+        String a = alias();
+        Object o1 = c.e1.accept(this);
+        Object o2 = c.e2.accept(this);
+        if (o1 instanceof Relation) {
+            Relation r0 = r != null ? r : newRelation(a);
+            return addDefinition(new Intersection(r0, (Relation) o1, relation(o2, c)));
+        }
+        return withAlias(a, FilterIntersection.get(set(o1, c), set(o2, c)));
+    }
+
```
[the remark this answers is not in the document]
xeren: Partially.  `checkNoCurrentOrNoAlias` checks for the structure invariant of the visitor.  `let rec` contexts pre-create all recursive relations and pass the respective to-be-defined one to the expression context.  `let` contexts pass just the desired name to avoid the creation of a new relation, in case aliasing will take place.  Both cannot be active at the same time.
Maybe it should call `verify`, instead of throwing `ParsingException`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
```diff
@@ -0,0 +1,358 @@
...
+    @Override
+    public Relation visitExprComposition(ExprCompositionContext c) {
+        checkNoCurrentOrNoAlias(c);
+        Relation r = current();
+        String a = alias();
+        Relation r0 = r != null ? r : newRelation(a);
+        Relation r1 = relation(c.e1);
+        Relation r2 = relation(c.e2);
+        return addDefinition(new Composition(r0, r1, r2));
+    }
```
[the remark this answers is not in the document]
xeren: I don't define neither composition nor inversion of sets.  I assert that the sub-expressions return relations.

ThomasHaas:
I suggest to clean up the `WmmEncoder` class a little bit.
I think we cannot have methods named like `encodeSAT`, `encodeIDL`, and `encodeEdgeSeq` within such a monolithic class. The methods are specifically made for coherence and read-from and should be named appropriately.
Also, the interleaving of helper methods and encoding methods in the program text is not nice.
I would prefer it we the code was systematically grouped with some comment lines to visualize the grouping.
There are multiple ways to do so. I think it is best to keep logically related code closely to each other so that, e.g., `visitMemoryOrder` is followed by `encodeIDL/SAT`.

Also, I'm not too sure about the naming convention. Why do we call the coherence order by memory order now? Wouldn't it be better to stay close to the standard names?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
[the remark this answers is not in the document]
xeren: I'd like to move those changes into #330, to avoid conflicts.  The logging time stamps were useful in debugging this and the other PR.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/ReadModifyWrites.java
xeren: Better move it to `RelationAnalysis`

xeren:
Regarding the order, I followed a field-constructor-method-first, accessibility-second approach.  I would like to inline those methods, since they are basically named blocks (one usage) and currently have about 20 similar lines of code.

As for the naming _Memory Order_ for `co`, I found it easier to explain its semantics to others by referring to it this way.  Also _Coherence_ was a synonym for the _SC-per-Location_ axiom in some literature that I consulted back then.

ThomasHaas:
> Regarding the order, I followed a field-constructor-method-first, accessibility-second approach. I would like to inline those methods, since they are basically named blocks (one usage) and currently have about 20 similar lines of code.

That order is fine on the macro-scopic level. But the ordering among the methods can still be smarter.
Also, inlining just because there is a single usage is not necessarily good. Code sharing always comes with coupling, so if you inline those functions to share code then whenever you want to adapt the SAT-encoding in the future, you will always have to think about the IDL encoding as well (and vice versa) due to their common code.
In other words: certain degrees of code duplication are actually good! It would be a different case if the common code of both methods needs to be consistent, i.e., updating one but not the other would break something. This is IMO not the case here.
Anyway, you can merge the functions if you want and if needed we will just separate them again.

> As for the naming _Memory Order_ for `co`, I found it easier to explain its semantics to others by referring to it this way. Also _Coherence_ was a synonym for the _SC-per-Location_ axiom in some literature that I consulted back then.

The name "memory order" is fine by itself, but it is unnecessarily inconsistent if we use "coherence" elsewhere.
We should consistently stick to one name in our code base and just add a comment somewhere that says "this is also sometimes called X in the literature".
People are not consistent with their naming in the literature, but that doesn't mean we shouldn't be either.
As far as I can tell, the name `co` seems to be more prevalent considering that most cat files use `co` rather than `mo`, so I would likely stay with "coherence" but that is not a "strict" preference.

xeren:
In #330, code distance in `WmmEncoder` was optimized and  `MemoryOrder` was renamed to `Coherence`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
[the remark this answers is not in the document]
xeren: #330 now includes these changes.

ThomasHaas:
Does this mean you do all the changes on the other branch instead of here? If so, then we can close this branch and directly review the other one.

xeren:
Yes.  The point is, that the missing changes would need to be done a second time when merging `development` into `xra`.

ThomasHaas:
I don't quite understand that. If we merge the changes into `development` and then merge `development` into `xra`, then the changes should end up just fine in `xra`, no? Do you expect merge conflicts that require you to manually merge the changes?

xeren:
Now they should.  Sorry, got fixated on my strategy.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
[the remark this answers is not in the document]
xeren: Now this PR also has them.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -14,8 +14,8 @@
 import com.dat3m.dartagnan.program.filter.FilterBasic;
 import com.dat3m.dartagnan.utils.dependable.DependencyGraph;
 import com.dat3m.dartagnan.verification.VerificationTask;
+import com.dat3m.dartagnan.wmm.Relation;
```
hernanponcedeleon, line 17: Unused import

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/DataRaceSolver.java
```diff
@@ -47,6 +47,7 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
 
 		memoryModel.configureAll(config);
 		preprocessProgram(task, config);
+		preprocessMemoryModel(task);
```
hernanponcedeleon, line 50: `program` variable is unused

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/IncrementalSolver.java
```diff
@@ -45,6 +45,7 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
 
         memoryModel.configureAll(config);
         preprocessProgram(task, config);
+        preprocessMemoryModel(task);
```
hernanponcedeleon, line 48: `program` variable is unused

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/TwoSolvers.java
```diff
@@ -46,6 +46,7 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
 
         memoryModel.configureAll(config);
     	preprocessProgram(task, config);
+        preprocessMemoryModel(task);
```
hernanponcedeleon, line 49: `program` variable is unused

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Empty.java
```diff
@@ -20,22 +21,25 @@ public Empty(Relation rel) {
...
     @Override
-    public BooleanFormula consistent(Set<Tuple> toBeEncoded, EncodingContext ctx) {
+    public BooleanFormula consistent(EncodingContext ctx) {
     	BooleanFormulaManager bmgr = ctx.getBooleanFormulaManager();
 		BooleanFormula enc = bmgr.makeTrue();
-        for (Tuple tuple : toBeEncoded) {
-            enc = bmgr.and(enc, bmgr.not(ctx.edge(rel, tuple)));
+        final RelationAnalysis ra = ctx.getAnalysisContext().get(RelationAnalysis.class);
+        final EncodingContext.EdgeEncoder edge = ctx.edge(rel);
+        for (Tuple tuple : ra.getKnowledge(rel).getMaySet()) {
```
hernanponcedeleon, line 35: I think `toBeEncoded` used to be the encodeTupleSet. Here we iterate over the may set. Is this correct?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Irreflexive.java
```diff
@@ -24,26 +27,27 @@ public Irreflexive(Relation rel) {
...
 
     @Override
-    public BooleanFormula consistent(Set<Tuple> toBeEncoded, EncodingContext ctx) {
+    public BooleanFormula consistent(EncodingContext ctx) {
     	BooleanFormulaManager bmgr = ctx.getBooleanFormulaManager();
 		BooleanFormula enc = bmgr.makeTrue();
-        for (Tuple tuple : toBeEncoded) {
+        final RelationAnalysis ra = ctx.getAnalysisContext().get(RelationAnalysis.class);
+        final EncodingContext.EdgeEncoder edge = ctx.edge(rel);
+        for (Tuple tuple : ra.getKnowledge(rel).getMaySet()) {
```
hernanponcedeleon, line 41: I think `toBeEncoded` used to be the encodeTupleSet. Here we iterate over the may set. Is this correct?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -81,93 +53,83 @@ public Set<Relation> getRelations() {
...
-            } else {
-                Constructor<?> constructor = cls.getConstructor(argClasses);
-                relation = (Relation) constructor.newInstance(args);
-                addRelation(relation);
-            }
-            return relation;
-        } catch (Exception e) {
-            e.printStackTrace();
+        if (!RelationNameRepository.contains(name)) {
             return null;
```
hernanponcedeleon, line 62: Isn't it dangerous to return `null`? Wouldn't be better that throw an exception if this is not expected?

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -118,7 +118,9 @@ private void buildVerdict(Program p, Result result, ProverEnvironment prover, So
         Model model = (result == FAIL && !p.getAss().getInvert()) || (result == PASS && p.getAss().getInvert()) ? prover.getModel() : null;
     	for(Axiom ax : wmm.getAxioms()) {
         	if(ax.isFlagged() && model != null && TRUE.equals(model.evaluate(CAT.getSMTVariable(ax, ctx)))) {
-        		sb.append("Flag " + (ax.getName() != null ? ax.getName() : ax.getRelation().getName())).append("\n");
+        		sb.append("Flag ")
+                        .append(ax.getName() != null ? ax.getName() : ax.getRelation().getNameOrTerm())
```
hernanponcedeleon, line 122: Optional.ofNullable(ax.getName()).orElse(ax.getRelation().getNameOrTerm())

hernanponcedeleon (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -208,17 +168,17 @@ public BooleanFormula visitDifference(Relation rel, Relation r1, Relation r2) {
...
+            final RelationAnalysis.Knowledge k1 = ra.getKnowledge(r1);
+            final RelationAnalysis.Knowledge k2 = ra.getKnowledge(r2);
+            for (Tuple tuple : encodeSets.get(rel)) {
                 BooleanFormula expr = bmgr.makeFalse();
-                if (minSet.contains(tuple)) {
+                if (k.getMustSet().contains(tuple)) {
                     expr = execution(tuple);
                 } else {
-                    for (Tuple t1 : r1Set.getByFirst(tuple.getFirst())) {
+                    for (Tuple t1 : k1.getMaySet().getByFirst(tuple.getFirst())) {
```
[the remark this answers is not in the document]
ThomasHaas: The `maySet` should be logically correct here. What is true, however, is that we should never iterate over a may-tuple that is unknown (no knowledge) and not encoded. The active constraints propagation should guarantee this, but maybe we could add an assertion to check this as well

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Empty.java
ThomasHaas: I think this is fine here, because `encodeSet=mayset` for the `Emptiness` axiom. It might be nice to have a comment that mentions this.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Irreflexive.java
ThomasHaas: The code does effectively iterate over the encode set (cause it looks for loops in the may set).
I think it would be nicer to iterate over `getEncodeTupleSet` to make the logic more clear, even if this creates a second set (which is only linear in the program, so the overhead is non-existent).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: This is intended behaviour if you query for a non-existent relation.
One could return `Optional<Relation>` instead, but I think returning `null` is also fine.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodeSets.java
```diff
@@ -0,0 +1,153 @@
...
+                Event e = t.getSecond();
+                if (lock.getCId() < e.getCId() && e.getCId() < unlock.getCId()) {
+                    queue.add(t);
+                }
+            }
+        }
+        return Map.of(rscs, queue.stream());
+    }
+
+    private Stream<Tuple> lower(Collection<Tuple> news, Relation relation) {
```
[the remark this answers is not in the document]
ThomasHaas: I agree. The method returns a stream of the unknwon entries in `news`.
So `extractUnknowns` or `extractUnknownTuples` would be a more descriptive name.
I think it would also be nice to have utility methods in `Knowledge` like `isUnknown` and maybe even `isStatic` (= in must-set) and `isPossible` (= in may-set).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
```diff
@@ -0,0 +1,347 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    public Void visitLetDefinition(LetDefinitionContext ctx) {
+        String n = ctx.n.getText();
+        alias = ctx.n.getText();
+        Object o = ctx.e.accept(this);
+        namespace.put(n, o);
```
ThomasHaas, line 67: I would name the local variables properly. I assume "n" stands for "name" but what does "o" stand for? "Object"?
That is too non-descriptive. You can call it "definedPred" or simply "pred" (I prefer the longer version) to at least indicate that we define a predicate (unary or binary) here.

Also, you can put `alias = n` instead of reusing the context (whose `getText` method has a non-descriptive name).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
```diff
@@ -0,0 +1,347 @@
...
+
+    private Relation current() {
+        Relation r = current;
+        current = null;
+        return r;
+    }
+
+    private void checkNoCurrentOrNoAlias(ExpressionContext c) {
+        verify(current == null || alias == null,
+                "Broken structure invariant at expression: {}", new Object() {@Override public String toString() { return c.getText(); }});
```
ThomasHaas, line 281: Why does this code generate an instance of an anonymous class?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Definition.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public abstract class Definition implements Constraint {
+
+    protected final Relation definedRelation;
+    final String term;
```
ThomasHaas, line 16: What is this `term` variable used for?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Definition.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private String getTerm(int depth) {
+        List<Relation> l = getConstrainedRelations();
+        int s = l.size() - 1;
+        Object[] o = new Object[s];
+        for (int i = 0; i < s; i++) {
+            Relation r = l.get(i + 1);
+            o[i] = depth == 0 ? r.definition.getTerm(depth - 1) : r.getName() != null ? r.getName() : "(" + r.definition.getTerm(depth - 1) + ")";
+        }
+        return String.format(term, o);
```
ThomasHaas, line 104: Something is wrong here ` o[i] = depth == 0 ? r.definition.getTerm(depth - 1) : ...`. It recursively descends if the depth is 0.
Also, rather than tracking some depth to unroll recursive definitions a bounded number of times, I think it might be better to use a `Set<Definition>` to track which definitions are recursively visited and place some placeholder name there to indicate that the definition can be unfolded there an unlimited amount of times.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -38,14 +89,24 @@ private RelationAnalysis(VerificationTask task, Context context, Configuration c
...
-        a.run(task, context);
+        a.run(task.getMemoryModel());
         return a;
     }
 
+    /**
+     * Fetches results of this analysis.
+     *
+     * @param relation Some element in the associated task's memory model.
+     * @return Local fraction of results.
```
ThomasHaas, line 104: "Computed knowledge of relation" or something like that would be better.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -51,7 +43,7 @@ public class WmmEncoder implements Encoder {
     private static final Logger logger = LogManager.getLogger(WmmEncoder.class);
 
     private final EncodingContext context;
-    private boolean isInitialized = false;
```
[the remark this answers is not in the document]
xeren: This can be removed, it was not reused.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodeSets.java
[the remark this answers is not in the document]
xeren: I'd prefer `filterUnknowns`.  'Extract' could be read as removing the tuples from `news`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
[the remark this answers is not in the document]
xeren: A `verify` is possible, the information is available there.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Irreflexive.java
xeren: Ok, although the coupled code was located 10 lines from here. dry

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
xeren: I could let it throw and provide a method `hasRelation(String)` for querying.  Most of the callers expect an object and would throw an ambiguous NullPointer later.

ThomasHaas: If you do this, you have to be careful about the semantics of `hasRelation(String)` for base relations that have not been "created" yet (i.e. those that would get created implicitly).

Honestly, I'm not really a fan of the implicit construction of named base relations upon querying to begin with. But I don't know how much code actually relies on this and how hard it is to get rid of that.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
xeren: To avoid calling `getText()` when the assertion holds.  I also misused log4j's formatting syntax here.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Definition.java
xeren: It is more a `termPattern` at this point.  It replaces overriding `toString()` in all specializations of `Definition`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
ThomasHaas: If you want to avoid calling `getText()`, I would rather do a normal if-conditional to check for the invariant (just like `checkNoCurrent` does).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Definition.java
xeren: That's indeed a bug.  I think a fixed-point algorithm is immoderate in this case.  Unnamed recursive definitions still should not occur in this project, but if they do, diagnostics should not fail.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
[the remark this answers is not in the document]
xeren: It might be expensive to check the union of `activeSet` and `mustSet`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
ThomasHaas: Your suggestion is also good.

In general though, I don't think that the java docs necessarily need to say more than the function name + signature itself (unless there is more to tell).
The reason is simply because well-designed function names may already tell the "whole story".
In that case, changing the docs just to avoid redundancy is not benefitial IMO.
I guess this is also a reason why I avoid writing docs and instead make the function itself more descriptive, if possible.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Definition.java
ThomasHaas: The "fixed-point" algorithm is simpler than arbitrarily bounding the depth, no?
I would just do something like `getTermRecursive(Set<Definition> alreadyVisited)` and then define `getTerm() = getTermRecursive(new HashSet())`.
The check whether to "descend" then changes from `depth == 0` to `alreadyVisited.contains(...)`.
I would say the code amount stays the same but the logic gets more straightforward and less arbitrary (the depth bound is kinda arbitrary).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
xeren: `getRelation(String)` now throws an InvalidArgument in those cases.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorBase.java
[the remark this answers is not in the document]
xeren: It now throws `VerifyException`, to emphasize that this is the programmer's fault.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -195,17 +180,24 @@ public BooleanFormula visitDifference(Relation rel, Relation r1, Relation r2) {
...
+                if (k.getMustSet().contains(tuple)) {
                     expr = execution(tuple);
                 } else {
-                    for (Tuple t1 : r1Set.getByFirst(tuple.getFirst())) {
-                        Tuple t2 = new Tuple(t1.getSecond(), tuple.getSecond());
-                        if (r2Set.contains(t2)) {
+                    Event x = tuple.getFirst();
+                    Event z = tuple.getSecond();
+                    Set<Tuple> a1 = Sets.union(encodeSets.get(r1), k1.getMustSet());
+                    Set<Tuple> a2 = Sets.union(encodeSets.get(r2), k2.getMustSet());
```
ThomasHaas, line 194: These sets can be constructed outside of the loop (the view construction is cheap, but still).

PR #318, opened 2022-09-30
Relation visitor

xeren:
Part 2 of #315 adds a visitor for relations and the `Constraint` class to unify axioms with the coming-soon definitions.  Different from the original, axioms are excluded from the visitor.


ThomasHaas:
The remaining bugs are likely due to the changes to `RelTrans` and `RelTransRef` (which is gone now) as all affected benchmarks involve PPC or LKMM.

ThomasHaas:
The last remaining bug affects only LKMM and only the eager encoding (Refinement gives correct results).
The fact that we get correct results under Power suggests that the encodings of recursion, `RelTrans`, and `RelTransRef` (its translation to `RelTrans` + `id`) are not fundamentally broken.
The fact that Refinement works means that the parsing of the memory model is also working.

So I cannot really pinpoint where the bug with LKMM is for now.

xeren:
My bad!  The implementation in `Relation` was missing.  You might just have missed the dependency edge from `RecursiveRelation`.  Your fix in `RelTrans` seems correct.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/relation/Relation.java
```diff
@@ -114,6 +121,16 @@ public boolean getIsNamed(){
...
+    @Override
+    public Collection<? extends Relation> getConstrainedRelations() {
+        return accept(new Visitor<>() {
+            @Override
+            public Collection<? extends Relation> visitDefinition(Relation rel, List<? extends Relation> dependencies) {
+                return dependencies;
+            }
+        });
+    }
+
```
ThomasHaas, line 133: I kinda dislike the use of a visitor here.
In general, I think the constrained relations can be assumed to be ordered and hence should be representable by a List<Relation>.
Then, for definitions we could simply maintain the invariant that the first or last entry is the defined relation and the other one's are the dependencies.
This simplifies the handling IMO.

xeren: The visitor always gets all constrained relations.  There is not need for an extra override-tree.  But the list could be helpful for other users.

ThomasHaas: I understand that and for now this might be ok.
But I'm thinking that eventually all definitions will have an explicit list of their constrained relations (for base definitions, it might be a singleton) which they simply can return.
Either way, for now this is ok.

ThomasHaas (approved):
LGTM

PR #322, opened 2022-10-06
Relation encoder

xeren:
Moves code around, such that `WmmEncoder` controls how the encode tuple sets translate into defining SMT propositions.  The remaining WMM encoding is still handled by `Axiom`.

The `RelCo.encodeLastCoConstraints()` is moved into `PropertyEncoder`.  Therefore, `WmmEncoder` does not need to consult the property.

Part of #315.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
@@ -15,11 +15,14 @@ import com.dat3m.dartagnan.utils.dependable.DependencyGraph;
 import com.dat3m.dartagnan.verification.VerificationTask;
 import com.dat3m.dartagnan.wmm.Wmm;
 import com.dat3m.dartagnan.wmm.relation.Relation;
-import com.dat3m.dartagnan.wmm.relation.base.memory.RelCo;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
+import org.sosy_lab.common.configuration.Configuration;
+import org.sosy_lab.common.configuration.InvalidConfigurationException;
+import org.sosy_lab.common.configuration.Option;
+import org.sosy_lab.common.configuration.Options;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 import org.sosy_lab.java_smt.api.Model;
 import org.sosy_lab.java_smt.api.SolverContext;
@@ -28,6 +31,7 @@ import java.math.BigInteger;
 import java.util.*;
 import java.util.stream.Collectors;
 
+import static com.dat3m.dartagnan.configuration.OptionNames.IDL_TO_SAT;
 import static com.dat3m.dartagnan.wmm.relation.RelationNameRepository.CO;
 import static com.dat3m.dartagnan.wmm.relation.RelationNameRepository.RF;
 import static com.dat3m.dartagnan.wmm.utils.Utils.coClockVar;
@@ -38,6 +42,7 @@ The ExecutionModel wraps a Model and extracts data from it in a more workable ma
 
 //TODO: Add the capability to remove unnecessary init events from a model
 // i.e. those that init some address which no read nor write accesses.
+@Options
 public class ExecutionModel {
 
     private final VerificationTask task;
@@ -86,9 +91,17 @@ public class ExecutionModel {
 
     private Map<BigInteger, List<EventData>> coherenceMapView;
 
+    // =========================== Configurables ===========================
+
+    @Option(
+            name=IDL_TO_SAT,
+            description = "Use SAT-based encoding for totality and acyclicity.",
+            secure = true)
+    private boolean useSATEncoding = false;
+
     //========================== Construction =========================
 
-    public ExecutionModel(VerificationTask task) {
+    private ExecutionModel(VerificationTask task) {
         this.task = task;
 
         eventList = new ArrayList<>(100);
@@ -108,6 +121,12 @@ public class ExecutionModel {
         coherenceMap = new HashMap<>();
 
         createViews();
+    }
+
+    public static ExecutionModel fromConfig(VerificationTask task, Configuration config) throws InvalidConfigurationException {
+        ExecutionModel m = new ExecutionModel(task);
+        config.inject(m);
+        return m;
     }
 
     private void createViews() {
@@ -485,7 +504,7 @@ public class ExecutionModel {
             final Set<EventData> writes = addrWrites.getValue();
 
             List<EventData> coSortedWrites;
-            if (co.usesSATEncoding()) {
+            if (useSATEncoding) {
                 // --- Extracting co from SAT-based encoding ---
                 Map<EventData, List<EventData>> coEdges = new HashMap<>();
                 for (EventData w1 : writes) {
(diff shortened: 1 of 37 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Axiom.java
```diff
@@ -80,7 +82,7 @@ public void setName(String name) {
 
     public abstract TupleSet getEncodeTupleSet();
 
-    public abstract BooleanFormula consistent(SolverContext ctx);
+    public abstract BooleanFormula consistent(Set<Tuple> toBeEncoded, Context analysisContext, SolverContext ctx);
```
ThomasHaas, line 85: I think the `toBeEncoded` parameter is weird, but so is our current solution :).
Eventually, this parameter will go, because there is simply no other entity in the code base that should change this parameter. Conceptually, it should always coincide with the return value of `getEncodeTupleSet` (right now it doesn't, cause our current code is conceptually wrong).

ThomasHaas (commented):
Overall, I'm fine with these changes for now.
I think some aspects like having rf/co/rmw be handled like other definitions in the WmmEncoder should eventually get changed. In particular, I think they should be modelled as Axioms/Constraints and also be Encoders, because their semantics cannot be defined point-wise like other definitions.
That being said, for now the PR is quite true to the current code base and I believe that these design decision can be deferred until we get the new analysis framework going.
I would still get rid of the `POSTFIX` option before we merge this though :)

ThomasHaas (approved):
LGTM

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Axiom.java
xeren: This is not the final signature, but it will ease the change later, when encode sets get moved.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -40,6 +46,42 @@ public static RelationAnalysis fromConfig(VerificationTask task, Context context
...
+            - Since co(w1, w3) is a must-edge, we have that w3 accesses the same address as w1 and w2,
+              and c(w1) < c(w3).
+            - Because addr(w2)==addr(w3), we must also have either co(w2, e3) or co(w3, w2).
+              The former is disallowed by assumption, so we have co(w3, w2) and hence c(w3) < c(w2).
+            - By transitivity, we have c(w1) < c(w3) < c(w2) as desired.
+            - Note that this reasoning has to be done inductively, because co(w1, w3) or co(w3, w2) may
+              not involve encoding a clock constraint (due to this optimization).
+        There is also a symmetric case where co(w3, w1) is impossible and co(w3, w2) is a must-edge.
+     */
+    public static Set<Tuple> findTransitivelyImpliedCo(Relation co, ExecutionAnalysis exec) {
```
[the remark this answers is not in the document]
xeren: Yes.  I should have left it where it was.  `Encoder` will be a better candidate.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -109,6 +123,12 @@ public ExecutionModel(VerificationTask task) {
         createViews();
     }
 
+    public static ExecutionModel fromConfig(VerificationTask task, Configuration config) throws InvalidConfigurationException {
+        ExecutionModel m = new ExecutionModel(task);
+        config.inject(m);
+        return m;
+    }
+
```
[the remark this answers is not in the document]
xeren: I'll note this for when `Encoder` is ready.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -338,31 +337,21 @@ private static Relation getCopyOfRelation(Relation rel, Wmm m) {
...
+        @Override public Relation visitComposition(Relation rel, Relation r1, Relation r2) { return new RelComposition(copy(r1), copy(r2)); }
+        @Override public Relation visitInverse(Relation rel, Relation r1) { return new RelInverse(copy(r1)); }
+        @Override public Relation visitDomainIdentity(Relation rel, Relation r1) { return new RelDomainIdentity(copy(r1)); }
+        @Override public Relation visitRangeIdentity(Relation rel, Relation r1) { return new RelRangeIdentity(copy(r1)); }
+        @Override public Relation visitTransitiveClosure(Relation rel, Relation r1) { return new RelTrans(copy(r1)); }
+        @Override public Relation visitIdentity(Relation rel, FilterAbstract filter) { return new RelSetIdentity(filter); }
+        @Override public Relation visitProduct(Relation rel, FilterAbstract f1, FilterAbstract f2) { return new RelCartesian(f1, f2); }
+        @Override public Relation visitFences(Relation rel, FilterAbstract type) { return new RelFencerel(type); }
+        private Relation copy(Relation r) { return getCopyOfRelation(r, targetModel); }
```
hernanponcedeleon, line 354: Does this somehow interfere with with the PR from @ThomasHaas for solving the problem related to cutting?

hernanponcedeleon (commented):
I like to see all the encoding stuff out of `Relation` :D

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
xeren: It should not be related, other than refactoring some part two lines before that.

xeren: Assuming you referred to #325.

xeren: Automatic merging could have missed this.  However, the visitor still does not add the copy to the model by itself.

(Not exactly refactoring: Trying to copy an unnamed relation that has neither of the listed definitions (i.e. `RecursiveRelation`) will now throw a `NullPointer` instead of `UnsupportedOperation`.)

hernanponcedeleon: I think the visitor will simply create a new instance of `Relation` and then the fix in #325 will take care of adding the relation to the memory model. I just wanted to double check that this would not again cause an issue that the CI did not noticed (because we don't have much tests for cutting).

I do not get your last comment about throwing a `NullPointer` exception instead of a `UnsupportedOperation` one.

xeren: The fallback branch in the replaced code threw an `UnsupportedOperationException`.  `Relation.Visitor` falls back to returning `null`.  In that case, nothing is checked (yet), so some parent relations will contain null references in the copied model.

hernanponcedeleon: Instead of checking if `copy` returned null, wouldn't it make more sense to override `visitDefinition` in `RelationCopier` and throw the exception here?

xeren: That would be even better

PR #323, opened 2022-10-06
Encoding Context

xeren:
Replaces most usages of `SolverContext` with `Encoder`, which starts off as basically a record of `VerificationTask`, `Context` and `SolverContext`.  It also provides general facilities for generating SMT formulas for fine-grained elements of the task, like expressions, execution variables and edges.
Previously initialized and cached in one swoop, those elements are now regenerated on each occasion.  This enables users to easily reuse the task when dealing with multiple `SolverContext` instances.  Caching can be reimplemented inside `Encoder`, later.

Requires #322.


hernanponcedeleon:
The CI has been running for almost 2 hs now, so probably there is some infinite loop going on

ThomasHaas:
I haven't checked this PR thoroughly yet, but I can tell you that you can get rid of the `allowPartialExecutions` option completely.
It's been unusable for a long time now and the idea of partial executions itself doesn't interact nicely with most of our reasoning we do.
For example, it is incompatible with all the branch-merging and the optimizations we do based on cf/exec-implications (forward implications do not hold in partial executions).

ThomasHaas:
This PR breaks refinement: The full memory model gets encoded so that Refinement degenerates to the basic assume solver and that is why it takes so much time.
This is likely because `Encoder` gets initialized with the `task` and not the `baselineTask`.
Changing it to `baselineTask` resulted in other problems though (unsound results).

I haven't checked the details, but my guess is that the introduction of `Encoder` made the whole architecture less flexible.
I also have the feeling that the design is inverted: The `Encoder` should actually be an `EncodingContext` that wraps the true encoders (`ProgramEncoder`, `WmmEncoder`, etc.) instead of the other way around. It can then dispatch/forward the calls to the appropriate encoders. The encoders themselves take as input the `EncodingContext` when doing encoding, so that you have a double dispatch structure (encoders -> encoding context -> encoders).
This would be my intuition at least (and also the solution I had in mind originally).
If we keep the current design, I would at least rename the class `Encoder` because it does not encode anything to be honest, it just provides services for encoding.

ThomasHaas:
Btw. I'm not at all a fan of the static `.of(...)` methods. By calling these factory methods `of` you lose one of the big advantages of static factory methods over constructors, namely the possibility to give them proper names.
`of` is only good for simple wrapper objects or collection types, where the intention of the method is clear.
For everything else, it is IMO bad naming convention as it is non-descriptive.

I chose `fromConfig` because it shows that you can provide a context which in principle would not be needed (i.e., you could imagine that there was an alternative version that does not take the config parameter as input). Obviously, we never implemented a "plain" version but the fact that it could exist is IMO sufficient to justify a distinct name.
Now that I think about it, `withConfig` would have been an even better name because the config is provided in addition to the needed parameters. `fromConfig` suggests that the config object replaces all the typical parameters.

xeren:
These method names originate from `.of(Encoder)`, where I initially viewed `Encoder` as a record with utility methods, and `ProgramEncoder` etc. as its decorators.  `fromConfig` or `withConfig` at least suggest that configuration takes place in these methods, even though the object is no longer directly passed to them, but the `throws` clauses already do this job.  `withConfig` sounds more like an operation that modifies an existing object of the respective class by inserting another configuration.

I think the current failure at `litmus/C11/manual/imm-E3.5.litmus` is a parsing misconception, where the memory object `y` actually has to be modeled with at least two fields.
- The test seems to intend an execution where event 4 (Load at line 11) looks up `y[1]`.
- The alias analysis proclaims that the only correct field is `y[0]` and inserts a *must-alias* between events 4 and 15 (Init of `y[0]`).  The proposition of equal addresses is replaced with `true`.  The model lets 4 read from event 13 (Init of `x[0]`), without checking the address.
- The former system returns `PASS`, as 4 cannot read from 15, and has therefore read from 13.  The address values are fixed to x=1 and y=2.  I suspect that if we would allow variable address values for memory objects again, this test would also fail.

I will remove the must-alias check for now.

ThomasHaas:
I agree that the test `litmus/C11/manual/imm-E3.5.litmus` is questionable because from our point of view it contains UB.
IIRC litmus has no notion of UB (just like assembler), but I didn't really expect to see a litmus test where this matters.
I guess @hernanponcedeleon didn't pay attention to UB when writing that test :).

That being said, the expected results for this litmus test is supposed to be PASS, so we accidentally gave a correct result here.
Maybe we should get rid of the litmus test or change it to make explicit that `y` represents a two-element array/struct (not sure if we can specify arrays in litmus).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -151,26 +136,20 @@ public BooleanFormula encodeRelations(SolverContext ctx) {
...
-        final AliasAnalysis alias = analysisContext.get(AliasAnalysis.class);
-        final SolverContext ctx;
         final BooleanFormulaManager bmgr;
-        final IntegerFormulaManager imgr;
-        public RelationEncoder(SolverContext c) {
-            ctx = c;
-            bmgr = c.getFormulaManager().getBooleanFormulaManager();
-            imgr = c.getFormulaManager().getIntegerFormulaManager();
+        public RelationEncoder() {
+            bmgr = context.getFormulaManager().getBooleanFormulaManager();
```
ThomasHaas, line 152: I have seen this access quite a few times now.
I think a direct getter `context.getBmgr()` would make it a little easier :).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -441,12 +417,9 @@ public BooleanFormula visitReadFrom(Relation rf) {
...
                 // In such cases we perform the transformation
-                Formula a1 = w.getMemAddressExpr();
-                Formula a2 = r.getMemAddressExpr();
-                BooleanFormula sameAddress = generalEqual(a1, a2, ctx);
-                Formula v1 = w.getMemValueExpr();
-                Formula v2 = r.getMemValueExpr();
-                BooleanFormula sameValue = generalEqual(v1, v2, ctx);
+                //FIXME litmus/C11/manual/imm-E3.5.litmus fails if mustAlias is checked here
+                BooleanFormula sameAddress = context.equal(context.address(w), context.address(r));
+                BooleanFormula sameValue = context.equal(context.value(w), context.value(r));
```
ThomasHaas, line 422: I would revert this. We shouldn't change the logic here to accommodate for an error we do somewhere else.
In that sense, it's actually good that you tried to use more of the alias analysis to optimize here.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -123,9 +110,9 @@ private ExecutionModel(VerificationTask task) {
         createViews();
     }
 
-    public static ExecutionModel fromConfig(VerificationTask task, Configuration config) throws InvalidConfigurationException {
-        ExecutionModel m = new ExecutionModel(task);
-        config.inject(m);
+    public static ExecutionModel fromConfig(EncodingContext context) throws InvalidConfigurationException {
```
ThomasHaas, line 113: Method should get renamed

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
```diff
@@ -33,46 +30,49 @@ private AssumeSolver(SolverContext c, ProverEnvironment p, VerificationTask t) {
         task = t;
     }
 
-    public static Result run(SolverContext ctx, ProverEnvironment prover, VerificationTask task)
+    public static AssumeSolver of(SolverContext ctx, ProverEnvironment prover, VerificationTask task)
             throws InterruptedException, SolverException, InvalidConfigurationException {
-        return new AssumeSolver(ctx, prover, task).run();
+        AssumeSolver s = new AssumeSolver(ctx, prover, task);
+        s.run();
+        return s;
```
ThomasHaas, line 37: Method should get renamed, in particular since the model checker not only gets created but it also runs!

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -124,12 +121,13 @@ private Result run() throws InterruptedException, SolverException, InvalidConfig
         performStaticWmmAnalyses(task, analysisContext, config);
         performStaticWmmAnalyses(baselineTask, baselineContext, config);
 
-        ProgramEncoder programEncoder = ProgramEncoder.fromConfig(program, analysisContext, config);
-        PropertyEncoder propertyEncoder = PropertyEncoder.fromConfig(program, baselineModel, analysisContext, config);
+        context = EncodingContext.of(task, analysisContext, ctx);
```
ThomasHaas, line 124: Constructing the encoding context with the baseline task should work, no?

hernanponcedeleon:
> I agree that the test `litmus/C11/manual/imm-E3.5.litmus` is questionable because from our point of view it contains UB. IIRC litmus has no notion of UB (just like assembler), but I didn't really expect to see a litmus test where this matters. I guess @hernanponcedeleon didn't pay attention to UB when writing that test :).
> 
> That being said, the expected results for this litmus test is supposed to be PASS, so we accidentally gave a correct result here. Maybe we should get rid of the litmus test or change it to make explicit that `y` represents a two-element array/struct (not sure if we can specify arrays in litmus).

We actually [do support arrays in litmus test](https://github.com/hernanponcedeleon/Dat3M/tree/master/dartagnan/src/test/resources/arrays). However, if I understood @xeren's comment correctly, we should get the correct result using this program (I checked and it is properly parsed as an array of size 2)
```
C imm-E3.5

(*
 * Bridging the Gap between Programming Languages and Hardware Weak Memory Models
 *)

{
  atomic_int y[2] = {0,0};
}

P0 (atomic_int* x, atomic_int* y) {
    int r0 = atomic_load_explicit(x, memory_order_relaxed);
    int r1 = atomic_load_explicit(y+r0, memory_order_relaxed);
    atomic_store_explicit(y, 1, memory_order_relaxed);
}

P1 (atomic_int* x, atomic_int* y) {
    int r0 = atomic_load_explicit(y, memory_order_relaxed);
}

exists (0:r0=1 /\ 1:r0=1)
```
However, even after reverting the FIX, I still get FAIL

ThomasHaas:
I discussed the problem with @hernanponcedeleon and the fixed version proposed by him may actually be correct.
Under `IMM` we get the expected result (`PASS`) but under `C11` we don't.
I suspect that failure under `C11` may indeed be correct, but we couldn't confirm this yet with Herd, because Herd gives qustionable results on the original benchmark (using our local repo version of 'C11'). And it does not seem to be able to parse arrays, so the fixed version cannot be tested with Herd.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
xeren: No problem, but I am opposed to abbreviations in public methods.

ThomasHaas: Sure, you can give it the longer name.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Flag.java
```diff
@@ -14,8 +12,8 @@ public enum  Flag {
     // TODO: Add linux when implemented
     public static ImmutableSet<Flag> all = ImmutableSet.of(ARM_UNPREDICTABLE_BEHAVIOUR);
 
-    public BooleanFormula repr(SolverContext ctx){
-    	return ctx.getFormulaManager().makeVariable(BooleanType, code());
+    public BooleanFormula repr(FormulaManager m){
```
ThomasHaas, line 15: Why the change to `FormulaManager` here?
Is it because you create some new variable that is not directly supported by the `EncodingContext`?
I would still provide the encoding context though.

xeren: This is a remainder of a change I missed to clean up.  This method would never need to create a prover environment, or fetch statistics of the context.  Neither would it need to react to elements of the verification task.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
xeren: It seems to work.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
xeren: I have changed the entry in `expected-C11.csv` *and* modified the litmus test by changing the address expression of the load in question.  I figured this test highlights a `rfe ; addr ; po-loc ; rfe ; po ; [crel]` cycle, which seemed to be allowed by C11, but would be disallowed by machines in one way or another.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Flag.java
xeren: I do not plan on reverting this.  Actually, with this argument even `EncodingContext` would not need a reference to `SolverContext`.  I yet decided against it, because it is unnecessary.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
hernanponcedeleon: I don't like the change to the litmus tests for at least two reasons
- the code you propose has a syntactic, but not semantic, address dependency. This tends to make things difficult in the formalizations of language level WMMs. Thus I would avoid such cases unless we really want to make a point about that particular case (which is not the case in here)
- the whole point of this litmus (and thus its name and the "documentation") was to reproduce the results from the IMM paper. By changing the code this is completely lost.

TBH, I don't even understand why you change the litmus. If we use the array version, we seem to get the correct result (assuming C11 accepts this test, which I think we mostly agree it is the case)

ThomasHaas: I agree that we should not change the test, but IMO the problem is actually not the difference between syntactic and semantics address dependency in this litmus test. What this litmus test highlights is that if `r0=0` holds, then the load and store operation to `y` alias and cannot be reordered with each other (since they go to the same address). If the value is `r0=1`, they access distinct memory and hence can get reordered. Once they are reordered, it is easy to imagine that the store to `y` can then be also moved to before the load from `x` since they are independent.

Well, all this cannot happen on hardware, because hardware needs to establish the value of `x` first, before it can understand if the subsequent operations are accessing different addresses and therefore can be reordered. But either way, none of them can be reordered before the load from `x`.
Strange enough, but not really surprising, C11 does allow this reordering, which requires a sort of speculation on the address of the store.
IMM is an intermediate model that is meant to be close to hardware models, so it intentionally shows the same semantics as hardware models, contrasting it to C11 and also the promising semantics.

So your proposed change makes this aliasing speculation a non-issue since the code will always alias. Indeed, I would expect that your code does not fail under C11 anymore (not sure though).
In any case, I agree totally with @hernanponcedeleon here: We should not touch litmus tests that are directly related to examples from the literature.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Flag.java
ThomasHaas: I'm not sure I can totally follow here. It is true that the full `EncodingContext` is not needed here, but that can also be said about the `FormulaManager`: the code just requires a `BooleanFormulaManager` to generate its flag variable.
So then why choose this "in-between" solution?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -29,6 +31,17 @@
...
 public abstract class ModelChecker {
 
+    protected Result res = Result.UNKNOWN;
+    protected EncodingContext context;
+
+    public final Result result() {
+        return res;
+    }
+
+    public EncodingContext encoding() {
```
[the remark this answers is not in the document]
ThomasHaas: You didn't change this :)

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
xeren: You're right.   I have missed the point.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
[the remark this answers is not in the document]
xeren: I forgot about this.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Flag.java
xeren: I used automatic refactoring.  Although it is not used in this particular place, `BooleanFormulaManager` does not provide methods like `escape(String)` which would be needed on other occasions.

hernanponcedeleon:
As a general comment, there are inconsistencies in the naming schemes for getters, i.e. this PR has things like this
```
    public SolverContext solverContext() {
        return solverContext;
    }

    public FormulaManager getFormulaManager() {
        return formulaManager;
    }
```
I personally think we should prefix every getter with `get`

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
```diff
@@ -304,32 +277,36 @@ private BooleanFormula consistentSAT(Set<Tuple> toBeEncoded, Context analysisCon
...
 
-    private BooleanFormula getSMTCycleVar(Tuple edge, SolverContext ctx) {
-        return edge(getName() + "-cycle", edge.getFirst(), edge.getSecond(), ctx);
+    private BooleanFormula cycleVar(Event event, FormulaManager m) {
+        return m.getBooleanFormulaManager().makeVariable(String.format("cycle %s %d", m.escape(getName()), event.getCId()));
+    }
+
+    private BooleanFormula getSMTCycleVar(Tuple edge, FormulaManager m) {
+        return m.getBooleanFormulaManager().makeVariable(String.format("cycle %s %d %d", m.escape(getName()), edge.getFirst().getCId(), edge.getSecond().getCId()));
     }
```
hernanponcedeleon, line 311: It seems the `BooleanFormulaManager` is available in every call to these methods. Why not to pass the `BooleanFormulaManager` instead of `FormulaManager` directly?

xeren:
> I personally think we should prefix every getter with `get`

No problem.  Note, however, that records will not follow this scheme, once we reach language level 14.  The `get` prefix tells me that the value is modifiable, i.e. that there should also be a setter (not necessarily public).
I kept `getFormulaManager()` because the name was already used, but I would have preferred `formulaManager()`.

I could take the extra step and add the prefix `encode` to the utility funtions in `EncodingContext`, as well.  Should I?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private static final Logger logger = LogManager.getLogger(EncodingContext.class);
+
+    private final VerificationTask verificationTask;
+    private final Context analysisContext;
+    private final ExecutionAnalysis executionAnalysis;
+    private final AliasAnalysis aliasAnalysis;
+    private final SolverContext solverContext;
+    private final FormulaManager formulaManager;
+    private final BooleanFormulaManager booleanFormulaManager;
```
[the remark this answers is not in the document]
xeren: Those two have multiple usages, in utility methods that are called thousands of times.  Since that class usually has one instance per runtime, the storage costs are almost non-existent.  We won't even save one import line by removing it.

hernanponcedeleon:
> > I personally think we should prefix every getter with `get`
> 
> No problem. Note, however, that records will not follow this scheme, once we reach language level 14. The `get` prefix tells me that the value is modifiable, i.e. that there should also be a setter (not necessarily public). I kept `getFormulaManager()` because the name was already used, but I would have preferred `formulaManager()`.

My preference for the get prefix is not that strong (TBH in some cases it event looks too verbose), but we you definitely try to be consistent with our naming scheme. 

> 
> I could take the extra step and add the prefix `encode` to the utility funtions in `EncodingContext`, as well. Should I?

You mean something like `sameAddress()` -> `encodeSameAddress()`? Again this might be too verbose. Not only because in many cases there is really no encoding involved (we just return a variable), but also because the "encoding notion" is already in the class.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
xeren: Because I use `escape` here.

xeren: I could assert `isValidName` instead.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
```diff
@@ -293,20 +294,20 @@ private BooleanFormula consistentSAT(Set<Tuple> toBeEncoded, EncodingContext con
...
-    private BooleanFormula cycleVar(Event event, FormulaManager m) {
-        return m.getBooleanFormulaManager().makeVariable(String.format("cycle %s %d", m.escape(getName()), event.getCId()));
+    private BooleanFormula cycleVar(Event event, FormulaManager fmgr, BooleanFormulaManager bmgr) {
+        return bmgr.makeVariable("cycle_" + fmgr.escape(getName()) + "_" + event.repr());
     }
 
-    private BooleanFormula getSMTCycleVar(Tuple edge, FormulaManager m) {
-        return m.getBooleanFormulaManager().makeVariable(String.format("cycle %s %d %d", m.escape(getName()), edge.getFirst().getCId(), edge.getSecond().getCId()));
+    private BooleanFormula getSMTCycleVar(Tuple edge, FormulaManager fmgr, BooleanFormulaManager bmgr) {
+        return bmgr.makeVariable("cycle_" + fmgr.escape(getName()) + "_" + edge.getFirst().repr() + "_" + edge.getSecond().repr());
```
hernanponcedeleon, line 311: I don't think passing a second argument was a good idea. The intention of my comment was "why do we need `fmgr` and not pass JUST `bmgr`?". Since we need `fmgr` for the `escape` and we can get `bmgr` from `fmgr`, I don't see the point in the second argument

xeren: Then I will revert this.

PR #324, opened 2022-10-06
Make dartagnan compatible with herd in the precedence order of operators

hernanponcedeleon:
Until now, dartagnan had a different precedence order than herd for certain operators, e.g. it interpreted relation `r1 ; r2 \ r3` as `(r1 ; r2) \ r3` while herd interprets this as `r1 ; (r2 \ r3)`.

The [CAT paper](https://arxiv.org/abs/1608.07531) does not specify the precedence order, but for compatibility reasons, we agreed that we should follow herd.

This PR makes dartagnan compatible with herd w.r.t the precedence order defined [here](https://github.com/herd/herdtools7/blob/4eaabb3ed445eab4c6b76174c3ad0e6ebec6d127/lib/modelParser.mly#L64-L71).


ThomasHaas (approved):
LGTM

PR #326, opened 2022-10-13
Cf-based symmetry breaking

ThomasHaas:
This PR adds the possibility to break symmetry on the control flow (technically execution flow) rather than relations.
This feature is enabled by setting the breaking relation to `_cf` (underscore to indicate that this is a special option and so that it does not interfere with relations named `cf` in the Wmm).


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
```diff
@@ -39,8 +41,9 @@ public class SymmetryEncoder implements Encoder {
     private final Relation rel;
 
     @Option(name = BREAK_SYMMETRY_ON_RELATION,
-            description = "The relation on which symmetry breaking should happen." +
-            "Empty, if symmetry shall not be encoded.",
+            description = "The relation which should be used to break symmetry." +
+            "Empty, if symmetry shall not be encoded." +
+            "Special value \"_cf\" allows for breaking on control flow rather than relations.",
             secure = true)
```
hernanponcedeleon, line 47: While implementation-wise, this is always a relation, at the abstract level it is not. I would rename the option name to `encoding.symmetry.breakOn`, and use `Concept (any better option) to break symmetry on. Allowed options ... (descriptions). Empry if symmetry shall not be used`.

What happen if someone uses a relation different than `rf` and `co`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
```diff
@@ -81,59 +90,56 @@ public static SymmetryEncoder withContext(EncodingContext context, Wmm memoryMod
...
-        if (rel == null || symmClass.getEquivalence() != symm) {
-            return enc;
-        }
-
-        Thread rep = symmClass.getRepresentative();
-        List<Thread> symmThreads = new ArrayList<>(symmClass);
+    private BooleanFormula encodeSymmetryClass(Relation rel, EquivalenceClass<Thread> symmClass) {
+        Preconditions.checkArgument(symmClass.getEquivalence() == symm,
+                "Symmetry class belongs to unexpected symmetry relation.");
+        Preconditions.checkState(rel != null, "Relation to break symmetry on is NULL");
```
hernanponcedeleon, line 107: relation -> concept (or whatever used in the description)

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
ThomasHaas: Regarding your question, you can use any relation you want. If the specified relation is not existent in the memory model, a warning is issued and the symmetry breaking gets disabled. In particular, for all non-sensical inputs, the symmetry breaking gets disabled.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
ThomasHaas: This is private code which does take a relation as input. The message is not intended to be seen by users.
But its good you pointed it out, because it should not be a state condition anymore :)

PR #327, opened 2022-10-13
Add fallback to Axiom.getName() to avoid null

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Axiom.java
```diff
@@ -71,7 +71,7 @@ public boolean isFlagged() {
     }
 
     public String getName() {
-        return name;
+        return name != null ? name : rel.getName();
```
ThomasHaas, line 74: You should not return the name of the underlying relations but at least "acyclic(rel.getName)".
Basically what `toString` would do.
This begs the question if we should rather have a dedicated `getNameOrTerm` method that returns either the name or the term.
That way, `getName` can still return `null` for unnamed relations.

ThomasHaas: No, I had a typo. I was talking about axioms, which can also be named like e.g. `acyclic(hb) as sc`. If they are not named, they still have a term like `acyclic(hb)` associated.

PR #330, opened 2022-11-03
Extended Relation Analysis

xeren:
Adds top-down propagation to the relation analysis, as well as multiple options to existing features.


ThomasHaas:
I tried to add push a commit to trigger the CI, but somehow it doesn't work.

xeren:
Thanks Thomas.  The failing test exposed another invalid premise of mine: The program contains a load event that must dereference a null pointer.  Therefore there is not even an initialization event it could read from.  The verifier concludes that there is no consistent execution that executes that load.

ThomasHaas:
Does this PR include all changes from #315 as well?

xeren:
> Does this PR include all changes from #315 as well?

It does.  #315 should be merged before this one.

ThomasHaas:
What is the intention behind changing `BooleanFormula encode(...)` to `List<BooleanFormula> encode(...)`?
Does it make the caller code in any way easier (which I doubt)? While I can see the callee using a `List<BooleanFormula>` for ease-of-implementation, I don't see the caller getting a benefit from getting a `List<BooleanFormula>` rather than a single formula (in fact, I don't think the caller needs to know that this list is supposed to be a conjunction).

If you expect any performance benefit from creating n-ary expressions instead of nested binary expressions, then I don't think you will see anything. The formula construction time might increase at best, but it is unlikely to affect solving in any way.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
```diff
@@ -92,35 +89,34 @@ public void add(Thread t) {
...
     public List<Thread> getThreads() {
         return threads;
     }
 
-	public List<Event> getEvents(){
-        // TODO: Why don't we use the cache if available?
+    /**
+     * Iterates all events in this program.
+     *
+     * @return {@code cId}-ordered complete sequence of all events in this program.
```
ThomasHaas, line 99: `globalId`, though I think the ids should not even be mentioned.
This is because the events are not ordered along the id, but the ids are assigned to match the order of the events.
If you want to be accurate, the event sequence is ordered by thread id and (syntactic) successor relationship.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
```diff
@@ -92,35 +89,34 @@ public void add(Thread t) {
...
+			events.addAll(t.getEvents());
 		}
 		return events;
 	}
 
+    /**
+     * Iterates a subset of events in this program.
+     *
+     * @param cls Class of events to be selected.
+     * @return {@code cId}-ordered complete sequence of all events of class {@code cls} in this program.
```
ThomasHaas, line 113: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -45,18 +42,20 @@ public class AndersenAliasAnalysis implements AliasAnalysis {
...
-
+    private final Map<Object, Set<Object>> edges = new HashMap<>();
+    private final Map<Object, Set<Location>> addresses = new HashMap<>();
+    private final Map<Register, Set<MemEvent>> events = new HashMap<>();
+    private final Map<Register, Set<Location>> targets = new HashMap<>();
     private final Map<MemEvent, ImmutableSet<Location>> eventAddressSpaceMap = new HashMap<>();
 
     // ================================ Construction ================================
 
     private AndersenAliasAnalysis(Program program) {
```
ThomasHaas, line 53: (NOTE: This comment refers to the whole class)

Is this just a reimplementation of the `AndersenAliasAnalysis` to be more similar to the field sensitive one?
Or does this rework involve any relevant changes (e.g., semantic change or performance improvement)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Event.java
```diff
@@ -227,16 +226,19 @@ public <T> T accept(EventVisitor<T> visitor) {
...
 
-	public BooleanFormula encodeExec(EncodingContext ctx) {
-		return ctx.getBooleanFormulaManager().makeTrue();
+	/**
+	 * Conditions related to the execution variable of this event.
+	 *
+	 * @return Executions of the associated program must meet all those assumptions.
+	 */
+	public Collection<BooleanFormula> encodeExec(EncodingContext ctx) {
+		return List.of();
```
ThomasHaas, line 242: You kept `BooleanFormula` for almost all methods, but this one returns a `Collection<BooleanFormula>`.
I think no (public) encoding function should return a collection of formulas if it already knows that those formulas must get conjoined.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
```diff
@@ -33,16 +49,159 @@ public Acyclic(Relation rel) {
...
+        public String toString() {
+            int i = 0;
+            for (Axiom a : axioms) {
+                if (a instanceof Acyclic) {
+                    i += ((Acyclic) a).getEncodeTupleSet(exec, ra).size();
+                }
+            }
+            return "Number of encoded tuples for acyclicity: " + i;
+        }
+    }
```
ThomasHaas, line 73: I find this class awkward. Is the only reason it is placed here to access `getEncodeTupleSet`?
If that is the case, then a simple `getEncodingSize` function would be easier and more flexibie, because the caller can decide which axioms to track and how to present/print it.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodeSets.java
```diff
@@ -6,21 +6,17 @@
...
+import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 // Collections of relationships in a verification task that need to be constrained in an SMT-based verification.
 final class EncodeSets implements Visitor<Map<Relation, Stream<Tuple>>> {
 
     private final RelationAnalysis ra;
     List<Tuple> news;
+    Map<Relation, Function<Event, Collection<Tuple>>> outSetMap = new HashMap<>();
```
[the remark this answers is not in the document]
xeren: Those functions indeed encapsulate closures with an eagerly-constructed `Map<Event,Collection<Event>>`.  I did not consider doing the collection lazily, because the full may set would have to be fully iterated on every update.
- At some point, I chose to outline expressions like `mayOutMap.getOrDefault(e, List.of())`, since a `null` return value does not serve any additional semantics.
- Although `getMayOut()` is currently implemented non-virtually, I consider to let `Knowledge` be extended for base relations.  When this happens, the `Function` interface will describe the structure of the result more precisely than `Map`.  In those cases, the function could be lazy (i.e. `id`,`po`).
- My choice came with the loss of `mayOutMap.keySet()` and `mayOutMap.values()`.  An additional method will be required to compute the domain and co-domain of the may set of a relation.

Maybe I should inject TODOs to tell code reviewers.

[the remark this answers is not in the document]
ThomasHaas: > I did not consider doing the collection lazily, because the full may set would have to be fully iterated on every update.

The whole intention of the caching mechanism is to avoid expensive recomputations, currently exactly by not being lazy but eager.

> Although getMayOut() is currently implemented non-virtually, I consider to let Knowledge be extended for base relations. When this happens, the Function interface will describe the structure of the result more precisely than Map. In those cases, the function could be lazy (i.e. id,po).

I agree that there is potential for laziness in some cases (like base relations) and this is also what we eventually need to do for large benchmarks where memory becomes a serious issue (especially computing `ext` explicitly sucks).
But this will also affect the whole set of edges itself, not just the in/out sets.

To handle this properly, we will likely need our own graph implementations (similar to `TupleSet` but just as an interface with many different actual implementations from fully eager to fully lazy).
One big problem is that due to backward propagations, even may and must sets may take unnatural forms (e.g. `maySet(id)` may not contain all identities anymore).
We will likely need to devise a semi-lazy structure like `Sets.difference(lazyIdBaseMaySet, eagerDisabledSet)` to handle this efficiently (`id` is not problem, but `ext` would be).

I guess one could model all the laziness by reusing the guava library for `Sets` and `Function` for in/out sets, but somehow I feel that this won't work in the long run.

Anyway, we can keep this for now, but I would put at least a comment that says that the actual Functions (or their results) are eagerly computed and that is why we cache them for performance reasons!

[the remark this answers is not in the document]
xeren: That being said, the construction of the function objects will be expensive most of the time.  Thus they should be cached for the duration of the lifetime of `EncodeSets`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Event.java
xeren: This is meant to be a optimization:  If `makeTrue()` is a native library call, then it does not have to be called thousands of times, just to be conjoined.  It would suffice to return an `Optional`, but `Collection` was just slightly easier to deal with.

ThomasHaas: It is no native call though (`JavaSMT` already caches true and false) and even if it was, it wouldn't matter.
The amount of events we encode is trivial in comparison to the amount of edges we encode.

ThomasHaas:
Is there a reason why you added the cache back in? I was totally fine with that change

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -249,12 +239,18 @@ public BooleanFormula encodeDataRaces() {
...
+                    if (!e1.hasFilter(Tag.WRITE) || e1.hasFilter(Tag.INIT)) {
+                        continue;
+                    }
                     MemEvent w = (MemEvent)e1;
                     if (!w.canRace()) {
                         continue;
                     }
-                    for(Event e2 : t2.getCache().getEvents(FilterMinus.get(FilterBasic.get(Tag.MEMORY), FilterBasic.get(INIT)))) {
+                    for(Event e2 : t2.getEvents()) {
+                        if (!e2.hasFilter(Tag.WRITE) || e2.hasFilter(Tag.INIT)) {
```
ThomasHaas, line 251: This was a check for `Memory \ Init` before, which seems to have changed to `Write \ Init`.

ui/src/main/java/com/dat3m/ui/result/ReachabilityResult.java
```diff
@@ -120,15 +120,18 @@ private void buildVerdict(Program p, Result result, ProverEnvironment prover, En
         StringBuilder sb = new StringBuilder();
         Model model = (result == FAIL && !p.getAss().getInvert()) || (result == PASS && p.getAss().getInvert()) ? prover.getModel() : null;
     	for(Axiom ax : wmm.getAxioms()) {
-        	if(ax.isFlagged() && model != null && TRUE.equals(model.evaluate(CAT.getSMTVariable(ax, ctx)))) {
+        	if(ax.isFlagged() && model != null && TRUE.equals(model.evaluate(CAT.getSMTVariable(ax, ctx.getFormulaManager())))) {
```
ThomasHaas, line 123: In some PR I changed `Property.getSMTVariable` to take an `EncodingContext` while you changed it to taking a `FormulaManager`. I don't mind either way for now, but this causes a compilation error at the moment.

ThomasHaas (commented):
I think this branch is fine now. @hernanponcedeleon if you find nothing anymore, feel free to approve and merge.

xeren:
> Is there a reason why you added the cache back in? I was totally fine with that change cause Caches suck unless they are definitely needed.

There was one cache field that I was actually fine with, and I wanted to check if it still showed some performance impact.  It did not on my machine.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/DataRaceSolver.java
```diff
@@ -49,18 +49,13 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
 		performStaticProgramAnalyses(task, analysisContext, config);
 		performStaticWmmAnalyses(task, analysisContext, config);
 
-		context = EncodingContext.of(task, analysisContext, ctx);
+		context = EncodingContext.of(task, analysisContext, ctx.getFormulaManager());
 		ProgramEncoder programEncoder = ProgramEncoder.withContext(context);
 		PropertyEncoder propertyEncoder = PropertyEncoder.withContext(context);
 		WmmEncoder wmmEncoder = WmmEncoder.withContext(context);
 		SymmetryEncoder symmetryEncoder = SymmetryEncoder.withContext(context, memoryModel, analysisContext);
```
hernanponcedeleon, line 56: This is never used. Do we want to add it to the solver?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/axiom/Acyclic.java
```diff
@@ -10,21 +9,38 @@
...
 
 	private static final Logger logger = LogManager.getLogger(Acyclic.class);
 
+    @Option(name = REDUCE_ACYCLICITY_ENCODE_SETS,
+            description = "Omit adding transitively implied relationships to the encode set of an acyclic relation." +
+                    " This option is only relevant if \"" + ENABLE_ACTIVE_SETS + "\" is set.",
+            secure = true)
+    private boolean reduceAcyclicityEncoding = true;
+
     public Acyclic(Relation rel, boolean negated, boolean flag) {
```
hernanponcedeleon, line 44: I get the following warning
```
Constructor does not receive Configuration instance and may not be able to inject configuration options of this class
```

hernanponcedeleon (commented):
The code LGTM. There are only two small comments. I will run some experiments on same large benchmarks to see there is nothing weird with performance.

After that it can be merged.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/DataRaceSolver.java
ThomasHaas: I don't see why we shouldn't add it (apart from the general soundness issues of symmetry breaking)

xeren: I accidentally added this in a previous commit 3804ded225682ff1682fd826e715a3f08db9a760.  I should remove this for now.  Once the issues have been solved, the symmetry encoder can be properly added.

ThomasHaas: Symmetry encoding only fails for CAAT. It is fine here

xeren: Then I'll add its encoding.

PR #331, opened 2022-11-16
C11

hernanponcedeleon:
This PR marks the compiled (when C11 is the target) events related to create/join/start/end as `Tag.C11.Atomic` (since they have Rel-Acq semantics).
With these changes, all C11 results makes sense (they are similar to the other models with the exception of linuxrwlock which returns FAIL; but this is possible because C11 is weaker than any other model we tested before).
Refinement might still require some work and thus, for now, we use the assume solver in the CI.
Two of the safe instances of LFDS take too long so they are not (yet) part of the CI.


ThomasHaas:
Looks good to me.

But I think we should at least add a comment to `linuxwrlock` in the test, cause we are not sure if that result is correct or not.
In particular, because the benchmark uses neither non-atomics nor SC-atomics, the expected result should likely be similar to what RC11 and IMM report, both of which say that the benchmark is safe.

ThomasHaas (approved):

PR #332, opened 2022-11-20
Support llvm IR as input

hernanponcedeleon:
The main purpose of this PR is to introduce support for LLVM as an input format. Changes include
- New LLVM events and implementation of their compilation scheme in the different visitors
- Simplification of the `Compilation` class (see 1. below)

As a biproduct, it "removes" the restriction of no supporting thread creation inside loops (the compiler takes care of unrolling this code when the bound is static, i.e. the most common use case). 

I was also able to simplify the modeling of create/start/end/join. Particularly, most of the ugly code in the `pthreadPool` class was to be able to match a join with its corresponding create. I noticed that whenever there is a join, there is a matching `Load`. I now use its address (which is also a parameter to the create) as the "communication channel (cc)" and can avoid all the "magic" to trace back the correct object we were using for the modeling. Since we already have the load to be sure the corresponding thread finished, the join event now compiles to a local which just assigns zero, meaning the join succeeds. This is not fully correct, but this is a limitation that we already had independently of this PR (see 2. below)

I marked the PR as a draft because there are several things we need to decide, and still need to be done. But it would be good to already get feedback in its current state. Open issues are listed below

- [x] The new code depends on a new llvm pass that "lifts" llvm-atomics to function calls. The PR for that pass will be open soon (hopefully this week)
- [x]  I think we should get rid of all the parsing code about C11 atomics ... now we can simply reason about the LLVM code. This also allows to remove our own `stdatomic.h` header. However we probably want to keep the C11 events and their compilation to be able to handle C11 litmus tests. Right now we can still parse the C11 atomics as C11 events because our custom header transform those to function calls. 
- [x]  Rewrite lock benchmarks
- [x]  Rewrite lfds benchmarks
- [x]  I think a lot of code in the svcomp module can be simplified. Specially the `SVCOMPSanitizer`


hernanponcedeleon:
> Am I right that with the new LLVM CAS, we do not support this (rather weird) case where the expected value is taken from a shared variable, right? Also, we do not store back the result to the expected register on failure?

It is possible to write in C this weird case and we do support it. For this program
```
atomic_int shared;
int expected;

int main()
{
    int old = atomic_compare_exchange_strong_explicit(&shared, &expected, 2, __ATOMIC_ACQ_REL, __ATOMIC_ACQ_REL);
    int new = atomic_load(&shared);
    assert(old == 1 && new == 2);
    return 0;
}
```
this is the IR
```
main
(2)      skip
(18)     0:$i0 = load(*expected)
(19)     (0:$p1(0), 0:$p1(1)) = llvm_cmpxchg_strong(*&mem1, 0:$i0, 2, ACQ_REL)  ### LLVM
(20)     0:$i2 <- 0:$p1(0)
(21)     0:$i3 <- 0:$p1(1)
(24)     if((! 0:$i3 == 1)); then goto 0:$bb3:
(25)     goto 0:$bb2:
(26)  0:$bb3:
(27)     if((! (! 0:$i3 == 1))); then goto END_OF_T1:
(28)     store(*expected, 0:$i2)
(30)  0:$bb2:
(31)     0:$i4 = llvm_load(*&mem1, SC)  ### LLVM
(32)     0:$i5 <- 0:$i4 == 2
(33)     0:$i6 <- (if 0:$i3 == 1 then 0:$i5 else 0)
(36)     if((! 0:$i6 == 1)); then goto 0:$bb5:
(37)     goto END_OF_0:
(38)  0:$bb5:
(39)     if((! (! 0:$i6 == 1))); then goto END_OF_T1:
(40)     assert_0 <- 0
(41)     if(assert_0 != 1); then goto END_OF_T1:
(42)     goto END_OF_T1:
(43)  END_OF_0:
(44)  END_OF_T1:
```
Notice that the compiler takes care of generating the store that updates expected.

If we make `expected` a local variable, the code gets simplified due to register promotion + CP

```
atomic_int shared;

int main()
{
    int expected = 0;
    int old = atomic_compare_exchange_strong_explicit(&shared, &expected, 2, __ATOMIC_ACQ_REL, __ATOMIC_ACQ_REL);
    int new = atomic_load(&shared);
    assert(old == 1 && new == 2);
    return 0;
}
```

```
main
(2)      skip
(18)     (0:$p0(0), 0:$p0(1)) = llvm_cmpxchg_strong(*&mem1, 0, 2, ACQ_REL)      ### LLVM
(19)     0:$i1 <- 0:$p0(1)
(20)     0:$i2 = llvm_load(*&mem1, SC)  ### LLVM
(21)     0:$i3 <- 0:$i2 == 2
(22)     0:$i4 <- (if 0:$i1 == 1 then 0:$i3 else 0)
(25)     if((! 0:$i4 == 1)); then goto 0:$bb2:
(26)     goto END_OF_0:
(27)  0:$bb2:
(28)     if((! (! 0:$i4 == 1))); then goto END_OF_T1:
(29)     assert_0 <- 0
(30)     if(assert_0 != 1); then goto END_OF_T1:
(31)     goto END_OF_T1:
(32)  END_OF_0:
(33)  END_OF_T1:
```

ThomasHaas:
I see. That is nice. So basically, LLVM_CAS works over an expected value rather than an address, but the compiler will add load/store from/to `expected` if necessary to mimic the C-CAS.

Btw. it seems that we do not yet mark the plain loads/stores as `NONATOMIC`. I think you added code to mark `llvm_load/llvm_store` but not plain memory ops.
Also, aren't `llvm_load/llbm_store` by definition atomic operations and hence have to have a non-empty `mo`, so they are never marked as `NONATOMIC`?
Talking about this, I think you should also change `mo.equals("")` to `mo.isEmpty()` for the check.

ThomasHaas:
I see. That is nice. So basically, LLVM_CAS works over an expected value rather than an address, but the compiler will add load/store from/to `expected` if necessary to mimic the C-CAS.

Btw. it seems that we do not yet mark the plain loads/stores as `NONATOMIC`. I think you added code to mark `llvm_load/llvm_store` but not plain memory ops.
Also, aren't `llvm_load/llbm_store` by definition atomic operations and hence have to have a non-empty `mo`, so they are never marked as `NONATOMIC`?
Talking about this, I think you should also change `mo.equals("")` to `mo.isEmpty()` for the check.

ThomasHaas:
I noticed that when compiling C11 CAS, we tag the load from expected as `atomic` (probably the store as well?) but that doesn't align with what Clang produces. Indeed, the documentation suggests that you cannot use the address of an atomic variable as `expected` so it should always be plain/non-atomic.

hernanponcedeleon:
If the CI passes, I think this PR is ready to be reviewed. Notice that I changed the expected result in one of the IMM tests.
Previously there was a comment that said we deviated from the result of the paper because we could not mark the store of a RMW as a strong one. The compilation now is doing exactly that, thus we can use the correct result from the paper.

hernanponcedeleon:
> Does the new LLVM pass work on all operating systems?

You need to build the library on your own (instructions are in the `README` of that folder), but this should be OS independent.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/VisitorBoogie.java
```diff
@@ -486,6 +479,11 @@ public Object visitAssume_cmd(Assume_cmdContext ctx) {
 			} else {
 				pairingLabel = pairLabels.get(currentLabel);
 			}
+			// "assume false" simple terminates the thread
+			// pairingLabel is guaranteed to be "END_OF_T"
+			if(ctx.proposition().expr().getText().equals("false")) {
+				programBuilder.addChild(threadCount, EventFactory.newGoto(pairingLabel));
+			}
```
ThomasHaas, line 486: Is this some kind of optimization for `assume(false)`? `assume(false)` does not simply terminate a thread, it is stronger than that. If you want to keep the assume semantics but also allow for DCE, I think you have to keep the assume and add the terminating goto afterwards (this goto will never get executed, but it makes sure that DCE can eliminate the subsequent code).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -259,5 +236,29 @@ public static final class IMM {
...
+        	switch(cMo) {
+        		case C11.MO_SC:
+        		case C11.MO_ACQUIRE_RELEASE:
+        			return C11.MO_RELEASE;
+        		case C11.MO_ACQUIRE:
+        			return C11.MO_RELAXED;
+        		default:
+        			return cMo;
+        	}
+        }
```
ThomasHaas, line 250: Why is this code now in IMM?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/llvm/LlvmStore.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public ImmutableSet<Register> getDataRegs(){
+        return value.getRegs();
+    }
+
+    @Override
+    public String toString() {
+        return "llvm_store(*" + address + ", " +  value + (mo != null ? ", " + mo : "") + ")\t### LLVM";
+    }
```
ThomasHaas, line 41: `mo` cannot be `null` anymore, can it? Together with the fact that `mo` has to be also non-empty (due to this being an atomic), the checks there are useless (probably the same for all other atomic events).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorBase.java
```diff
@@ -97,52 +98,52 @@ public List<Event> visitUnlock(Unlock e) {
...
     @Override
 	public List<Event> visitStoreExclusive(StoreExclusive e) {
-		throw new IllegalArgumentException("Compilation for " + e.getClass().getName() + " is not supported by " + getClass().getName());
+		throw new IllegalArgumentException("Compilation for " + e.getClass().getSimpleName() + " is not supported by " + getClass().getSimpleName());
 	};
 
 	@Override
 	public List<Event> visitRMWAbstract(RMWAbstract e) {
-		throw new IllegalArgumentException("Compilation for " + e.getClass().getName() + " is not supported by " + getClass().getName());
+		throw new IllegalArgumentException("Compilation for " + e.getClass().getSimpleName() + " is not supported by " + getClass().getSimpleName());
```
ThomasHaas, line 106: Maybe it would be nicer to add a method `throwError(Event)` that generates the error message rather than copying it over and over again.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
```diff
@@ -160,41 +154,106 @@ public List<Event> visitAtomicXchg(AtomicXchg e) {
...
+
+    @Override
+    public List<Event> visitLlvmFence(LlvmFence e) {
+            return tagList(eventSequence(
+                            newFence(e.getMo())));
+    }
+
+    private List<Event> tagList(List<Event> in) {
+        return in.stream().map(e -> tagEvent(e)).collect(Collectors.toList());
+    }
```
ThomasHaas, line 250: You shouldn't abuse `map` like this.
You can just do `in.forEach(this::tagEvent); return in;` and define `tagEvent` to not have a return value.

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/VisitorBoogie.java
hernanponcedeleon: Yes, it was intended to be an optimization. I think this code is quite old (before all our discussion about assume semantics). I agree it does not match our current status of assume semantics. I don't fully understand the second part of your comment. Do you mean adding an `Assume` event just before the `Goto` in line 485?

ThomasHaas: Yes. The `Assume` event will give the intended semantics. The `goto END_OF_T` will just be there so that our control-flow analysis can understand that the subsequent code is actually dead (we could also extend our cf-analysis to reason about `Assume` but that seems a lot harder).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
hernanponcedeleon: There are 3 C-like models: C11, RC11 and IMM. The first two understand all memory orders (RC11 does not really understand `CON`), thus it is only IMM which needs to convert SC to something else. We are doing pretty much the same for hardware models. Maybe `extractXMoFromCMo` would be a better name to be consistent with the rest.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/VisitorBoogie.java
hernanponcedeleon: There seems to be a mismatch of what we expect of the boogie code `assume(false)` and what smack generates.
The problem with the CI is that the original C code [calls `pthread_exit()`](https://github.com/sosy-lab/sv-benchmarks/blob/2e1723fde6aa65a250dcb677efa45edaa4b6b631/c/pthread/fib_bench-2.c#L32) which should terminate only the calling thread [according to the documentation](https://www.ibm.com/docs/en/i/7.2?topic=ssw_ibm_i_72/apis/users_18.html).

The code generated by smack looks like this
```
t2
(88)     skip
(89)     start_thread()
(90)     5:$p0 <- 0
(92)     5:$i1 <- 0
(94)  5:$bb1:
(95)     5:$i2 <- 5:$i1 < 5
(98)     if((! 5:$i2 == 1)); then goto 5:$bb3:
(99)     begin_atomic()
(100)    5:$i3 = load(*i)
(101)    5:$i4 = load(*j)
(102)    5:$i5 <- (5:$i4 + 5:$i3)
(103)    store(*j, 5:$i5)
(104)    end_atomic()
(107)    5:$i6 <- (5:$i1 + 1)
(108)    5:$i1 <- 5:$i6
(109)    goto 5:$bb1:
(110) 5:$bb3:
(111)    if((! (! 5:$i2 == 1))); then goto END_OF_T3:
(112)    assume(False)
(114) END_OF_T3:
```
which will execute the assume event 112 when the bound is large enough and thus blocking the execution which should fail.

hernanponcedeleon: From [here](https://github.com/smackers/smack/blob/c7d0694f08cefb422ebcc67c23824727b06b370e/lib/smack/SmackInstGenerator.cpp#L329-L333) it is clear that smack converts any unreachable instruction into an `assume(false)` and thus it differs from our semantics for assume events

benchmarks/lfds/safe_stack.h
```diff
@@ -5,7 +5,7 @@
 #endif
 
 typedef struct {
-    atomic_int value;
+    volatile value;
```
ThomasHaas, line 8: `volatile int`

ThomasHaas (approved):
LGTM

PR #334, opened 2022-11-21
Memory order cannot be null

hernanponcedeleon:
We recently changed to the empty string as a default memory order (before we were using `null`).
There was one remaining use of null that caused a null pointer when checking if some event could race.
This PR fixes this issue.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemEvent.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemEvent.java
@@ -15,7 +15,7 @@ public abstract class MemEvent extends Event {
     	Preconditions.checkNotNull(mo, "The memory ordering cannot be null");
         this.address = address;
         this.mo = mo;
-        if(mo != null){
+        if(!mo.isEmpty()){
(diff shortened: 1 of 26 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemEvent.java
```diff
@@ -46,16 +47,16 @@ public String getMo(){
...
 
     public void setMo(String mo){
-    	Preconditions.checkNotNull(mo, "Only the parser can set the memory ordering to null");
-    	if(this.mo != null) {
+    	Preconditions.checkNotNull(mo, "The memory ordering cannot be null");
+    	if(!this.mo.isEmpty()) {
             removeFilters(this.mo);    		
     	}
         this.mo = mo;
         addFilters(mo);
```
ThomasHaas, line 55: Only add filter if it is non-empty!

ThomasHaas (approved):
It seems after 11 commits, you finally got all of the null checks :).
I think you should add a comment to `MemEvent` that says that we use the empty string to mean "no memory order" and then you can merge.

PR #339, opened 2022-12-07
Add call stack to violation witnesses

hernanponcedeleon:
Debugging large violation witnesses is hard when method calls are spawned over different files (or the same but large file). Assigning to each event the call stack that leads to its execution makes the debugging task easier.

This PR adds the stack call information to such events and the generated witness graphs.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CallStackAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CallStackAnalysis.java
@@ -1,0 +1,68 @@
+package com.dat3m.dartagnan.program.processing;
+
+import com.dat3m.dartagnan.program.Program;
+import com.dat3m.dartagnan.program.Thread;
+import com.dat3m.dartagnan.program.event.core.*;
+import com.dat3m.dartagnan.program.event.core.annotations.FunCall;
+import com.dat3m.dartagnan.program.event.core.annotations.FunRet;
+import org.sosy_lab.common.configuration.Configuration;
+import org.sosy_lab.common.configuration.InvalidConfigurationException;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Stack;
+
+public class CallStackAnalysis  implements ProgramProcessor {
+	
+	private Map<Event, String> callStackMapping = new HashMap<>(); 
+
+    private CallStackAnalysis () { }
+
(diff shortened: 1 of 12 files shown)


ThomasHaas:
Do we need all this call stack tracking during the parsing?
Can't we just iterate once through the program and collect the FunCall/FunRet annotations to reconstruct the call stack?
That seems like a more generic solution.
In fact, I would probably create a `CallStackAnalysis` that computs a map `Map<Event, String>` in a single pass, a just generate it on demand when creating the witness.
Seems like it would be more modular, involve less code, and do not require adaption of so much existing code.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CallStackAnalysis.java
```diff
@@ -0,0 +1,68 @@
...
+import com.dat3m.dartagnan.program.event.core.annotations.FunRet;
+import org.sosy_lab.common.configuration.Configuration;
+import org.sosy_lab.common.configuration.InvalidConfigurationException;
+
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Stack;
+
+public class CallStackAnalysis  implements ProgramProcessor {
```
ThomasHaas, line 16: We usually do not inherit from `ProgramProcessor` for analyses, but only for transformation code.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
```diff
@@ -34,12 +35,14 @@ public class ExecutionGraphVisualizer {
     private static final Logger logger = LogManager.getLogger(ExecutionGraphVisualizer.class);
 	
     private final Graphviz graphviz;
+    private final Map<Event, String> callStackMapping;
     private BiPredicate<EventData, EventData> rfFilter = (x, y) -> true;
     private BiPredicate<EventData, EventData> coFilter = (x, y) -> true;
     private Map<BigInteger, IExpr> addresses = new HashMap<BigInteger, IExpr>();
 
-    public ExecutionGraphVisualizer() {
+    public ExecutionGraphVisualizer(Map<Event, String> callStackMapping) {
```
ThomasHaas, line 43: I think it would be nicer if the map is default-initialized to be an empty `Map` and then you can add a fluent setter `setCallStackMapping` (so it aligns with the way we set the other options).

ThomasHaas (approved):
LGTM

Issue #336, opened 2022-11-29
Compilation of Lock events to hardware architecture


hernanponcedeleon:
We parse `pthread_mutex` as lock/unlock events. However we only "compile" them with `VisitorBase`. This is enough for language level memory models (e.g. what we use in SVCOMP), but it fails when `pthread_mutex` is used together with a  hardware model. One problem (there might be more) is that the resulting store and load have `SC` tag which does not guarantee proper rel/acq propagation on hardware models. To solve this we need to implement a proper compilation of such events for each hardware architecture (the easiest is probably to treat them as a test-and-set lock).

PR #340, opened 2022-12-18
336 (2)

hernanponcedeleon:
This PR implements the second part of #336, i.e., it allows to compile LKMM locks to hardware.
Now there is no need to exclude the litmus tests containing locks from the compilation tests.

The hardware implementation is more from a modeling perspective (i.e., I don't expect bugs in real simple spinloops implementation in the kernel) meaning that this would allow us to target verification of real kernel code which uses simple spinloops under hardware architectures.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
@@ -547,10 +547,12 @@ class VisitorArm8 extends VisitorBase {
 	public List<Event> visitLKMMLock(LKMMLock e) {
 	Register dummy = e.getThread().newRegister(GlobalSettings.ARCH_PRECISION);
         // Spinlock events are guaranteed to succeed, i.e. we can use assumes
+        // With this we miss a ctrl dependency, but this does not matter
+        // becase the load is an acquire one.
 	return eventSequence(
                 newRMWLoadExclusive(dummy, e.getLock(), ARMv8.MO_ACQ),
                 newAssume(new Atom(dummy, COpBin.EQ, IValue.ZERO)),
-                newRMWStoreExclusive(e.getLock(), IValue.ONE, ARMv8.MO_RX, true)
+                newRMWStoreExclusive(e.getLock(), IValue.ONE, "", true)
         );
 	}
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
@@ -1,7 +1,9 @@
 package com.dat3m.dartagnan.program.processing.compilation;
 
+import com.dat3m.dartagnan.GlobalSettings;
 import com.dat3m.dartagnan.expression.*;
 import com.dat3m.dartagnan.expression.op.BOpUn;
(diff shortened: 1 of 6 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -528,4 +542,23 @@ public List<Event> visitRMWOpAndTest(RMWOpAndTest e) {
...
 	};
+
+        @Override
+	public List<Event> visitLKMMLock(LKMMLock e) {
+	Register dummy = e.getThread().newRegister(GlobalSettings.ARCH_PRECISION);
+        // Spinlock events are guaranteed to succeed, i.e. we can use assumes
+	return eventSequence(
+                newRMWLoadExclusive(dummy, e.getLock(), ARMv8.MO_ACQ),
+                newAssume(new Atom(dummy, COpBin.EQ, IValue.ZERO)),
+                newRMWStoreExclusive(e.getLock(), IValue.ONE, ARMv8.MO_RX, true)
```
ThomasHaas, line 553: Using `Assume` avoids ctrl-dependencies. It doesn't matter, because the `load` is acquire, but maybe we should put a comment about the missing ctrl-dep.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -751,6 +765,30 @@ public List<Event> visitRMWOpAndTest(RMWOpAndTest e) {
...
+	@Override
+	public List<Event> visitLKMMLock(LKMMLock e) {
+	Register dummy = e.getThread().newRegister(GlobalSettings.ARCH_PRECISION);
+	Label label = newLabel("FakeDep");
+    // Spinlock events are guaranteed to succeed, i.e. we can use assumes
+	return eventSequence(
+				newRMWLoadExclusive(dummy, e.getLock(), ""),
+                newAssume(new Atom(dummy, COpBin.EQ, IValue.ZERO)),
+                Power.newRMWStoreConditional(e.getLock(), IValue.ONE, "", true),
+				// Fake depedency to guarantee acquire semantics
```
ThomasHaas, line 777: typo -> "dependency"

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
hernanponcedeleon: Following the `extractXMoFromLKMo` pattern, we should make it non-atomic

ThomasHaas: What even is `ARMv8.MO_RX`? Does ARMv8 even have a notion of relaxed atomic? I thought there is only REL, ACQ and plain (~non-atomic/relaxed).

hernanponcedeleon: Good question. The cat file does not refer to the `RX` tag. I could not find anything in arm documentation. Neither in `herd7 code` (they have something called `AArch64.N` which might simply be "non-atomic"). 

Our litmus parser creates such tag, but I think we can simply change it to `""`. I will change this in a different PR.

ThomasHaas (approved):

PR #341, opened 2022-12-19
Guarantee always correct acquire semantics for loads related to pthread_join

hernanponcedeleon:
With #332 we broke the acquire semantics of loads related to joins under LKMM.
This is because we were always using `C11.ACQ` as a tag which LKMM does not understand.
I also simplified the compilation of create/start/end/join events: we don't need that many fences / `SC` semantics ... `ACQ/REL` is enough.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorTso.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorTso.java
@@ -24,49 +24,45 @@ class VisitorTso extends VisitorBase {
                 super(forceStart);
         }
 
-	@Override
-	public List<Event> visitCreate(Create e) {
-        Store store = newStore(e.getAddress(), e.getMemValue(), e.getMo());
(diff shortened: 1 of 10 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorTso.java
```diff
@@ -24,49 +24,45 @@ protected VisitorTso(boolean forceStart) {
...
-        return eventSequence(
-        		load,
-        		newJumpUnless(new Atom(resultRegister, EQ, IValue.ZERO), (Label) e.getThread().getExit())
-        );
-	}
+        @Override
+        public List<Event> visitEnd(End e) {
+                return eventSequence(
+                                newStore(e.getAddress(), IValue.ZERO, ""),
+                                X86.newMemoryFence());
```
ThomasHaas, line 41: Aren't TSO stores always release by default (and loads are acquire). If so, you don't need any fences here (neither in `visitCreate`).

ThomasHaas (approved):
LGTM

PR #343, opened 2022-12-19
Avoid deadlock in Compilation.runCmd

hernanponcedeleon:
I was experiencing a deadlock when creating a process to take care of compiling the file. I found out that this is possible when the input and output streams are promptly written and read respectively of the sub process (which seems to happen when trying to compile large files).

This PR solves this problem.


ThomasHaas (approved):
LGTM

PR #344, opened 2022-12-20
Read error msg from log file

hernanponcedeleon:
The changes in #343 redirect the error msg to the log file. This makes the input stream (from which we used to read the error msg) empty, thus making it hard to understand from the console what the problem was.

This PR makes the error msg be read from the log file.


ThomasHaas (approved):
LGTM

PR #345, opened 2022-12-20
Improvements

hernanponcedeleon:
This PR makes some small improvements
- After #341, we ended up with two `Load` events for each `pthread_join`, one created in `VisitorBoogie` and the other in `PthreadProcedures`. I removed the later.
- Nodes from non-atomic events in witnesses where having labels like `W(x, 1, )`. Now we use `W(x,1)`. 
- Barriers should be visible in witness graphs
- When setting the `sourceCFile`, we assumed that there was a path containing `/` rather than simply a file name. This is now fixed

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
@@ -42,7 +42,8 @@ public class PthreadsProcedures {
 			pthread_create(visitor, ctx);
 			break;
 		case "pthread_join":
-			pthread_join(visitor, ctx);
+			// VisitorBoogie already took care of creating the join event
+			// when it parsed the previous load.
 			break;
 		case "pthread_cond_init":
 		case "pthread_cond_wait":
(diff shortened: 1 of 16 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
```diff
@@ -131,7 +128,7 @@ private ExecutionGraphVisualizer addAllThreadPos(ExecutionModel model) {
 
     private ExecutionGraphVisualizer addThreadPo(Thread thread, ExecutionModel model) {
         List<EventData> threadEvents = model.getThreadEventsMap().get(thread)
-            .stream().filter(e -> e.isMemoryEvent()).collect(Collectors.toList());
+            .stream().filter(e -> e.isMemoryEvent() || e.isFence()).collect(Collectors.toList());
```
ThomasHaas, line 131: You can simply check for `Tag.VISIBLE` I think. IIRC that was also the code before you changed it in a previous PR.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
```diff
@@ -42,7 +42,8 @@ public static void handlePthreadsFunctions(VisitorBoogie visitor, Call_cmdContex
 			pthread_create(visitor, ctx);
 			break;
 		case "pthread_join":
-			pthread_join(visitor, ctx);
+			// VisitorBoogie already took care of creating the join event
+			// when it parsed the previous load.
```
ThomasHaas, line 46: How does this work now? If the load is generated by Clang only, then where do we make sure that it is an acquire load?

hernanponcedeleon: When visiting `pthread_create` in `PthreadsProcedures`, we add certain expressions to `visitor.allocations`. When creating a Load in `VisitoBoogie`, we check if the value is one of such expressions and create a Load or a Join event accordingly
```
if(allocations.contains(value)) {
    // These loads corresponding to pthread_joins
    child = EventFactory.Pthread.newJoin(register, (IExpr)value);
 else {
    child = EventFactory.newLoad(register, (IExpr)value, "");
}
```
With the code before this PR, we ended up creating two Join events, the one from the code above and the one from the code which this PR removes in `PthreadsProcedures`.

ThomasHaas (approved):
LGTM

PR #346, opened 2022-12-22
Addresses must be even

hernanponcedeleon:
[Most hardware architectures use even values for addresses](https://stackoverflow.com/questions/23315939/why-2-lsbs-of-32-bit-arm-instruction-address-not-used). This allows certain algorithms to make use (i.e., their correctness depends on this) of the least significant bit as a flag.

This PR modifies our fixed allocation to comply with such assumption.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
@@ -55,7 +55,11 @@ public class Memory {
             BigInteger i = BigInteger.ONE;
             for(MemoryObject a : program.getMemory().objects) {
                 a.address = i;
-                i = i.add(BigInteger.valueOf(a.size()));
+                // Addresses are tipically at least two byte aligned
+                //      https://stackoverflow.com/questions/23315939/why-2-lsbs-of-32-bit-arm-instruction-address-not-used
+                // Many algorithms rely on this assumption for correctness
+                int padding = (a.size() % 2);
+                i = i.add(BigInteger.valueOf(a.size() + padding));
             }
         }
     }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
@@ -52,7 +52,7 @@ public class Memory {
 
         @Override
         public void run(Program program) {
-            BigInteger i = BigInteger.ONE;
+            BigInteger i = BigInteger.TWO;
             for(MemoryObject a : program.getMemory().objects) {
                 a.address = i;
                 // Addresses are tipically at least two byte aligned
--- a/dartagnan/src/test/resources/arrays/ok/C-array-ok-17.litmus
+++ b/dartagnan/src/test/resources/arrays/ok/C-array-ok-17.litmus
@@ -15,4 +15,4 @@ P0(atomic_t* arr2) {
   WRITE_ONCE(*(r1 + 1), 5);
 }
 
-forall (0:r0 = 4 /\ 0:r1 = arr1[0] /\ 0:r2 = 1 /\ arr1[1] = 5);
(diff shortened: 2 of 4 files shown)


hernanponcedeleon:
The CI is failing in this test
```
C C-array-ok-18.litmus
/*
    Note: The weird syntax "*(&b)" is for compatibility with Herd,
    which threats right side variables as pointers by default
 */
{
  a = 1; b = 2;
  arr2[2] = {&a, *(&b)};
}

P0(atomic_t* arr2) {
  int r0 = READ_ONCE(*arr2);
  int r1 = READ_ONCE(**arr2);
  int r2 = READ_ONCE(*(arr2 + 1));
}

forall (0:r0 = a /\ 0:r1 = 1 /\ 0:r2 = 2);
```
I think the test is confusing: at first glance, it seems that `0:r2 = 2` is referring to the initial value in `b`, but it actually refers to the address of `b` (in the way we were generating memory allocation up to now, the address was exactly 2). My claim that it refers to the address matches the comment in [this litmus](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/test/resources/arrays/ok/C-array-ok-17.litmus).

Thus, I think the correct litmus test should be
```
C C-array-ok-18.litmus
/*
    Note: The weird syntax "*(&b)" is for compatibility with Herd,
    which threats right side variables as pointers by default
 */
{
  a = 1; b = 2;
  arr2[2] = {&a, *(&b)};
}

P0(atomic_t* arr2) {
  int r0 = READ_ONCE(*arr2);
  int r1 = READ_ONCE(**arr2);
  int r2 = READ_ONCE(*(arr2 + 1));
}

forall (0:r0 = a /\ 0:r1 = 1 /\ 0:r2 = b);
```

With such changes the current PR returns the expected result.

I tried to run both litmus tests with herd but its parser complains despite I tried several changes (the syntax of herd for handling arrays might have changed since WE wrote this litmus tests several years ago).

hernanponcedeleon:
BTW, to guarantee that addresses are even, the first address should start at `BigInteger.TWO` rather than `BigInteger.ONE`. I will soon push this change.

ThomasHaas:
I agree that the litmus test seems wrong (more precisely, it oddly relies on a very specific memory layout). I don't understand the confusing syntax `*(&b)` though. According to the comments, it is some weird workaround that we are doing?
The comments suggest that for herd, a simple `b` would suffice (but then `&a` becomes weird again).

hernanponcedeleon:
I don't fully understand this either (this test was created by Natalia and today was the first time I tried to run it with herd). The comment suggests that the code should be accepted by herd as it is, but it is not the case with the version I have (`7.56.2, Rev: exported`). I also tried their online version, and it fails too.

For me the comment suggests the opposite: we need the annotation for herd to work. I actually tried using `b` with dartagnan and it works, thus another reason to believe the weird annotations has nothing to do with us.

ThomasHaas:
> For me the comment suggests the opposite: we need the annotation for herd to work. I actually tried using `b` with dartagnan and it works, thus another reason to believe the weird annotations has nothing to do with us.

The comment says that `herd` treats rhs variables as pointers, i.e. a simple `b` would actually refer to the address of `b`.
But this is exactly what you think the test should do and also why the check `0:r2 = b` actually uses the address of `b` rather than its value (btw. this should mean that the actual value of  `b` is never used, right? I.e. you could change it to anything and it will still work, no?).

For example, in this test [LKMM-ABA](https://github.com/hernanponcedeleon/Dat3M/blob/master/litmus/LKMM/lockfree/aba-simple.litmus), you see that a linked list is initialized like this
```
{
    d = -1;
    c = d;
    b = c;
    a = b;
    ToS = a;
}
```
Here all variables contain a pointer to the rhs variable (cause herd treats them as pointers by default).
So it seems to me that we did a workaround in that test, because we did not treat the rhs as a pointer by default.
Maybe there is some weird interaction with the way arrays work that needs a different syntax...

ThomasHaas (approved):
LGTM

PR #347, opened 2022-12-30
Regenerate bpl files for LFDS using more optimization flags

hernanponcedeleon:
The flags `-sroa` and `-early-cse` seem to simplify some of our benchmarks quite a bit, thus we add those to the required flags in `ATOMIC_REPLACE_FLAGS`.

I also regenerated the bpl files for those benchmarks where the passes have effect.


ThomasHaas (approved):
LGTM

PR #348, opened 2023-01-01
Remove unused passes

hernanponcedeleon:
After #332 was merged, there is no longer a need to promote plain accesses accessing atomic variables to SC. Also, the pass that replaced atomic blocks with locks was both unused and in general unsound due the fact that atomic blocks work in isolation.

This PR removes both passes.


ThomasHaas (approved):
LGTM

PR #353, opened 2023-01-05
Log flagged pairs

hernanponcedeleon:
When we flag an axiom (e.g., a data race was found), we only report that the axiom was violated, but we don't explain why.
This PR shows the user (via the log in debug mode) which are the pairs that belong to the relation of the axiom that was flagged.

Since I needed access to the `EncodingContext`, I moved the part that flags the violation from `Dartagnan` main class to each solver (`Refinement` cannot deal with such violations and thus it is omitted there). This results in some code duplication. I explored the idea of moving the logging part to `ModelChecker` (the common parent of all solvers), but this required moving there the logger and the `EncodingContext` so it did not look that much like an improvement to me.


ThomasHaas:
You could also refactor the code into a method in `ModelChecker` and simply pass both the `Logger` instance as well as the `EncodingContext` as parameters.
Just because the code is in `ModelChecker`, it doesn't mean its dependencies also have to be there :).

ThomasHaas (approved):
LGTM

PR #354, opened 2023-01-05
Improved Id handling

ThomasHaas:
This PR improves the way we use Event ids.
- Events now have two fluent ids (i.e., one's that may change during processing), namely `localId` (within thread) and `globalId` (within program).
- The ids `oId`, `uId` and `cId` are now snapshots of the `globalId` at different points of the processing:
   - `oId` is the global id directly after program construction and never changes from thereon.
   - `uId` is the global id directly **BEFORE** unrolling. To find out if two events are copies of the same one but in different iterations, checking the equality of `uId` will do it now.
   - `cId` is the global id directly **BEFORE** compilation. Similarly to above, two events stem from the same compiled event if their `cIds` match.

All passes and analyses now just work over `globalId`. This, in particular, also makes these passes agnostic to when they are run in the program processing pipeline, so we can more easily rearrange passes (there are still precondition checks of course, so arbitrary reordering may not work).
The `localId` and `globalId` just give the following guarantees:
- They are non-negative (unless the ids were not assigned yet, which would be a bug)
- They are unique at any time, meaning every event has a different id. Within a program processing pass, this invariant may get temporarily violated of course, but the pass is responsible for restoring it then.
- The ids follow the program syntax, in particular, for acyclic programs an event with a higher id cannot be before an event with a smaller id (the usual property we expected from `cId` as well).

I also reworked the way we print the program for debugging. There are now 5 different print flags and all of them are inside `ProcessingManager`.

**Note:** This PR is based on #350 and hence should be reviewed after #350.


hernanponcedeleon:
Some general comments / questions based on the description of the PR.
- `localId` simply replaces `fId` and it is only used in the symmetry pass. Do we really need this? Can't we simply rely on `globalId / #threadEvents` when inserting `(id,event)` in the map?
- the names `oId`, `uId`, `cId` make less sense now and do not represent the fact that the later two refers to ids **BEFORE**. What about using `apId` (after parsing), `buId` (before unrolling) and `bcId` (before compiling)?

ThomasHaas:
>     * `localId` simply replaces `fId` and it is only used in the symmetry pass. Do we really need this? Can't we simply rely on `globalId / #threadEvents` when inserting `(id,event)` in the map?

It is possible to eliminate this id, yes. The `ThreadSymmetry` class can compute a `Map<Event, Integer> localIdMap` instead.
If you want me to, I can rewrite `ThreadSymmetry` and remove `localId`.

>     * the names `oId`, `uId`, `cId` make less sense now and do not represent the fact that the later two refers to ids **BEFORE**. What about using `apId` (after parsing), `buId` (before unrolling) and `bcId` (before compiling)?

You mean they make less sense now because the first one is "after" and the other two are "before"? Or do you think that, e.g., `uId` is naturally interpreted as "after unrolling"? In either case, I would argue against that. Our former notions of the ids where all broken. For example, `oId` was mutated by passes (so in what way was it "original"?) and so was `cId`... they were ids "somewhere in the process" and not at a specific point in time. E.g., `cId` was the id somewhere after compilation (unspecific when exactly) rather than directly after. Not that "directly after" would be any helpful at all, since what you cared about is the ids directly before compilation (which is the reason why you had to look at `uid` to see which events came from the same compiled event!?). In fact, we used `oId` with the meaning of "id before unrolling" and `uId` with the meaning of "before compilation", which, to be honest, is even more confusing.

Rather than thinking of "before" and "after" (which are time spans, btw.), you can think of "at the time of". "At the time of unrolling/compilation" IMO naturally reads as "immediately before unrolling/compiling" and "at the time of parsing" naturally reads as "immediately after being parsed" ("before being parsed" doesn't even make sense).
Long story short, I think the the meaning of those ids now matches their most natural interpretation, i.e., the naming is now better than before.

Whether `apId`, `buId` and `bcId` are even better... I don't think so. They look very cryptic for barely any benefit in my opinion.
They also somehow suggest that something like `auId` should exist, but for none of the ids there is a counterpart (and likely never will, due to lack of usability).

hernanponcedeleon:
I took a look to the code now (most passes look much cleaner now!). 

I would still prefer not to have the local id, but by removing it we only get rid of ~5 lines of code, so I would say "yes I prefer you re-write the symmetry pass" only if the change would also be ~5 lines of code. If we ended up with a more complex implementation of that class, then the use of a local id is well justified.

About the ids meaning names. You are right that once we started introducing passes their original meaning was completely destroyed. In my mind I still interpret `uId` as *"the id generated by the unrolling"* but I might be simply biased for many years of using it like this. Also, the use of e.g., `getCid` is minimal now and mostly replaced by `getGlobalId`, so I think the naming and the differences between `before` and `after` are not much of a problem.

Other than the local id thing, the PR looks good.

ThomasHaas:
Btw. I think `oId`and `uId` are completely unused now, so they could even be deleted (I didn't cause of potential use-cases). And the only use-case of `cId` is some alias analysis unit test, IIRC.

The rewriting of `ThreadSymmetry` to get rid of `localId` is no problem. Codewise, it won't be much more. I can do that.

PR #356, opened 2023-01-07
getModel() as a local variable

hernanponcedeleon:
I was consistently experiencing some library crashes when trying to extract the model after a data race was found. I assume that from the JavaSMT side, calling twice `prover.getModel()` should be safe. However, when I avoid the double call (by saving the model in a local variable), the crash disappears.


ThomasHaas (approved):
LGTM

PR #359, opened 2023-01-12
Fixed liveness encoding bug

ThomasHaas:
Problem: Our liveness encoding forced all loads inside a loop body to read from co-maximal stores, even those loads that did not execute!

Fix: Only actually executed(!) loads need to read from co-maximal stores.


hernanponcedeleon:
LGTM

PR #362, opened 2023-01-13
Fix Idd

xeren:
The internal dependency pairs were encoded as must-tuples.  This seems to be relevant in language-level edge cases (maybe just release-consume semantics), when a program on weak memory relaxes the ordering by overwriting the result of a load.


hernanponcedeleon:
@xeren  can you merge `development` (so only the real changes are listed) and fix the conflicts?

hernanponcedeleon:
I don't really understand the description of this PR. 

- What does "were encoded" mean? It used to be like that and now we change it? 
- What does it mean that is relevant to language level edges? 
- Is there an example where the encoding is broken and this PR intends to fix the problem?

ThomasHaas:
>     * Is there an example where the encoding is broken and this PR intends to fix the problem?
Good point. I remember that there was an issue with `Idd` (or `idd+`?) always being encoded as if it were static, i.e., as if all edges were must-tuples. This was the issue we had with XRA shortly before we submitted to PLDI, and this is the issue this PR should resolve (as far as I understand it).
However, this issue definitely did not exist in development, so it was only introduced in the XRA branch.
In that sense, I don't quite understand why this PR fixes anything unless the recently merged PR #315 contained this mistake already and we did not catch it.

That being said, this issue has nothing to do with language-level wmms; it applies to all dependency-tracking models.

hernanponcedeleon:
Just to check, I run safe stack in `development` (now that this is fast, we should add it to the unit tests) under RISCV (this was the one causing the xra-bug) and we get the correct result there.

ThomasHaas:
The issue in XRA only appeared in conjunction with everything else that happened, no (the new backpropagations)?
I think it is best if @xeren explains more in this issue.

xeren:
This bug is indeed hard to encounter.  Too many actors that decide on whether it happens, or not.  A litmus test would require an unknown `idd` (or `addrDirect`) pair to be relevant to the memory model.  In such cases, the relation analysis always marks those pairs as *must*.  That makes it difficult to find an execution where two events happen, but do actually **not** participate in `idd`, because meanwhile the register got overwritten.

The following is an unchecked prototype to be differentiated by this PR:
```
acyclic addr | data | po-loc | co | rf | fr
```
```
T1(int* x, int* y)
{
    int r = *x;    // E0
    int s = r;
    if(r >= 0)
        s = *y;    // E1, does not dominate E0
    *z = s;        // E2, may be data-related from E0 or E1
}
T2(int* x, int* y, int* z)
{
    *y = 1;
    int t = *z;
    *x = t + 1;
}
not exists r = 2
```

ThomasHaas:
Hmm, by your explanation, this is the issue we had with XRA back then, no?
And this does not require any extremely weak model. The reason why you might be thinking this is because your example requires that ctrl-deps are not tracked. But you can have this simple case:

```
int r = load(x);
int s = load(y);
if (r > 0) 
    s = 2;
store (x, s);
```
After the load from `x`, the hardware may decide to perform `store(x, 2)` before `load(y)` if it realizes that `r > 0`.
Indeed, I would assume that ARMv8 and Power allow this reordering.

hernanponcedeleon:
What would be the assertion to be checked in your example @ThomasHaas ? Can we write a litmus test showing the problem that this PR fixes and add it to the PR as a regression test?

ThomasHaas:
> What would be the assertion to be checked in your example @ThomasHaas ? Can we write a litmus test showing the problem that this PR fixes and add it to the PR as a regression test?

Something like this:
```
// All accesses relaxed atomics on the C-level
// For ARM/Power, just plain accesses
T1: 
int r = load(x);
int s = load(y);
assert (s != 42);
if (r == 0) 
    s = 42;
store (x, s);
// -------
T2:
int a = load(x);
store(y, a)
```
You NEED(!) to test this with Assume/Incremental, because Refinement may give a wrong result here (it also assumes dependencies to be static). That being said, I cannot give 100% guarantee that this is indeed possible behaviour.

hernanponcedeleon:
Does this means that this PR might solve the problem when using assume/incremental solver, but not caat?

ThomasHaas:
Since for CAAT we usually do not encode dependency relations, it doesn't affect CAAT at all.
That being said, since the very first version of CAAT, it always treated dep-relations as static when generating reasons (not when exploring the execution graph though) and we have not encountered a single wrong result so far.
The reason is simply because we never encoded `data`, `addr`, and `ctrl` as base relations so CAAT couldn't produce reasons over those relations and instead replaced them by pairs of event literals.

hernanponcedeleon:
For the program below, both `development` and this branch return PASS (using the assume solver)

```
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <stdatomic.h>

atomic_int x, y;

void *thread_1(void *unused)
{
	int r = atomic_load_explicit(&x, memory_order_relaxed);
	int s = atomic_load_explicit(&y, memory_order_relaxed);
	assert (s != 42);
	if (r == 0) 
    	s = 42;
	atomic_store_explicit(&x, s, memory_order_relaxed);
	return NULL;
}

void *thread_2(void *unused)
{
	int a = atomic_load_explicit(&x, memory_order_relaxed);
	atomic_store_explicit(&y, a, memory_order_relaxed);
	return NULL;
}

int main()
{
	pthread_t t1, t2;

	pthread_join(t1, NULL);
	pthread_join(t2, NULL);

	return 0;
}
```

ThomasHaas:
The above version has too strong ctrl-deps because the assertion `assert(s != 42)` produces one.
The below version fixes this.
```
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <stdatomic.h>

atomic_int x, y;

void *thread_1(void *unused)
{
	int r = atomic_load_explicit(&x, memory_order_relaxed);
	int s = atomic_load_explicit(&y, memory_order_relaxed);
        int u = s;
	if (r == 0) {
    	      s = 42;
    	}
	atomic_store_explicit(&x, s, memory_order_relaxed);
	assert (u != 42);
	return NULL;
}

void *thread_2(void *unused)
{
	int a = atomic_load_explicit(&x, memory_order_relaxed);
	atomic_store_explicit(&y, a, memory_order_relaxed);
	return NULL;
}

int main()
{
	pthread_t t1, t2;

	pthread_join(t1, NULL);
	pthread_join(t2, NULL);

	return 0;
}
```

According to @hernanponcedeleon, we return `PASS` in development for both `Assume` and `CAAT` under `IMM`.
It seems that we also return `PASS` in this branch.
However, GenMC reports `FAIL` (which is IMO the correct result).

hernanponcedeleon:
I also tried this test in `xra` to see if the larger set of changes fix it, but I still get `PASS`.

The question is: is the problem still related to the changes in this PR or completely unrelated?

ThomasHaas:
I think the problem is related, but let's say it is not, then we still should find SOME test that distinguishes this PR from development. We cannot really have a "fix" if we do not even have a "bug".

ThomasHaas:
I could verify that this PR indeed fixes the benchmark I have mentioned above.
However, a slight adaption has to be made to fix a miscompilation(?!)

```
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <stdatomic.h>

atomic_int x, y, z;

void *thread_1(void *unused)
{
	int r = atomic_load_explicit(&x, memory_order_relaxed);
	int s = atomic_load_explicit(&y, memory_order_relaxed);
        int u = s;
	if (r == 0) {
	      atomic_store_explicit(&z, 42, memory_order_relaxed); // This line is just there to avoid compiler tricks.
    	      s = 42;
    	}
	atomic_store_explicit(&x, s, memory_order_relaxed);
	assert (u != 42);
	return NULL;
}

void *thread_2(void *unused)
{
	int a = atomic_load_explicit(&x, memory_order_relaxed);
	atomic_store_explicit(&y, a, memory_order_relaxed);
	return NULL;
}

int main()
{
	pthread_t t1, t2;

	pthread_join(t1, NULL);
	pthread_join(t2, NULL);

	return 0;
}
```

The issue with the original code was that the conditional
```
if (r == 0) {
    s = 42;
 }
```
was compiled to a `ITE`
```
s = (if r == 0 then 42 else resultOfLoad)
```
For this code we will always generate a data-dependency to the load, even if the second branch is not taken.
This caused all our solvers to return PASS.
The version with the inserted store results in proper branching where the difference of this PR vs. development can be seen.

This raises another question though: How shall we handle dependencies involving ITE expression?
I assume the ITE expressions on the llvm level are the equivalent to conditional moves on hardware, but I do not know how the dependencies of those are defined and preserved.

hernanponcedeleon:
Yes, those instructions result in branchless code. Using such instructions is one of the possible fixes to the spectre vulnerability that exploits branch predictors.

Can you add a unit test and we merge this PR?

benchmarks/c/miscellaneous/idd_dynamic.c
```diff
@@ -0,0 +1,43 @@
...
+// This test shows that data-dependencies have to be treated dynamically
+// Expected result: FAIL under ARM8 (possibly under other weak models as well)
+
+atomic_int x, y, z;
+
+void *thread_1(void *unused)
+{
+	int r = atomic_load_explicit(&x, memory_order_relaxed);
+	int s = atomic_load_explicit(&y, memory_order_relaxed);
+    int u = s;
```
hernanponcedeleon, line 15: missing tab

ThomasHaas: Had to enable whitespace rendering to see the difference locally.
Do we want to use actual tabs or regular spaces instead?
Any preferences about that? Or does it not matter as long as it is consistent in a single file?

hernanponcedeleon: I don't have preferences about those, but we should be consistent.

That being said, the whole format of the project is a mess. If you check the diff of a PR (or even the content of a file from the web browser) the spacing format is all broken even if from the IDE looks nice (at least mine).

I will take a look if there is an easy way to automatically keep the format properly done without introducing much burden for development.

dartagnan/src/test/java/com/dat3m/dartagnan/c/MiscellaneousTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+	@Parameterized.Parameters(name = "{index}: {0}, target={1}")
+    public static Iterable<Object[]> data() throws IOException {
+    	return Arrays.asList(new Object[][]{
+	            {"idd_dynamic", ARM8, FAIL},
+		});
+    }
+
+    @Test
+	//@CSVLogger.FileName("csv/assume")
```
hernanponcedeleon, line 45: If we don't need to log the csv file, remove (it also applies to the one below)

PR #363, opened 2023-01-14
Clean-up cat folder

hernanponcedeleon:
This PR does some housekeeping of the `/cat` folder by removing the following files:
- `arm.cat`: not really interesting since it is pretty much that same as `power.cat` and we don't have `ARM7` as a compilation target anymore.
- `cut-*.cat`: these were only used for the OOSPLA evaluation. There are no real uses cases for them.
- `ocaml.cat`: never really used. We also lack the corresponding compilation target and even test cases.


ThomasHaas (approved):
LGTM

PR #365, opened 2023-01-18
Enable and fix CLKMMTest

hernanponcedeleon:
The latest updates to LKMM modify two expected results in the CLKMM class.
However the error was not detected because the test was not run in the CI.
This PR makes the following changes:
- enables those tests
- updates the expected results
- documents why the changes in results are sound


ThomasHaas (approved):
LGTM

PR #374, opened 2023-01-23
Simple performance improvement in alias analysis

ThomasHaas:


ThomasHaas:
The CI has failed with a timeout of CAAT on `qspinlock` (`CLKMMTest`).
Locally, I solve it instantly though.
I'm wondering if there is a rare bug in CAAT where the refinement does not progress properly (i.e. it gets stuck in a loop) or if this was just some random error in the CI.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/FieldSensitiveAndersen.java
```diff
@@ -209,12 +209,19 @@ private static final class Location {
...
         @Override
         public boolean equals(Object o) {
-            return this == o || o instanceof Location && base.equals(((Location)o).base) && offset==((Location)o).offset;
+            if (this == o) {
+                return true;
+            } else if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            Location other = (Location) o;
+            // Can we check reference-equality on the MemoryObject?
```
hernanponcedeleon, line 223: Is this a kind of "TODO question"?

hernanponcedeleon (commented):
The code LGTM, but to be honest I don't fully understand where the performance improvement comes from. Is it because we avoid collisions or because you explicitly iterate using `for` rather than the stream functions?

Performance if much better now. Here are some numbers using the bpl file from #359 and another version having one more thread in the client code. 

```
development; 2 threads
[24.01.2023] 09:30:53 [INFO] AliasAnalysis.fromConfig - Selected Alias Analysis: FIELD_SENSITIVE
[24.01.2023] 09:33:04 [INFO] AliasAnalysis.fromConfig - Finished Alias Analysis

development; 3 threads
[24.01.2023] 09:33:50 [INFO] AliasAnalysis.fromConfig - Selected Alias Analysis: FIELD_SENSITIVE
[24.01.2023] 09:42:56 [INFO] AliasAnalysis.fromConfig - Finished Alias Analysis

PR; 2 threads
[24.01.2023] 09:45:27 [INFO] AliasAnalysis.fromConfig - Selected Alias Analysis: FIELD_SENSITIVE
[24.01.2023] 09:45:52 [INFO] AliasAnalysis.fromConfig - Finished Alias Analysis

PR; 3 threads
[24.01.2023] 09:46:12 [INFO] AliasAnalysis.fromConfig - Selected Alias Analysis: FIELD_SENSITIVE
[24.01.2023] 09:48:07 [INFO] AliasAnalysis.fromConfig - Finished Alias Analysis
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/FieldSensitiveAndersen.java
ThomasHaas: Almost, but it is not really important at all. I was just wondering if we ever create duplicated `MemoryObjects`, cause if not, then we could get away with reference equality checks. But this is no performance bottleneck, so it doesn't matter.

ThomasHaas:
> The code LGTM, but to be honest I don't fully understand where the performance improvement comes from. Is it because we avoid collisions or because you explicitly iterate using `for` rather than the stream functions?

The collisions are just a minor detail I have not really profiled, but it was a no-brainer so I added it.
The performance improvement is just avoiding the streams, in particular `flatMap`. This made the `fields` method 5-6x faster.

PR #377, opened 2023-01-24
Specifications

ThomasHaas:
This PR refactors a lot of code related to `Property` and `Assert` as well as the way multiple properties get encoded and tracked.
All properties are formulated in a positive manner, e.g., "DATARACES" became "DATARACEFREEDOM".
Also "REACHABILITY" is now called "PROGRAM_SPEC", by the fact that litmus tests define a custom query that can be reachability or safety, while C-code actually only ever specifies safety via user-given assertions.
The output generated is also streamlined between command line and UI now.

A crucial change is that the default property to check for is only `PROGRAM_SPEC`, while previously we also had `CAT_SPEC`.
This is because this combination of properties cannot be checked in a single SMT-query for (some) litmus tests, and we will now generate an error message for this. For C-Code, this combination is always valid and so the verification goes through.
Also, it is now possible for a solver to return `PASS` + a model, namely if the program specification was an existential query (e.g., does some execution exists that satisfies P). Only litmus tests have these queries for now.

There are still potential things to improve, but the branch should be in a good shape.


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -223,13 +197,119 @@ public static void main(String[] args) throws Exception {
...
+						summary.append("Flag ")
+								.append(Optional.ofNullable(violatedAx.getName()).orElse(violatedAx.getNameOrTerm()))
+								.append("\n");
+					}
+				}
+			} else if (hasPositiveWitnesses) {
+				if(TRUE.equals(model.evaluate(PROGRAM_SPEC.getSMTVariable(encCtx)))) {
+					// The check above is just a sanity check: the program spec has to be true
+					// because it is the only property that got encoded.
+					summary.append("Program specification witness found.").append("\n");
```
hernanponcedeleon, line 253: Shouldn't we add
```
summary.append("User assertion: ").append(p.getSpecification().toStringWithType()).append("\n");
```
for consistency with the violation case?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -223,13 +197,119 @@ public static void main(String[] args) throws Exception {
...
+			// Litmus-specific output format that matches with Herd7 (as good as it can)
+			if (p.getFilterSpecification() != null) {
+				summary.append("Filter ").append(p.getFilterSpecification().toStringWithType()).append("\n");
+			}
+
+			// NOTE: We cannot produce an output that matches herd7 when checking for both program spec and cat properties.
+			// This requires two SMT-queries because a single model is unlikely to witness/falsify both properties
+			// simultaneously.
+			// Instead, if we check for multiple safety properties, we find the first one and
+			// generate output based on its type (Program spec OR CAT property)
```
hernanponcedeleon, line 267: We should add a TODO for using two smt queries. This would still be cheap for litmus tests.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -223,13 +197,119 @@ public static void main(String[] args) throws Exception {
...
+						summary.append("Flag ")
+								.append(Optional.ofNullable(violatedAx.getName()).orElse(violatedAx.getNameOrTerm()))
+								.append("\n");
+					}
+				}
+			} else {
+				// We have neither a witness nor a violation ...
+				if (result == UNKNOWN) {
+					// Sanity check: UNKNOWN is not an expected result for litmus tests
+					summary.append(result).append("\n");
```
hernanponcedeleon, line 295: can we log a warning? I think this would be easier to spot for us that an UNKNOWN result when we are running litmus tests

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
```diff
@@ -5,51 +5,81 @@
...
 
 import com.dat3m.dartagnan.wmm.axiom.Axiom;
 
 public enum Property implements OptionInterface {
-	REACHABILITY, RACES, LIVENESS, CAT;
+	PROGRAM_SPEC, 		// Litmus queries OR assertion safety in C-code
+	LIVENESS,			// Liveness property
+	CAT_SPEC,			// CAT-spec defined via flagged axioms in .cat file (~bug specification)
+	DATARACEFREEDOM; 	// Special option for data-race detection in SVCOMP only
+
```
hernanponcedeleon, line 19: Left alignment of comments

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
```diff
@@ -5,51 +5,81 @@
...
+        	case DATARACEFREEDOM:
+        		return "Data-race freedom (SVCOMP only)";
 			case LIVENESS:
 				return "Liveness";
-			case CAT:
-				return "CAT properties";
+			case CAT_SPEC:
+				return "CAT specification";
 			default:
 				throw new UnsupportedOperationException("Unrecognized property: " + this);
```
hernanponcedeleon, line 40: align case statements

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
```diff
@@ -5,51 +5,81 @@
...
+			return Type.SAFETY;
+		} else {
+			return Type.REACHABILITY;
+		}
+	}
+
+	public static Type getCombinedType(EnumSet<Property> properties, VerificationTask context) {
+		return properties.stream().map(p -> p.getType(context))
+				.reduce((x, y) -> x == y ? x : Type.MIXED)
+				.orElse(Type.SAFETY); // In the odd case that the properties are empty, we consider it a safety spec
```
hernanponcedeleon, line 74: Is it even possible to have an empty property? I don't think so, but if so, we should add a check (probably somewhere) to detect this

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -75,32 +103,66 @@ public BooleanFormula encodeBoundEventExec() {
...
-        if(property.contains(RACES)) {
-            specViolation = bmgr.or(specViolation, encodeDataRaces());
-        }
-        if(property.contains(CAT)) {
-            specViolation = bmgr.or(specViolation, encodeCATPropertyViolations());
+    public BooleanFormula encodeProperties(EnumSet<Property> properties) {
+        Property.Type specType = Property.getCombinedType(properties, context.getTask());
+        if (specType == Property.Type.MIXED) {
+            final String error = String.format(
+                    "The set of properties %s are of mixed type (safety and reachability properties)." +
```
hernanponcedeleon, line 110: Missing space after the dot

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -48,10 +51,35 @@ public class PropertyEncoder implements Encoder {
     private final AliasAnalysis alias;
     private final RelationAnalysis ra;
 
+    // We may want to make this configurable or just keep the option fixed.
+    private final boolean doWeakTracking = true;
```
hernanponcedeleon, line 55: do you expect any performance difference between the two?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -350,9 +423,8 @@ public BooleanFormula encodeDeadlocks() {
                 allStuckOrDone = bmgr.and(allStuckOrDone, bmgr.or(isStuck, isTerminatingNormally));
             }
 
-            final BooleanFormula deadlockVariable = LIVENESS.getSMTVariable(context);
             final BooleanFormula hasLivenessViolation = bmgr.and(allStuckOrDone, atLeastOneStuck);
```
hernanponcedeleon, line 426: Maybe `deadlockViolation` makes more sense now

dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
```diff
@@ -16,11 +16,14 @@ public class GitInfo {
...
                 .getResourceAsStream("git.properties")) {
-            properties.load(is);
-            logger.info("Git branch: " + properties.getProperty("git.branch"));
-            logger.info("Git commit ID: " + properties.getProperty("git.commit.id"));
+		   if (is == null) {
+			   return;
+		   }
+		   properties.load(is);
+		   logger.info("Git branch: " + properties.getProperty("git.branch"));
+		   logger.info("Git commit ID: " + properties.getProperty("git.commit.id"));
```
hernanponcedeleon, line 26: Does this fix the annoying issue that every now and then after compiling you get a `null stream` (or something of the sort) exception? Have you seen this? I mostly got it when I was using a proxy and I had some problem on my proxy configuration

dartagnan/src/test/java/com/dat3m/dartagnan/compilation/AbstractCompilationTest.java
```diff
@@ -129,14 +132,20 @@ public void testIncremental() throws Exception {
...
+                // We found no model showing a specific behaviour (either positively or negatively),
+                // so the compiled code should also not exhibit that behaviour, unless we
+                // know the compilation is broken
+                boolean compilationIsBroken = getCompilationBreakers().contains(path);
+                final Result expected = compilationIsBroken ? s1.getResult().invert() : s1.getResult();
                 IncrementalSolver s2 = IncrementalSolver.run(context2Provider.get(), prover2Provider.get(), task2Provider.get());
         		assertEquals(expected, s2.getResult());
+
+                // We could also check if the second solver has a model, instead of comparing results.
+                //assertEquals(s2.hasModel(), compilationIsBroken);
```
hernanponcedeleon, line 148: why is this line commented?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: Yes, seems I forgot it.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: Will add

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
ThomasHaas: I don't think it is possible to have an empty spec. Maybe if the user puts `--property=""` or something?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
ThomasHaas: Hard to tell. The weak tracking is technically a slightly more simple encoding in the case we check for multiple properties. I wouldn't be surprised if there is no difference in performance, but who knows :).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
ThomasHaas: You mean `hasDeadlock` maybe? `deadlockViolation` sounds kinda wrong.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
ThomasHaas: Yes, I got null streams sometimes when running from the IDE I believe. Since it didn't seem to be crucial output, I just added the simple null-check-and-return.

dartagnan/src/test/java/com/dat3m/dartagnan/compilation/AbstractCompilationTest.java
ThomasHaas: Just an alternative way of checking the assertion. I forgot to remove it after testing around with it I guess.

hernanponcedeleon:
@ThomasHaas can you fix the conflicts? After that I can merge

Issue #254, opened 2022-05-20
Failed atomic_cmpxchg


hernanponcedeleon:
We currently report a wrong result for this test
```
C C-atomic-cmpxchg-failure-02

{
  atomic_t x = ATOMIC_INIT(0);
  atomic_t z = ATOMIC_INIT(0);
}

P0(atomic_t *x, atomic_t *z) {
  int r0; int r1;
  r0 = atomic_cmpxchg_acquire(z,2,1);
  r1 = atomic_read(x);
}

P1(atomic_t *x, atomic_t *z) {
  atomic_set(x,1);
  atomic_set_release(z,1);
}

exists (0:r0 = 1 /\ 0:r1 = 0)
```

[Here](https://github.com/herd/herdtools7/issues/384) there is a discussion about the issue. The problem is that right now we don't support dynamic tagging, i.e. non-static unary predicates.

PR #382, opened 2023-01-26
Better compilation for Linux CAS and AddUnless

ThomasHaas:
- Updated Compilation of Linux CAS and AddUnless
- Removed Linux Cond Events
- Enabled previously unsupported Litmus test

This PR solves issues #254 and #200.
It also opens up better support for standard control-flow/data-flow reasoning over Linux code, because we have no more odd conditional events.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLockRead.java
```diff
@@ -10,12 +10,12 @@
...
 	public LKMMLockRead(Register register, IExpr lock) {
-		super(register, lock, Tag.Linux.MO_ACQUIRE);
+		super(register, lock, Linux.MO_ACQUIRE);
 		addFilters(RMW, Linux.LOCK_READ);
 	}
 
 	@Override
 	public String toString() {
-		return String.format("spin_lock_R(*%s)", address);
+		return String.format("%s <- spin_lock_R(*%s)", resultRegister, address);
```
hernanponcedeleon, line 19: I would keep the old string representation which matches how the instruction is used in the litmus format, see e.g. [here](https://github.com/hernanponcedeleon/Dat3M/blob/master/litmus/LKMM/manual/locked/W%2BRW%2BRR%2BWR%2Bonces%2Blocked.litmus)

ThomasHaas: But it results in code like this:
```
spin_lock_R(...)
assume(DUMMY_REG == 0);
```
where the assume is working on a register that you never see in the code.
I found it very confusing to be honest until I checked the spinlock event.

hernanponcedeleon: Ohhh this is the read part (generated during compilation) of a spinlock. Then  it is ok (I thought it was the complete spinlock event). I will merge once the CI passes

PR #387, opened 2023-01-29
Add long type version of linux macros

hernanponcedeleon:
This PR adds all linux macros with long type.


ThomasHaas:
I assume the long version works over 64-bit types rather than 32 bit one's? I know that we internally do not differentiate, so the implementation of the macros is identical (for now at least).

hernanponcedeleon:
Exactly ... so right now the only problem is that without this change we cannot parse code that make use of long types.

ThomasHaas:
I understand that. LGTM

ThomasHaas (approved):

PR #392, opened 2023-01-31
Lkmm races

hernanponcedeleon:
This PR adds a unit test to check that we correctly detect data-races in LKMM (`LitmusLinuxTest` only check `PROG_SPEC`).

I did not find a nice way to handle the prefix of the file both in the `buildLitmusTests` method (which is static) and for the `expectedResultProvider` (this is why I marked as a DRAFT in case someone has a suggestion). However the later seems to be related to the `CSVLogger` which I am not even sure we really use anymore.


ThomasHaas:
The `CSVLogger` is irrelevant: you need to define the expected result per test, hence it must be non-static.
I think what is awkward is that `ResourceHelper.getExpectedResults` performs caching based on the name of the results file, so you need to remember that name throughout all tests.

I think you can decouple these aspects:
- Let `AbstractLitmusTest` have a `protected static Map<String, Result> expectedResults`.
- Let `buildLitmusTests` parse whatever file it needs to parse via `ResourceHelper` and assign the returned map to `expectedResults`.
- Let `expectedResultProvider` use the static map instead of relying on caching done by `ResourceHelper` (I think `ResourceHelper` does not need to do caching at all!).

ThomasHaas (approved):
LGTM

PR #396, opened 2023-02-03
[Draft] Fix CP and DCE

ThomasHaas:
This PR fixes Constant Propagation to make it work on BVs(*) as well as preserve dependencies.
It also fixes DeadAssignmentElimination.
Both passes are enabled by default now.

(*) It does not respect overflow behavior, which is UB most of the time anyway.

The reason why this PR is marked as "Draft" is because
- the old CP code is not deleted yet
- the code was not yet tested on actual BV benchmarks, so it may still be faulty.

@hernanponcedeleon Can you maybe test this PR on some BV benchmarks? Do you remember on which benchmark the old CP was definitely unsound?


hernanponcedeleon:
The only benchmark I remember being buggy probably due to CP was #225, but that is not reproducible anymore. 

Have you tried setting `GlobalSetting = 64` and run all the tests locally? If we pass the whole unit tests using CP + BVs, I think it should be good enough.

ThomasHaas:
I guess I can push a version with `ARCH_PRECISION=64` so the CI can do all the checks for me :)

hernanponcedeleon:
If you think doing twice the changes and push is easier than running `mvn install` locally, be my guest :P

ThomasHaas:
There is a big issue with CP and dynamic loop cutting.
CP can propagate local side-effects of loops, causing single iterations to appear side-effect free which are then getting cut of early.
I will need to think about how to solve this problem.

hernanponcedeleon:
I don't get it. Do you have a snippet of code that shows the problem?

ThomasHaas:
> I don't get it. Do you have a snippet of code that shows the problem?
```
cnt = 0;
while (cnt++ < 2) { }
assert (cnt == 2);
```
After unrolling:
```
cnt = 0;
Itr1:
if (cnt >= 2) goto loop_end;
   cnt++;
Itr2:
if (cnt >= 2) goto loop_end;
    cnt++
itr3:
if (cnt >= 2) goto loop_end;
   cnt++;
loop_end:
assert(cnt==2);
```
After CP + DCE
```
Itr1:
Itr2:
Itr3:
loop_end:
assert(2==2);
```
DPLC
```
Itr1:// Itr1 does nothing -> always side-effect-free
goto END_OF_T ### Spinloop
Itr2:
goto END_OF_T ### Spinloop
Itr3:
goto END_OF_T ### Spinloop
loop_end:
assert(2==2);
```

The issue can be resolved for now if the dynamic loop cutting runs before CP to observe the side effects (or run it before unrolling, but this has some other issues).
But there is still the problem that DCE can eliminate events that are tracked by the loop cutting (vie `ExecutionStatus`).
I think this requires something like the very old "listener" system we had: We shouldn't delete events if some other event is referencing it.

ThomasHaas:
Now all the fixable issues are fixed (for now).
The last benchmark that fails is because we perform a i32 load but the init event is i64. This is unrelated to the content of this PR and also happens on `development` with `ARCH_PRECISION = 64` (or any other value except -1 and 32).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NewConstantPropagation.java
```diff
@@ -0,0 +1,238 @@
...
+import com.dat3m.dartagnan.program.event.core.*;
+import com.dat3m.dartagnan.program.event.core.utils.RegWriter;
+import com.dat3m.dartagnan.program.event.visitors.EventVisitor;
+import com.google.common.base.Preconditions;
+
+import java.math.BigInteger;
+import java.util.HashMap;
+import java.util.Map;
+
+public class NewConstantPropagation implements ProgramProcessor {
```
hernanponcedeleon, line 19: Remove the other class and rename this one

hernanponcedeleon (commented):
The code looks good. After my minor comments are fixed, we can merge

hernanponcedeleon:
Would it make sense to add the code snippet from above (related to the issue with CP and dynamic loop cutting) as a new unit test?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/Dependency.java
```diff
@@ -148,66 +143,70 @@ private void process(Thread thread, ExecutionAnalysis exec) {
...
                     jumps.keySet());
-            for(Set<Writer> j : jumps.values()) {
+            for (Set<Writer> j : jumps.values()) {
                 state.addAll(j);
             }
         }
-        for(Register register : thread.getRegisters()) {
-            finalWriters.put(register, process(state, register, exec));
+        for (Register register : thread.getRegisters()) {
+            finalWriters.put(register, process(null, state, register, exec));
```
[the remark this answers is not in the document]
ThomasHaas: I added the `reader` as a new parameter to be able to check if the `reader` is mutual exclusive wrt. some `writer`.
Final writes have no `reader` though, so I use `null` to represent that.

ThomasHaas:
I will add a unit test.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ConstantPropagation.java
```diff
@@ -1,313 +1,264 @@
...
-		@Override
-		public <T> T visit(ExpressionVisitor<T> visitor) { throw new UnsupportedOperationException(); }
-	};
 
-    private ConstantPropagation() { }
+    @Option(name = PROPAGATE_COPY_ASSIGNMENTS,
+            description = "Propagates copy assignments of the form 'reg2 := reg1' to eliminate " +
+                    "intermediate register 'reg2'. Can only be active if " + CONSTANT_PROPAGATION + " is enabled.",
+            secure = true)
+    private boolean propagateCopyAssignments = false;
```
hernanponcedeleon, line 34: Why do we want to have fine grain control over this?

ThomasHaas: It is not clear if this improves or degrades performance, so I didn't enable it by default.
It may be helpful to reduce the analysis time of the alias analysis though.

hernanponcedeleon: The alias analysis time can change +-10% (in rt_mutex) even when it is run using exactly the same options. Enabling or not this option falls in the same range and thus I would consider it noise. For the time being, I would enable this option since it is the one that allows more simplifications in CP.

PR #397, opened 2023-02-03
Support for SRCU

hernanponcedeleon:
This PR adds support for Sleepable RCU (SRCU) in LKMM.

Following the changes in `linux-kernel.def`, `srcu_read_lock / srcu_down_read` are modeled as loads and `srcu_read_unlock / srcu_up_read` as stores. `synchronize_srcu` is modeled as a `MemEvent` because it needs to contribute to the `loc` relation (see definition of `srcu-rscs`). Because of this, I had to implement `SrcuSync.getMemValue()`. The documentation of that method explains why the used value (which I selected randomly) should not have any impact on the verification.


ThomasHaas (approved):
LGTM

PR #398, opened 2023-02-08
Fix XRA

xeren:
Disables more tuples in XRA by computing the full transitive must-set in acyclicities.  Formerly, they disabled inv(T) with T = (must(R) | (T ;; T)) & inv(may(R)).  Intended was T = must(R) | (T ;; T).


ThomasHaas (approved):
LGTM

PR #399, opened 2023-02-09
395

### LINKED ISSUE #395: Problem in compilation C11 -> TSO
The current compilation from C11 to TSO is broken for (at least) RMW instructions as shown by the litmus test below.
```
C SB-RMW

(*
 * A variant of fhe SB pattern using RMW instructions
 *)

{}

P0 (atomic_int* y, atomic_int* x) {
    int r0 = atomic_fetch_add_explicit(x, 1, memory_order_seq_cst);
    int r1 = atomic_load_explicit(y, memory_order_seq_cst);
}

P1 (atomic_int* y, atomic_int* x) {
    int r0 = atomic_fetch_add_explicit(y, 1, memory_order_seq_cst);
    int r1 = atomic_load_explicit(x, memory_order_seq_cst);
}

exists (0:r1=0 /\ 1:r1=0)
```
Under C11 the behavior is now allowed
```
java -jar dartagnan/target/dartagnan-3.1.1.jar litmus/C11/manual/SB-RMW.litmus --method=assume cat/c11.cat

Condition exists (0:r1==0 && 1:r1==0)
No
Total verification time(ms): 354
```
but under TSO it is
```
java -jar dartagnan/target/dartagnan-3.1.1.jar litmus/C11/manual/SB-RMW.litmus --method=assume cat/tso.cat --target=tso

Condition exists (0:r1==0 && 1:r1==0)
Ok
Total verification time(ms): 458
```

hernan-poncedeleon:
This PR fixes #395 and adds the corresponding litmus test to the unit tests.


ThomasHaas:
If you tag all instructions generated by atomics as `A`, then there is no point in placing an `mfence` when compiling atomic stores.
We have to decide whether we place a fence or mark the instruction as atomic, but doing both is unnecessary.

Intuitively, I would assume that only RMW instructions get atomic tags, but never simple/plain memory accesses.
The reason is that on the assembly level, there are only RMW instructions, plain accesses, and fences, but no idea of "atomic load/store".
So if you compile from C down to X86 assembly, you either use genuine RMW instructions (which are tagged, but fenceless) or you use simple memory accesses (untagged) + fences.

hernan-poncedeleon:
I think we should then only keep the `tagList` for `CAS, xchg` and `RMW` (both in the LLVM and C11 case)

ThomasHaas:
Yes.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorTso.java
```diff
@@ -56,203 +55,200 @@ public List<Event> visitStart(Start e) {
...
+                Register dummyReg = e.getThread().newRegister(resultRegister.getPrecision());
+
+                IExpr address = e.getAddress();
+                String mo = e.getMo();
+                Load load = newRMWLoad(resultRegister, address, mo);
+
+                return eventSequence(
+                                load,
+                                newLocal(dummyReg, new IExprBin(resultRegister, e.getOp(), (IExpr) e.getMemValue())),
+                                newRMWStore(load, address, dummyReg, mo));
```
ThomasHaas, line 205: missing tags.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorTso.java
```diff
@@ -29,14 +27,14 @@ public List<Event> visitCreate(Create e) {
                 Store store = newStore(e.getAddress(), e.getMemValue(), "");
                 store.addFilters(C11.PTHREAD);
 
-                return eventSequence(
-                                store);
+                return tagList(eventSequence(
+                                store));
```
ThomasHaas, line 31: Do you want to keep atomic tags for the pthread events?
I think an mfence (if even necessary) is more natural.

hernan-poncedeleon: Power (which is the closest architecture in the sense that it relies in fences to implement act-rel) uses fences so I will do the same (not 100% they are needed, but I prefer to play it safe than smart)

ThomasHaas (approved):
LGTM

PR #400, opened 2023-02-10
Imm

hernanponcedeleon:
The original IMM does not support SC accesses. The same benchmarks as in #395 shows that we currently lack support for SC accesses in IMM. However, genMC supports them in a similar way to how it is done in RC11. This PR fixes this. We now support IMM as implemented in genMC rather than as it is presented in the paper.


cat/imm.cat
```diff
@@ -1,11 +1,19 @@
-(* CAT for IMM *)
+(* CAT for IMM (as implemented by GenMC) *)
+
+(*
+ * GenMC does not follow the published IMM model to the letter
+ * https://github.com/MPI-SWS/genmc/issues/44
+ * It also uses part of RC11 to deal with SC accesses
+ * According to Viktor Vafeiadis, this is a faithfull model
```
ThomasHaas, line 7: faithful

ThomasHaas (approved):
LGTM

PR #401, opened 2023-02-13
Add wsq benchmark

hernanponcedeleon:
This PR adds a new work-stealing queue benchmark (coming from `svcomp-complex`). There is a safe and unsafe (`-DFAIL`) version. The benchmark requires BVs (i.e. `GlobalSettings.ARCH = 64`) and thus I did not add it to the unit tests. Why that change, the caat solver needs ~3min to find the bug with IMM and ~5 to prove the safe version.

### DIFF
--- a/benchmarks/lfds/wsq.LICENSE.txt
+++ b/benchmarks/lfds/wsq.LICENSE.txt
@@ -1,0 +1,23 @@
+Copyright (c) 2016, Truc Nguyen Lam All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+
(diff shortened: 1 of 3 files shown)


ThomasHaas (approved):
LGTM.

I'm just wondering: did you add `STATICSIZE` to avoid dynamic allocations? If so, it might be good to add a comment describing the necessity of `STATICSIZE`.

hernanponcedeleon:
I just took the code as it was and removed the redefinition of `CAS` and `xchg` using `__VERIFIER_atomic_X`

ThomasHaas:
I see. 
I think it would be a good idea to add a comment on where this benchmark is from (a link to the svcomp file?) before we merge it.

hernanponcedeleon:
I just noticed that even the correct version fails under hardware models. This is because locks are not correctly compiled to hardware (see #336), but I think this is independent of this PR.

ThomasHaas:
Yeah, you can still merge.

PR #402, opened 2023-02-13
Add race definition to RC11

hernanponcedeleon:
This PR adds the data-race axiom to RC11 as defined in the cat file [here](https://plv.mpi-sws.org/scfix/) (with minimal modifications due to the fact that we treat init events as non-atomic).

Signed-off-by: Hernan Ponce de Leon <hernanl.leon@huawei.com>

### DIFF
--- a/cat/rc11.cat
+++ b/cat/rc11.cat
@@ -30,3 +30,10 @@ irreflexive (myrmw; eco) as coherencermw
 empty (myrmw & (fr; co)) as atomicity
 acyclic psc as SC
 acyclic (po | rf) as no-thin-air
+
+(* data_races *)
+let conflict = (((W * _) | (_ * W)) & loc)
+// We treat initial events as non-atomic thus we need to explicit remove them
+let race = ext & (conflict \ hb \ (hb^-1) \ (A * A) \ ((IW * _) | (_ * IW)))
+
+flag ~empty race as racy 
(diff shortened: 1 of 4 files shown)


cat/rc11.cat
```diff
@@ -30,3 +30,10 @@ irreflexive (myrmw; eco) as coherencermw
 empty (myrmw & (fr; co)) as atomicity
 acyclic psc as SC
 acyclic (po | rf) as no-thin-air
+
+(* data_races *)
+let conflict = (((W * _) | (_ * W)) & loc)
+// We treat initial events as non-atomic thus we need to explicit remove them
+let race = ext & (conflict \ hb \ (hb^-1) \ (A * A) \ ((IW * _) | (_ * IW)))
```
ThomasHaas, line 37: I find it more natural to move ` ((IW * _) | (_ * IW))` into the definition of `conflict` (as well as the intersection with `ext`).
Cause it sounds odd to talk about a store conflicting with an init event or with another store of the same thread (the lattercould be argued though).

Also, it might be more efficient to do `(conflict \ AxA) \ hb \ hb^-1` (i.e., remove `AxA` as early as possible).
Not sure if this is true though, so take this with a grain of salt :P.

ThomasHaas (approved):
LGTM

PR #403, opened 2023-02-17
Log program spec violation

hernanponcedeleon:
This PR logs the concreate assertions (event + file / line) that is violated one the program specification is not guaranteed. I marked as a DRAFT because there are several things we might want to improve
- do we want to log this or printed in the console as the rest of the result output we give?
- if we log, do we want to do it only at the debug level (not only for program_spec violations but also for flagged axioms)?
- I guess we don't want this information for litmus tests

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -243,9 +243,9 @@ public class Dartagnan extends BaseOptions {
                 }
                 if (props.contains(LIVENESS) && FALSE.equals(model.evaluate(LIVENESS.getSMTVariable(encCtx)))) {
                     summary.append("============ Liveness violation found ============\n");
-                    for(Event e : p.getEvents(CondJump.class)) {
+                    for(CondJump e : p.getEvents(CondJump.class)) {
                         if(e.is(Tag.SPINLOOP) && TRUE.equals(model.evaluate(encCtx.execution(e)))
-                            && TRUE.equals(model.evaluate(encCtx.jumpCondition((CondJump)e)))) {
+                            && TRUE.equals(model.evaluate(encCtx.jumpCondition(e)))) {
                             summary
                                 .append("\t").append(e.getGlobalId())
                                 .append(":\t(").append(e.getSourceCodeFile()).append("#").append(e.getCLine())
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -11,6 +11,7 @@ import com.dat3m.dartagnan.program.Program.SourceLanguage;
 import com.dat3m.dartagnan.program.analysis.CallStackComputation;
 import com.dat3m.dartagnan.program.event.Tag;
 import com.dat3m.dartagnan.program.event.core.Event;
+import com.dat3m.dartagnan.program.event.core.Label;
 import com.dat3m.dartagnan.program.event.core.Load;
 import com.dat3m.dartagnan.program.event.core.Local;
 import com.dat3m.dartagnan.utils.Result;
@@ -241,8 +242,15 @@ public class Dartagnan extends BaseOptions {
                     summary.append(" ================================================= \n");
                 }
                 if (props.contains(LIVENESS) && FALSE.equals(model.evaluate(LIVENESS.getSMTVariable(encCtx)))) {
-                    summary.append(" ===== Liveness violation found ===== \n");
-                    // TODO
+                    summary.append("============ Liveness violation found ============ \n");
+                    for(Event e : p.getEvents(Label.class)) {
+                        if(e.is(Tag.SPINLOOP) && TRUE.equals(model.evaluate(encCtx.execution(e)))) {
+                            summary
+                                .append("\t").append(e.getGlobalId())
+                                .append(":\t(").append(e.getSourceCodeFile()).append("#").append(e.getCLine())
+                                .append(")\n");
+                        }
+                    }
                     summary.append(" ================================================= \n");
                 }
                 final List<Axiom> violatedCATSpecs = task.getMemoryModel().getAxioms().stream()
@@ -250,7 +258,7 @@ public class Dartagnan extends BaseOptions {
                         .filter(ax -> props.contains(CAT_SPEC) && FALSE.equals(model.evaluate(CAT_SPEC.getSMTVariable(ax, encCtx))))
                         .collect(Collectors.toList());
                 if (!violatedCATSpecs.isEmpty()) {
-                    summary.append(" ===== CAT specification violation found ===== \n");
+                    summary.append(" ======= CAT specification violation found ======= \n");
                     // Computed by the model checker since it needs access to the WmmEncoder
                     summary.append(modelChecker.getFlaggedPairsOutput());
                     summary.append(" ================================================= \n");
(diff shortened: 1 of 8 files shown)


ThomasHaas:
>     * do we want to log this or printed in the console as the rest of the result output we give?

I don't mind either way. But if we print it as output, then it should not be part of the `ModelChecker` but instead be generated in `Dartagnan`. Then it would also automatically work in our UI.

>     * if we log, do we want to do it only at the debug level (not only for program_spec violations but also for flagged axioms)?

I think debug level might be the best for now. While I don't think info level is wrong per say, I think that we print too much on the info level right now and we need to cut down on that.

>     * I guess we don't want this information for litmus tests

I agree here. But if one really wanted to do it, I think a possible way would be to transform the program spec into CNF/DNF and report on the clauses/co-clauses that were violated/satisfied (more generally: some sort fault identification in the AST of the spec).

hernanponcedeleon:
> I don't mind either way. But if we print it as output, then it should not be part of the `ModelChecker` but instead be generated in `Dartagnan`. Then it would also automatically work in our UI.

Now that I thought more about this, we should definitely print this. It gives the user much more information than the current printed output, i.e., why do I care about something like this?
```
User assertion: forall (assert_0 && assert_1 && assert_2 && assert_3 && assert_4 && assert_5 && assert_6 && assert_7 && assert_8 && assert_9 && assert_10 && assert_11)
```
When you are dealing with programs having several assertions, or checking for data races, you are more interested in knowing exactly where in your code the problem is so you can fix it.

I think we can also get rid of the `PASS/FAIL` since (i) we already have the e.g., `Program specification violation found` when there is a violation, and (ii) we have `[INFO] RefinementSolver.run - Verification finished with result FAIL`

> I think debug level might be the best for now. While I don't think info level is wrong per say, I think that we print too much on the info level right now and we need to cut down on that.

I agree we log too much. During the weekend I will try to clean this up in another PR.

> I agree here. But if one really wanted to do it, I think a possible way would be to transform the program spec into CNF/DNF and report on the clauses/co-clauses that were violated/satisfied (more generally: some sort fault identification in the AST of the spec).

Litmus are small and easy to spot where the problem is. Also, we are generally not interested in fixing litmus programs.

What I also want to do in this PR is give information about dead locks ... however I am not sure how good the debugging information coming from llvm will be to track the spinloop back to the real code.

ThomasHaas:
I think we still should keep FAIL/PASS though. 
First, I think that no crucial output should be logging output (except error/warn level maybe), so I would disregard  `[INFO] RefinementSolver.run - Verification finished with result FAIL` as a valid reason for not printing the result (the user might have logging disabled after all).
Secondly, I think FAIL/PASS is very easy to spot for the user to immediately understand the outcome of the verification.
Lastly, it is also easier for other tools to parse our verification result, because the output FAIL/PASS is irrespective of the actual property that gets checked, while a message like `Program specification violation found` is property-specific.

I agree with the rest though: (i) no need for detailed output on litmus tests, and (ii) we need better bug descriptions for C-code.

Regarding the dead locks: I would assume that the loop header's label has the proper C-line attached.
It might be a littler harder to figure out which dead lock belongs to which loop though.

hernanponcedeleon:
> I think we still should keep FAIL/PASS though. First, I think that no crucial output should be logging output (except error/warn level maybe), so I would disregard `[INFO] RefinementSolver.run - Verification finished with result FAIL` as a valid reason for not printing the result (the user might have logging disabled after all). Secondly, I think FAIL/PASS is very easy to spot for the user to immediately understand the outcome of the verification. Lastly, it is also easier for other tools to parse our verification result, because the output FAIL/PASS is irrespective of the actual property that gets checked, while a message like `Program specification violation found` is property-specific.

I was thinking what would need to be done for SVCOMP and the integration or dartagnan into vsyncer. You are right and those two uses cases are already sufficient to keep the FAIL/PASS.

This is how the output looks like now
```
===== Program specification violation found ===== 
        89:     (benchmarks/lkmm/qspinlock.c#26)
FAIL
Total verification time(ms): 925
```
```
============ Liveness violation found ============ 
        62:     (benchmarks/lkmm/qspinlock-liveness.c#42)
FAIL
Total verification time(ms): 1236
```
```
======= CAT specification violation found ======= 
Flag data-race
        43 -> 24        (benchmarks/lkmm/qspinlock.c#26 -> benchmarks/lkmm/qspinlock.c#11)
FAIL
Total verification time(ms): 1304
```

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -228,30 +230,44 @@ public static String generateResultSummary(VerificationTask task, ProverEnvironm
...
+                    summary.append("============ Liveness violation found ============\n");
+                    for(Event e : p.getEvents(Label.class)) {
+                        if(e.is(Tag.SPINLOOP) && TRUE.equals(model.evaluate(encCtx.execution(e)))) {
+                            summary
+                                .append("\t").append(e.getGlobalId())
+                                .append(":\t(").append(e.getSourceCodeFile()).append("#").append(e.getCLine())
+                                .append(")\n");
+                        }
+                    }
+                    summary.append("=================================================\n");
```
ThomasHaas, line 254: This won't work for two reasons:
- For dynamic spinloops, the label is not tagged.
- Even if the label is tagged, just executing it is not a violation because it's just the header of a potentially spinning loop.

You need to look for jumps that are tagged by `SPINLOOP` and then you need to check if those got executed AND if their condition was also true (for `CondJump` we set the `exec` variable to true no matter if the condition holds or not).
The metadata of those jumps might not be accurate, cause some of them get generated by the loop cutting pass.
I think you either need to somehow find the loop-header and use its metadata or you might check the loop cutting pass and add attach the correct metadata whenever it inserts a new event.

hernanponcedeleon: The fact that the `SPINLOOP` is executed and its condition is true means that it is stuck as defined in `PropertyEncoder.isStuck`, right?

We only need to propagate the meta data from the label to `assumeSideEffect` in `DynamicPureLoopCutting` if I understood it correctly. This should make it work both for statically and dynamically detected loops.

ThomasHaas: Technically, execution with a true condition does only mean the loop did not terminate in a single iteration, but it does not mean it is stuck. To be stuck, it also needs to read co-maximal stores.
However, if you know that liveness is violated, then it must be stuck for otherwise we would not have reported a violation to begin with. So it is indeed enough to do that simple check.

Adding the correct metadata to `assumeSideEffect` should do the trick. But you might want to only copy over the `cLine`. If you copy all metadata, you would also copy over `oId` and `cId`, which might be misleading.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -243,8 +243,9 @@ public static String generateResultSummary(VerificationTask task, ProverEnvironm
                 }
                 if (props.contains(LIVENESS) && FALSE.equals(model.evaluate(LIVENESS.getSMTVariable(encCtx)))) {
                     summary.append("============ Liveness violation found ============\n");
-                    for(Event e : p.getEvents(Label.class)) {
-                        if(e.is(Tag.SPINLOOP) && TRUE.equals(model.evaluate(encCtx.execution(e)))) {
+                    for(Event e : p.getEvents(CondJump.class)) {
+                        if(e.is(Tag.SPINLOOP) && TRUE.equals(model.evaluate(encCtx.execution(e)))
+                            && TRUE.equals(model.evaluate(encCtx.jumpCondition((CondJump)e)))) {
```
ThomasHaas, line 248: You can directly do `for (CondJump e : p.getEvents(CondJump.class))` (the method returns a correctly typed list of events), so you don't need to do the cast anymore.

ThomasHaas (approved):
LGTM

PR #404, opened 2023-02-19
Memory allocation after processing

ThomasHaas:
This PR defers the handling of mallocs to the end of processing.
This allows us to have mallocs in loops and mallocs whose sizes can be reduced to constants by CP.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -74,29 +75,16 @@ public static String extractLoadMoFromLKMo(String lkMo) {
...
-			case C11.MO_RELEASE:
-				return MO_REL;
-			case C11.MO_ACQUIRE_RELEASE:
-			case C11.MO_SC:
-				return MO_ACQ_REL;
-			default:
-				return "";
-			}
-        }
-
```
hernanponcedeleon, line 99: Why are these gone?

ThomasHaas: It was never called.

PR #408, opened 2023-02-21
Add support for llvm.ctlz

hernanponcedeleon:
This PR adds support for `llvm.ctlz` intrinsic which counts the leading zeros of a value. This resulted in a new `IOpUn` operator. While it might be possible to properly encode its semantics as a big ITE, in most of the cases I have seen, CP should completely remove such instructions so I currently do not see the need to implement this.


ThomasHaas (approved):
LGTM.

PR #409, opened 2023-02-22
Arch precision as a proper option from cmd line

hernanponcedeleon:
This PR allows to configure the arch precision as a proper option rather than having to change it in the code.


ThomasHaas:
LGTM. I guess you have tried running it locally to see if it works?

ThomasHaas (approved):

PR #410, opened 2023-02-24
Logging of general program statistics

ThomasHaas:
This PR adds a pass that logs (info-level) program statistics after all processing.

```
// Example safe_stack
15:04:18 [INFO] LogProgramStatistics.run - 
======== Program statistics ========
#Threads: 4
#Events: 1019
	#Stores: 95
	#Loads: 125
	#Fences: 3
	#Init: 77
#Allocated bytes: 77
========================================
```

The `allocated bytes` are only accurate for boogie programs. For Litmus programs, the number aligns more with the number of variables rather than the total size of the memory, because for litmus programs we have all `MemoryObjects` of size 1 (except arrays maybe?).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LogProgramStatistics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LogProgramStatistics.java
@@ -1,0 +1,61 @@
+package com.dat3m.dartagnan.program.processing;
+
+import com.dat3m.dartagnan.program.Program;
+import com.dat3m.dartagnan.program.event.core.*;
+import com.dat3m.dartagnan.program.memory.MemoryObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LogProgramStatistics.java
```diff
@@ -0,0 +1,61 @@
...
+
+        int totalEventCount = 0;
+        int storeCount = 0;
+        int loadCount = 0;
+        int initCount = 0;
+        int fenceCount = 0;
+        for (Event e : program.getEvents()) {
+            totalEventCount++;
+            if (e instanceof Store) {
+                storeCount++;
```
hernanponcedeleon, line 31: Here we count `Init` events twice, right? Do you really want that?

ThomasHaas: `Init` events are no `Store` events. Nothing is counted twice.

PR #411, opened 2023-02-24
Improved memory allocation

ThomasHaas:
This PR allows us to distinguish between statically and dynamically allocated memory.
For statically allocated memory, we exploit the static initialization generated by Smack to create `Init` events for only a subset of the statically allocated address space.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/VisitorBoogie.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/VisitorBoogie.java
@@ -96,12 +96,27 @@ public class VisitorBoogie extends BoogieBaseVisitor<Object> {
 	protected BeginAtomic currentBeginAtomic = null;
 	protected Call_cmdContext atomicMode = null;
 	 
-	private ExprSimplifier exprSimplifier = new ExprSimplifier();
-	
-	private final List<String> smackDummyVariables = Arrays.asList("$M.0", "$exn", "$exnv", "$CurrAddr", "$GLOBALS_BOTTOM", "$EXTERNS_BOTTOM", "$MALLOC_TOP", "__SMACK_code", "__SMACK_decls", "__SMACK_top_decl", "$1024.ref", "$0.ref", "$1.ref", ".str.1", "env_value_str", ".str.1.3", ".str.19", "errno_global", "$CurrAddr");
+	private final ExprSimplifier exprSimplifier = new ExprSimplifier();
+
 
 	public VisitorBoogie(ProgramBuilder pb) {
 		this.programBuilder = pb;
+	}
+
+	private final List<String> smackDummyVariables =
+			Arrays.asList("$M.0", "$exn", "$exnv", "$CurrAddr", "$GLOBALS_BOTTOM",
+			"$EXTERNS_BOTTOM", "$MALLOC_TOP", "__SMACK_code", "__SMACK_decls", "__SMACK_top_decl",
+			"$1024.ref", "$0.ref", "$1.ref", "env_value_str", "errno_global", "$CurrAddr");
+	private boolean doIgnoreVariable(String varName) {
+		// We ignore some smack-generated dummy variables
+		if (smackDummyVariables.contains(varName)) {
+			return true;
+		}
+		// We also ignore all kinds of strings for now.
+		if (varName.contains(".str")) {
+			return true;
+		}
+		return false;
 	}
 	
     @Override
@@ -178,7 +193,7 @@ public class VisitorBoogie extends BoogieBaseVisitor<Object> {
 			if(ctx.getText().contains(":treadLocal")) {
 				threadLocalVariables.add(name);
 			}
-			if(ctx.getText().contains("ref;") && !procedures.containsKey(name) && !smackDummyVariables.contains(name)) {
+			if(ctx.getText().contains("ref;") && !procedures.containsKey(name) && !doIgnoreVariable(name)) {
 				int size = ctx.getText().contains(":allocSize")
 					? Integer.parseInt(ctx.getText().split(":allocSize")[1].split("}")[0])
 					: 1;
@@ -202,7 +217,7 @@ public class VisitorBoogie extends BoogieBaseVisitor<Object> {
         for(Attr_typed_idents_whereContext atiwC : ctx.typed_idents_wheres().attr_typed_idents_where()) {
             for(ParseTree ident : atiwC.typed_idents_where().typed_idents().idents().Ident()) {
                 String name = ident.getText();
-                if(!smackDummyVariables.contains(name)) {
+                if(!doIgnoreVariable(name)) {
                     programBuilder.newObject(name,1);
                 }
             }
@@ -412,7 +427,7 @@ public class VisitorBoogie extends BoogieBaseVisitor<Object> {
 	        	continue;
 	        }		
 			String name = ctx.Ident(i).getText();
-	        if(smackDummyVariables.contains(name)) {
+	        if(doIgnoreVariable(name)) {
 	        	continue;
 	        }
 			if(constantsTypeMap.containsKey(name)) {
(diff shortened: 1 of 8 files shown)


ThomasHaas:
It seems that there are some circumstances under which `Smack` fails to generate the static initialization code (large arrays?).
This PR fails in such cases.

If it only happens on large arrays, I can add fallback code that initializes each byte,
For example, we would expect that every 8 bytes there must be at least one initialization (since the largest primitive type is usually 8 bytes). If we find less initialization events in the Boogie code, we can assume Smack failed to generate the necessary code and then fall back to generate the init events ourselves.

But this heuristic is only valid if Smack fails on large arrays. If there are other circumstances that generate small allocations without inits, then we have a problem.
@hernanponcedeleon Our unit tests work just fine. Can you somehow reduce the problematic code you have found? Maybe it is enough to cut out everything but the array declaration.

hernanponcedeleon:
The problem is definitely in this line
```
fieldsToInit = memObj.getInitializedFields();
```
The boogie code has this constant declaration
```
const {:allocSize 256} {:count 64} array: ref;
```
but `__SMACK_static_init()` makes no reference to `array`.

I took a look to some other bpl files and it seems whenever a constant is defined with this `count` annotation, the variable is properly initialized by `__SMACK_static_init()` so I am not sure why it is not the case in the case where I see problems.

Adding the following to `VisitorBoogie.visitConst_decl` after creating the object seems to fix the problem, but I am not sure if it may cause problems in those cases where variables are properly initialize by `__SMACK_static_init()`
```
if(size > 1 && ctx.getText().contains(":count")) {
  int offsets = Integer.parseInt(ctx.getText().split(":count")[1].split("}")[0]);
  for(int i = 0; i < offsets; i++) {
    programBuilder.getObject(name).setInitialValue(i, IValue.ZERO);
  }
}
```

ThomasHaas:
The code makes a distinction between static and dynamic allocation.
If the static cases fails because explicit initialization is missing, we can just remove the "static case" altogether and it would work.
A different approach would be more of a guessing approach: If the allocation is large but it does not have at least one initialization every 8 byte, then smack failed to generate the initialization and we can go ahead and do it ourselves.
This would not interfere with correctly initialized code and might be good enough in practice

hernanponcedeleon:
The final proposed LGTM. 

However, I still don't see what would be the consequences of removing the "static case".

ThomasHaas:
In the dynamic case, the code has no extra information and just initializes each single address. We can just do this for the static case as well and then we are back to exactly what `development` does right now, with the only difference that the `MemoryAllocation` pass creates all init events rather than half of them being created during parsing (the static inits) and half of them later on (the dynamic ones).
So in the worst case, this PR is just a refactoring to move all allocation/init code into a single pass.

hernanponcedeleon:
But then we would miss the performance boost from the current code which is what makes it cool :D 

I would prefer to go with the solution of using the `count` annotation to initialize the values when parsing constants. We are already making use of the `allocSize` annotation, so I don't see why not using that one too.

ThomasHaas:
Sure, we want to keep the number of init events small, that's why I suggested "guessing" whether Smack initialized a static object properly.

>
```
if(size > 1 && ctx.getText().contains(":count")) {
  int offsets = Integer.parseInt(ctx.getText().split(":count")[1].split("}")[0]);
  for(int i = 0; i < offsets; i++) {
    programBuilder.getObject(name).setInitialValue(i, IValue.ZERO);
  }
}
```
I don't quite understand this code. If `count=N` then you initialize the first `N` addresses but no more?
Allocating in steps of `allocSize / count` is also not going to work (fails on arrays of non-primitive type).
So I don't really understand how you want to exploit `count` annotations in any way.

hernanponcedeleon:
You are right. I made my usual mistake of thinking that the step would be magically correct

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/VisitorBoogie.java
```diff
@@ -96,13 +96,28 @@ public class VisitorBoogie extends BoogieBaseVisitor<Object> {
...
+			Arrays.asList("$M.0", "$exn", "$exnv", "$CurrAddr", "$GLOBALS_BOTTOM",
+			"$EXTERNS_BOTTOM", "$MALLOC_TOP", "__SMACK_code", "__SMACK_decls", "__SMACK_top_decl",
+			"$1024.ref", "$0.ref", "$1.ref", "env_value_str", "errno_global", "$CurrAddr");
+	private boolean doIgnoreVariable(String varName) {
+		// We ignore some smack-generated dummy variables
+		if (smackDummyVariables.contains(varName)) {
+			return true;
+		}
+		// We also ignore all kinds of strings for now.
+		if (varName.contains(".str")) {
```
hernanponcedeleon, line 116: Maybe `varName.startsWith(".str")` to minimize changes of matching a real variable?

ThomasHaas: We also skip user-declared string variables, no?
And name-wise, I don't think there are clashes because of the `.` in the front (I don't think real variables have a dot in them).

hernanponcedeleon: I mean, if the user declares e.g., `int no.string`, we would skip this real variable.

ThomasHaas: I don't think that this is a valid name. The only allowed special symbol is `_` I believe.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
```diff
@@ -53,12 +53,38 @@ private int getSize(Malloc malloc) {
...
+        // Addresses are typically at least two byte aligned
+        //      https://stackoverflow.com/questions/23315939/why-2-lsbs-of-32-bit-arm-instruction-address-not-used
+        // Many algorithms rely on this assumption for correctness.
+        // Many objects have even stricter alignment requirements and need up to 8-byte alignment.
+        final BigInteger alignment = BigInteger.valueOf(8);
+        BigInteger nextAddr = alignment;
+        for(MemoryObject memObj : memory.getObjects()) {
+            memObj.setAddress(nextAddr);
+
+            // nextAddr += memObjSize + ((-memObjSize) mod alignment)
```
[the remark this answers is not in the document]
ThomasHaas:
```
nextAddr + size + padding = k*alignment   // Alignment requirement
=> padding = k*alignment - nextAddr - size 
=> padding mod alignment = (-size) mod alignment // k*alignment and nextAddr are 0 mod alignment.
=> padding = (-size) mod alignment // Because padding < alignment
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -22,113 +26,114 @@ public class MemoryObject extends IConst {
...
+    public String getCVar() { return cVar; }
+    public void setCVar(String name) { this.cVar = name; }
+
+    public boolean isStaticallyAllocated() { return isStatic; }
+    public boolean isDynamicallyAllocated() { return !isStatic; }
 
-	public void setCVar(String name) {
-		this.cVar = name;
-	}
+    // Should only be called for statically allocated objects.
```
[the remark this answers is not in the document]
ThomasHaas: Done.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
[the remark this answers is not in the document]
hernanponcedeleon: can you add this to the comment? Otherwise that equation looks like black magic to me

hernanponcedeleon (approved):

PR #414, opened 2023-03-01
Show fr edges in witness

hernanponcedeleon:
This PR adds `fr` edges to the graph witness (which is useful to understand the scheduling points). Both `fr` and `co` edges only bring information about ordering in general. This is not needed between events of the same thread (since the ordering is given by `po`) and thus we filter some edges in the graph. However, `rf` edges also explain how the data flows, and thus keeping `rfi` edges is valuable.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
@@ -35,9 +35,9 @@ public class ExecutionGraphVisualizer {
     private final Graphviz graphviz;
     private Map<Event, String> callStackMapping;
     // By default we do not filter anything
-    private BiPredicate<EventData, EventData> rfFilter = (x, y) -> false;
-    private BiPredicate<EventData, EventData> frFilter = (x, y) -> false;
-    private BiPredicate<EventData, EventData> coFilter = (x, y) -> false;
+    private BiPredicate<EventData, EventData> rfFilter = (x, y) -> true;
+    private BiPredicate<EventData, EventData> frFilter = (x, y) -> true;
+    private BiPredicate<EventData, EventData> coFilter = (x, y) -> true;
     private Map<BigInteger, IExpr> addresses = new HashMap<BigInteger, IExpr>();
 
     public ExecutionGraphVisualizer() {
@@ -97,7 +97,7 @@ public class ExecutionGraphVisualizer {
             EventData r = rw.getKey();
             EventData w = rw.getValue();
 
-            if (ignore(r) || ignore(w) || rfFilter.test(w, r)) {
+            if (ignore(r) || ignore(w) || !rfFilter.test(w, r)) {
                 continue;
             }
 
@@ -119,15 +119,11 @@ public class ExecutionGraphVisualizer {
                 continue;
             }
 
-            for (List<EventData> co : model.getCoherenceMap().values()) {
-                for (int i = 2; i < co.size(); i++) {
-                    // We skip the init writes
-                    EventData w1 = co.get(i - 1);
-                    EventData w2 = co.get(i);                    
-                    if (ignore(w1) || ignore(w2) || !w.equals(w1) || frFilter.test(r, w2)) {
-                        continue;
-                    }
-                    
+            List<EventData> co = model.getCoherenceMap().get(w.getAccessedAddress());
+            // Check if exists w2 : co(w, w2)
+            if(co.indexOf(w) + 1 < co.size()) {
+                EventData w2 = co.get(co.indexOf(w) + 1);
+                if (!ignore(w2) && frFilter.test(r, w2)) {
                     appendEdge(r, w2, model, "label=fr");
                 }
             }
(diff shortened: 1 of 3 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
```diff
@@ -34,8 +34,10 @@ public class ExecutionGraphVisualizer {
 	
     private final Graphviz graphviz;
     private Map<Event, String> callStackMapping;
-    private BiPredicate<EventData, EventData> rfFilter = (x, y) -> true;
-    private BiPredicate<EventData, EventData> coFilter = (x, y) -> true;
+    // By default we do not filter anything
+    private BiPredicate<EventData, EventData> rfFilter = (x, y) -> false;
+    private BiPredicate<EventData, EventData> frFilter = (x, y) -> false;
+    private BiPredicate<EventData, EventData> coFilter = (x, y) -> false;
```
ThomasHaas, line 40: Any reason you inverted the meaning here? Originally, the meaning was "keep the edge if the filter returns true".
This is in line with, e.g., the meaning of `stream.filter`.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
```diff
@@ -99,17 +107,46 @@ private ExecutionGraphVisualizer addReadFrom(ExecutionModel model) {
...
+                    EventData w1 = co.get(i - 1);
+                    EventData w2 = co.get(i);                    
+                    if (ignore(w1) || ignore(w2) || !w.equals(w1) || frFilter.test(r, w2)) {
+                        continue;
+                    }
+                    
+                    appendEdge(r, w2, model, "label=fr");
+                }
+            }
+        }
```
ThomasHaas, line 134: This is unnecessarily complicated. `model.getCoherenceMap` returns a map which gives you the coherence per address.
You can just do `model.getCoherenceMap().get(w.getAccessedAddress())` to get the total order that contains "w".
This list is ordered, so you just need to find the index of `w` in this list and then go over all successors.

```
List<EventData> co = model.getCoherenceMap().get(w.getAccessedAddress());
List<EventData> coLaterWrites = co.subList(co.indexOf(w) + 1, co.size());
for (EventData w2 : coLaterWrites) {
     if (!ignore(w2) && !frFilter.test(r, e2)) {
          appendEdge(r, w2, model, "label=fr");
     }
}
```

That being said, it might be good to leave out the transitive edges and just put the edge to the immediate co-successor (this avoids a quadratic number of edges).

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -426,9 +426,11 @@ private static void generateGraphvizFiles(VerificationTask task, ExecutionModel
         String directoryName = String.format("%s/refinement/%s-%s-debug/", System.getenv("DAT3M_OUTPUT"), programName, task.getProgram().getArch());
         String fileNameBase = String.format("%s-%d", programName, iterationCount);
         // File with reason edges only
-        generateGraphvizFile(model, iterationCount, edgeFilter, directoryName, fileNameBase, new HashMap<>());
+        generateGraphvizFile(model, iterationCount, edgeFilter, edgeFilter, edgeFilter, directoryName, fileNameBase,
+                new HashMap<>());
         // File with all edges
-        generateGraphvizFile(model, iterationCount, (x,y) -> true, directoryName, fileNameBase + "-full", new HashMap<>());
+        generateGraphvizFile(model, iterationCount, (x, y) -> true, (x, y) -> true, (x, y) -> true, directoryName,
+                fileNameBase + "-full", new HashMap<>());
```
ThomasHaas, line 433: You inverted the meaning of filters, but you didn't do it here (not sure if we ever use this code again, but still).

ThomasHaas (changes requested):

dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
hernanponcedeleon: I initially found the meaning confusing, but better to follow the "standard" API meaning.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
hernanponcedeleon: Reverting to previous meaning.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
hernanponcedeleon: The graph is unreadible with the transitive ones, so I will only keep the edge to the immediate co-successor

ThomasHaas (approved):
LGTM

PR #415, opened 2023-03-06
Coverage

hernanponcedeleon:
This PRs adds the option to print a verification coverage report (when `--method=assume`). 

The idea is to collect the events that were executed in any violating execution and compare those wrt the events of the program (before compilation / unrolling, i.e., by comparing `oId`). The report also shows which lines of code were not covered.

Here is a sample output
```
Property-based coverage: 
   -- Events executed by at least one property-violating behavior (including inconsistent behaviors): 99% 
   -- Missing events: 
        T2 / T3 / T4 -> ../cna-verification/kernel/locking/qspinlock.c#204
        T2 / T3 / T4 -> ../cna-verification/kernel/locking/qspinlock.c#335
        T2 / T3 / T4 -> ../cna-verification/kernel/locking/qspinlock.c#536
        T2 / T3 / T4 -> cna-verification/include/asm-generic/qspinlock.h#63
        T2 / T3 / T4 -> cna-verification/include/asm-generic/qspinlock.h#68
```

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionNames.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionNames.java
@@ -10,6 +10,7 @@ public class OptionNames {
 	public static final String SOLVER = "solver";
 	public static final String TIMEOUT = "timeout";
 	public static final String VALIDATE = "validate";
+	public static final String COVERAGE = "coverage";
 
 	// Modeling Options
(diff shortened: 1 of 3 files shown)


ThomasHaas:
I assume the error in the CI is because of the coverage reports that get generated.
I think we should either keep it on info logging level or set the coverage report variable to `false` by default.
Or at least use a no-report configuration in the unit tests.

PR #418, opened 2023-03-07
Round

hernanponcedeleon:
I noticed that due to rounding errors, the coverage report could report 100% coverage even if some events where not covered.
This PR fixes this by using `BigDecimal`s to compute the coverage percentage.


ThomasHaas:
This seems both wrong and overkill :P.
First, you likely want to round down when reporting coverage. Integer division already does this, so `branchCoveragePercentage` should be correct.
For `eventCoveragePercentage` you have this subtraction `100 - X` so you actually want to round `X` up to get a lower bound about coverage. And to me it seems that your BigDecimal implementation still rounds down.

To get consistent rounding behavior, you could just do the full computation with `double` and convert to integer afterwards.
```
P = (int)(100d - ((double)A \ (double)B))
```
Or just create a helper method
```
private static int ceilDiv(int a, int b) {
    int div = a \ b;
    return div * b == a ? div : div + 1;
}
```

PR #419, opened 2023-03-08
Propagate oId, cId, uId to freshly created events

hernanponcedeleon:
Some of our code (e.g. the coverage report) relies on the assumption that events have a proper ` oId, cId, uId` (i.e. different than `-1`) . This PR guarantees that this assumption is fulfilled by coping (from where depends on the pass) those ids when new events are created.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
@@ -64,8 +64,8 @@ public class ComplexBlockSplitting implements ProgramProcessor {
             final Label blockLabel = EventFactory.newLabel(newLabelName);
             final CondJump gotoLabel = EventFactory.newGoto(blockLabel);
 
-            blockLabel.copyIds(condJump);
-            gotoLabel.copyIds(condJump);
+            blockLabel.copyMetadataFrom(condJump);
+            gotoLabel.copyMetadataFrom(condJump);
 
             condJump.insertAfter(gotoLabel);
             gotoLabel.insertAfter(blockLabel);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
@@ -43,9 +43,6 @@ public class ComplexBlockSplitting implements ProgramProcessor {
 
     @Override
     public void run(Program program) {
-        // This might not a strong requirement, but if it is weaken, then events created by
-        // this class need to not only get the corresponding oId, but also cId.
-        Preconditions.checkArgument(!program.isCompiled(), "ComplexBlockSplitting should be run before compilation.");
         final int numBlockSplittings = program.getThreads().stream().mapToInt(this::run).sum();
 
         logger.info("Split {} complex blocks.", numBlockSplittings);
@@ -67,9 +64,8 @@ public class ComplexBlockSplitting implements ProgramProcessor {
             final Label blockLabel = EventFactory.newLabel(newLabelName);
             final CondJump gotoLabel = EventFactory.newGoto(blockLabel);
 
-            // ComplexBlockSplitting is run before compilation, thus we only need to assign oId
-            blockLabel.setOId(condJump.getOId());
-            gotoLabel.setOId(condJump.getOId());
+            blockLabel.copyIds(condJump);
+            gotoLabel.copyIds(condJump);
 
             condJump.insertAfter(gotoLabel);
             gotoLabel.insertAfter(blockLabel);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
@@ -5,8 +5,6 @@ import com.dat3m.dartagnan.program.Thread;
 import com.dat3m.dartagnan.program.event.EventFactory;
 import com.dat3m.dartagnan.program.event.core.CondJump;
 import com.dat3m.dartagnan.program.event.core.Label;
-import com.google.common.base.Preconditions;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -43,9 +41,6 @@ public class ComplexBlockSplitting implements ProgramProcessor {
 
     @Override
     public void run(Program program) {
-        // This might not a strong requirement, but if it is weaken, then events created by
-        // this class need to not only get the corresponding oId, but also cId.
-        Preconditions.checkArgument(!program.isCompiled(), "ComplexBlockSplitting should be run before compilation.");
         final int numBlockSplittings = program.getThreads().stream().mapToInt(this::run).sum();
 
         logger.info("Split {} complex blocks.", numBlockSplittings);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
@@ -5,8 +5,6 @@ import com.dat3m.dartagnan.program.Thread;
 import com.dat3m.dartagnan.program.event.EventFactory;
 import com.dat3m.dartagnan.program.event.core.CondJump;
 import com.dat3m.dartagnan.program.event.core.Label;
-import com.google.common.base.Preconditions;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
@@ -5,6 +5,8 @@ import com.dat3m.dartagnan.program.Thread;
 import com.dat3m.dartagnan.program.event.EventFactory;
 import com.dat3m.dartagnan.program.event.core.CondJump;
 import com.dat3m.dartagnan.program.event.core.Label;
+import com.google.common.base.Preconditions;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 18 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ComplexBlockSplitting.java
```diff
@@ -41,6 +43,9 @@ private ComplexBlockSplitting() {}
 
     @Override
     public void run(Program program) {
+        // This might not a strong requirement, but if it is weaken, then events created by
+        // this class need to not only get the corresponding oId, but also cId.
+        Preconditions.checkArgument(!program.isCompiled(), "ComplexBlockSplitting should be run before compilation.");
```
ThomasHaas, line 48: I think this is a bad idea. We went a long way to get rid of the unnecessary preconditions.
I don't want them back in.
You can copy ALL metadata with `event.copyMetaData` (not sure about the name). This way you don't need to care about when it is actually run. This holds for all the cases you added (I won't mark every single one explicitly).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
```diff
@@ -137,6 +141,13 @@ private void unrollLoop(CondJump loopBackJump, int bound) {
                 endOfLoopMarker.addFilters(Tag.NOOPT);
                 boundEvent.getPredecessor().insertAfter(endOfLoopMarker);
 
+                // Unrolling is currently run after compilation, thus we need to assign both oId and cId.
+                // The final step of run takes care of assigning uId.
+                for(Event e : Arrays.asList(threadExit, boundEvent, endOfLoopMarker)) {
+                    e.setOId(loopBackJump.getOId());
+                    e.setCId(loopBackJump.getCId());
```
ThomasHaas, line 148: It doesn't make sense to copy metadata to `threadExit` cause that event does not get created here: it already exists.

dartagnan/src/test/java/com/dat3m/dartagnan/miscellaneous/AnalysisTest.java
```diff
@@ -73,8 +73,8 @@ public void dependencyMustOverride() throws InvalidConfigurationException {
         b.addChild(0, e5);
 
         Program program = b.build();
-        LoopUnrolling.newInstance().run(program);
         Compilation.newInstance().run(program);
+        LoopUnrolling.newInstance().run(program);
```
ThomasHaas, line 77: This reordering will be unnecessary if you remove the preconditions again (also holds for all the other tests).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
```diff
@@ -130,6 +133,11 @@ private void insertSideEffectChecks(IterationData iter) {
             trackingRegs.add(trackingReg);
 
             final Event execCheck = EventFactory.newExecutionStatus(trackingReg, sideEffect);
+            // DynamicPureLoopCutting is run after unrolling, thus we need to assign all
+            // three of oId, cId and uId.
+            execCheck.setOId(insertionPoint.getOId());
+            execCheck.setCId(insertionPoint.getCId());
+            execCheck.setUId(insertionPoint.getUId());
```
ThomasHaas, line 140: I'm not a big fan of this. Unlike the other cases where you copied metadata over, in this case the inserted event does not have any reasonable association with the metadata you attach to it.
This totally breaks the meaning of what the different Ids even mean!

I know that you want to have this to do "coverage" so that each branch representative has reasonable metadata that connects it to the parsed code.
But rather than destroying the meaning of our metadata to fix this problem for coverage reports, you should rather fix the coverage reporting itself. For example, if the canonical representative of a branch has no metadata (cause we created it via some passes), then maybe just use a different event in the branch to distinguish it from other branches (i.e. compute a mapping `Branch->Event` that gives you "good" representatives).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
```diff
@@ -141,6 +149,11 @@ private void insertSideEffectChecks(IterationData iter) {
         assumeSideEffect.addFilters(Tag.SPINLOOP, Tag.EARLYTERMINATION, Tag.NOOPT);
         final Event spinloopStart = iterInfo.getIterationStart();
         assumeSideEffect.setCFileInformation(spinloopStart.getCLine(), spinloopStart.getSourceCodeFile());
+        // DynamicPureLoopCutting is run after unrolling, thus we need to assign all
+        // three of oId, cId and uId.
+        assumeSideEffect.setOId(spinloopStart.getOId());
+        assumeSideEffect.setCId(spinloopStart.getCId());
+        assumeSideEffect.setUId(spinloopStart.getUId());
```
ThomasHaas, line 156: Same as above

ThomasHaas (changes requested):

dartagnan/src/test/java/com/dat3m/dartagnan/miscellaneous/AnalysisTest.java
hernanponcedeleon: I will keep them simply to be consistent to the order the `ProcessorManager` applies transformations

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
hernanponcedeleon: "Same as above" as "copy all meta-data" or "does not have any reasonable association with the metadata you attach to it"? If the later, I don't agree. We are even setting all the `CFileInformation` due to the association

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
hernanponcedeleon: I agree that the association between `execCheck` and `insertionPoint` is less clear than in other cases (to the point that I initially had a hard time to decide from which event to copy the metadata). 

That being said, I disagree that this destroys the meaning of the metadata. I don't think we have any case where we created an event `a` that is not related in any form with any other event `b` (as soon as we have `b`, we can copy its metadata). For me is actually the opposite: allowing several events to have some id with value `-1` after all processing destroys the meaning of the metadata. My interpretation of ids is that they are unique identifiers and the meaning of `-1` is "it has not yet been assigned". 

Based on all the above, I would consider that `execCheck` is related to `spinloopStart` and copy metadata from it.

ThomasHaas: Regarding the definition of the `ids`: they are defined as the unique global id an event had at some previous point in time, i.e, they track the history of an event over time. The nonexistance of an id then means that the event simply did not exist in any shape or form. In this sense, the value "-1" actually has meaning. because it indicates nonexistence.
Now how do you distinguish multiple events with the same "-1" entries? Well, you just don't compare them on these id's but rather on `globalId` (this the only unique id!). 
If you compare events on, e.g., their `oIds`, then if exactly one of the two events has an `-1` entry, then the events have no common history. If both events have `-1` entries then you, need to check their other ids to see if they have anything in common.

Regarding the "metadata" aspect: Metadata is "additional/optional information", i.e., under ideal circumstance it should be droppable without affecting semantic/functional correctness (our code certainly has places that strictly require metadata, but I think that is a fault in the code). In the case of coverage reports for example, one possible way to achieve this is to put events with `oId=-1` into separate groups(*), i.e., singleton groups. 
Then even if we dropped all our metadata, the report would still be working, but just more fine-grained as now everything is distinguished from one another.
While I'm on this topic: Why don't we just filter out all `oId=-1` and `cLine=-1` events? The former indicates that the event did not exist during parsing, so it surely has no syntactic counterpart in the code that it represents.

>Based on all the above, I would consider that `execCheck` is related to `spinloopStart` and copy metadata from it.

`spinloopStart` is a reasonable choice. Intuitively, I would say that the most appropriate choice (if any), would be to match the metadata of the extra events (both the `assumeSideEffect` and the `execCheck`) to the backjump of the loop.
Unfortunately, I'm not too sure if we retain the backjump metadata anywhere except possibly the very last iteration where we insert the bound. The second best choice is the loop header label (`spinloopStart`) and I can live with that :).


(*) If you want to be more precise in such cases, you could group events by their other ids: if `oId` fails, try `cId`. If that fails, try `uId`, and if that fails try `globalId` (which will always be able to distinguish events by its uniqueness).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
ThomasHaas: Both I guess. if you copy all ids and the source information, then you are copying everything, so just do `copyMetadata`.
The argument about already associating `CFileInformation` is only semi-valid.
First, one may argue that associating the `CFileInformation` is questionable to begin with. Indeed, I was surprised that we already did (cause I didn't add this when writing the code!), but I remember that you added it to track which loops ended up in deadlocks. This is a totally fine reason, and I agree there.
However, I don't think that matching `Cline` in any way correlates with the other ids also matching. I mean, it is totally common that two same-Cline events differ in all their ids.
In any case, matching this event with the loop back jump or the loop header is reasonable (as discussed above), so I'm fine with this.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -442,16 +443,22 @@ private static CharSequence generateCoverageReport(Set<Event> coveredEvents, Pro
         final Set<Integer> coveredOIds = new HashSet<>();
         final Set<Integer> coveredBranches = new HashSet<>();
         for (Event e : coveredEvents) {
-            if (e.getCLine() > 0) {
+            if (e.getCLine() > 0 && e.getOId() != -1) {
```
ThomasHaas, line 446: Maybe we should add `hasOId`, `hasCId`, `hasUid` and `hasCSourceInformation` helper functions to `Event` which do the check against `-1`, so we can get rid of the magic constants.
Just the fact that we have the check `getCline() > 0` but `getOId() != -1` shows how inconsistently we query for the (non)-existence of metadata.

ThomasHaas (approved):
LGTM

PR #422, opened 2023-03-09
Use call stack for everything we report to the user

hernanponcedeleon:
This PR makes sure we consistently report the same kind of information (something link `event: callStack -> file#line`) whenever we report anything to the user (witness graph / violation found / coverage report).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Event.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Event.java
@@ -24,8 +24,8 @@ public abstract class Event implements Encoder, Comparable<Event> {
 	protected int uId = -1;		// Global ID right before unrolling
 	protected int cId = -1;		// Global ID right before compilation
 
-	protected int cLine = -1;				// line in the original C program
-	protected String sourceCodeFile = "";	// filename of the original C program
+	protected int cLine = -1;				    // line in the original C program
+	protected String sourceCodeFilePath  = "";	// path of the original C program
 
 	protected Thread thread; // The thread this event belongs to
 
@@ -49,7 +49,7 @@ public abstract class Event implements Encoder, Comparable<Event> {
 		this.uId = other.uId;
 		this.cId = other.cId;
 		this.cLine = other.cLine;
-		this.sourceCodeFile = other.sourceCodeFile;
+		this.sourceCodeFilePath = other.sourceCodeFilePath;
 	}
 
 	public int getGlobalId() { return globalId; }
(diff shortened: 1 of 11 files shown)


ThomasHaas:
I think we should generalize this even more and compute per event not only the callstack but also the iteration numbers (multiple, if there is nesting).
I added a very rudimentary version of this in the scheduling branch, but I think adding one such proper analysis to development would be great.

hernanponcedeleon:
That seems to be orthogonal (i.e. yet another improvement) to the changes in this PR. Can we tackle that in a different PR?

hernanponcedeleon:
BTW ... assuming you are mostly talking about the coverage report, adding information about loop iterations would gives us path coverage rather that branch coverage as we currently have

ThomasHaas:
> That seems to be orthogonal (i.e. yet another improvement) to the changes in this PR. Can we tackle that in a different PR?

We can tackle this in a different PR.

> BTW ... assuming you are mostly talking about the coverage report, adding information about loop iterations would gives us path coverage rather that branch coverage as we currently have.

No, I'm actually talking about all other use cases besides coverage. In the witness graph of a violation, it would be totally helpful to have a loop iteration number attached to each event. Similarly, when generating scheduling information, we want to distinguish between loop iterations.

And I think all this information about the "context" of some event (i.e., the callstack, the iteration numbers, the surrounding branch, and whatever you can think of), would be nice to have generated by a single analysis.

ThomasHaas:
I think there is one simple but nice feature you should add: report not only the name of the function calls in the call stack, but also the Cline from where they are called (maybe not the source file, cause that would get too large).

hernanponcedeleon:
I thought about this earlier today when I was checking some reports having functions calls, but the line number alone felt weird. I ended up going for a solution that also keeps the file name, but without the whole path. I tried it in several examples (for all, coverage report, witness, violations) and it LGTM

ThomasHaas:
Then I think we can merge the PR.


*As a note for future changes:* I think it is not ideal to mix up the computation of the call stacks and the string representation of that call stack into a single component.
A cleaner and more general design is to have an analysis that computes the call stacks and represents them as true stacks, rather than strings (I can easily imagine cases where one would be interested in the call stacks without actually wanting to print them!).
A second component (a "pretty printer") would then use this information to generate some unified string representation, which we use in all the common use-cases.
This separation of concerns is IMO easy to achieve when writing new components (almost as easy as writing the combined version) but can get hard to recover down-the-line when other components start to built on it.
We can save a lot of headache by doing it the right way directly upfront :).

ThomasHaas (approved):

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Event.java
```diff
@@ -74,7 +74,7 @@ public String getSourceCodeFile() {
 
 	public Event setCFileInformation(int line, String file) {
 		this.cLine = line;
-		this.sourceCodeFile = file;
+		this.sourceCodeFile = file.contains("/") ? file.substring(file.lastIndexOf("/") + 1) : file;
 		return this;
 	}
```
ThomasHaas, line 79: I don't think this is a good idea. You are just dropping information here just because you do not need/want it at some other component. We should never drop free information :).

We should rename this to `sourceCodeFilePath` (it's a whole path after all).
And then we either provide an additional `getSourceCodeFileName` which returns just the file name rather than the whole path, or you add this feature directly into the ` CallStackComputation` class, i.e., it retrieves the file path, extracts the file name and then adds it to the call stack string.

ThomasHaas (approved):
Now I'm happy :)

PR #423, opened 2023-03-12
Add clh_mutex to lock benchmarks

hernanponcedeleon:
This PR adds a new benchmark based on a CLH mutex. The code comes from the [ConcurrencyFreaks repo](https://github.com/pramalhe/ConcurrencyFreaks/blob/master/C11/locks/clh_mutex.c), but I had to make one barrier stronger (otherwise mutual exclusion is violated). Both the correct and incorrect version were added to our usual benchmarks for the CI.


benchmarks/locks/clh_mutex.h
```diff
@@ -0,0 +1,169 @@
...
+ */
+void clh_mutex_lock(clh_mutex_t * self)
+{
+    // Create the new node locked by default, setting islocked=1
+    clh_mutex_node_t *mynode = clh_mutex_create_node(1);
+    clh_mutex_node_t *prev = atomic_exchange(&self->tail, mynode);
+
+    // This thread's node is now in the queue, so wait until it is its turn
+    // The original benchmark had a relaxed load, but this violates mutual exclusion
+    int prev_islocked = atomic_load_explicit(&prev->succ_must_wait, memory_order_relaxed);
```
ThomasHaas, line 137: It's still relaxed here...
Why don't you something like `#IFDEF FAIL` to have both versions in code?

ThomasHaas:
The original version claims to be correct with minimal synchronisation. On the other hand, the claims it makes in the comments do not seem to match the code (e.g., it says the unlock does a "release", but the code actually uses SC accesses).
Is it known that the original version is indeed buggy?

hernanponcedeleon:
I think we cannot trust the comment. While this load is relaxed, the one that updates the variable in the spinning is a SC one (i.e. an overprotected way of giving the acq needed for the synchronization). So if this load could really be relaxed, probably also the one in the spinning.

I also checked with the version of libvsync and indeed it uses an acquire load.

ThomasHaas (approved):
LGTM

PR #424, opened 2023-03-12
Ticket awnsb

hernanponcedeleon:
Yet another benchmark from [ConcurrencyFreaks](https://github.com/pramalhe/ConcurrencyFreaks/blob/master/C11/locks/ticketawn/ticket_awnsb_mutex.c). Barriers were manually relaxed (once again the comments are not correct). It needs `--bound=4` to give any meaningful result (because of the loop in `ticket_awnsb_mutex_init`) and thus I did not add it to our unit tests.


ThomasHaas:
You could add a `_VERIFIER_loop_bound` annotation to unroll that one loop.

ThomasHaas (approved):
LGTM

PR #427, opened 2023-03-15
More consistent way of computing coverage

hernanponcedeleon:
I noticed some unexpected results in the verification coverage (100% event coverage but <100% branch coverage).

Instead of comparing representatives of events, this PR compares the original events (i.e., "is there any event in the program that was not covered?"). Only after that it computes the symmetry / branch representative to add to the sets keeping track of coverage.

I tried the current code on several example and the coverage results make more sense now.

This is a rework of #425.


ThomasHaas (approved):
LGTM

PR #428, opened 2023-03-16
Avoid ConcurrentModificationException in Label.delete()

hernanponcedeleon:
When calling `Label.delete()`, it was possible that the `jumpSet` was modified while iterating over it, resulting in a `ConcurrentModificationException`.

This PR solves the problem by iterating over a snapshot of the set.


ThomasHaas (approved):
I thought I fixed this a long time ago already... I'm super confused.

hernanponcedeleon:
I mentioned this problem to you a few weeks back and you told me this was the way to solve it, but it was never implemented

PR #429, opened 2023-03-16
Memcpy

hernanponcedeleon:
Thie PR adds a new event to handle `memcpy` operations.

There is a new transformation pass that converts this event into a sequence of `Load/Store`s.

Supporting `memcpy` is tricky because the copy happens at the byte granularity. However, we will always (irrespectively of the size) store / initialize the value at the first corresponding offset and do the following loads accordingly.

E.g. the following C code
```
int src[5] = {0,1,2,3,4};
```
results in this static init in the boogie code
```
  $M.0 := $store.i32($M.0, src, 0);
  $M.0 := $store.i32($M.0, $add.ref($add.ref(src, $mul.ref(0, 20)), $mul.ref(1, 4)), 1);
  $M.0 := $store.i32($M.0, $add.ref($add.ref(src, $mul.ref(0, 20)), $mul.ref(2, 4)), 2);
  $M.0 := $store.i32($M.0, $add.ref($add.ref(src, $mul.ref(0, 20)), $mul.ref(3, 4)), 3);
  $M.0 := $store.i32($M.0, $add.ref($add.ref(src, $mul.ref(0, 20)), $mul.ref(4, 4)), 4);
```
meaning that we have init events for `src + {0,4,8,12,16}`.

The actual call to `memcpy` is `$memcpy.i8($M.0, $M.0, dest, src, 20, false);` which we will unroll 20 times.
This means our modeling is not optimal (we will have e.g. accesses `src + {1,2,3,...19}`) instead of the expected `src + {0,4,8,12,16}`, but this should still be sound in the absence of mix-size accesses (which we always assume anyway).


benchmarks/c/miscellaneous/memcpy_int.c
```diff
@@ -0,0 +1,21 @@
...
+
+#ifdef FAIL
+    int bytesToCopy = sizeof(src) - sizeof(int);
+#else
+    int bytesToCopy = sizeof(src);
+#endif
+    memcpy(dest, src, bytesToCopy);
+    for(int i = 0; i < sizeof(src); i++) {
+        assert(dest[i] == src[i]);
+    }
```
ThomasHaas, line 18: Is this even correct? I think `sizeof(src)` returns `20` rather than `5`. So this program has UB.
I think it should be `sizeof(src) / sizeof(src[0])` to get the number of indices (or just use `#define size 5`);

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/StdProcedures.java
```diff
@@ -70,8 +69,11 @@ public static void handleStdFunction(VisitorBoogie visitor, Call_cmdContext ctx)
             // TODO: Implement this
             return;
         }
-        if (name.startsWith("memcpy") | name.startsWith("$memcpy")) {
-            // TODO: Implement this
+        if (name.equals("$memcpy.i8")) {
+            IExpr dst = (IExpr)ctx.call_params().exprs().expr(2).accept(visitor);
+            IExpr src = (IExpr)ctx.call_params().exprs().expr(3).accept(visitor);
+            IExpr length = (IExpr)ctx.call_params().exprs().expr(4).accept(visitor);
+            visitor.programBuilder.addChild(visitor.threadCount, EventFactory.Std.newMemCpy(dst, src, length));
```
ThomasHaas, line 76: What are the parameters 0 and 1 that get skipped here? Can you add a comment saying what gets skipped?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/std/MemCpy.java
```diff
@@ -0,0 +1,75 @@
...
+    public IExpr getSrc() { return src; }
+    public IExpr getBytes() { return bytes; }
+    public void setBytes(IExpr bytes) { this.bytes = bytes; }
+
+    @Override
+    public ImmutableSet<Register> getDataRegs() {
+        Set<Register> res = new HashSet<Register>();
+        res.addAll(dst.getRegs());
+        res.addAll(src.getRegs());
+        return ImmutableSet.copyOf(res);
```
ThomasHaas, line 47: Technically, those regs are `addr` regs, no?
Fortunately, we do compile the code early now, so messed up register dependencies are no problem.
But maybe we should just throw an exception here to make sure that no code relies on this.

EDIT: also see comment about `MemCpyUnrolling`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -93,6 +93,7 @@ private ProcessingManager(Configuration config) throws InvalidConfigurationExcep
                 RemoveDeadCondJumps.fromConfig(config),
                 reduceSymmetry ? SymmetryReduction.fromConfig(config) : null,
                 MemoryAllocation.newInstance(),
+                MemCpyUnrolling.newInstance(),
```
ThomasHaas, line 96: We cannot do this so late. Not unrolling`MemCopy` during/after compilation will mess up at least spinloop detection.
At the same time, I think you want to do it after CP, so that you know how much to unroll. 
This also stands in contrast to what I said earlier about the data regs, i.e. that you want to throw an exception there.

I think fixing this issue requires a more general change in our infrastructure:
- `MemCpy` needs to be a `MemEvent`, likely tagged as both `Write` and `Load`
- `MemEvents` need to be have an array/list of addresses they access (`MemCpy` accesses 2 addresses)
- We may want to have a `getAddrRegs` method in addition to `getDataRegs`
- We need to revise all passes/analysis to make sure that they can deal with this.

ThomasHaas (commented):
Unfortunately, this feature is not so trivial to add :(

hernanponcedeleon:
This will handle in the rework. Closing.

PR #431, opened 2023-04-25
Missing changes from lkmm.2023.04.07a

hernanponcedeleon:
This PR contains some missing changes to `to-r` and `ppo` from LKMM v6.4 in [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=406037351e08dea03735178bf11046da85f00125).

### DIFF
--- a/cat/linux-kernel.cat
+++ b/cat/linux-kernel.cat
@@ -97,8 +97,8 @@ let dep = addr | data
 let rwdep = (dep | ctrl) ; [W]
 let overwrite = co | fr
 let to-w = rwdep | (overwrite & int) | (addr ; [Plain] ; wmb)
-let to-r = addr | (dep ; [Marked] ; rfi)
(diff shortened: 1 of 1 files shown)


ThomasHaas (approved):
LGTM.

I don't really get Paul's change in `ppo` from `fence` to `fence & int` though.
He says it "makes ppo a subrelation of po", but all the fence relations are already subrelations of `po` by construction.

ThomasHaas:
I took our lkmm.cat (the one you updated) as reference and there `po-unlock-lock-po` is not part of the fence relation. It is part of `ppo` and already intersected with `int` to remove the edge case mentioned by Jonas.
Maybe Paul's comment refers to this change which we already have, but it is still a mystery to me why he intersected `fence` with `int` now. Maybe it's just a safety thing and he expects to make changes in the future that could accidentally make `fence` not a subset of `po`.

PR #432, opened 2023-04-26
Fix MacOS x86 profile

ThomasHaas:


hernanponcedeleon:
I don't have a machine to test this. @ThomasHaas I assume you tested it in your Mac with Intel processor.

ThomasHaas:
Yes, the setting was simply wrong and it didn't pick up the profile.
If you do `mvn --version` you can see what the architecture string should be, and it was `x86_64` rather than just `x86`.
I assume no one is running a 32bit version anymore, so I didn't look into that.

PR #434, opened 2023-04-26
Expression encoder

xeren:
This PR moves the code of encoding boolean and integer expressions into a visitor.  It removes the respective interface methods from `ExprInterface` and the encoding methods from the operator classes.

This is a design-PR.  It adds no new features and should have no impact the performance of the application.  It will, however, ease the planned extension of the expression type system.  It also allows to better configure the encoding of expressions.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
@@ -503,7 +503,7 @@ class ExpressionEncoder implements ExpressionVisitor<Formula> {
 
     @Override
     public Formula visit(Location location) {
-        checkState(event != null, "Cannot evaluate %s at event %s.", location, event);
+        checkState(event == null, "Cannot evaluate %s at event %s.", location, event);
         return getLastMemValueExpr(location.getMemoryObject(), location.getOffset(), formulaManager);
     }
 }


xeren:
The point that bugged me here is that `CBinOp.encode` is used both by `Atom` and `AssertBasic`.  We could merge `AbstractAssert` into `BExpr` by adding `Final` and `EventLocal` operators, thus adding the feature to express any local state of any thread, as well as last values, from anywhere.

ThomasHaas:
`EventLocal` is basically an SSA expression, right? Just a unary operator that takes an event as additional input (the point of evaluation). If so, what is the value of this expression if the event does not get executed?

What if we build assertions with our regular expressions, and every reference to a memory address or register is always assumed to mean the final value? We could probably give unique register names for the few `AssertInline` usages we have, so that the final register value coincides with the one produced by a certain event

Anyway, I'm not against merging most of the assertion language into `BExpr`.
But if we go with very generic `EventLocal`/`SSA` operators, we might need to have `UserLists` in `Event` as we have discussed before.
Cause judging if an event is deletable becomes harder without this feature: not only do you need to traverse the program to find usages of this event, but you also have to traverse assertions and possibly other "external" expressions for any references.
Alternatively, we can also allow for generic `EventLocal` operators and just internally decide (and document) under what circumstances we make use of them, and under what circumstances we can (temporarily) break them.
For example, currently it is no problem to delete `Local`s that are referenced by `AssertInline` because we know that we are going to recompute the whole assertion after processing, so that outdated/broken `AssertInline` instances do not matter (indeed, I think we might actually compute them only once after all processing is done).

xeren:
Some `EventLocal(e1, x)` inside an expression of another event `e2` would indeed be undefined, if `e1` does not imply the execution of `e2`.  So users would have to be careful where to place them.  They also introduce an unwanted dependency from expressions to events.  Like you said, it would not be an _immediate_ problem, since those operations only occur after the program transformations.

ThomasHaas:
> Some `EventLocal(e1, x)` inside an expression of another event `e2` would indeed be undefined, if `e1` does not imply the execution of `e2`. 

I think you mean if `e2` does not imply the execution of `e1`. 

> So users would have to be careful where to place them. They also introduce an unwanted dependency from expressions to events. Like you said, it would not be an _immediate_ problem, since those operations only occur after the program transformations.

If we can guarantee that they only appear after program transformation, then everything is fine I guess.

So do we want to do it like this?
- We can have special expressions that reference events.
- There is no generic framework for dealing with those expressions when it comes to maintaining them: they are handled case-by-case with special rules.
  - If they are only created after program processing, they are always fine.
  - If we want to use them during program processing, we are responsible for adding special-case handling.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -108,12 +110,24 @@ public BooleanFormulaManager getBooleanFormulaManager() {
...
 
+    public BooleanFormula encodeBooleanExpression(Event event, ExprInterface expression) {
+        return new ExpressionEncoder(formulaManager, event).encodeBoolean(expression);
+    }
+
+    public Formula encodeIntegerExpression(Event event, ExprInterface expression) {
+        return new ExpressionEncoder(formulaManager, event).encodeInteger(expression);
+    }
+
+    public BooleanFormula encodeAtom(COpBin op, Formula lhs, Formula rhs) {
```
hernanponcedeleon, line 121: `Atom` is just another `BExpr`, why not simply make `encodeBooleanExpression` cover this case? Also, why do the other two take an event as input but not this one? Where is the difference?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -0,0 +1,444 @@
...
+    private final BitvectorFormulaManager bitvectorFormulaManager;
+    private final Event event;
+
+    public ExpressionEncoder(FormulaManager formulaManager, Event event) {
+        this.formulaManager = formulaManager;
+        booleanFormulaManager = formulaManager.getBooleanFormulaManager();
+        integerFormulaManager = formulaManager.getIntegerFormulaManager();
+        bitvectorFormulaManager = formulaManager.getBitvectorFormulaManager();
+        this.event = event;
+    }
```
hernanponcedeleon, line 28: Be consistent with the use of `this`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -0,0 +1,444 @@
...
+                : integerFormulaManager.makeNumber(iValue.getValue());
+    }
+
+    @Override
+    public Formula visit(IExprBin iBin) {
+        Formula lhs = encodeInteger(iBin.getLHS());
+        Formula rhs = encodeInteger(iBin.getRHS());
+        // Some SMT solvers do not support certain theories.
+        // Calling the constructor of the manager in such solvers results in an Exception.
+        // Thus we initialize the manager inside the branches
```
hernanponcedeleon, line 174: These lines seem to be outdated and make me wonder if the initializer in `ExpressionEncoder` would work for those cases there the solver does not support the theory

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -0,0 +1,444 @@
...
+                case OR:
+                    return bitvectorFormulaManager.toIntegerFormula(bitvectorFormulaManager.or(bitvectorFormulaManager.makeBitvector(32, i1), bitvectorFormulaManager.makeBitvector(32, i2)), false);
+                case XOR:
+                    return bitvectorFormulaManager.toIntegerFormula(bitvectorFormulaManager.xor(bitvectorFormulaManager.makeBitvector(32, i1), bitvectorFormulaManager.makeBitvector(32, i2)), false);
+                case L_SHIFT:
+                    return bitvectorFormulaManager.toIntegerFormula(bitvectorFormulaManager.shiftLeft(bitvectorFormulaManager.makeBitvector(32, i1), bitvectorFormulaManager.makeBitvector(32, i2)), false);
+                case R_SHIFT:
+                    return bitvectorFormulaManager.toIntegerFormula(bitvectorFormulaManager.shiftRight(bitvectorFormulaManager.makeBitvector(32, i1), bitvectorFormulaManager.makeBitvector(32, i2), false), false);
+                case AR_SHIFT:
+                    return bitvectorFormulaManager.toIntegerFormula(bitvectorFormulaManager.shiftRight(bitvectorFormulaManager.makeBitvector(32, i1), bitvectorFormulaManager.makeBitvector(32, i2), true), false);
```
hernanponcedeleon, line 201: Why 32 and not arch-precision?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -0,0 +1,444 @@
...
+                case SEXT164:
+                case SEXT816:
+                case SEXT832:
+                case SEXT864:
+                case SEXT1632:
+                case SEXT1664:
+                case SEXT3264:
+                    return inner;
+                default:
+                    // TODO add support for CTLZ. Right now we assume constant propagation gor rid
```
hernanponcedeleon, line 316: gor -> got

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Local.java
```diff
@@ -61,12 +61,26 @@ public String toString() {
...
+                BitvectorFormulaManager bvmgr = context.getFormulaManager().getBitvectorFormulaManager();
+                enc = bmgr.and(enc,
+                        bvmgr.greaterOrEquals((BitvectorFormula) expression, bvmgr.makeBitvector(((INonDet) expr).getPrecision(), min), signed),
+                        bvmgr.lessOrEquals((BitvectorFormula) expression, bvmgr.makeBitvector(((INonDet) expr).getPrecision(), max), signed));
+            } else {
+                IntegerFormulaManager imgr = context.getFormulaManager().getIntegerFormulaManager();
+                enc = bmgr.and(enc,
+                        imgr.greaterOrEquals((NumeralFormula.IntegerFormula) expression, imgr.makeNumber(min)),
+                        imgr.lessOrEquals((NumeralFormula.IntegerFormula) expression, imgr.makeNumber(max)));
+            }
```
hernanponcedeleon, line 81: Too many repeated cast. Extract them to a variable

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -464,11 +467,10 @@ private void extractReadsFrom() {
             for (EventData read : addressedReads.getValue()) {
                 for (EventData write : addressWritesMap.get(address)) {
                     BooleanFormula rfExpr = rf.encode(write.getEvent(), read.getEvent());
-                    // The null check is important: Currently there are cases where no rf-edge between
+                    // The null check in isTrue is important: Currently there are cases where no rf-edge between
                     // init writes and loads get encoded (in case of arrays/structs). This is usually no problem,
                     // since in a well-initialized program, the init write should not be readable anyway.
-					Boolean rfVal = model.evaluate(rfExpr);
-					if (rfVal != null && rfVal) {
+					if (isTrue(rfExpr)) {
```
hernanponcedeleon, line 473: Isn't the comment outdated? It seems you are removing the null check

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
xeren: This method is provided for `AssertBasic`, which requires no event.  Also, it accepts both operands in their encoded form.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
xeren: It would not work properly.  `getIntegerFormulaManager()` will throw an `UnsupportedOperation`.  I am going to fix this.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
xeren: This seems to be an older issue, since I copied it from `IOpBin`.  It requires the arch precision to be positive / bounded.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
xeren: The null-check is implicit: `TRUE.equals(null) == false`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
ThomasHaas: Maybe you should rename the methods with a suffix "At": `encodeBooleanExpressionAt`.
If you then reverse the order of parameters, I think it fits nicely.

You also might want to rename `encodeAtom`. The issue is that we use `Atom` to refer to an expression class, yet this function does not take an `Atom` as input.
Maybe rename it to `encodeAtomicFormula`, `encodeFormulaAtom`, or even `encodeFormulaComparison`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -0,0 +1,488 @@
...
+
+    private IntegerFormulaManager integerFormulaManager() {
+        return formulaManager.getIntegerFormulaManager();
+    }
+
+    private BitvectorFormulaManager bitvectorFormulaManager() {
+        return formulaManager.getBitvectorFormulaManager();
+    }
+
+    public BooleanFormula encodeBoolean(ExprInterface expression) {
```
ThomasHaas, line 35: I would rename this to `encodeAsBoolean` to make it more explicit that the expression itself need not be a boolean one.

Same for `encode(As)Integer`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -0,0 +1,488 @@
...
+
+    private final FormulaManager formulaManager;
+    private final BooleanFormulaManager booleanFormulaManager;
+    private final Event event;
+
+    public ExpressionEncoder(FormulaManager formulaManager, Event event) {
+        this.formulaManager = formulaManager;
+        this.booleanFormulaManager = formulaManager.getBooleanFormulaManager();
+        this.event = event;
+    }
```
ThomasHaas, line 25: I somewhat dislike that this Encoder works differently than the other one's, namely, that it is *used by* the `EncodingContext` while the other Encoders *use* the `EncodingContext`.

The fact that this Encoder is package-private remedies the issue though, so I'm fine with that for now.
(Feel free to mark this comment as resolved)

ThomasHaas (changes requested):
The PR looks almost ready to be merged. Just some more minor changes are needed.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: Arch precision cannot be used here, because this code is only executed when `arch-precision = -1` holds.
More precisely, the code runs when we do an integer-based encoding but run into BV operations.
In that case, we need to convert the integers to BV, but have no idea what bit-width we should use (since arch-precision=-1), though I think it woukd be more safe to default to 64 bits rather than 32.

On the other hand, we can keep it like this for now as the issue will go away with the coming rework.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/IConst.java
```diff
@@ -18,12 +16,7 @@ public abstract class IConst extends IExpr implements LastValueInterface {
...
-	@Override
-    public Formula toIntFormula(Event e, FormulaManager m) {
-		return toIntFormula(m);
-	}
-
-	public Formula toIntFormula(FormulaManager fmgr) {
+    public Formula toIntFormula(FormulaManager fmgr) {
 		return getPrecision() > 0
 				? fmgr.getBitvectorFormulaManager().makeBitvector(getPrecision(),getValue())
 				: fmgr.getIntegerFormulaManager().makeNumber(getValue());
```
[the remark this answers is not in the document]
xeren: `IConst.toIntFormula(FormulaManager)` is still used by `getLastValueExpr(FormulaManager)`.  I could let `ExpressionEncoder` also accept `event = null` and then let it cover this method, as well.  This could also be a step into merging `AbstractAssert` into `ExprInterface`.  Just trying to locally evaluate `Location` will throw an exception.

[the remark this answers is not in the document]
xeren: There is also still `Register.toIntFormulaResult(Event, FormulaManager)`.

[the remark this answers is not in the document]
xeren: Update: `LastValueInterface` has been removed and there is no usage of `org.sosy_lab.java_smt.api` inside `com.dat3m.dartagnan.expression` or any implementation of `ExprInterface` anymore.

ThomasHaas (approved):
LGTM

PR #435, opened 2023-04-26
Hash table

hernanponcedeleon:
This PR adds a new benchmark based on a (simple) hash table. The library code is correct event if all atomic instructions are relaxed (see comment in `hash_table.h` for an explanation). However, because of the relaxed accesses, the user might need to provide barriers on certain client code, e.g., if the hash table is used to store the flag in the MP pattern.


benchmarks/lfds/hash_table.h
```diff
@@ -0,0 +1,118 @@
...
+	// h ^= h >> 16;
+	// h *= 0x85ebca6b;
+	// h ^= h >> 13;
+	// h *= 0xc2b2ae35;
+	// h ^= h >> 16;
+
+    // The above causes problems to the SMT solver.
+    // Since this does not affect correctness, we simply use identity.
+	return h;
+}
```
ThomasHaas, line 51: It would be interesting to see how uninterpreted functions perform in this case.
Maybe once we reworked Dartagnan`s internals, it could be easy to just treat extern functions as uninterpreted (assuming no implementation is given).

ThomasHaas (commented):
Can you double check the formatting in the c files?
It seems there is a mixture of spaces and tabs again. If so, can you change everything to spaces?
Apart from that, it looks good to me

ThomasHaas (approved):
LGTM

PR #436, opened 2023-04-27
Set LD_LIBRARY_PATH in docker container

hernanponcedeleon:
The `atomic-replace` pass was not able to find the library in the docker container. This PR fixes the problem.
Since @ThomasHaas seems to have problems running docker, @xeren can you please test this?


ThomasHaas (approved):
LGTM

PR #437, opened 2023-04-27
Implicit knowledge

xeren:
Reduces the memory usage of the Relation Analysis.  The initial knowledge of base relations like `ext`, `int`, `po` and `X*Y` is not explicitly stored.  This comes with hopefully-negligible runtime costs.


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -1864,8 +2149,8 @@ private static void update(Map<Event, List<Event>> map, Collection<Tuple> set) {
         }
     }
 
-    private boolean implies(Event y, Event z) {
-        return exec.isImplied(z, y);
+    private boolean implies(Event premised, Event concluded) {
+        return exec.isImplied(concluded, premised);
```
ThomasHaas, line 2153: I think the order is wrong: `exec.isImplied(x, y)` already means `exec(x) => exec(y)`.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/SetUtil.java
```diff
@@ -84,4 +86,125 @@ public void clear() {
 
         }
     }
+
+    // Manages a distinct union of a modifiable and an immutable set.
+    public static final class GroundedSet<T> extends AbstractSet<T> {
+
+        private final Set<? extends T> ground;
+        private final Set<T> added;
+        private final Set<Object> removed;
```
ThomasHaas, line 95: You can make this a `Set<T>`. Notice that you only ever add an element to `removed` if it is contained in `ground`, which guarantees that the removed element must be compatible with type `T`.
Therefore, you can soundly do a type cast in the respective code, i.e., you can do `removed.add((T)o)`.
You can then also add a `@SuppressWarnings("unchecked")` annotation to silence the compiler.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/SetUtil.java
```diff
@@ -84,4 +86,125 @@ public void clear() {
...
+
+        public GroundedSet(Set<? extends T> groundSet, Set<T> addedSet) {
+            this(groundSet, addedSet, new HashSet<>());
+        }
+
+        public GroundedSet(Set<? extends T> groundSet, Set<T> addedSet, Set<Object> removedSet) {
+            ground = groundSet;
+            added = addedSet;
+            removed = removedSet;
+        }
```
ThomasHaas, line 109: These constructor can violate the (class) invariants. I assume that `RelationAnalysis` will guarantee that this does not happen, but in the general case, this is problematic.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/TwoSolvers.java
```diff
@@ -97,5 +100,6 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
         // For Safety specs, we have SAT=FAIL, but for reachability specs, we have SAT=PASS
         res = Property.getCombinedType(task.getProperty(), task) == Property.Type.SAFETY ? res : res.invert();
         logger.info("Verification finished with result " + res);
+        System.out.printf("%s;%s;%d;%d%n", task.getProgram().getName(), res, t2 - t1, System.nanoTime() - t0);
```
ThomasHaas, line 103: You need to remove the debug printing code.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -205,11 +210,11 @@ public Set<Tuple> findTransitivelyImpliedCo(Relation co) {
             final MemoryEvent z = (MemoryEvent) t.getSecond();
             final boolean hasIntermediary = mustOut.apply(x).stream().map(Tuple::getSecond)
                     .anyMatch(y -> y != x && y != z &&
-                            (exec.isImplied(x, y) || exec.isImplied(z, y)) &&
+                            (implies(y, x) || implies(y, z)) &&
```
ThomasHaas, line 213: The intention is to check `exec(x) => exec(y) || exec(z) => exec(y)`, i.e., that either of the edge events implies some intermediary event `y`. 
The method name "implies" suggests that you try to check the wrong direction though.
However, as I already mentioned, you also inverted the code in "implies" which causes this to be correct again...
In short: you should avoid this double inversion and revert to essentially the original code (you can keep the shorter helper function though).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -47,20 +53,19 @@
...
-import static java.util.stream.Collectors.toSet;
-import static java.util.stream.IntStream.iterate;
 
 @Options
 public class RelationAnalysis {
 
     private static final Logger logger = LogManager.getLogger(RelationAnalysis.class);
 
-    private static final Set<Tuple> EMPTY_SET = new HashSet<>();
+    private static final HashSet<Tuple> EMPTY_SET = new HashSet<>();
```
ThomasHaas, line 68: You should keep using `Set<Tuple>` unless the more concrete type exposes more functionality (which it does not IIRC). The JIT compiler is aware of the true type, so you will not get any performance benefit whatsoever when using a concrete type.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -220,75 +225,125 @@ public Set<Tuple> findTransitivelyImpliedCo(Relation co) {
...
-                qGlobal.computeIfAbsent(r, x -> new ArrayList<>(1))
-                        .add(new Delta(k.may, k.must));
+        final Initializer initializer = new Initializer();
+        final Propagator propagator = new Propagator();
+        for (final Set<Relation> scc : getComponents(memoryModel.getRelations(), Relation::getDependencies)) {
+            // Determine initialization order.
+            final var initializationOrder = new ArrayList<Relation>(scc.size());
+            final var feedbackVertexSet = new HashSet<Relation>();
+            if (scc.size() == 1) {
+                initializationOrder.add(scc.iterator().next());
```
ThomasHaas, line 242: Just use `addAll(scc)`

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -220,75 +225,125 @@ public Set<Tuple> findTransitivelyImpliedCo(Relation co) {
...
+                for (final Relation r : scc) {
+                    (r.getName().isEmpty() ? unnamedRelations : feedbackVertexSet).add(r);
+                }
+                for (final Relation r : feedbackVertexSet) {
+                    knowledgeMap.put(r, new Knowledge(new HashSet<>(), new HashSet<>()));
+                }
+                for (final Set<Relation> set : getComponents(unnamedRelations,
+                        r -> Iterables.filter(r.getDependencies(), unnamedRelations::contains))) {
+                    verify(set.size() == 1, "Recursive group with cycles of unnamed relations.");
+                    initializationOrder.add(set.iterator().next());
```
ThomasHaas, line 255: `addAll(set)`

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -505,59 +860,189 @@ public Knowledge visitIdentity(Relation rel, Filter set) {
             return new Knowledge(must, enableMustSets ? new HashSet<>(must) : EMPTY_SET);
         }
 
+        private final class ExternalSet extends AbstractSet<Tuple> {
+
+            @Override
+            public Iterator<Tuple> iterator() {
+                // No test for exec.areMutuallyExclusive, since that currently does not span across threads.
+                // This stream-based approach might be more expensive than writing an own iterator.
```
ThomasHaas, line 868: This comment is not true since I reworked how threading works, however, events between threads are only considered mutual exclusive if their creator events are mutually exclusive, which is rarely the case in practice.
So you can keep the implementation, but change the comment into a "TODO" or "FIXME".

hernanponcedeleon:
> Can we close this? This PR is over a year old and recent changes like #743 are going in the same direction.

@xeren ?

xeren:
Yes, this can be closed.

PR #440, opened 2023-05-09
Tag cleanup

ThomasHaas:
This PR cleans up our usage of Tags:
(1) Dubious usages of tags got replaced (e.g., `e.is(Tag.JUMP)` got replaced by `e instanceof CondJump`)
(2) Unused Tags got removed
(3) Tags that are used only internally and should not be exposed to CAT are now prefixed by "__" (e.g. "__BOUND").

In order to implement change (1), some code changes have been made, but none of which should affect functionality.
There is a slight change in the semantics of `ctrlDirect` in that it does not relate unconditional jumps to any events anymore (this should not affect the semantics of `ctrl = idd+;ctrlDirect & (RxV)` since unconditional jumps never had incoming `idd` edges).


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -117,8 +114,8 @@ private C11() {
         public static final String ATOMIC               = "A";
         public static final String NONATOMIC            = "NA";
 
-        public static final String PTHREAD              = "PTHREAD";
-        public static final String LOCK                 = "LOCK";
+        public static final String PTHREAD              = "__PTHREAD";
+        public static final String LOCK                 = "LK";
```
hernanponcedeleon, line 118: I am not too sure about this (i.e. should be `__LK`?). LKMM does not use this tag (it splits the lock into the read an write parts `LKR, LKW`). In SVCOMP we use the tag as an internal one. [C11 mentions it, but we have that part of the model commented out](https://github.com/hernanponcedeleon/Dat3M/blob/abc74af479b39b3005ed75dbe5bd2757dcf3e64b/cat/c11.cat#L46). Also, it is not clear to me where this comes from in the original model. There is this `include c11_los.cat`, but [that one refers to `LS`](https://github.com/herd/herdtools7/blob/master/herd/libdir/c11_los.cat)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -1,39 +1,36 @@
...
     private Tag() { }
 
-    public static final String ANY              = "_";
+    public static final String VISIBLE          = "_";
     public static final String INIT             = "IW";
     public static final String READ             = "R";
     public static final String WRITE            = "W";
     public static final String MEMORY           = "M";
     public static final String FENCE            = "F";
+    public static final String STRONG           = "STRONG"; // TODO: Maybe move to C11 or IMM?
```
hernanponcedeleon, line 18: Shouldn't it be `__STRONG`? If so, do we need this now that you added the `s.isStrong` field?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/rmw/RMWStoreExclusive.java
```diff
@@ -11,26 +11,38 @@
...
     public boolean cfImpliesExec() {
-        return is(Tag.STRONG); // Strong RMWs always succeed
+        return isStrong; // Strong RMWs always succeed
     }
 
     @Override
     public String toString() {
-        String tag = is(Tag.STRONG) ? " strong" : "";
+        String tag = isStrong ? " strong" : "";
+        tag += requiresMatchingAddresses ? " addrmatch" : "";
```
hernanponcedeleon, line 45: Do you think this one is really useful?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
ThomasHaas: I initially removed it, but `IMM` actually refers to this Tag, so we need it.
I would still keep the field instead of checkin for `e.is(Tag.STRONG)` in the code. One could argue that the Tag should only be generated when C11 (IMM) is the target and otherwise serves no purpose, while the field is always meaningful.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/rmw/RMWStoreExclusive.java
ThomasHaas: The extra "addrmatch" tag here? It's only useful for debugging in case you accidentally generate the wrong one's for some architecture. I can remove it again if you want.

Issue #82, opened 2021-08-14
Proper type system for Expressions and Registers


ThomasHaas:
This is an extension of #39.

Context: We currently have 2 types of expressions BExpr and IExpr. IExpr can either represent a mathematical integer or a bit vector expression. However, we only have one type of Register that always holds IExpr values.

Problems: It is easy to mix up operations of incompatible types. Currently, a lot of implicit conversions happen which somewhat circumvent the type mismatch problems. Proper typing can make the code more save, can avoid the implicit conversions and can make the encoding more efficient.

Possible solution: 
   - Add BVExpr and maybe use a common supertype ValueExpr for BVExpr and IExpr.
   - For all Expr types add a corresponding Register type (BRegister, IRegister, BVRegister and possibly ValueRegister as a supertype of IRegister and BVRegister).
   - Add toBoolFormula, toIntegerFormula and toBVFormula methods (possibly with a precision parameter for the latter) that allow conversions if necessary and avoid implicit conversions as much as possible (with properly typed registers, conversions should only be necessary when mixing theories).
   - Do the same for getLastValueExpr (this one in particular can introduce a lot of implicit conversions right now)

The reason why a ValueExpr as a supertype for IExpr and BVExpr might be reasonable is because our domain of computation is always split into Booleans + the value domain of Integers, BitVectors or both.

hernanponcedeleon:
@thomas is there anything missing in our current implementation from this?

ThomasHaas:
We can probably close it. Though I'm not entirely sure how well the types are actually used in the encoding.

Issue #39, opened 2021-04-26
Distinguishing boolean and integer registers


ThomasHaas:
When compiling C code to LLVM IR, booleans may end up as i8 (8-bit integers). Encoding them as such is unnecessary and puts more reasoning onto the theory-solver. Encoding them as pure propositonal values may give significant benefit.

Similarly, instrumentations of C code/LLVM code may use suboptimal types for introduced variables (e.g. int32 instead of bool). Automatic detection of bools via static analysis also addresses this problem as a side-effect.

hernanponcedeleon:
Fixed in #478

PR #442, opened 2023-05-12
Types

xeren:
Fundamentally changes how Expressions of a program are constructed.  Introduces the `Type` class.  Tries to tackle issues  #82 and #39.

Implemented:
- Adds `TypeFactory` for constructing and reusing types in a way that is configurable by runtime options.  For now, it uses a common static instance.  I plan to decorate `Program` with its own `getTypeFactory()` in the future.
- Adds `ExpressionFactory` for eagerly constructing simplified expressions.  Unifies `ExprSimplifier`, `IExpr.reduce()` and `SparseConditionalConstantPropagation.ConstantPropagator`, as well as any similar planned process.  This could be made optional.  Like `TypeFactory`, a static instance is accessible for the moment.
- Removes `IExpr` as a boolean-exclusive requirement for expressions.  This means that boolean expressions, as well as other types, can potentially be used for `Register`, `INonDet`, program constants and as operands in other expressions.

Planned:
- Allow registers and memory accesses of different types.  Make use of this in the parsers.


ThomasHaas:
> Adds ExpressionFactory for eagerly constructing simplified expressions. Unifies ExprSimplifier, IExpr.reduce() and SparseConditionalConstantPropagation.ConstantPropagator, as well as any similar planned process. This could be made optional. Like TypeFactory, a static instance is accessible for the moment.

You cannot do "aggressive" simplification during construction time. In particular, you are not allowed to remove any register expressions for otherwise you break dependencies. I suspect that this causes most of the errors you get in the CI (I confirmed some already).
Basically, you are only allowed to do constant folding or perform operations that retain the register accesses (e.g. `a + a -> 2*a`).

dartagnan/src/main/java/com/dat3m/dartagnan/program/expression/type/NumberType.java
```diff
@@ -0,0 +1,5 @@
+package com.dat3m.dartagnan.program.expression.type;
+
+public final class NumberType implements Type {
+    NumberType() {}
+}
```
ThomasHaas, line 5: `NumberType` is a bad name because it is highly unspecific (literally everything can be called a number).
If the separation of types is important, then I would rather use `IntegerType` and `BitVectorType` for mathematical vs. programmatical integers. This is close to logic terminology.
If one wants to stay closer to the programmatic side, i.e., use `IntegerType` for bounded integers, then we should rename `NumberType` to something like `MathIntegerType`.
I kinda dislike the usage of "Integer" in both names, but it is still better than `NumberType`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/expression/type/TypeFactory.java
```diff
@@ -0,0 +1,55 @@
...
+
+public final class TypeFactory {
+
+    private static final TypeFactory instance = new TypeFactory();
+    private final BooleanType booleanType = new BooleanType();
+    private final NumberType numberType = new NumberType();
+    private final PointerType pointerType = new PointerType();
+    private final Map<Integer, IntegerType> integerTypeMap = new HashMap<>();
+    private final Map<Type, Map<Integer, ArrayType>> arrayTypeMap = new HashMap<>();
+    private final Map<List<Type>, AggregateType> aggregateTypeMap = new HashMap<>();
```
ThomasHaas, line 17: Rather than using the constructor parameters as key, it is easier to use the type object itself as key.
As long as there is only a single constructor with a single parameter, it might be fine, but if that changes, then your approach doesn't work nicely (the nested map for array types is already quite ugly IMO).
Just construct the instance and then use it to check if there is a canonical instance in the map.
Indeed, I would probably introduce a `Canonicalizer` class that does precisely that: it takes an object as input and returns a canonical version of it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -35,13 +33,14 @@ public class MemoryObject extends IConst {
...
     MemoryObject(int index, int size, boolean isStaticallyAllocated) {
+        super(TypeFactory.getInstance().getPointerType());
         this.index = index;
         this.size = size;
         this.isStatic = isStaticallyAllocated;
 
         if (isStaticallyAllocated) {
             // Static allocations are default-initialized
-            initialValues.put(0, IValue.ZERO);
+            initialValues.put(0, ExpressionFactory.getInstance().makeZero(TypeFactory.getInstance().getPointerType()));
```
ThomasHaas, line 43: Integer type

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -73,7 +72,7 @@ public Set<Integer> getStaticallyInitializedFields() {
      */
     public IConst getInitialValue(int offset) {
         checkArgument(offset >= 0 && offset < size, "array index out of bounds");
-        return initialValues.getOrDefault(offset, IValue.ZERO);
+        return initialValues.getOrDefault(offset, ExpressionFactory.getInstance().makeZero(TypeFactory.getInstance().getPointerType()));
```
ThomasHaas, line 75: IntegerType

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
```diff
@@ -122,22 +127,24 @@ private void insertSideEffectChecks(IterationData iter) {
 
         final List<Register> trackingRegs = new ArrayList<>();
         Event insertionPoint = iterInfo.getIterationEnd();
+        Type type = TypeFactory.getInstance().getPointerType();
         for (int i = 0; i < sideEffects.size(); i++) {
             final Event sideEffect = sideEffects.get(i);
             final Register trackingReg = thread.newRegister(
                     String.format("Loop%s_%s_%s", loopNumber, iterNumber, i),
-                    GlobalSettings.getArchPrecision());
+                    type);
```
ThomasHaas, line 135: Okay, this shows that you just used the `PointerType` everywhere.
I won't keep pointing it out then, but please add something like `TypeFactory.getPointerSizedIntegerType` and use that instead.
Having `PointerType` for memory-unrelated values just feels too wrong to me.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicPureLoopCutting.java
```diff
@@ -122,22 +127,24 @@ private void insertSideEffectChecks(IterationData iter) {
...
 
-        final BExpr atLeastOneSideEffect = trackingRegs.stream()
-                .map(reg -> (BExpr) new Atom(reg, COpBin.EQ, IValue.ZERO))
-                .reduce(BConst.FALSE, (x, y) -> new BExprBin(x, BOpBin.OR, y));
-        final CondJump assumeSideEffect = EventFactory.newJumpUnless(atLeastOneSideEffect, (Label) thread.getExit());
+        ExpressionFactory expressionFactory = ExpressionFactory.getInstance();
+        Expression noSideEffect = trackingRegs.stream()
+                .map(reg -> expressionFactory.makeBinary(reg, COpBin.NEQ, expressionFactory.makeZero(type)))
+                .reduce(expressionFactory.makeTrue(), (x, y) -> expressionFactory.makeBinary(x, BOpBin.AND, y));
+        final CondJump assumeSideEffect = EventFactory.newJump(noSideEffect, (Label) thread.getExit());
```
ThomasHaas, line 147: Is there any reason why you removed the convenience function `EventFactory.newJumpUnless`?
A big use-case of a factory is to have these convenience methods.

ThomasHaas:
This branch produces a lot of type mismatch warnings.

Also, the newest version is considerably slower in verification (5-6x times slower on LFDS benchmarks) as `refinement` needs more iterations than usual.
This seems to be caused by breaking side-effect detection, which is confirmed by the fact that some lock benchmarks return `UNKNOWN` rather than `PASS` (indicating that some spinloop is not detected).
You will see this once the CI is through, but locally I can produce these issues.

Lastly, I noticed that pointer type expressions give bad string outputs, for example:
```
// From the DGLM unit test
672:     Loop1_2_0 <- status(store(*&mem1, 14:$i7))
673:     if(Loop1_2_0 != com.dat3m.dartagnan.program.expression.type.PointerType@6b4ef459(0)); then goto END_OF_T4	### SPINLOOP
```

ThomasHaas:
I haven't checked it, but do you happen to always use BV encoding now, even if `arch_precision = -1`?
If so, then this also explains a slowdown, especially in the litmus tests which should not contain loops.
Still, the wrong results on the C-benchmarks are related to side-effect/spinloop detection being broken.

ThomasHaas:
There is a tiny regression in constant folding: 
```
// safe_stack TSO.
181:     if(0 != 1); then goto END_OF_T2 // Gets eliminated in development, but not in this branch
```
The performance regressions are gone though, and the branch seems to finally work :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/SparseConditionalConstantPropagation.java
```diff
@@ -87,19 +86,22 @@ private void run(Thread thread) {
...
+                    Register register = ((RegWriter) cur).getResultRegister();
+                    // Loads may generate replacements for register expressions.
+                    // We treat all other register writers as non-constant.
+                    Expression expr = cur instanceof Local ? ((Local) cur).getExpr() : null;
+                    boolean doPropagate = expr != null && expr.getRegs().isEmpty() ||
+                            propagateCopyAssignments && expr instanceof Register;
+                    propagationMap.compute(register, (k, v) -> doPropagate ? expr : null);
+                    if (propagateCopyAssignments) {
+                        propagationMap.values().remove(register);
+                    }
```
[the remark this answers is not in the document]
xeren: If `propagateCopyAssignments` is true, `propagationMap` also receives register-register mappings `r`-`s`, leading to occurrences of `r` being replaced by the expression `s`, until `r` is overwritten again.  This is allowed, as long as both registers hold the same value.  After `s := t`, the equality should break, as well.
Consider a program like
```
load s := *x;
r := s;
s := 1;
assert r == 0;
```

You could reproduce this on `development`:
```
@Test
public void constantPropagationWithCopyAssignment() {
    ProgramBuilder b = new ProgramBuilder(Program.SourceLanguage.LITMUS);
    b.initThread(0);
    Register register0 = b.getOrCreateRegister(0, "r0", getArchPrecision());
    Register register1 = b.getOrCreateRegister(0, "r1", getArchPrecision());
    MemoryObject object = b.newObject("x", 1);
    Load event0 = newLoad(register0, object, "");
    b.addChild(0, event0);
    Local event1 = newLocal(register1, register0);
    b.addChild(0, event1);
    Store event2 = newStore(object, new IExprBin(register0, IOpBin.PLUS, IValue.ONE), "");
    b.addChild(0, event2);
    Load event3 = newLoad(register0, object, "");
    b.addChild(0, event3);
    Store event4 = newStore(object, register1, "");
    b.addChild(0, event4);
    Program program = b.build();
    LoopUnrolling.newInstance().run(program);
    SparseConditionalConstantPropagation.newInstance().run(program);
    assertEquals(event4.getMemValue(), register1);
}
```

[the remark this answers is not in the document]
ThomasHaas: Makes sense. That is why this option caused problems many times.

ThomasHaas:
I feel like this PR is getting quite out-of-hand. It seems that every change breaks something and then is followed by yet another change that breaks something else.
There are over 100 commits, none of which result in a working version, so it is not even possible to split the commit history into chunks.

I think that this PR should get reorganized:
- (1) A Type PR, that replaces all `precisions` by types and possibly adds new expression classes. It should not introduce expression factories, do proper type checking, nor care about the actual types in the encoding yet (i.e., just use the `arch_precision` as we do now). It should also not introduce any kind of new expression optimizations.
- (2) A PR that introduces expression factories, possibly with some optimizations (e.g., constant folding). Though I would even avoid doing any optimizations in the factory at all.
- (3) A PR that introduces type factories (if needed). This PR might come before (2), if it makes more sense.
Also, I'm not sure if we really need instantiable factories or if we should simply stick with static factories for now.
- (4) A PR that tries to enforce proper typing, possibly by inserting explicit casts wherever needed.
- (5) Once the types are there and the program is properly typed, we can create a PR that exploits this typing for the encoding. Up until then, the encoding just works like it used to: it just uses `arch_precision` for all integers.

This is just a rough sketch and we actually might need some other PRs inbetween. But the point is, that I can see this PR split into at least 5 smaller PRs that we can properly review and discuss.
Not only that, it also allows us to split work better. I haven't touched anything since last week, because for pretty much all the future changes I need a `Type` class, which I hoped would be introduced by this PR. But apart from Types existing (PR (1)), there is nothing else required, i.e., PRs (2)-(5) are not even needed. In fact, exploiting the types could (in principle) be delayed until we have switched over to LLVM and our new internal IR, so it is extremely orthogonal to most of the rework.

xeren:
I then suggest splitting (1) into (1.1) Adding Type as an alternative, and (1.2) Replace precision entirely.  Then do (1.2) after (3).  It involves over 200 places, that each will have to be modified multiple times, otherwise.

ThomasHaas:
Sure, more PRs are better :)

ThomasHaas:
Can we close this?

hernanponcedeleon:
Works for me. 

@xeren ?

I also want to do some cleaning of all the branches once we get the rework branch merged.

xeren:
These changes have been implemented over time from #469 to #493.  This stale branch has nothing more in store.

PR #445, opened 2023-05-16
Add Metadata

ThomasHaas:
This PR adds `Metadata` to `Event` and moves `oId`, `uId`, `cId`, and `cLine/sourceFile` into `Metadata` classes.

There are some changes left to do, for example, `Event.setCFileInformation` can be replaced by `Event.setMetadata(new SourceLocation(fileName, cLine))` (internally, the former function now just calls the latter). However, currently there are around 40 calls to `Event.setCFileInformation` and another (concurrent) PR (#442) heavily reduces this number. I think it is easier to wait until that PR is merged to make the changes.

*Other changes:* `oId` has been renamed to `ParseId` (maybe rename to `ParsingId`?) to more clearly communicate that this Id was assigned at/after parsing time.

*Potential issues:* The CI does not test coverage reports nor witness generation and those make most use of the metadata. Some manual testing is needed to make sure that everything works as expected. I tested coverage reports over the LFDS benchmarks and it seemed fine, but I have not compared with development.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/metadata/MetadataMap.java
```diff
@@ -0,0 +1,36 @@
+package com.dat3m.dartagnan.program.event.metadata;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MetadataMap {
+
+    //private final HashMap<Class<? extends Metadata>, Metadata> metadataMap = new HashMap<>();
```
hernanponcedeleon, line 8: if not needed, remove

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -441,12 +443,13 @@ private static CharSequence generateCoverageReport(Set<Event> coveredEvents, Pro
...
 
         final Set<Event> programEvents = program.getEvents(MemEvent.class).stream()
-                .filter(e -> e.hasCLine() && e.hasOId()).collect(Collectors.toSet());
+                // TODO: Can we have events with source information but without parse id?
+                .filter(e -> e.hasMetadata(SourceLocation.class) && e.hasMetadata(ParseId.class))
+                .collect(Collectors.toSet());
         
         // Track (covered) events and branches via oId
-        final Set<Integer> branches = new HashSet<>();
-        final Set<Integer> coveredOIds = new HashSet<>();
```
hernanponcedeleon, line 448: Were the `coveredOIds` really never used?

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessBuilder.java
```diff
@@ -109,22 +112,22 @@ public WitnessGraph build() {
...
 				Event e = execution.get(i);
 				if (i + 1 < execution.size()) {
 					Event next = execution.get(i + 1);
-					if (e.getCLine() == next.getCLine() && e.getThread() == next.getThread()) {
+					if (e.hasEqualMetadata(next, SourceLocation.class) && e.getThread() == next.getThread()) {
 						continue;
 					}
 				}
 
+				int cLine = e.hasMetadata(SourceLocation.class) ? e.getMetadata(SourceLocation.class).getLineNumber() : -1;
```
hernanponcedeleon, line 127: why not `@unknown` instead of -1?

dartagnan/src/test/java/com/dat3m/dartagnan/miscellaneous/AnalysisTest.java
```diff
@@ -426,6 +427,7 @@ private void assertList(List<?> results, Object... expected) {
     }
 
     private Event findMatchingEventAfterProcessing(Program p, Event orig) {
-        return p.getEvents().stream().filter(e -> e.getOId() == orig.getOId()).findFirst().get();
+        return p.getEvents().stream().filter(e -> e.hasEqualMetadata(orig, ParseId.class)).findFirst().get();
+        //return p.getEvents().stream().filter(e -> e.getOId() == orig.getOId()).findFirst().get();
```
hernanponcedeleon, line 431: Remove if not needed

hernanponcedeleon (commented):
I manually tested witness generation (both for graphviz and svcomp) + coverage report and things seem ok

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: Yes, they are unused in development.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessBuilder.java
ThomasHaas: This is the original code in the `WitnessBuilder`. I have not checked what it is used for and if it has to be an int, so I kept it as `-1`.
If we can use arbitrary strings, then I agree that `unknown` is the better option.

hernanponcedeleon: According to [the witness format](https://github.com/sosy-lab/sv-witnesses/blob/main/README-GraphML.md): 

> The attr.type attribute of this key is int

Thus `-1` is correct

hernanponcedeleon:
LGTM

PR #449, opened 2023-05-29
Add PTX support for dartagnan

tonghaining:
This PR adds support for PTX memory model. The main updates:
1. Introducing four new base relations: `sr`(same scope), `scta`(same cta), `alias`(accessed locations alias each other), and `syncbar`(barriers with same id).
2. Updating the encoding of coherence for PTX beceuse it is not total-ordered in PTX memory model.
3. Adding `alias` releveant members to MemoryObject, and update the relevant encoding for PTX memory model.

Reference:
- [Repository of ptx-v6.0 (proxy-less)](https://github.com/NVlabs/ptxmemorymodel), the corresponding [paper](https://github.com/NVlabs/ptxmemorymodel/blob/master/PTXMemoryModelASPLOS2019.pdf).
- [Repository of ptx-v7.5 (with-proxy)](https://github.com/NVlabs/mixedproxy), the corresponding [paper](https://dl.acm.org/doi/pdf/10.1145/3470496.3533045).
- [Formal document of PTX from NVIDIA](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#memory-consistency-model)


hernanponcedeleon:
I already reviewed the code in Haining's fork, but I assume @ThomasHaas will have some further comments and suggestions

ThomasHaas:
I will certainly take a look. I did occasionally follow your discussions at @tonghaining's branch, so I already got some ideas about what is waiting for me :)

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -204,7 +204,22 @@ private BooleanFormula encodeLastCoConstraints() {
...
+                    for (MemEvent w2 : program.getEvents(Store.class)) {
+                        if (!alias.mayAlias(w1, w2)) {
+                            continue;
+                        }
+                        BooleanFormula sameAddressOther = context.sameAddress(w1, w2);
+                        BooleanFormula lastCoExprOther = lastCoVar(w2);
+                        BooleanFormula sameValueOther = context.equal(context.value(w2), v2);
+                        other = bmgr.or(other, bmgr.and(lastCoExprOther, sameAddressOther, sameValueOther));
+                    }
+                    enc.add(bmgr.implication(bmgr.and(lastCoExpr, sameAddress), bmgr.or(sameValue, other)));
```
ThomasHaas, line 222: Maybe we can make this encoding conditional on the fact whether `co` is total or not to avoid overhead `co` is total.
```
final boolean coIsTotal = (program.getArch() != Arch.PTX);
if (coIsTotal) {
  // usual encoding
} else {
  // new encoding
}
```
Apart from that, this encoding seems to be unnecessarily large. There are 3 nested loops but 2 should be sufficient (I think you create redundant constraints when `w1` and `w2` are swapped).
I think you can do roughly this:
```
for (Init init : program.getEvents(Init.class)) {
  BooleanFormula lastValueEnc = bmgr.mkFalse();
  BooleanFormula lastStoreExistsEnc = bmgr.mkFalse();
  for (Store w : program.getEvents(Store.class)) { // TODO: restrict to mayAlias only
      BooleanFormula isLast = ...;
      BooleanFormula sameAddr = ...;
      BooleanFormula sameVal = ...;
      lastValueEnc = bmgr.or(lastValueEnc, bmgr.and(isLast, sameAddr, sameVal));
      lastStoreExistsEnc = bmgr.or(bmgr.and(isLast, sameAddr));
  }
  enc.add(bmgr.ITE(lastStoreExists, lastValueEnc, context.equal(initValue, 0))); // Pseudo-code
}
```
This resembles your encoding, but leaves out the write of the outer loop (the code should therefore get moved outside of the outer loop). Also, I think you need to include `Init` events for otherwise if there is no single store to that address, the encoding will leave the final address value unconstrained rather than matching it with the init event.

ThomasHaas (commented):
It would be nice if you or @hernanponcedeleon could give a short overview about the unique aspects of the PTX model like, e.g., the meaning of scopes, contexts, virtual addresses/aliases, non-total coherences etc.
That will make it easier for me to follow the code changes (and your design decisions) and give more constructive feedback.

cat/ptx-v6.0.cat
```diff
@@ -0,0 +1,62 @@
...
+let fence-sc = [F & SC] ; sr; [F & SC]
+
+// The ASPLOS'19 paper uses the definition below
+// let rec observation = (morally-strong & rf) | (observation; rmw; observation)
+// but the ISCA'22 paper claims the one below is equivalent, thus we avoid the recursion
+let observation = (morally-strong & rf) | rmw
+
+let release-pattern = ([W & REL]; po-loc?; [strong-write]) | ([F & ACQ_REL]; po; [strong-write])
+
+let acquire-pattern = ([strong-read]; po-loc?; [R & ACQ]) | ([strong-read]; po; [F & ACQ_REL])
```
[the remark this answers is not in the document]
tonghaining: 1. Yes the release/acquire  `M`s only guarantee ordering to the same-address strong accesses. I'm not too familar with the standard semantics of release/acquire in other models, but in PTX, the order of weak accesses or different-address accesses are guaranteed through causality order (which is named `cause` in this cat).
2. ACQ_REL fences ensure communication of all address accesses inside its specified scope. @hernanponcedeleon could you give the comparasion with the typical C11 here?
3. There seems to be only `fence.acq_rel` in the PTX memory model. (here is the supported [operations](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#operation-types) in ptx)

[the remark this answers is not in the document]
ThomasHaas: I just noticed that the direction of the rel/acq relations is opposite to the usual semantics.
Usually, a release store relates with po-before memory operations, but your `release-pattern` relates with po-after memory operations. I guess I need to check out the .cat in more detail to understand why this is.

cat/ptx-v7.5.cat
```diff
@@ -0,0 +1,91 @@
...
+(*******************)
+
+(* Events *)
+let strong-write = W & (RLX | REL)
+let strong-read = R & (RLX | ACQ)
+let strong-m = strong-write | strong-read
+let strong-operation = strong-m | F
+
+(* Proxy *)
+let same-proxy = GEN * GEN | SUR * SUR | TEX * TEX | CON * CON
```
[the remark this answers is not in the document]
tonghaining: Thanks for the suggestion, comments for proxy names and scope names are added.

cat/ptx-v7.5.cat
```diff
@@ -0,0 +1,91 @@
...
+// let rec observation = (morally-strong & rf) | (observation; rmw; observation)
+// We opt to avoid the recursion
+let observation = (morally-strong & rf) | rmw
+
+let release-pattern = ([W & REL]; po-loc?; [strong-write]) | ([F & ACQ_REL]; po; [strong-write])
+
+let acquire-pattern = ([strong-read]; po-loc?; [R & ACQ]) | ([strong-read]; po; [F & ACQ_REL])
+
+let sync = morally-strong & (release-pattern; observation+; acquire-pattern)
+let cause-base = (po*; ((sync | sync_barrier | fence-sc); po*)+) | po
```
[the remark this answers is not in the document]
tonghaining: 1. Yes they are the same. The ptx-v6.0 was following the paper's style, it's updated to the same as here now for consistency.
2. The `cause` is meant to be the causality order relation, the name is following the [alloy model](https://github.com/NVlabs/mixedproxy/blob/main/alloy/ptx.als#L141) of NVIDIA's repository. 
3. Causality order captures how memory operations become visible across threads\proxies via synchronizing operations (release/acquire M and all kinds of fences or barriers). This relation is used in two axioms of the PTX memory model: in coherence axiom it restricts two writes in causality order must be in coherence order; in causality axiom it restricts the communication order(`rf`/`fr`) must be in causality order. Could you give an example how is it forming cycles?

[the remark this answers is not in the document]
ThomasHaas: Regarding point 3: I assume the same-scope relation `sr` is an actual equivalence relation, i.e., it is symmetric.
Then a code as simple as 
```
P0
e1: sc-fence;
e2: someEvent;
e3: sc-fence;
```
causes `e2` to be related to itself via `e2 --po--> e3 --fence-sc-> e1 --po--> e2`.
Basically as soon as you place two SC-fences into some thread, all events inbetween are related to each other. Similar things happen with `sync_barrier` I believe.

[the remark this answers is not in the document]
ThomasHaas: Btw. I was suggesting to change `po*` to `po?` in version 7.5, rather than the other way around.
The v6.0 paper also uses `po?`, so it matches better with the reference source.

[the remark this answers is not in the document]
hernanponcedeleon: Thomas is right, we are missing these axioms from the the alloy models
- v6.0
```
//fence
fact wf_sc {
  //all s: Scope | total[sc, scoped[s] & inside[s] & FenceSC]
  all disj f1, f2 : FenceSC | f1->f2 in strong_r => (f1->f2 in sc or f2->f1 in sc)
  sc in typed[strong_r, FenceSC]
  acyclic[sc]
}
``` 
- v7.5
```
//sc
fact {
  all s: Scope | total[sc, scoped[s] & inside[s] & FenceSC]
}
```
In the later they go even further and require `sc` to be total. This is probably "wrong", i.e., not matching the documentation.

We need to make `sc` a base relation. It would be great if we could add test cases that show the current version if buggy.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/RMWAbstract.java
```diff
@@ -16,14 +15,14 @@ public abstract class RMWAbstract extends MemEvent implements RegWriter, RegRead
     protected final Register resultRegister;
     protected ExprInterface value;
 
-    RMWAbstract(IExpr address, Register register, ExprInterface value, String mo) {
+    protected RMWAbstract(IExpr address, Register register, ExprInterface value, String mo) {
```
ThomasHaas, line 18: You shouldn't move this class to the `core` package. I would rather have it in `lang` (at top-level, not in any of its subpackages). The reason is that `RMWAbstract` is always getting compiled down to more primitive events and the `core` package should contain only those primitives.
Alternatively, you could keep the class in `lang.linux` and just create a new PTX-specific version (`PTXRMWAbstract`) in the ptx package.

We need to refactor the event package at some point to give a better place for common events like `RMWAbstract`, but for now I think the above two solutions are the best one's.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
```diff
@@ -81,6 +81,10 @@ private void createInitEvents(Program program) {
 
         int nextThreadId = threads.get(threads.size() - 1).getId() + 1;
         for(MemoryObject memObj : program.getMemory().getObjects()) {
+            // We do not create Init events for virtual addresses
+            if(memObj.isVirtual()) {
+                continue;
+            }
```
ThomasHaas, line 87: A method `memory.getPhysicalObjects()` would be nice. Together with the aforementioned `memory.allocateVirtual`, the `Memory` class provides a good interface to properly separate physical and virtual addresses.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/WmmAnalysis.java
```diff
@@ -28,7 +28,10 @@ public class WmmAnalysis {
             name= LOCALLY_CONSISTENT,
             description="Assumes local consistency for all created wmms.",
             secure=true)
-    private boolean assumeLocalConsistency = true;
+    // Set to false because local consistency does not hold in PTX
+    // When location accessed via different proxies but not properly synchronized,
+    // they can form intra-thread data races.
+    private boolean assumeLocalConsistency = false;
```
ThomasHaas, line 34: I think we should make the default case somehow aware of the source-language and then choose `true` for non-PTX programs, and `false` for PTX.
I think it is possible to check with `config.hasProperty()` whether the user specified the local consistency assumption. If so, we take whatever the chose (even if it is wrong), and if not, we default to a source-language-aware default value.
`config.hasProperty` is deprecated, but I think this is a valid use-case here.
Alternatively we just override the user-specified value with `false` for PTX (i.e., we ignore the user input in that case).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -1452,4 +1584,47 @@ private long countMaySet() {
...
+            assert (py != null);
+            return x.equals(py.getAlias()) || py.equals(px);
+        } else if (virtual1) { // virtual1 && !virtual2
+            // Virtual addresses always alias to a physical one
+            assert (px != null);
+            return y.equals(px.getAlias()) || px.equals(py);
+        } else { // !virtual1 && !virtual2
+            return x.equals(py) || y.equals(px);
+        }
+    }
```
ThomasHaas, line 1629: This code is very similar to `AliasAnalysis.virtualAlias`. 
Can this be somehow unified?
Also, the method name is bad. Checks like `alias.mayAlias(e1 ,e2) || alias(e1, e2)` look awkward and hard to understand.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/SameScope.java
```diff
@@ -0,0 +1,24 @@
...
+import com.dat3m.dartagnan.wmm.Definition;
+import com.dat3m.dartagnan.wmm.Relation;
+
+public class SameScope extends Definition {
+    private final String specificScope;
+
+    public SameScope(Relation r) {
+        super(r);
+        this.specificScope = null;
+    }
```
ThomasHaas, line 12: You can call the other constructor `this(r, null)`.

cat/ptx-v7.5.cat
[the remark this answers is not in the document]
hernanponcedeleon: I don't think you latest commit solves the problems mentioned by Thomas. You just moved the same definition to the code rather than the cat file. I see two problems:
- you need to add to the cat file the fact that `sc` is acyclic (this should make some of your tests break; if not, we need more tests)
- the visitor that computes knowledge is adding both directions to the must set, meaning that both direction are always part of the relation (if the events execute). What you need to actually do is to add them to the may set, this + the acyclicity constraint will force the solver to add only one on the directions in many cases (making `sc` acyclic as the documentation states)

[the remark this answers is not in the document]
ThomasHaas: Not only is it acyclic, it should also be total on morally strong sc fences:
```
The Fence-SC order is an acyclic partial order, determined at runtime, that relates every pair of morally strong fence.sc operations.
```
I think this requires a similar encoding to `co`, but possibly with different integer clock variables for different scopes/proxies (can `fence.sc` also be applied on different scope/proxy levels?).

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopedThread/ScopedThread.java
```diff
@@ -0,0 +1,31 @@
...
+import com.dat3m.dartagnan.program.event.core.Event;
+
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class ScopedThread extends Thread {
+
+    // There is a hierarchy of scopes, the order of keys
+    // is important, thus we use a LinkedHashMapeMap
```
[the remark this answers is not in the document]
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopedThread/ScopedThread.java
```diff
@@ -0,0 +1,31 @@
...
+        super(id, entry);
+    }
+
+    public ArrayList<String> getScopes() {
+        return new ArrayList<>(scopeIds.keySet());
+    }
+
+    public int getScopeIds(String scope) {
+        return scopeIds.getOrDefault(scope, -1);
+    }
```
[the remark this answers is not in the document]
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -41,6 +42,50 @@ static AliasAnalysis fromConfig(Program program, Configuration config) throws In
...
+    // (4) - both addresses are physical: the traditional alias analysis handles this
+
+    static boolean virtuallyAlias(MemEvent e1, MemEvent e2) {
+        // TODO: Add support for pointers, i.e. if `x` and `y` virtually alias, 
+        // then `x + offset` and `y + offset` should too
+        if (!(e1.getAddress() instanceof MemoryObject) || !(e2.getAddress() instanceof MemoryObject)) {
+            return false;
+        }
+        MemoryObject add1 = (MemoryObject) e1.getAddress();
+        MemoryObject add2 = (MemoryObject) e2.getAddress();
```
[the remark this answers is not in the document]
tonghaining: Updated.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -41,6 +42,50 @@ static AliasAnalysis fromConfig(Program program, Configuration config) throws In
...
+            assert(add1.getAlias() != null);
+            assert(add2.getAlias() != null);
+            // add1, add2 should virtually alias to the same physical Address
+            return (add1.getAlias().equals(add2.getAlias()));
+        } else if (!isAdd1Virtual && isAdd2Virtual) {
+            // Case (2)
+            // Virtual addresses always have an alias
+            assert(add2.getAlias() != null);
+            // add2 should virtually alias to physical add1
+            return add1 == add2.getAlias();
```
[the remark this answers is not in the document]
tonghaining: Updated `.equals` to `==`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -604,4 +607,29 @@ public static Fence newLwSyncBarrier() {
...
-}
+    // =============================================================================================
+    // ============================================ PTX ============================================
+    // =============================================================================================
+    public static class PTX {
+        private PTX() {}
+
+        public static AtomOp newAtomOp(IExpr address, Register register, IExpr value,
+                                             IOpBin op, String mo, String scope) {
+            AtomOp atom = new AtomOp(address, register, value, op, mo); // mo = ACQ_REL || RLX
```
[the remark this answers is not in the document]
tonghaining: There is a mo checker in the `VisitorLitmusPTX` where this is invoked, is it necessory to add checkArgument here as well?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/RMWAbstract.java
tonghaining: `RMWAbstract` moved to `lang`

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/SameScope.java
tonghaining: Updated

cat/ptx-v7.5.cat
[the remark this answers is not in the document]
ThomasHaas: I don't think your semantics of `sync_fence` is right. You won't get away with modelling the properties of `sync_fence` inside the CAT file itself, because some of those properties are not expressible.
What you have right now is that `sync_fence` can take any value as long as it is acyclic.
But that also allows it to be simply empty in all executions, which is clearly wrong.

You will need to define a proper semantics in the `WmmEncoder` class. 
In particular, `sync_fence` has to be total over morally strong fences (which I think are fences in the same scope?), and this requires you to do something similar to how `co` is encoded.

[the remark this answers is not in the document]
tonghaining: After adding a co-like encoder for sync_fence, there is a [test](https://github.com/tonghaining/Dat3M/blob/ptx-proxy/litmus/PTX/SB%2Bsc-gpu-multiFence.litmus) with multi SC getting different result on ptx-v6.0 and ptx-v7.5. Not sure if it's from incorrect encoder implementation or the model's difference?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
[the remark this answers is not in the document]
tonghaining: Thanks! Comment updated.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -604,4 +607,29 @@ public static Fence newLwSyncBarrier() {
...
+            atom.addFilters(scope);
+            return atom;
+        }
+
+        public static RedOp newRedOp(IExpr address, Register register, IExpr value,
+                                           IOpBin op, String mo, String scope) {
+            RedOp red = new RedOp(address, register, value, op, mo); // mo = ACQ_REL || RLX
```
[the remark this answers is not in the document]
tonghaining: Updated.

cat/ptx-v7.5.cat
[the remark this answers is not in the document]
tonghaining: I tried on the alloy model with removing the `po` of `cause_base` and it got same result as ptx-v6.0. So I guess the difference comes from the models.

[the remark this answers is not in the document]
ThomasHaas: So are you saying that the addition of `po` to `cause_base` in ptx-v7.5 might indeed give different semantics?
IIRC you were missing a concrete example that shows the semantics to be inconsistent with each other (according to what I got from the discussions on the NVIDIA github).
Fortunately, this should be rather easy to confirm. One of the models gives you a concrete execution which should or should not exist (not sure). You can take this exact execution and check it against the other model by hand.
It might take some effort to get the execution printed though (Dartagnan can do it, but it will only show the po, rf, and co, but not sync_fence or any derived relations).

ThomasHaas:
I want to talk about a more general issue with this PR. I think the notions of `loc` and `alias` have been unnecessarily swapped, which causes more changes (and confusion) than needed.
You see, for all models we actually operate over the assumption that `rf` and `co` are subsets of `loc`.
This holds true even with virtual aliases as long as `loc` relates events based on the actual physical address they access. This also makes sense from a naming point of view: The NVIDIA docs say that a "memory *location*" is a physical address. Indeed, I think even the Alloy model respects this (with a minor odd exception (*))

Furthermore, with this interpretation you do not need to change the encoding of `rf` and `co` directly.
A fully correct approach would rather update the definition of `context.sameAddress` to check for actually identical physical locations (i.e., it would be responsible for resolving aliases (**)).
The alias analysis would then still be responsible for approximating the `loc` relation as it does right now.
Indeed, with this approach it should suffice to simply update the alias analysis (via the wrapper I proposed) to respect virtual aliases and then everything else works out-of-the-box (***).

(*) The alloy model defines `po-loc = po  & alias` (`fun po_loc : Op->Op { same_alias[^po] }`). This is awkward, but we can add a .cat definition like `po-loc = po & alias` to redefine the usual meaning of `po-loc`.

(**) We should (eventually) rename `sameAddress` to `sameLocation` to make this even more clear. You don't have to do this in this PR, but I put this here just a reminder for me and @hernanponcedeleon.

(***) at least as long as you only have aliasing that you can statically resolve. For dynamic aliasing there is no other way than to also update the encoding produced by `context.sameAddress` to be alias-aware, but for now you can ignore that.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
tonghaining: Updated

cat/ptx-v7.5.cat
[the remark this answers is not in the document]
hernanponcedeleon: I just tried the current vresion of the code. v7.5 says the `forall` is guarantee and I think this is because the order between the stores in the same threads is guaranteed (due to `po` being part of `cause-base`). However I am not too sure v6.0 should guarantee this property (i.e. I could believe this tests show differences between the two models).

@tonghaining are you saying the following?
- in their alloy tool (for v7.5), the test passes, and
- in their alloy tool, but removing the `po` (we do this to mimic v6.0 which lacks tool support), the test also passes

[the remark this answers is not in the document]
ThomasHaas: I don't understand how `forall` could fail to hold. To have a violating execution, you need both loads to read from the earlier store (or even the initial one). This gives you `fr`-edges from the loads to the later stores (those that store `2`). Since `sync_fence` is total, it will relate the two later fences in some direction (let's say from P0 to P1).
But then you get a `po;sync_fence;po` path from `st.weak x, 2` to `ld.weak r2, x`, i.e, you get a `cause` edge.
In the opposite direction you have the aforementioned `fr` edge, and so you get a `fr;cause` cycle which should be prohibited by `(* Axiom 6 Causality *) irreflexive ((rf | fr); cause) as axiom-Causality`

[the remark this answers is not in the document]
hernanponcedeleon: The final part of your argument focus on the case where you read from the stores with value 2. What if you read from the initial ones? You would need a `cause` path from the init to the store with value 2 to forbid the execution, and I am not sure we have this.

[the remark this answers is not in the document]
ThomasHaas: No, I say that you do NOT read from the store with value 2 and hence you get a `fr`-edge towards that store.
The fences guarantee that there is always a cause-path from one of the "value 2" stores to its corresponding load.
It shouldn't matter if there is a cause-path between the different stores.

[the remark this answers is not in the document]
hernanponcedeleon: The `rf` edge would hold if not only you have the `rf` edge, but also the `co` one. I think this is one of the cases (because both stores are weak) where PTX does not guarantee you to have a `co` edge.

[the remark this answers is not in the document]
ThomasHaas: You are right. If those two stores are not `co`-related, you indeed get different behaviour. But they have to be related according to the documentation:
```
Two overlapping write operations are related in coherence order if they are morally strong or if they are related in causality order
```
Both writes are overlapping (accessing the same address) and are morally strong (again, because they access the same address and are in program order). The SC-per-loc axiom prohibits `co` to oppose `po-loc`, so you get the forward coherences that are necessary to get the `fr;cause` cycle.

I think that our encoding of `co` is wrong and fails to relate morally strong accesses, and that is why we get different results here.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -631,6 +632,67 @@ public Void visitSyncBarrier(Relation rel) {
...
+        @Override
+        public Void visitSyncFence(Relation syncFence) {
+            boolean idl = !context.useSATEncoding;
+            List<Fence> allFenceSC = program.getEvents(Fence.class).stream()
+                    .filter(e -> e.is(Tag.PTX.SC))
+                    .sorted(Comparator.comparingInt(Event::getGlobalId))
+                    .collect(toList());
+            EncodingContext.EdgeEncoder edge = context.edge(syncFence);
+            RelationAnalysis.Knowledge k = ra.getKnowledge(syncFence);
+            Set<Tuple> transSC = idl ? ra.findTransitivelyImpliedRel(syncFence) : null;
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -631,6 +632,67 @@ public Void visitSyncBarrier(Relation rel) {
...
+            IntegerFormulaManager imgr = idl ? context.getFormulaManager().getIntegerFormulaManager() : null;
+            if (idl) {
+                // ---- Encode clock conditions (fence >= 0) ----
+                NumeralFormula.IntegerFormula zero = imgr.makeNumber(0);
+                for (Fence fence : allFenceSC) {
+                    NumeralFormula.IntegerFormula clock = context.clockVariable(fence.getName(), fence);
+                    enc.add(imgr.greaterOrEquals(clock, zero));
+                }
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -631,6 +632,67 @@ public Void visitSyncBarrier(Relation rel) {
...
+                    Tuple zx = xz.getInverse();
+                    boolean forwardPossible = k.containsMay(xz);
+                    boolean backwardPossible = k.containsMay(zx);
+                    if (!forwardPossible && !backwardPossible) {
+                        continue;
+                    }
+                    BooleanFormula pairingCond = execution(xz);
+                    BooleanFormula coF = forwardPossible ? edge.encode(xz) : bmgr.makeFalse();
+                    BooleanFormula coB = backwardPossible ? edge.encode(zx) : bmgr.makeFalse();
+                    enc.add(bmgr.equivalence(pairingCond, bmgr.or(coF, coB)));
```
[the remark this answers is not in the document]
tonghaining: Moved the scope check to the encoding part.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -631,6 +632,67 @@ public Void visitSyncBarrier(Relation rel) {
...
+                    if (idl) {
+                        enc.add(bmgr.implication(coF, transSC.contains(xz) ? bmgr.makeTrue() :
+                                imgr.lessThan(context.clockVariable(x.getName(), x), context.clockVariable(z.getName(), z))));
+                        enc.add(bmgr.implication(coB,transSC.contains(zx) ? bmgr.makeTrue() :
+                                imgr.lessThan(context.clockVariable(z.getName(), z), context.clockVariable(x.getName(), x))));
```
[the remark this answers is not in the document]
tonghaining: Updated

cat/ptx-v7.5.cat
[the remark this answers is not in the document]
hernanponcedeleon: Now I am confused. We are arguing that according to our reasoning, the `forall` should hold in both versions 6 and 7.5.

Dartagnan says it does not hold for v6, but we already identified what the problem is.

Are you saying their tool is also giving a wrong result? Have you added the missing axiom on top of removing `po` from `cause-base`?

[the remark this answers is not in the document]
tonghaining: Okay it was from the missing `sc_per_location` axiom. I got the result from their main branch version (without the axiom). With adding it, both `v6` and `v7.5` will pass as tested.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
```diff
@@ -30,6 +30,9 @@ public static EqualityAliasAnalysis fromConfig(Program program, Configuration co
 
     @Override
     public boolean mustAlias(MemEvent a, MemEvent b) {
+        if (AliasAnalysis.virtuallyAlias(a, b)) {
+            return true;
+        }
```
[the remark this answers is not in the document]
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -34,10 +45,13 @@ public class MemoryObject extends IConst {
 
     private final HashMap<Integer, IConst> initialValues = new HashMap<>();
 
-    MemoryObject(int index, int size, boolean isStaticallyAllocated) {
+    MemoryObject(int index, int size, boolean isStaticallyAllocated, boolean virtual, MemoryObject alias) {
+        checkArgument(!virtual || alias != null, "Virtual MemoryObject must have alias target");
```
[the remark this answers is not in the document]
tonghaining: This `alias` was used to discriminate `generic` and `non-generic` address. This member variable has been removed in the new commit.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
```diff
@@ -16,9 +16,9 @@ public class Memory {
      * @return
      * Points to the created location.
      */
-    public MemoryObject allocate(int size, boolean isStatic) {
+    public MemoryObject allocate(int size, boolean isStatic, boolean virtual, MemoryObject alias) {
         Preconditions.checkArgument(size > 0, "Illegal malloc. Size must be positive");
-        MemoryObject address = new MemoryObject(nextIndex++, size, isStatic);
+        MemoryObject address = new MemoryObject(nextIndex++, size, isStatic, virtual, alias);
```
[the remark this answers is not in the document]
tonghaining: The GPU-relevant parameters has been removed out of MemoryObject

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
tonghaining: This special case has been removed in the new commit

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
tonghaining: This function has been rewritten as `samePhysicalAddress`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
ThomasHaas: Your changes look good, but you still miss the case where there exists no store to the same address as the init event. In that case, the initial value never gets overwritten so you should have `lastValue = initialValue`.
Your current encoding leaves the `lastValue` unconstrained, so it can take any arbitrary value.

cat/ptx-v7.5.cat
```diff
@@ -0,0 +1,105 @@
...
+(******************************)
+
+(* Axiom 1 Coherence *)
+let cause-write = ([W]; cause; [W]) & loc
+empty (cause-write \ co) as axiom-Coherence
+
+(* Axiom 2 FenceSC *)
+// This is equivalent to: irreflexive (sync_fence ; cause) as axiom-FenceSC
+empty (([F & SC]; cause; [F & SC]) \ sync_fence) as axiom-FenceSC
+acyclic sync_fence as axiom-FenceSC-Acyclic
```
ThomasHaas, line 90: We don't need this anymore since the encoding guarantees acyclicity already.

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Arch.java
```diff
@@ -33,9 +35,21 @@ public static Arch getDefault() {
...
+    public static boolean coIsTotal(Arch arch) {
+        Arch[] coNotTotal = {PTX};
+        return !Arrays.asList(coNotTotal).contains(arch);
+    }
+
+    // used to check if supports virtual addressing.
+    public static boolean supVirAdd(Arch arch) {
+        Arch[] supportVirtualAddress = {PTX};
+        return Arrays.asList(supportVirtualAddress).contains(arch);
+    }
```
ThomasHaas, line 54: Please use the full name `supportsVirtualAddressing`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -204,8 +205,32 @@ private BooleanFormula encodeLastCoConstraints() {
                     BooleanFormula sameAddress = context.sameAddress(init, w1);
                     Formula v2 = ExpressionEncoder.getLastMemValueExpr(init.getBase(), init.getOffset(), fmgr);
                     BooleanFormula sameValue = context.equal(context.value(w1), v2);
-                    enc.add(bmgr.implication(bmgr.and(lastCoExpr, sameAddress), sameValue));
+                    if (Arch.coIsTotal(program.getArch())) {
+                        enc.add(bmgr.implication(bmgr.and(lastCoExpr, sameAddress), sameValue));
+                    }
```
ThomasHaas, line 210: You can move the check upwards into the first conditional: `if (doEncodeFinalAddrssValues && Arch.coIsTotal(program.getArch()))`.
This is because all the code inbetween the two conditionals does nothing if `Arch.coIsTotal(program.getArch())` fails to hold.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -204,8 +205,32 @@ private BooleanFormula encodeLastCoConstraints() {
...
+                    if (!alias.mayAlias(w, init)) {
+                        continue;
+                    }
+                    BooleanFormula isLast = lastCoVar(w);
+                    BooleanFormula sameAddr = context.sameAddress(init, w);
+                    BooleanFormula sameValue = context.equal(context.value(w), v2);
+                    lastValueEnc = bmgr.or(lastValueEnc, bmgr.and(isLast, sameAddr, sameValue));
+                    lastStoreExistsEnc = bmgr.or(lastStoreExistsEnc, bmgr.and(isLast, sameAddr));
                 }
+                enc.add(bmgr.implication(lastStoreExistsEnc, lastValueEnc));
```
ThomasHaas, line 233: As I commented before, you need to make sure that if no store gets executed, then the last value matches the initial value!

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -574,7 +578,10 @@ public Void visitCoherence(Relation co) {
                     BooleanFormula pairingCond = bmgr.and(execPair, sameAddress);
                     BooleanFormula coF = forwardPossible ? edge.encode(xz) : bmgr.makeFalse();
                     BooleanFormula coB = backwardPossible ? edge.encode(zx) : bmgr.makeFalse();
-                    enc.add(bmgr.equivalence(pairingCond, bmgr.or(coF, coB)));
+                    // Coherence is not total for weak stores in PTX
+                    if(!x.is(Tag.PTX.WEAK) || !z.is(Tag.PTX.WEAK)) {
+                        enc.add(bmgr.equivalence(pairingCond, bmgr.or(coF, coB)));
+                    }
```
ThomasHaas, line 584: We also need to make sure that we have coherences between `morally-strong` stores.
Since `morally-strong` is a derived relation, we have to somehow formulate the coherence condition inside the cat file.
Maybe something like this works:
```
// Make sure that all morally strong same-loc writes are related by co in either direction
empty ([W];morally-strong;[W] & loc) \ (co | co^-1)
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopedThread/ScopedThread.java
```diff
@@ -0,0 +1,54 @@
...
+        // i = 0 is global, every thread will always have the same id, so start from i = 1
+        for (int i = 1; i <= validIndex; i++) {
+            if (!sameAtSingleSameScope(thread, scopes.get(i))) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean sameAtSingleSameScope(ScopedThread thread, String scope) {
```
ThomasHaas, line 49: Should it be `sameAtSingleScope` (it sounds odd to have "same" twice in the name)?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/VirtualAliasAnalysis.java
```diff
@@ -0,0 +1,39 @@
...
+        this.wrappedAnalysis = analysis;
+    }
+
+    public static AliasAnalysis wrap(AliasAnalysis analysis) {
+        return new VirtualAliasAnalysis(analysis);
+    }
+
+    @Override
+    public boolean mayAlias(MemEvent e1, MemEvent e2) {
+        return sameGenericAdd(e1, e2) || wrappedAnalysis.mayAlias(e1, e2); // + version for mustAlias
```
ThomasHaas, line 19: You can remove the comment :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/VirtualAliasAnalysis.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+    @Override
+    public boolean mustAlias(MemEvent e1, MemEvent e2) {
+        return sameGenericAdd(e1, e2) || wrappedAnalysis.mustAlias(e1, e2);
+    }
+
+    // GPU memory models make use of virtual addresses.
+    // This models same_alias_r from the PTX Alloy model
+    // Checking address1 and address2 hold the same generic address
+    public boolean sameGenericAdd(MemEvent e1, MemEvent e2) {
```
ThomasHaas, line 29: Use the full name `sameGenericAddress`

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -854,6 +890,83 @@ public Knowledge visitSameAddress(Relation rel) {
...
+            Set<Tuple> must = new HashSet<>();
+            List<MemEvent> events = program.getEvents(MemEvent.class);
+            for (MemEvent e1 : events) {
+                for (MemEvent e2 : events) {
+                    if (samePhysicalAddress(e1, e2) && !exec.areMutuallyExclusive(e1, e2)) {
+                        must.add(new Tuple(e1, e2));
+                    }
+                }
+            }
+            return new Knowledge(must, new HashSet<>(must));
```
ThomasHaas, line 967: I would change this definition with the definition of `loc`. I think `loc` should be based on the physical address while `alias` is based on the virtual one.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
tonghaining: readFromInit added.

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Arch.java
tonghaining: Changed

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopedThread/ScopedThread.java
tonghaining: Updated (Yes it was an absurd name..)

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/VirtualAliasAnalysis.java
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
tonghaining: Updated to `visitLoc`, but I'm afraid is it a bit confusing for Alias.java class to visitLoc?

ThomasHaas: I'm not saying that you should just rename the function (that is indeed confusing).
I'm talking more generally that the semantics of `loc` and `alias` should be swapped around.
This also implies that you need to change the cat file and swap all occurences of `loc` and `alias`.
I've written a post about why this is better (just scroll up). But in short: this aligns better with the typical notion of `loc` in other models, it allows you to preserve the invariants `rf \subseteq loc` and `co \subseteq loc`, and it better aligns with the Alloy model!

tonghaining: The `same_physical_address` has been moved into the `VirtualAliasAnalysis` (meanwhile removed from `co`, `rf` relations). The usage of `loc` and `alias`(now named `vloc`) in cat are swapped.  The `loc` relation now signifies the same physical address, whereas the `vloc` relation represents the same virtual address, which is stricter than `loc`.

ThomasHaas: Perfect, that's what I had in mind.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
hernanponcedeleon: The current implementation forces to have one of the co-orders if one store is weak, but the other is not. I don't think this is correct since they might still not be morally strong.

Using an && in line 582 and adding the axiom that Thomas proposes should solve the problem AFAIS.

tonghaining: Is it possible that utilizing `&&` might still include morally-weak writes, like two `st.rlx.cta` from different ctas?

ThomasHaas: Do you even need the check at all anymore? With the new Axiom in cat, you force a coherence between all morally strong writes. Axiom 1
```
(* Axiom 1 Coherence *)
let cause-write = ([W]; cause; [W]) & loc
empty (cause-write \ co) as axiom-Coherence
```
forces coherences between causally related writes as well. This should cover all cases, so that your encoding doesn't need to enforce any coherence at all, does it? I can imagine that transitivity is not yet given, so you might need to add `let co = co+` in your cat.

hernanponcedeleon: We definitely need line 583 for the other models. The axiom you propose + axiom 1 guarantee that anything that is causal or morally strong should be part of `co`. But we still need to tell the solver that "if the are not morally strong and they are not causal, then you are free not to add any `co` edge". Or am I missing something?

ThomasHaas: > We definitely need line 583 for the other models. 

Of course, we would skip encoding the equivalence only for PTX.

> The axiom you propose + axiom 1 guarantee that anything that is causal or morally strong should be part of `co`. But we still need to tell the solver that "if the are not morally strong and they are not causal, then you are free not to add any `co` edge". Or am I missing something?

The solver is free to not choose a coherence, because he is unconstrained (we do not encode totality anymore).
However, the solver could also choose some coherence, which is wrong according to the documentation: 
"Two overlapping writes are unrelated in coherence order if they are in a data-race".
We would further need to add something like:
```
let data-race = (M x (W \ IW) | (W \ IW) x M) \ morally-strong \ cause
empty data-race & co
```

More generally, the only properties of `co` that we can enforce in the anarchic semantics are that `co` relates only same-address writes, that it is transitive, and that it is acyclic. All other properties are related to derived relations and hence need to be defined in the cat. We could, in principle, formulate the former properties also in the .cat so that the anarchic encoding is basically empty:
```
acyclic co // acyclicity
empty co \ (WxW & loc) // co only relates same-address writes
let co = co+ // transitivity
```

tonghaining: The cause-write and morally-strong-write seems cover all cases. But isn't the `co`  in dartagnan transitive by default? Or we need to add it because PTX is kept out from this constraints.

ThomasHaas: I'm not sure if you get transitivity for free. Suppose you had only weak stores (a)-(c) related as follows `a -cause-> b -poloc-> c`, then causality gives you `a -co-> b` and morally strong gives you `b -co-> c`. The pair `(a, c)` is neither morally strong nor causally related, yet you should have `a -co-> c` by transitivity.

The encoding of `co` is only transitive in the non-PTX models because of totality. This is because if you have `a -co-> b` and `b -co-> c`, the solver must (by totality) choose a coherence between `a` and `c`, however `c -co-> a` would lead to cycle, so the solver has to choose `a -co-> c`. Without totality, the solver could simple choose neither of the two options!
You could do this in the .cat though:
```
empty ([W];(morally-strong | cause)[W] & loc)+ \ (co | co^-1);
```
I think this will make sure that `co` also relates stores over transitive chains over `cause` and `morally-strong` relations. Interestingly, it will not force `co` to be transitive, so it might be kinda wrong (it could still be non-transitive on racy stores).

cat/ptx-v6.0.cat
```diff
@@ -62,7 +64,7 @@ let dep = addr | data | ctrl
 acyclic (rf | dep) as axiom-NoThinAir
 
 (* Axiom 5 SC-per-Location *)
-acyclic ((morally-strong & (rf | co | fr)) | po-loc) as axiom-ScPerLocation
+acyclic ((morally-strong & (rf | co | fr)); po-loc) as axiom-ScPerLocation
```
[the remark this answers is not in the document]
tonghaining: I forgot to change it back after testing, updated now

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/WmmAnalysis.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
ThomasHaas: This won't cover the case I mentioned above where you get a transitive edge by the interaction between causally related stores and morally-strong stores.
The easiest way should be to just define `let co = co+` in the beginning of the cat and then just do the two axioms without transitive closures, i.e.
```
empty (([W]; cause; [W]) & loc) \ co as axiom-Coherence
empty (([W]; morally-strong; [W]) & loc)+ \ (co | co^-1) as axiom-Coherence2
```
I think `id` is not necessary in either case, because in the former you would have a store causally related to itself (which is surely not allowed) and in the latter I think `morally-strong` should already make sure to never relate an event to itself (at least the documentation suggests this).

ThomasHaas: Yeah, I forgot to remove the transitive closure there. It was meant to be
```
empty (([W]; morally-strong; [W]) & loc) \ (co | co^-1) as axiom-Coherence2

```

ThomasHaas:
I think the PR starts to be in a good shape.
The only concern I have on the semantic level is that we still allow for coherence edges between racy stores.
The documentation cleary states that this should not be allowed.
We do not have an example where this breaks though.

The other thing is: Do we want to keep the `SC-per-loc` axiom in `ptx-v7.5` or remove it?

hernanponcedeleon:
> I think the PR starts to be in a good shape. The only concern I have on the semantic level is that we still allow for coherence edges between racy stores. The documentation cleary states that this should not be allowed. We do not have an example where this breaks though.

Isn't this test checking exactly that?
```
PTX Co-Total-3-threads-scope-same
"cyclic weak rf when not sync"
{
x=0; f1=0; f2=0;
P2:r0=0; P2:r1=0; P2:r2=0;
P3:r0=0; P3:r1=0; P3:r2=0;
}
 P0@cta 0,gpu 0         | P1@cta 0,gpu 0         | P2@cta 0,gpu 0         ;
 st.weak x, 1           | st.weak x, 2           | ld.acquire.sys r0, f1  ;
 st.release.sys f1, 1   | st.release.sys f2, 1   | ld.acquire.sys r1, f2  ;
                        |                        | ld.weak r2, x   ;
                        |                        | ld.weak r3, x   ;
exists
(P2:r0 = 1 /\ P2:r1 = 1 /\ P2:r2 = 1 /\ P2:r3 = 2)
```

> The other thing is: Do we want to keep the `SC-per-loc` axiom in `ptx-v7.5` or remove it?

I think we should remove it to be in sync with the alloy model. The NVIDIA guys claims it is not necessary, and by removing it, we would be continuosly testing this claim every time we add a new litmus test.

ThomasHaas:
I don't think your example shows this. We allow for no coherence between racy accesses, so the outcome is fine.
The question is this: Is there an example where adding a `co`-edge between racy accesses enables additional behavior?
I guess this can never be the case, because the model is monotonic in `co` (minus the few axioms that define `co` itself).
That means the inaccuracy in our modelling is unobservable, so we can keep it like that.

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/WmmFromArchitectureProvider.java
```diff
@@ -32,6 +32,7 @@ public class WmmFromArchitectureProvider extends AbstractProvider<Wmm> {
         ARCH_WMM_MAP.put(Arch.RISCV, () -> new ParserCat().parse(new File(CAT_RESOURCE_PATH + "cat/riscv.cat")));
         ARCH_WMM_MAP.put(Arch.LKMM, () -> new ParserCat().parse(new File(CAT_RESOURCE_PATH + "cat/linux-kernel.cat")));
         ARCH_WMM_MAP.put(Arch.IMM, () -> new ParserCat().parse(new File(CAT_RESOURCE_PATH + "cat/imm.cat")));
+        ARCH_WMM_MAP.put(Arch.PTX, () -> new ParserCat().parse(new File(CAT_RESOURCE_PATH + "cat/ptx.cat")));
```
hernanponcedeleon, line 35: This is not needed and should be removed. This is used to generally avoid having to implement the `getWmmProvider` method, but since we have two different cat files having the same target `PTX`, we have to implement it anyway.

tonghaining: Removed

cat/ptx-v6.0.cat
```diff
@@ -0,0 +1,72 @@
...
+let strong-read = R & (RLX | ACQ)
+let strong-m = strong-write | strong-read
+let strong-operation = strong-m | F
+
+(* Relations *)
+let sr-fence = ([strong-operation]; sr; [F]) | ([F]; sr; [strong-operation])
+let po-fence = ([strong-operation]; po; [F]) | ([F]; po; [strong-operation])
+let sr-m = ([strong-m]; (sr & loc \ id); [strong-m])
+let po-m = ([M]; ((po | po^-1) & loc); [M])
+let morally-strong = sr-fence | po-fence | sr-m | po-m
```
ThomasHaas, line 33: `po`-related events are always in the same scope, right?
Then `po-fence \subseteq sr-fence`, so the `po-fence` relation is pointless.

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Arch.java
```diff
@@ -33,9 +35,29 @@ public static Arch getDefault() {
...
+
+    // Set to false for architectures don't hold local consistency e.g. PTX
+    // When location accessed via different proxies but not properly synchronized,
+    // they can form intra-thread data races.
+    public static boolean archLocallyConsistent(Arch arch) {
+        Arch[] supportVirtualAddress = {PTX};
+        return !Arrays.asList(supportVirtualAddress).contains(arch);
```
ThomasHaas, line 61: I think this method can be removed because `WmmAnalysis` should do this check for us.
Technically speaking, the architecture itself is insufficient to decide if local consistency holds.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -120,6 +121,10 @@ public BooleanFormula encodeBooleanExpressionAt(ExprInterface expression, Event
     }
 
     public Formula encodeIntegerExpressionAt(ExprInterface expression, Event event) {
+        if (expression instanceof VirtualMemoryObject) {
+            return new ExpressionEncoder(formulaManager, event).encodeAsInteger(
+                    ((VirtualMemoryObject) expression).getPhysicalAddress());
+        }
```
ThomasHaas, line 127: I think you should move this code into the `ExpressionEncoder`.
It has a `visit(MemoryObject)` method that could do the check and extract the physical address if necessary.

cat/ptx-v6.0.cat
hernanponcedeleon: While events in the same thread can have different scope tags, my understanding of `sr` is that it tells you which events can eventually synchronize. Events in the same thread should always be able to synchronize, thus it should be the case that `po \subseteq sr`. If this is true, then indeed the relation can be removed (I tried it locally and all tests pass).

@tonghaining can you please confirm if `po \subseteq sr`?

tonghaining: The `sr` relation tells whether two events specifie the same scope (and the same scope index). I think for example in this thread P0:
`P0@cta 0,gpu 0; 
 st.release.cta x, 1 ; 
 fence.acq_rel.gpu ;`
 `W` and `F` are `po` but not `sr`. Removing this line will not include them in `morally-strong` relation. I think we don't have a test to show this in our test set.

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Arch.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
tonghaining: This code seems not needed for the new physical memory implementation, think I'll just remove it.

cat/ptx-v6.0.cat
hernanponcedeleon: I would go with this definition that follows closely the English documentation
```
// The operations are related in program order, or each operation is strong and 
// specifies a scope that includes the thread executing the other operation.
let ms1 = po | ([strong-operation]; sr; [strong-operation])
// Both operations are performed via the same proxy.
let ms2 = same-proxy
// If both are memory operations, then they overlap completely
let ms3 = ((M * M) & vloc) | ((_ * _) \ (M * M))
let morally-strong = (ms1 & ms2 & ms3) \ id
```

Something similar would need to be done for `v6.0`

tonghaining: Updated

hernanponcedeleon:
LGTM

ThomasHaas (approved):
LGTM

PR #450, opened 2023-06-01
Register reads (rework)

ThomasHaas:
This PR introduces the `Register.Read` class which allows events to specify *why* they read from a register (via a `UsageType`).
As a consequence, the `RegReaderData` has been updated to a more general `RegReader` interface that returns a set of this new `Register.Read` objects.
`MemEvent` is now also a `RegReader` as it reads from registers to determine its address (it has `UsageType=ADDR`).
All the user code of the old `RegReaderData` interface were updated accordingly.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/lisa/RMW.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/lisa/RMW.java
@@ -4,15 +4,15 @@ import com.dat3m.dartagnan.expression.ExprInterface;
 import com.dat3m.dartagnan.expression.IExpr;
 import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.event.core.MemEvent;
-import com.dat3m.dartagnan.program.event.core.utils.RegReaderData;
 import com.dat3m.dartagnan.program.event.core.utils.RegWriter;
 import com.dat3m.dartagnan.program.event.visitors.EventVisitor;
-import com.google.common.collect.ImmutableSet;
+
+import java.util.Set;
 
 import static com.dat3m.dartagnan.program.event.Tag.RMW;
 import static com.dat3m.dartagnan.program.event.Tag.*;
 
-public class RMW extends MemEvent implements RegWriter, RegReaderData {
+public class RMW extends MemEvent implements RegWriter {
 
     private final Register resultRegister;
     private final IExpr value;
@@ -40,10 +40,10 @@ public class RMW extends MemEvent implements RegWriter, RegReaderData {
         return value;
     }
 
-	@Override
-	public ImmutableSet<Register> getDataRegs() {
-		return value.getRegs();
-	}
+    @Override
+    public Set<Register.Read> getRegisterReads() {
+        return Register.collectRegisterReads(value, Register.UsageType.DATA, super.getRegisterReads());
+    }
 
 	@Override
 	public Register getResultRegister() {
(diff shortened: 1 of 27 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/lisa/RMW.java
```diff
@@ -40,10 +40,10 @@ public ExprInterface getMemValue(){
         return value;
     }
 
-	@Override
-	public ImmutableSet<Register> getDataRegs() {
-		return value.getRegs();
-	}
+    @Override
+    public Set<Register.Read> getRegisterReads() {
+        return Register.collectRegisterReads(value, Register.UsageType.DATA, super.getRegisterReads());
```
hernanponcedeleon, line 45: The address dependencies are collected in the last parameter by calling `super.getRegisterReads()`, right?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/Xchg.java
```diff
@@ -32,8 +33,8 @@ public Register getResultRegister(){
     }
 
     @Override
-    public ImmutableSet<Register> getDataRegs(){
-        return ImmutableSet.of(resultRegister);
+    public Set<Register.Read> getRegisterReads(){
+        return Set.of(new Register.Read(resultRegister, Register.UsageType.DATA));
```
hernanponcedeleon, line 36: why data and not address? we are simply exchanging the values in two addresses, aren't we?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -391,69 +388,75 @@ private void trackDependencies(Event e) {
...
-                deps.add(eventMap.get(tracked));
+        if (e instanceof RegReader regReader) {
+            final Set<EventData> dataDeps = new HashSet<>();
+            final Set<EventData> addrDeps = new HashSet<>();
+            final Set<EventData> ctrlDeps = new HashSet<>();
+            for (Register.Read regRead : regReader.getRegisterReads()) {
+                final Register reg = regRead.register();
+                final Set<EventData> visibleRootDependencies = lastRegWrites.get(reg);
+                if (visibleRootDependencies == null) {
+                    // FIXME: This should never happen, but our parser is buggy and produces ill-formed code.
```
hernanponcedeleon, line 405: This is not the first time we have missing dependencies. We should issue a warning to be sure we don't have other instances (besides the `pthread_join`), and once we get rid of that parser, use an assertion

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/lisa/RMW.java
ThomasHaas: Yes. I did it like this so can you either create a new set with the `RegisterReads` or collect them into an existing set.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/Xchg.java
ThomasHaas: From what I understood is that this event exchanges the data between an address and a register (hence the reg is a data dependency).
But you are right that the address dependencies are not getting collected.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
ThomasHaas: Sure, I will add a `logger.warn`.

hernanponcedeleon:
LGTM. Will merge once the CI passes

PR #451, opened 2023-06-01
Fix order of parameters of makeBitvector in ExecutionStatus.encodeExec()

hernanponcedeleon:
The current order of parameters for `makeBitvector` is `(value, length)` instead of the intended `(length, value)`


ThomasHaas:
That's interesting. The code of `ExecutionStatus` was already buggy for always using BVs (I mentioned this in #442) and I didn't understand why that never caused exceptions before. The reason is that we had yet another bug there that hid the first one....
Since you fixed one, you gotta fix the other one as well :P.

ThomasHaas (approved):
LGTM

PR #452, opened 2023-06-02
Event as interface (rework)

ThomasHaas:
This PR just renames the abstract `Event` class to `AbstractEvent` and introduces a new `Event` interface instead.
`RegReader` and `RegWriter` now extend the `Event` interface (this allows us to simplify some code, but I didn't do that yet).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AbstractEvent.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AbstractEvent.java
@@ -1,6 +1,5 @@
 package com.dat3m.dartagnan.program.event.core;
 
-import com.dat3m.dartagnan.encoding.Encoder;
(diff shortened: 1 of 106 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AbstractEvent.java
```diff
@@ -0,0 +1,242 @@
...
+
+	protected AbstractEvent(){
+		tags = new HashSet<>();
+	}
+
+	protected AbstractEvent(AbstractEvent other){
+		copyAllMetadataFrom(other);
+        this.tags = other.tags; // TODO: Dangerous code! A Copy-on-Write Set should be used (e.g. PersistentSet/Map)
+        this.thread = other.thread;
+    }
```
hernanponcedeleon, line 36: formatting

PR #453, opened 2023-06-02
Program.getConstants() (rework)

xeren:
Moves the encoding of bounds of `INonDet` from `Local` into `ProgramEncoder`.  This means that `Program` is now responsible of managing those instances.


dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
```diff
@@ -9,22 +9,21 @@
...
-    private final BigInteger min;
-    private final BigInteger max;
+    private BigInteger min;
+    private BigInteger max;
+    private String sourceName;
 
-    // Should only be accessed from Program
-    public INonDet(int id, IntegerType type, boolean signed, BigInteger min, BigInteger max) {
+    // Should only be accessed from ProgramBuilder
+    public INonDet(int id, IntegerType type, boolean signed) {
```
ThomasHaas, line 19: Is this comment reasonable? What if some passes generate non-deterministic choices like, e.g., the compilation of Linux CAS?
I think it is appropriate to have the `Program` be able to generate new constants and the `ProgramBuilder` just invokes that functionality.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -63,13 +64,38 @@ public static ProgramEncoder withContext(EncodingContext context) throws Invalid
...
         return context.getBooleanFormulaManager().and(
+                encodeNondeterministicBounds(),
                 encodeMemory(),
                 encodeControlFlow(),
                 encodeFinalRegisterValues(),
                 encodeFilter(),
                 encodeDependencies());
     }
 
+    public BooleanFormula encodeNondeterministicBounds() {
```
ThomasHaas, line 75: Wouldn't it be more appropriate to name it `encodeProgramConstants`, not matter if they are non-deterministic or not?

xeren: You could argue that, since the method is public and thus may abstract from what its implementation actually does.  On the other hand, if encoding constants would include something else or different, i.e. sets of distinct constants, a new `encodeProgramConstants` could also be appropriate.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
xeren: I agree, it should be going to read `Program or ProgramBuilder`, but currently, this is correct as it is.  I plan to tinker with the life time of `Program` as in #442 in a future PR.  We should avoid two `INonDet` managers.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: Well yes, but so far you use `INonDet` interchangably with the notion of constants.
If you had classes like `NonDet` and `Constant` then I would totally agree that you can have two different methods.
Right now, the situation is that "constants" get created, but "nondeterministic values" get encoded, i.e., there is a sort of implicit conversion from constants to nondets going on.
I think this is unnecessarily confusing (for now).

dartagnan/src/main/java/com/dat3m/dartagnan/expression/INonDet.java
ThomasHaas: We should definitely avoid two `INonDet` managers, but then why not have it only in the `Program`? All functionality on `ProgramBuilder` "dies" (becomes inaccessible) as soon as the initial program is constructed. So by default I would rather put such functionality into the `Program` if there is reason to access it later.
Linux CAS compilation does exactly that (well, it uses `BNonDet`, but still).

Anyway, the code can stay as it is, but I would have just changed/removed the comment (you don't need to, it's just a minor concern).

ThomasHaas (approved):
LGTM

PR #455, opened 2023-06-05
MemoryEvent as interface

ThomasHaas:
This PR renames `MemEvent` to `AbstractMemoryEvent` and introduces an interface `MemoryEvent`.
The method `MemEvent.canRace()` was removed (it had two very specific use-cases, so I inlined the method there).


hernanponcedeleon:
LGTM

PR #456, opened 2023-06-06
Add MemoryAccess and MemoryCoreEvents

ThomasHaas:
This PR generalizes the `MemoryEvent` interface and introduces a `MemoryCoreEvent` interface for memory events we support at the core (i.e., in analyses and encoding).
There is a new `MemoryAccess` class that describes a single memory access. A `MemoryEvent` can have multiple accesses, but all core events perform a single access.
Core events do not have a memory order anymore (only in the form of tags), but can have a `MemoryOrder` metadata attached. The metadata is used just for printing/debugging but serves no semantical purpose (\*).
The `EventFactory` now has `XYZwithMo` methods that create a core event, add a tag (based on the `mo`) and attach the `MemoryOrder` metadata.
As a result, most `CompilationVisitors` were updated to call the new methods if they provided a `mo`.
While adding these changes, I took the liberty to reformat the `CompilationVisitors` (replacing tabs by spaces etc.) and also fixed some compiler mappings that produced memory orderings even for architectures that don't have them (e.g., Power and TSO).
Lastly, there is now a `event/common` package which contains abstract classes that provide typical implementations (`StoreBase`, `LoadBase` and `SingleAccessMemoryEvent`). Most language-level events that previously inherited from core events (e.g. `Store` or `Load`) now instead inherit from these new classes instead (they are no core events after all!).
I added a new annotation `NoInterface` which we can attach to abstract classes to indicate that those are meant for sharing implementation but NOT for providing a common interface!
There are still some language-level events that inherit from the core events to provide custom printing methods, but this will be fixed in a different PR.

(\*) This is not quite true yet as we check the metadata in the SVCOMP data race detection to understand which accesses can be racing.
A better way would be to assign proper `NONATOMIC` tags beforehand, but this goes beyond this PR.

*--------------------- Outdated -----------------------*

_This PR adds the `MemoryAccess` class which describes single memory accesses.
A `MemoryEvent` is an event that performs memory accesses(\*).
The `AbstractMemoryEvent` was renamed to `SingleAddressMemoryEvent` to clearly indicate that this base class represents events accessing a single address (Load, Store, RMWs, and abstract events like SRCU).
This class is NOT intended to be an interface (i.e., our code should NEVER check if an event is a `SingleAddressMemoryEvent`), but just serves to share common implementation._

_There are some more TODOs remaining, but I haven't decided yet which of those to tackle within this PR (cause it might get to complex)._

_(\*) In the current state of the PR, `MemoryEvent` performs just a single `MemoryAccess`. I'm still wondering about whether we should have something like a `SimpleMemoryEvent` that performs exactly one memory access and if we then require that after all processing only simple memory events remain. This would be sufficient for what we do right now.
However, if we ever want to model AMOs with single events, then those events would not be "simple".
Apart from AMOs, I don't see a reason to have complex memory events after processing._

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
@@ -20,6 +20,7 @@ import com.dat3m.dartagnan.program.event.lang.llvm.*;
 import com.dat3m.dartagnan.program.event.lang.pthread.*;
 import com.dat3m.dartagnan.program.event.lang.std.Malloc;
 import com.dat3m.dartagnan.program.event.lang.svcomp.*;
+import com.dat3m.dartagnan.program.event.metadata.MemoryOrder;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
 
 import java.util.*;
@@ -70,11 +71,21 @@ public class EventFactory {
     // ------------------------------------------ Memory events ------------------------------------------
 
     public static Load newLoad(Register register, IExpr address, String mo) {
-        return new Load(register, address, mo);
+        Load load = new Load(register, address);
+        if (mo != null && !mo.isEmpty()) {
+            load.setMetadata(new MemoryOrder(mo));
+            load.addTags(mo);
+        }
+        return load;
     }
 
     public static Store newStore(IExpr address, ExprInterface value, String mo) {
-        return new Store(address, value, mo);
+        Store store = new Store(address, value);
+        if (mo != null && !mo.isEmpty()) {
+            store.setMetadata(new MemoryOrder(mo));
+            store.addTags(mo);
+        }
+        return store;
     }
 
     public static Fence newFence(String name) {
@@ -159,17 +170,27 @@ public class EventFactory {
     }
 
     public static RMWStore newRMWStore(Load loadEvent, IExpr address, ExprInterface value, String mo) {
-        return new RMWStore(loadEvent, address, value, mo);
+        RMWStore store =  new RMWStore(loadEvent, address, value);
+        if (mo != null && !mo.isEmpty()) {
+            store.setMetadata(new MemoryOrder(mo));
+            store.addTags(mo);
+        }
+        return store;
     }
 
     public static Load newRMWLoadExclusive(Register reg, IExpr address, String mo) {
-        Load load = new Load(reg, address, mo);
+        Load load = newLoad(reg, address, mo);
         load.addTags(Tag.RMW, Tag.EXCL);
         return load;
     }
 
     public static RMWStoreExclusive newRMWStoreExclusive(IExpr address, ExprInterface value, String mo, boolean isStrong) {
-        return new RMWStoreExclusive(address, value, mo, isStrong, false);
+        RMWStoreExclusive store = new  RMWStoreExclusive(address, value, isStrong, false);
+        if (mo != null && !mo.isEmpty()) {
+            store.setMetadata(new MemoryOrder(mo));
+            store.addTags(mo);
+        }
+        return store;
     }
 
     public static RMWStoreExclusive newRMWStoreExclusive(IExpr address, ExprInterface value, String mo) {
@@ -516,8 +537,12 @@ public class EventFactory {
         }
 
         public static RMWStoreExclusive newRMWStoreConditional(IExpr address, ExprInterface value, String mo, boolean isStrong) {
-            RMWStoreExclusive store = new RMWStoreExclusive(address, value, mo, isStrong, true);
+            RMWStoreExclusive store = new RMWStoreExclusive(address, value, isStrong, true);
             store.addTags(Tag.RISCV.STCOND);
+            if (mo != null && !mo.isEmpty()) {
+                store.setMetadata(new MemoryOrder(mo));
+                store.addTags(mo);
+            }
             return store;
         }
 
@@ -588,7 +613,12 @@ public class EventFactory {
         }
 
         public static RMWStoreExclusive newRMWStoreConditional(IExpr address, ExprInterface value, String mo, boolean isStrong) {
-            return new RMWStoreExclusive(address, value, mo, isStrong, true);
+            RMWStoreExclusive store = new RMWStoreExclusive(address, value, isStrong, true);
+            if (mo != null && !mo.isEmpty()) {
+                store.setMetadata(new MemoryOrder(mo));
+                store.addTags(mo);
+            }
+            return store;
         }
 
         public static Fence newISyncBarrier() {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
@@ -629,13 +629,8 @@ public class EventFactory {
         private Power() {
         }
 
-        public static RMWStoreExclusive newRMWStoreConditional(Expression address, Expression value, String mo, boolean isStrong) {
-            RMWStoreExclusive store = new RMWStoreExclusive(address, value, isStrong, true);
-            if (mo != null && !mo.isEmpty()) {
-                store.setMetadata(new MemoryOrder(mo));
-                store.addTags(mo);
-            }
-            return store;
+        public static RMWStoreExclusive newRMWStoreConditional(Expression address, Expression value, boolean isStrong) {
+            return new RMWStoreExclusive(address, value, isStrong, true);
         }
 
         public static Fence newISyncBarrier() {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
@@ -51,7 +51,7 @@ public class EventFactory {
             if (obj == null) {
                 continue;
             }
-            if (obj instanceof AbstractEvent) {
+            if (obj instanceof Event) {
                 retVal.add((Event) obj);
             } else if (obj instanceof Collection<?>) {
                 retVal.addAll((Collection<? extends Event>) obj);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
@@ -21,7 +21,6 @@ import com.dat3m.dartagnan.program.event.lang.llvm.*;
 import com.dat3m.dartagnan.program.event.lang.pthread.*;
 import com.dat3m.dartagnan.program.event.lang.std.Malloc;
 import com.dat3m.dartagnan.program.event.lang.svcomp.*;
-import com.dat3m.dartagnan.program.event.metadata.MemoryOrder;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
 
 import java.math.BigInteger;
@@ -556,10 +555,7 @@ public class EventFactory {
         public static RMWStoreExclusive newRMWStoreConditional(Expression address, Expression value, String mo, boolean isStrong) {
             RMWStoreExclusive store = new RMWStoreExclusive(address, value, isStrong, true);
             store.addTags(Tag.RISCV.STCOND);
-            if (mo != null && !mo.isEmpty()) {
-                store.setMetadata(new MemoryOrder(mo));
-                store.addTags(mo);
-            }
+            store.setMemoryOrder(mo);
             return store;
         }
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
@@ -70,21 +70,23 @@ public class EventFactory {
 
     // ------------------------------------------ Memory events ------------------------------------------
 
-    public static Load newLoad(Register register, IExpr address, String mo) {
-        Load load = new Load(register, address);
-        if (mo != null && !mo.isEmpty()) {
-            load.setMetadata(new MemoryOrder(mo));
-            load.addTags(mo);
-        }
+    public static Load newLoad(Register register, IExpr address) {
+        return new Load(register, address);
+    }
+
+    public static Load newLoadWithMo(Register register, IExpr address, String mo) {
+        Load load = newLoad(register, address);
+        load.setMemoryOrder(mo);
         return load;
     }
 
-    public static Store newStore(IExpr address, ExprInterface value, String mo) {
-        Store store = new Store(address, value);
-        if (mo != null && !mo.isEmpty()) {
-            store.setMetadata(new MemoryOrder(mo));
-            store.addTags(mo);
-        }
+    public static Store newStore(IExpr address, ExprInterface value) {
+        return new Store(address, value);
+    }
+
+    public static Store newStoreWithMo(IExpr address, ExprInterface value, String mo) {
+        Store store = newStore(address, value);
+        store.setMemoryOrder(mo);
         return store;
     }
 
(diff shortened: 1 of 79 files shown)


hernanponcedeleon:
I don't yet understand the changes in this PR. I think they are related to two things, but which I find orthogonal (number of core events and number of accessed addresses). 

Can you please describe which of the 5cases below are:
- intended to be covered (maybe in the future)
- currently covered and which class represent it

(1) There are instructions that will result in a single core event, access a single address and have a single access mode, e.g. a `Load`.
(2) There are instructions that will result in a single core event, access a single address and have more than an access mode in the current implementation (see *), `AMO` is the only example.
(3) There are instructions that will result in more than one core event, access a single address. Each core event will eventually get  a single access mode; the current event has (should have as marked by the TODOs) more than an access mode in the current implementation , e.g. a `RMW`.
(4) There are instructions that will result in several core event, access several addresses and have a single access mode, e.g. a `MemSet`.
(5) There are instructions that will result in several core event, access several addresses and have "several" access mode, e.g. a `MemCopy`.

(*) Why not add `Mode.RMW`?

ThomasHaas:
The intention of this PR was to add support for complex memory events, the most complex being `MemCpy` which performs multiple different accesses. Now let me get through your points:

(1) Yes, there will be core events that perform a single memory access ("simple/primitive memory events") such as `Load` and `Store`. I think it might be best to add a third core memory event that is neither `Load` nor `Store` but still has an address. This third event could be tagged and/or subclassed to produce special abstract events like SRCU, Lock/Unlock (as abstract events), or hazard pointers.

(2) I had the intend of supporting `AMO` as a single core event in principle, but after doing some changes, I'm not so sure about this anymore. First of all, we could not assign a "memory value" to events anymore, but would rather have to assign `loadValue` and `storeValue` separately (not a big deal). However, what concerns me about `AMOs` is that their result cannot be expressed as an `Expression` in our framework. The reason is that the load part of the `AMO` does not assign to any register and this makes it impossible to reference that value in the store part. 
We need special case handling for `AMO` in our analyses then. If we also want to support single event CAS, we would need even more special case handling.
Right now, I'm not to convinced that this added complexity is the way to go.

Regarding the `Mode.RMW`: I just had this discussion with René. At first I was under the impression that we need to support multiple `MemoryAccesses` anyhow, so why bother with a `Mode.RMW` if you can just have two separate accesses. Also, most code would likely end up doing checks like these `bool isStore = (mode == STORE || mode == RMW)` and there would be rarely (if ever) a case where you would check for just `RMW.Mode`.
Anyway, after playing around with these memory accesses, I believe that it is not really feasible to support core events that have multiple accesses in any generic form except RMW/AMO-type events (which still has issues as mentioned above). Given that, I think it is best to make all core events always be single access and if we want RMW/AMO support, then treating them as a single access (of type RMW) is the way to go. If we do not want `AMO` as a core event, then I would keep modelling them as `Mode.Load` + `Mode.Store` (cause that is what they eventually do!).

(3) - (5) All of these non-core events would be captured by the most general `MemoryEvent` interface.
The `SingleAddressMemoryEvent` class I added was just to share some common implementation, rather than creating "groups" of similar events.
I think there will be only two groups of memory events: those that are getting compiled/lowered eventually (these can be arbitarily complex) and those that we can handle natively (core events). I would probably have two interfaces `MemoryEvent` and `CoreMemoryEvent` or something like that to make this distinction clear.

In short: 
- I will add a separation between arbitrary memory events and (primitive) core memory events.
- Core memory events are always "single access": they access a single location with a single type and a single mode. After processing, a well-formed program contains no non-core memory events anymore.
- If we want to support `AMO` in the core, we add `Mode.RMW` to conform to the above point.
- Non-core memory events can return any number of memory accesses. 
Q: Do we have any instances apart from `MemCpy` that would require such functionality? I'm wondering if it might be better to give `MemCpy` its special treatment and don't bother too much with a general `MemoryEvent` interface at all?

hernanponcedeleon:
> * I will add a separation between arbitrary memory events and (primitive) core memory events.

It is not clear if you will add this as part of this PR or in the future.

>   Q: Do we have any instances apart from `MemCpy` that would require such functionality? I'm wondering if it might be better to give `MemCpy` its special treatment and don't bother too much with a general `MemoryEvent` interface at all?

What about atomic blocks / transactions?

hernanponcedeleon:
I mentioned a function pointer just to say: "tell this instruction what need to be done atomically". That function will definitely access memory (there is no point in making it atomic unless there are potential conflicts).

I would even argue that memcpy is also just a function (which is not even atomic) which we simply decided to treat natively for being a standard function. 

Memset is yet another function we would want to treat natively.

ThomasHaas:
Yes, we can treat all these as calls to intrinsic functions. The question is more about how much you need to understand about this function to treat it properly.
A function like `do_atomic(funcPtr)` is hard to analyze: even if it was a `MemoryEvent` you would know nothing about its memory accesses until you resolve `funcPtr` to a concrete function. At that point, you would be better off inlining the target function and placing atomic marker around it (i.e., lowering the intrinsic to actual core-level code).

In comparison, `MemCpy` behaves differently. You first need to resolve its range for otherwise you don't know how to lower it to sequence of Load/Store instructions. To figure out the range, the `MemCpy` event must survive compliation and unrolling until `ConstantPropagation` can compute the range.
Even then, you might not want to lower immediately, because depending on which addresses it accesses, you might do it more efficiently (e.g., if it copies 3 ints, you need only 3 loads, but if it copies 12 bytes, you need 12 loads).
And for this point you need to understand what addresses it accesses and how it accesses them, i.e., you need to understand the event as more than just a function call.
The intention of having a more general `MemoryEvent` was to allow this "better understanding of the event" before lowering it to core events. 
Right now I don't see if there is any other event besides `MemCpy` and `MemSet` that would benefit from such detailed modelling. If it is only those two, we might as well treat them as intrinsics and give special handling to those.
Indeed, after all these discussions I now think that this handling as intrinsic functions is the way to go. The `MemoryEvent` framework around this feels like a too general solution for too special of a problem (i.e. it is not worth it).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorIMM.java
```diff
@@ -24,98 +25,102 @@
...
                 store
         );
-	}
+    }
 
-	@Override
-	public List<Event> visitEnd(End e) {
+    @Override
+    public List<Event> visitEnd(End e) {
         //TODO boolean
```
hernanponcedeleon, line 64: what is this TODO?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorIMM.java
```diff
@@ -24,98 +25,102 @@
...
         Local casCmpResult = newLocal(resultRegister, expressions.makeEQ(regValue, regExpected));
         CondJump branchOnCasCmpResult = newJump(expressions.makeNEQ(resultRegister, one), casFail);
         CondJump gotoCasEnd = newGoto(casEnd);
-        Load loadValue = newRMWLoad(regValue, address, extractLoadMo(mo));
-        Store storeValue = newRMWStore(loadValue, address, e.getMemValue(), extractStoreMo(mo));
+        Load loadValue = newRMWLoadWithMo(regValue, address, extractLoadMo(mo));
+        Store storeValue = newRMWStoreWithMo(loadValue, address, e.getMemValue(), extractStoreMo(mo));
 
         return eventSequence(
                 // Indentation shows the branching structure
```
hernanponcedeleon, line 126: If you applied automatic formatting, this note is probably outdated. This recently cause problems to Rene and I was always avoid automatic formatting in these files exactly for this. I would fully remove such comment.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorLKMM.java
```diff
@@ -23,33 +23,33 @@ public class VisitorLKMM extends VisitorBase {
...
-
-        @Override
-        public List<Event> visitEnd(End e) {
-            //TODO boolean
-            return eventSequence(
-                    newStore(e.getAddress(), expressions.makeZero(types.getArchType()), Tag.Linux.MO_RELEASE)
-            );
-        }
+    public List<Event> visitEnd(End e) {
+        //TODO boolean
```
hernanponcedeleon, line 40: What is this boolean?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -858,114 +854,117 @@ public List<Event> visitRMWOpReturn(RMWOpReturn e) {
...
         CondJump branchOnCauCmpResult = newJump(expressions.makeEQ(dummy, zero), cauEnd);
 
         Fence optionalMemoryBarrierBefore = mo.equals(Tag.Linux.MO_MB) ? Power.newSyncBarrier()
                 : mo.equals(Tag.Linux.MO_RELEASE) ? Power.newLwSyncBarrier() : null;
         Fence optionalMemoryBarrierAfter = mo.equals(Tag.Linux.MO_MB) ? Power.newSyncBarrier()
-				: mo.equals(Tag.Linux.MO_ACQUIRE)? Power.newISyncBarrier() : null;
+                : mo.equals(Tag.Linux.MO_ACQUIRE) ? Power.newISyncBarrier() : null;
 
         return eventSequence(
                 // Indentation shows the branching structure
```
hernanponcedeleon, line 922: Remove

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -403,132 +405,132 @@ public List<Event> visitAtomicXchg(AtomicXchg e) {
...
-		Register dummy = e.getThread().newRegister(e.getResultRegister().getType());
-		Register statusReg = e.getThread().newRegister(e.getResultRegister().getType());
+        Register dummy = e.getThread().newRegister(e.getResultRegister().getType());
+        Register statusReg = e.getThread().newRegister(e.getResultRegister().getType());
         Label casEnd = newLabel("CAS_end");
         CondJump branchOnCasCmpResult = newJump(expressions.makeNEQ(dummy, e.getCmp()), casEnd);
-        
-        Load load = newRMWLoadExclusive(dummy, address, "");
+
+        Load load = newRMWLoadExclusive(dummy, address); // No mo on the load?
```
hernanponcedeleon, line 494: Add TODO so IDE warning would just us this

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -610,36 +614,36 @@ public List<Event> visitRMWFetchOp(RMWFetchOp e) {
...
+        Expression zero = expressions.makeZero(type);
+        IOpBin op = e.getOp();
+        Expression value = e.getMemValue();
+        Expression address = e.getAddress();
+        String mo = e.getMo();
+
+        Register dummy = e.getThread().newRegister(type);
+        Register statusReg = e.getThread().newRegister(type);
+
+        Load load = newRMWLoadExclusive(dummy, address); // No mo on the load?
```
hernanponcedeleon, line 637: Add TODO

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -649,33 +653,35 @@ public List<Event> visitRMWOpReturn(RMWOpReturn e) {
...
+        Expression address = e.getAddress();
+        Expression value = e.getMemValue();
+        String mo = e.getMo();
 
         Register regValue = e.getThread().newRegister(type);
-		Register statusReg = e.getThread().newRegister(type);
+        Register statusReg = e.getThread().newRegister(type);
 
-		Load load = newRMWLoadExclusive(regValue, address, "");
+        Load load = newRMWLoadExclusive(regValue, address);
```
hernanponcedeleon, line 676: Add TODO as similar instances

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -685,42 +691,44 @@ public List<Event> visitRMWAddUnless(RMWAddUnless e) {
...
+
+        Register dummy = e.getThread().newRegister(type);
+        Register statusReg = e.getThread().newRegister(type);
         Register retReg = e.getThread().newRegister(type);
         Local localOp = newLocal(retReg, expressions.makeBinary(dummy, op, value));
         Local testOp = newLocal(resultRegister, expressions.makeEQ(retReg, expressions.makeZero(type)));
 
-        Load load = newRMWLoadExclusive(dummy, address, "");
-        Store store = newRMWStoreExclusive(address, retReg, mo.equals(Tag.Linux.MO_MB) ? Tag.RISCV.MO_REL : "", true);
+        Load load = newRMWLoadExclusive(dummy, address);
```
hernanponcedeleon, line 725: Add TODO

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorTso.java
```diff
@@ -17,272 +17,272 @@
...
+        Local casCmpResult = newLocal(resultRegister, expressions.makeEQ(regValue, regExpected));
+        Label casFail = newLabel("CAS_fail");
+        CondJump branchOnCasCmpResult = newJump(expressions.makeNEQ(resultRegister, one), casFail);
+        Store storeValue = newRMWStore(loadValue, address, value);
+        Label casEnd = newLabel("CAS_end");
+        CondJump gotoCasEnd = newGoto(casEnd);
+        Store storeExpected = newStore(expectedAddr, regValue);
+
+        return tagList(eventSequence(
+                // Indentation shows the branching structure
```
hernanponcedeleon, line 216: Remove

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -521,8 +554,12 @@ private RISCV() {
         }
 
         public static RMWStoreExclusive newRMWStoreConditional(Expression address, Expression value, String mo, boolean isStrong) {
-            RMWStoreExclusive store = new RMWStoreExclusive(address, value, mo, isStrong, true);
+            RMWStoreExclusive store = new RMWStoreExclusive(address, value, isStrong, true);
             store.addTags(Tag.RISCV.STCOND);
+            if (mo != null && !mo.isEmpty()) {
```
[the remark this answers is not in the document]
ThomasHaas: Just double protection. We could either throw an exception or allow `null` to mean no memory order.
Our frontend tries to never generate `null`, but I don't see the benefit of forcing the backend to not accept `null`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -521,8 +554,12 @@ private RISCV() {
...
+                store.setMetadata(new MemoryOrder(mo));
+                store.addTags(mo);
```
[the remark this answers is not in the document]
ThomasHaas: Good catch. I guess I missed to update this code to use `setMemoryOrder`. 
I will double check that all `EventFactory` methods use `setMemoryOrder` properly.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLoad.java
```diff
@@ -8,8 +8,13 @@
 
 public class LKMMLoad extends Load {
```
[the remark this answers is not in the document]
ThomasHaas: This is because LKMMLoad is unforunately not compiled when verifying under LKMM, so it has to be a core event.
If I add compilation to a normal Load, we lose the custom printing as `READ_ONCE` after processing.
If you are willing to lose the custom printing, I can make it a non-core event. If you want to keep the custom printing, you will need to wait for a follow-up PR that fixes this :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/llvm/LlvmStore.java
```diff
@@ -37,32 +38,36 @@ public Set<Register.Read> getRegisterReads() {
...
     }
-    
-    @Override
+
     public void setMemValue(Expression value){
         this.value = value;
     }
 
+    @Override
+    public MemoryAccess getMemoryAccess() {
```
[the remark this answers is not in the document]
ThomasHaas: Yes, but not for `SingleAccessMemoryEvent`. There it is sufficient to implement a single memory access method

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorIMM.java
ThomasHaas: This is a TODO from the #457 PR. 
I think it says that we eventually want to use boolean variables to communicate thread creation (since we only use values `0` and `1` anyway).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorIMM.java
ThomasHaas: Yeah, I did automatic formatting. I forgot about the indentations there. I will add them back in unless you are fine without the indentations.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorLKMM.java
ThomasHaas: See above

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
ThomasHaas: Will do

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLoad.java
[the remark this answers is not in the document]
ThomasHaas: More precisely, it will be a `LoadBase`, it will always get compiled to a core-level `Load` and if compiled to Linux, we add custom printing.
Keep in mind that `SingleAccessMemoryEvent` and `LoadBase` are just classes to share implementation, so it would also be feasible (and semantically identical) to have `LKMMLoad` inherit from neither of those.

Issue #446, opened 2023-05-17
[Rework] Overview


ThomasHaas:
This issue is for documentation purposes.

There is a new branch called `development_rework`.
PRs that are directly related to the upcoming rework should target this new branch.
The branch is set to be compatible with Java 17 so new language features can be used.
Once we are done with the rework, we will merge `development_rework` into `development`.

*NOTES:* 

(1) There is a `prototype` package in this new branch. The prototype package contains code that should be used as a guide/reference when it comes to implementing PRs that affect the actual code base. The prototype code itself is highly non-functional.
Changes to the prototype are also welcomed, but those should be done in a separate PR that is named like `[Prototype] XYZ`.

(2) Actual implementation will inevitably differ from the prototype. There are (at least) three reasons for this:
  - (2.1) Due to small implementation details that the prototype (intentionally) did not cover, but that are needed in practice. Such differences are non-issues.
  - (2.2) Due to intermediate PRs that are not able to match the actual code with the prototype yet (it is hard to do all the necessary changes in a single PR). If the PR is meant to be just a stepping stone towards the prototype, then the PR should clearly document this.
  - (2.3) Due to important practical demands that were not (correctly) modelled/covered in the prototype. In this case, we should first create a PR that fixes/adapts the prototype. This also opens up discussions about the design goal.

(3) Adapting the prototype: The prototype does not only provide documentation via a rough implementation in code, but also via comments that show things such as design intention, possible issues (in practice), and possible solutions.
These comments are as essential as the code itself, so PRs that just add or clarify comments are perfectly fine.

ThomasHaas:
This serves no purpose anymore (and it never really did).

PR #460, opened 2023-06-09
[Prototype] Core memory events

ThomasHaas:
This PR updates only the prototype code (according to #446 ).

This PR adds a `CoreMemoryEvent` interface that has essentially 4 implementations: `Load`, `Store`, `ReadModifyWrite`, and `GenericMemoryEvent`.
The former two are self-explanatory. `ReadModifyWrite` is for single-event RMWs/AMOs and essentially amounts to a `RMWOp` event (read -> local op -> write).
`GenericMemoryEvent` is used for more abstract kinds of memory events like "SRCU", "Hazard Pointers (SMR)", or abstract locking. This event has an address but no access type (since it neither loads nor stores values).

I want to use this PR for discussions about how memory events should be modelled, what we want to support in the core, etc.
Depending on the result of this discussion, I will update PR #456 accordingly.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/prototype/program/Register.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/prototype/program/Register.java
@@ -25,7 +25,7 @@ public class Register extends LeafExpressionBase<Type, ExpressionKind.Leaf> {
         return new Register(type, name, owner);
     }
 
-    public static void collectRegisterReads(Expression expr, UsageType depType, Set<Read> collector) {
+    public static Set<Read> collectRegisterReads(Expression expr, UsageType depType, Set<Read> collector) {
         if (expr instanceof Register reg) {
             collector.add(new Read(reg, depType));
         } else {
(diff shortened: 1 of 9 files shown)


hernanponcedeleon:
I will only comment about the "idea" of the PR and not the "implementation" (there are still implementation details missing as shown by the use of `abstract` in most of the classes).

The design implies that when we need to reason about / encode the WMM, we only need to reasons about these 4 events. I like this. The 4 cases cover everything that comes to my mind (`GenericMemoryEvent` might have a few subclasses with special semantics, but having that event as the abstraction of all those seem the right level).

My only comment: I believe it is now clear that we can get rid of the `mo` in the final events. When we reach this point, we will only care about the tags that the events have.

ThomasHaas:
From what you said I assume that you approve of these design-level changes.

Regarding the `mo`: While it is true that we do not require any `mo` at the end anymore, it is used for presentation purposes (e.g. printing).
Maybe we can treat `mo` as metadata in this case and have a `MemoryOrder` metadata class, so that we still get proper debug printing abilities + nicer witness generation.
If you agree with this, then I will remove `mo` from the core classes (maybe I keep `getMo`/`setMo` but just as nice functions to attach/retrieve the metadata).

PR #462, opened 2023-06-19
Event printer & reduced core

ThomasHaas:
This PR adds a new `CustomPrinting` metadata (naming is subject to change) that allows us to add custom printing code to any event (in particular core-level events).
Towards this, `AbstractEvent.toString` has been made final but a new `defaultString` method has been added.
`toString` will now first check for `CustomPrinting` metadata and use that if available before falling back to `defaultString`.

There is a new `GenericMemoryEvent` core event that neither stores nor loads but just "abstractly" accesses memory.
`SrcuSync` has been removed (\*) for now and `EventFactory.newSrcuSync` just creates a `GenericMemoryEvent` with appropriate tagging.

~~What remains to be done is to make many of the language-level events stop inheriting from core-level events (e.g. `LKMMLoad`) and instead compile them down while attaching custom printing~~

There is also a new `CoreCodeVerification` pass that throws exceptions if the program contains non-core events.
~~The pass is disabled for now, because there are still many LKMM events that are not compiled away (work-in-progress).~~

(\*): We have no compilation for `SrcuSync`, so we do not need a class to represent it. If we ever want to add compilation, we would add this class back in and compile it down to `GenericMemoryEvent` for LKMM.


dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/SyntacticContextAnalysis.java
```diff
@@ -207,7 +207,7 @@ private Map<Event, LoopMarkerTypes> getLoopMarkerTypesMap(Thread thread, LoopAna
 
     public static String getSourceLocationString(Event ev) {
         SourceLocation loc = ev.getMetadata(SourceLocation.class);
-        return loc != null ? String.format("@%s#%s", loc.getSourceCodeFileName(), loc.getLineNumber()) : "@unknown";
+        return loc != null ? String.format("@%s#%s", loc.getSourceCodeFileName(), loc.lineNumber()) : "@unknown";
```
hernanponcedeleon, line 210: why `get` for the source code file name, but not for the line?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CoreCodeVerification.java
```diff
@@ -0,0 +1,50 @@
...
+            Load.class, Store.class, Init.class, GenericMemoryEvent.class, Fence.class,
+            CondJump.class, IfAsJump.class, ExecutionStatus.class, Label.class, Local.class,
+            Skip.class, Assume.class, RMWStore.class, RMWStoreExclusive.class
+    ));
+
+    @Override
+    public void run(Program program) {
+
+        final List<Event> nonCoreEvents = program.getEvents().stream().
+                filter(e -> !(e instanceof CodeAnnotation) && !CORE_CLASSES.contains(e.getClass())).toList();
```
hernanponcedeleon, line 40: Why do you need to treat `CodeAnnotation` differently?

hernanponcedeleon (commented):
The current changes LGTM, but not sure if you want to implement what is missing (and enable the new pass) as part of this PR

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CoreCodeVerification.java
ThomasHaas: In a sense, `CodeAnnotations` live on a different level than both language-level events and core-level events and so it is easier to also treat them differently (also, `CodeAnnotations` are never compiled!).

ThomasHaas:
I would like to still do some changes, that's why this is work-in-progress.
I created the PR just to get proper CI support :)

hernanponcedeleon:
I think that if you use the [DRAFT] prefix, even GitHub recognizes this is WIP in at might have some builtin features

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorLKMM.java
```diff
@@ -264,15 +253,60 @@ public List<Event> visitRMWXchg(RMWXchg e) {
...
+        return load;
+    }
+
+    private static Store newCoreStore(Expression addr, Expression value, String mo) {
+        Store store = EventFactory.newStoreWithMo(addr, value, mo);
+        store.setMetadata(LKMMStore.CUSTOM_CORE_PRINTING);
+        return store;
+    }
+
+    private static Load newLockRead(Expression lock, Register dummy) {
```
hernanponcedeleon, line 293: For consistency with all other `Load`s, the register should be the first parameter

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CoreCodeVerification.java
```diff
@@ -30,7 +32,8 @@ public static CoreCodeVerification fromConfig(Configuration config) {
     private static final Set<Class<? extends Event>> CORE_CLASSES = new HashSet<>(Arrays.asList(
             Load.class, Store.class, Init.class, GenericMemoryEvent.class, Fence.class,
             CondJump.class, IfAsJump.class, ExecutionStatus.class, Label.class, Local.class,
-            Skip.class, Assume.class, RMWStore.class, RMWStoreExclusive.class
+            Skip.class, Assume.class, RMWStore.class, RMWStoreExclusive.class,
+            BeginAtomic.class, EndAtomic.class // We add SVCOMP atomic blocks here as well
```
hernanponcedeleon, line 36: The comment is useless. It is obvious that those were added. Either explain why we need to add them or simply remove the comment

hernanponcedeleon:
LGTM

PR #463, opened 2023-06-19
Barrier

hernanponcedeleon:
We have been treating compiler barriers as no ops in LKMM. [This is wrong](https://github.com/hernanponcedeleon/Dat3M/blob/b3c7d9a7ba5cd5230414bdcdfa3880abd44d9f84/cat/linux-kernel.cat#L75). For the architectures we support, the barrier is indeed a no op, thus the compilation to those archs has been updated.


ThomasHaas (approved):
LGTM

Issue #260, opened 2022-05-26
Issues with Duplicated Labels


ThomasHaas:
Code that contains multiple labels with the same name causes various problems:
- The parsing generates bogus code if the same label name is used in _different_ threads. This should not be a problem in principle.
- If the same thread contains duplicate labels, the parser generates programs with cyclic `getSuccessor`-relation and runs into an infinite loop trying to find the exit event.

The underlying issues are 
- The `ProgramBuilder` tracks labels globally instead of per thread (causing issue 1).
- The method `getOrCreateLabel` is used throughout. This causes creation of labels in cases where none should get created and retrieval of existing labels when new ones should get created.

hernanponcedeleon:
Fixed in #470

PR #470, opened 2023-06-22
Functions (Part 1)

ThomasHaas:
This PR introduces a `Function` class. Events are now part of a `Function` rather than a `Thread`.
However, a `Thread` inherits from `Function`, so events can be part of a thread if the function happens to be a thread.

While adding/parsing functions, I noticed that the boogie parser is a nightmare I needed to fix first. So most of this PR is now dedicated to cleaning up `VisitorBoogie` and its related components.
An overview of the changes:
- `VisitorBoogie` and its related components are far simpler now
- `VisitorBoogie` can parse in inline mode and no-inline mode (the former mode matches with what we are doing right now)
-  The parser currently runs once in inline mode and once without, generating proper functions rather than threads. 
However, the generated functions are incomplete (contain placeholders) and unused in this PR. A future PR (Part 2?) will change this. For the purpose of reviewing this PR, code inside `!inlineMode`-conditionals can be disregarded and parts marked by "// TODO ---- Test Code ----" can be assumed to be commented out (those are for the future PR).
- Issue #260 should be fixed.
- Scope-prefixes reset per thread now, i.e., they start from `0:` per thread. This was previously not the case because of the above issue with duplicate label names.
- The parser creates no more unused dummy registers in functions/threads.


dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
```diff
@@ -0,0 +1,118 @@
...
+        int paramIndex = 0;
+        for (Type paramType : functionType.getParameterTypes()) {
+            final Register paramReg = newRegister(parameterNames.get(paramIndex++), paramType);
+            parameterRegs.add(paramReg);
+        }
+
+        Event cur = entry;
+        while (cur != null) {
+            cur.setFunction(this);
+            exit = cur;
```
hernanponcedeleon, line 43: do we really need to update in every iteration? why not simply at the end?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AbstractEvent.java
```diff
@@ -258,8 +259,8 @@ public int compareTo(Event e) {
     }
 
     @Override
-    public Event getCopy() {
-        throw new UnsupportedOperationException("Copying is not allowed for " + getClass().getSimpleName());
+    public void transformExpressions(ExpressionVisitor<? extends Expression> exprTransformer) {
+        // WARNING: This should get overwritten by any event that reads Expressions
```
hernanponcedeleon, line 263: Why not to add an exception then? Right now we will simply miss if an implementation is missing

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/pthread/Lock.java
```diff
@@ -31,21 +31,17 @@ public String defaultString() {
         return "pthread_mutex_lock(&" + name + ")";
     }
 
+    //TODO: Why does this have a ResultRegister without inheriting from RegWriter?
+    // Why is this even needed?
```
hernanponcedeleon, line 35: It is used in the compilation to create the corresponding Load. We should probably make it inherit from `RegWriter`

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/pthread/Unlock.java
```diff
@@ -31,21 +31,17 @@ public String defaultString() {
         return "pthread_mutex_unlock(&" + name + ")";
     }
 
+    //TODO: Why does this have a ResultRegister without inheriting from RegWriter?
```
hernanponcedeleon, line 34: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/FunctionType.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * returnType.hashCode() + Arrays.hashCode(parameterTypes);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%s (%s)", returnType,
```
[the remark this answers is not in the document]
ThomasHaas: I used C-like syntax, but I can change it.  Just to be sure, you would then be fine with a type like  `(t1, ..., tn) -> void`?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -65,19 +71,27 @@ public Program build(){
         return program;
     }
 
-    public void initThread(String name, int id){
-        if(!threads.containsKey(id)){
+    public void initThread(String name, int id) {
+        if(!functions.containsKey(id)){
             Skip threadEntry = EventFactory.newSkip();
```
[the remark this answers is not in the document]
ThomasHaas: > entry sounds like a better name since this is not necessary a thread.

It is necessarily a thread that gets created. There is a separate `initFunction` method that creates functions.

> Not sure if it make sense in the current state of this PR, but shouldn't we at some point add FunctionEntry and FunctionExit nodes (i.e. events)?

Why? I don't think functions need any of that, because (1) you can have a empty (= no body) functions and (2) we have a `Return` event to exit functions.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
ThomasHaas: You mean `exit`? At the end of the loop `cur == null`, so you do not know what the exit event was, unless you track it, which is exactly what the code does.
If you talk about `setFunction`, then this needs to get executed for every event of course. 
The whole loop is just to make sure that if you create a function with an entry event that already has successors, then all the successors are added to the function as well.
That being said, I should probably make sure that `entry.getPredecessor() == null` holds.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AbstractEvent.java
ThomasHaas: A lot of events actually have an empty implementation for that, for example `Fence`, `CodeAnnotation`,`Label`, and the SVCOMP atomic block events. I can add the exception, but then I need to provide an empty implementation for all of these events separately.
An alternative is to move this method into `RegReader`, assuming that all expression-dependent events are actually `RegReaders` (which I think is a fair assumption). I will move it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/pthread/Lock.java
ThomasHaas: But the register a pure dummy register, no? The compilation can simply create a fresh one instead of using some pre-defined one.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/FunctionType.java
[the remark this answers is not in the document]
hernanponcedeleon: Works for me

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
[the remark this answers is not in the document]
hernanponcedeleon: In a CFG you would normally have entry/exit nodes. The `Return` event is exactly what I had in mind (I wrote the comment before going over all changes). Not sure if having a specialized `Entry extends Skip` would make sense or bring any advantage.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
hernanponcedeleon: I was talking about `exit` so your first comment answers my question already

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/pthread/Lock.java
hernanponcedeleon: You are right. We can fully remove the register from lock/unlock events.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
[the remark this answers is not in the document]
ThomasHaas: Currently, all function bodies start with a `$bb0` label. And I think this will also holds true for LLVM code.
At least for now, a dedicated entry node just has disadvantages if you allow for empty-bodied functions. Because on an empty function, `f.appendEvent(e)` would need to implicitly create an entry event before appending `e`.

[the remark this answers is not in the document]
ThomasHaas: Actually, now that I think about it, we will probably need a marker of sorts to distinguish between "empty body" and "no body". Maybe a dedicated `FunctionEntry` event could serve this purpose. But I would do this in a PR that does add proper support for functions (i.e. part 2 or 3, depending on how difficult it gets :)).

hernanponcedeleon:
LGTM

PR #471, opened 2023-06-22
Generic Registers (Rework)

xeren:
This PR rewrites more static expression class requirements into dynamic type requirements.  It also removes `Register extends IExpr`, which should now allow registers of any type, including booleans.  On the other hand, the intermediate representation of the program can now be more strictly-typed:
- In any assignment, the type of the result register should equal the type of the value expression.
- Any `ExprTransformer` should map to expressions of equal type.
- `ExpressionFactory.makeCast(Expression,Type)` should be used as a last resort to enforce the type.
- The value type of a `Store` is the type of its value expression.  The value type of a `Load` is the type of its result register.


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -111,20 +113,50 @@ public BooleanFormulaManager getBooleanFormulaManager() {
...
 
-    public Formula encodeFinalIntegerExpression(Expression expression) {
-        return new ExpressionEncoder(formulaManager, null).encodeAsInteger(expression);
+    public Formula encodeFinalExpression(Expression expression) {
+        return new ExpressionEncoder(this, null).encode(expression);
     }
 
     public BooleanFormula encodeBooleanExpressionAt(Expression expression, Event event) {
-        return new ExpressionEncoder(formulaManager, event).encodeAsBoolean(expression);
+        return new ExpressionEncoder(this, event).encodeAsBoolean(expression);
```
ThomasHaas, line 121: Does this function convert integer expressions to boolean one's, or does it perform a type-check to make sure only boolean expressions get through?
If it is the former, I think the function should get renamed to something like `encodeExpressionAsBooleanAt`.
Otherwise it is fine.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -304,11 +243,13 @@ public Formula visit(IExprBin iBin) {
...
     public Formula visit(IExprUn iUn) {
-        //TODO if inner is a boolean formula, we still want to encode it as boolean in case of conversions
-        Formula inner = encodeAsInteger(iUn.getInner());
+        Formula inner = encode(iUn.getInner());
         Type targetType = iUn.getType();
         switch (iUn.getOp()) {
             case MINUS -> {
+                if (inner instanceof BooleanFormula bool) {
+                    return bool;
+                }
```
ThomasHaas, line 252: Kinda fishy. Are there cases where this code actually runs, i.e., do we ever produce somthing like `-false`?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -21,163 +26,207 @@ public static ExpressionFactory getInstance() {
...
     public IValue makeZero(IntegerType type) {
         return makeValue(BigInteger.ZERO, type);
     }
 
+    @Deprecated
+    public IValue makeZero(Type type) {
+        if (type instanceof IntegerType t) {
+            return makeZero(t);
+        }
+        throw new IllegalArgumentException("Non-integer type " + type);
```
ThomasHaas, line 69: You should use `Preconditions.checkArgument` instead.
This is also true for the other methods like that raise `IllegalArgumentExceptions` in a similar way.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/IExprUn.java
```diff
@@ -41,7 +41,9 @@ public String toString() {
 
     @Override
     public IConst reduce() {
-        IntegerType innerType = b.getType();
+        if (!(b.getType() instanceof IntegerType innerType)) {
+            throw new IllegalStateException(String.format("Non-integer operand %s.", b));
+        }
```
ThomasHaas, line 46: Use `Preconditions.checkState` and manually cast to `IntegerType`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
```diff
@@ -62,54 +63,34 @@ public Expression visit(Atom atom) {
...
-        if(!(innerExpr instanceof BExpr)) {
-            return bUn;
-        }
-        BExpr inner = (BExpr) innerExpr;
-        if (inner instanceof BConst) {
-            return expressions.makeValue(!inner.isTrue());
+    public Expression visit(BExprUn bUn) {
+        Expression inner = bUn.getInner().visit(this);
+        if (inner instanceof BConst constant) {
+            return expressions.makeValue(!constant.getValue());
```
ThomasHaas, line 90: Maybe assert that the unary operator is indeed negation just to be sure that we do not accidentally break this in the future.

ThomasHaas (commented):
Our CI does not run any tests with non-mathematical integers, so if you make changes in types and their encodings, make sure to do some manual testing.
That being said, I tested this branch locally with `arch_precision=32` and didn't find any problems.

PR #474, opened 2023-06-24
Avoid explicit casts by using java17

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -458,8 +458,7 @@ public Object visitNreC11Store(LitmusCParser.NreC11StoreContext ctx) {
     public Object visitNreAssignment(LitmusCParser.NreAssignmentContext ctx){
         Expression variable = (Expression)ctx.varName().accept(this);
         if(ctx.Ast() == null){
-            if(variable instanceof Register){
-                returnRegister = (Register)variable;
+            if(variable instanceof Register returnRegister){
                 ctx.re().accept(this);
                 return null;
```
ThomasHaas, line 463: This is wrong, because the original code wrote into an instance variable, while the new code creates a fresh and unused local.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -176,7 +176,7 @@ public IExpr makeRSH(Expression leftOperand, Expression rightOperand, boolean si
     }
 
     public IExpr makeBinary(Expression leftOperand, IOpBin operator, Expression rightOperand) {
-        checkArgument(leftOperand instanceof IExpr && rightOperand instanceof IExpr,
+        checkArgument(leftOperand instanceof IExpr left && rightOperand instanceof IExpr right,
                 String.format("Non-integer operands for %s %s %s.", leftOperand, operator, rightOperand));
         return new IExprBin((IExpr) leftOperand, operator, (IExpr) rightOperand);
```
ThomasHaas, line 181: This doesn't do anything :)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -539,8 +539,8 @@ public IExpr visitVarName(LitmusCParser.VarNameContext ctx){
 
     private IExpr getAddress(LitmusCParser.ReContext ctx){
         Expression address = (Expression)ctx.accept(this);
-        if(address instanceof IExpr){
-           return (IExpr)address;
+        if(address instanceof IExpr add){
+           return add;
```
ThomasHaas, line 543: Use `addr` as abbreviation. I dislike `add` because it conflicts with "addition": if you just read `IExpr add` without context, it looks like it refers to a binary addition.
In general, I think we should aim to name variables in way that minimizes the amount of context needed to understand their meaning.

ThomasHaas (commented):
LGTM

ThomasHaas (approved):

PR #475, opened 2023-06-24
Rmw base classes & refactor

ThomasHaas:
This PR adds 4 RMW-based template classes into the `event.common` package: `RMWOpBase`, `RMWOpResultBase`, `RMWCmpXchgBase` and `RMWXchgBase`.

These base classes are used to implement all most language/hardware-level atomics (C11, Linux, Llvm, Tso, Lisa, and PTX). To avoid name clashes, the Linux atomics are now prefixed with `LKMM` rather than `RMW` (e.g. `RMWFetchOp` -> `LKMMFetchOp`). Hardware events are now also prefixed (`Xchg -> TSOXchg`, `AtomOp` -> `PTXAtomOp`, `RMW -> LISARMW`).


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -793,17 +787,15 @@ public List<Event> visitRMWXchg(RMWXchg e) {
...
-    public List<Event> visitRMWOp(RMWOp e) {
-        Register resultRegister = e.getResultRegister();
-        IOpBin op = e.getOp();
-        Expression value = e.getMemValue();
+    public List<Event> visitRMWOp(LKMMOpNoReturn e) {
         Expression address = e.getAddress();
         String mo = e.getMo();
 
-        Register dummy = e.getThread().newRegister(resultRegister.getType());
+        Register dummy = e.getThread().newRegister(types.getArchType());
```
hernanponcedeleon, line 794: Are you sure this does not cause problems when we enable BVs?

ThomasHaas: If we enable BVs, then `getArchType` will return a BV type as expected.

`RMWOp` was non-returning: the `resultRegister` did not make any sense to begin with.
In the boogie code, there is no register, so we artificially created one which naturally got the `archType` (what else would we choose?), i.e., the new code produces the exact same result as before (just more directly).

ThomasHaas:
I just noticed that I did not rename the `EventVisitor` methods to match with the new event names. I will fix this.

hernanponcedeleon:
LGTM

PR #478, opened 2023-06-30
Type Safety (rework)

xeren:
This PR lets compilers use boolean registers if appropriate.  It also strengthens the type requirements for `Local` assignments, such that the internal representation does not rely on implicit type casts.  Finally, it replaces the deprecated methods in `ExpressionFactory` that were added in #457.


ThomasHaas:
In case you didn't find the source of the error yet: 
LitmusC tests have code like `r2 = (r2 == r1)` where `r2` is of integer type.
You need to change the parser to insert explicit casts in such cases.

PR #479, opened 2023-07-02
Parsing & thread creation

ThomasHaas:
This PR mostly reworks (refactor + cleanup) the boogie parser and changes the way pthread events are handled and threads are created to more closely resemble the correct `pthread_create/join` semantics.

- There are no more fake matcher events. 
- The addresses of `pthread_t` variables are no longer abused to start/end threads. Instead, threads get a proper communication address allocated.
- The parser does a simple form of constant propagation specific to thread ids stored by `pthread_create` and read by `pthread_join` to statically match `pthread_join` with a specific thread. This trick roughly amounts to what we did in `development` already, but follows a more systematic way. Once we delay thread creation to the actual processing stage, we can get rid of this "magic".
- `FunctionType` is now created via `TypeFactory`. All threads have a `FunctionType` now.
- Overall cleanup/refactoring, especially of `VisitorBoogie` and `ProgramBuilder`.
- `VisitorBoogie` no longer abuses registers to parse constants.


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLitmusAArch64.java
```diff
@@ -21,12 +20,10 @@ public Program parse(CharStream charStream) {
         LitmusAArch64Parser parser = new LitmusAArch64Parser(tokenStream);
         parser.addErrorListener(new DiagnosticErrorListener(true));
         parser.addErrorListener(new ParserErrorListener());
-        ProgramBuilder pb = new ProgramBuilder(SourceLanguage.LITMUS);
         ParserRuleContext parserEntryPoint = parser.main();
-        VisitorLitmusAArch64 visitor = new VisitorLitmusAArch64(pb);
+        VisitorLitmusAArch64 visitor = new VisitorLitmusAArch64();
 
         Program program = (Program) parserEntryPoint.accept(visitor);
-        program.setArch(Arch.ARM8);
```
hernanponcedeleon, line 29: This clashes with #480.

I just saw `ProgramBuilder.forArch()` method. Just to be sure, they arch is eventually set for litmus tests, right?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLlvm.java
```diff
@@ -20,7 +20,7 @@ public Program parse(CharStream charStream) {
 
         LLVMIRParser parser = new LLVMIRParser(tokenStream);
         parser.setErrorHandler(new BailErrorStrategy());
-        ProgramBuilder pb = new ProgramBuilder(SourceLanguage.LITMUS);
+        ProgramBuilder pb = ProgramBuilder.forLanguage(SourceLanguage.LITMUS);
```
hernanponcedeleon, line 23: Shouldn't it be `SourceLanguage.BOOGIE`? If so, the fact that this was there before and did not cause problems suggests that this is completely useless.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -25,272 +27,307 @@
...
+    }
+
+    // ----------------------------------------------------------------------------------------------------------------
+    // Threads and Functions
+
+    public Thread newThread(String name, int tid) {
+        if(id2FunctionsMap.containsKey(tid)) {
+            throw new MalformedProgramException("Function or thread with id " + tid + " already exists.");
         }
+        final Thread thread = new Thread(name, DEFAULT_THREAD_TYPE, List.of(), tid, EventFactory.newSkip());
```
hernanponcedeleon, line 106: Add a comment that we assume the main function does not have parameters. I am not sure we will ever get rid of this assumption, but better to have this properly documented.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -25,272 +27,307 @@
...
-        if (child.getThread() != null) {
-            //FIXME: This is a bad error message, but our tests require this for now.
-            final String error = String.format(
-                    "Trying to reinsert event %s from thread %s into thread %s",
-                    child, child.getThread().getId(), fid);
-            throw new MalformedProgramException(error);
+        return (Thread) id2FunctionsMap.get(tid);
+    }
+
+    public boolean functionExists(int fid) {
```
hernanponcedeleon, line 133: not used anywhere. However this class has several instances where this could be used

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -28,23 +30,24 @@
...
+    public VisitorLitmusC(){
     }
 
     // ----------------------------------------------------------------------------------------------------------------
     // Entry point
 
     @Override
     public Program visitMain(LitmusCParser.MainContext ctx) {
+        //FIXME: We should visit thread declarations before variable declarations
+        // because variable declaration refer to threads.
```
hernanponcedeleon, line 50: Why is this not causing problems right now? It is because of line 85?

Also, don't we have the same issue with other litmus parsers?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
```diff
@@ -65,6 +68,33 @@ public static void handlePthreadsFunctions(VisitorBoogie visitor, Call_cmdContex
...
+            visitor.addEvent(EventFactory.newFunctionCall("dummy pthread_join()"));
+            // TODO: Create a proper function call, not just an annotation
+            return;
+        }
+        // ----- TODO: Test code end -----
+        // reg := pthread_join(tid);
+        final Register reg = visitor.getScopedRegister(ctx.call_params().Ident(0).getText());
+        Expression threadToJoinWith = (Expression) ctx.call_params().exprs().expr(0).accept(visitor);
+        //FIXME: The first parameter should be a proper register, but we do not know its constant value here
+        // so we loop it up in a map.
```
hernanponcedeleon, line 84: loop -> look

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
```diff
@@ -65,6 +68,33 @@ public static void handlePthreadsFunctions(VisitorBoogie visitor, Call_cmdContex
...
+        threadToJoinWith = visitor.expr2tid.getOrDefault(threadToJoinWith, threadToJoinWith);
```
hernanponcedeleon, line 85: If the expression is not in the map, we return the same expression. I assume this is the case where we do not know its constant value. What are the implications of this?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
```diff
@@ -65,6 +68,33 @@ public static void handlePthreadsFunctions(VisitorBoogie visitor, Call_cmdContex
...
+
+        if (!(threadToJoinWith instanceof IConst constId)) {
+            throw new UnsupportedOperationException("Cannot handle pthread_join with dynamic thread parameter.");
```
hernanponcedeleon, line 88: I guess this answer my question

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Local.java
```diff
@@ -19,6 +20,8 @@ public class Local extends AbstractEvent implements RegWriter, RegReader {
     protected Expression expr;
 
     public Local(Register register, Expression expr) {
+        Preconditions.checkNotNull(register);
+        Preconditions.checkNotNull(expr);
```
hernanponcedeleon, line 24: Can't we rewrite this as
```
this.register = Preconditions.checkNotNull(register);
this.expr = Preconditions.checkNotNull(expr);
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLitmusAArch64.java
ThomasHaas: It is actually set immediately inside `ProgramBuilder.forArch()`. `ProgramBuilder` now starts from an empty `Program` instance and adds functions/threads/events to that instance.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLlvm.java
ThomasHaas: This is a snippet of the `LLVMParser` which is not used right now. It should neither be Boogie nor Litmus but `LLVM` instead.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
ThomasHaas: Actually, the parser does handle main with multiple input parameters, treating them all as unassigned registers (which are eventually treated as either 0 or non-deterministic, not sure).
I will add a comment that says that the thread type does not matter for now (I could set it to anything really).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
ThomasHaas: The other Litmus parser work properly by first visiting all thread declarations.
I guess the LitmusC parser doesn't do this, because its C-like syntax is different to the usual Litmus syntax?
Anyway, the functions of the LitmusC parser are written so that they create a thread if they encounter a variable/register that refers to that thread.
I didn't have the nerves to also fix the LitmusC parser in this PR, so I just put comments on all its weird behaviors :).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
ThomasHaas: The implications are that the subsequent check for a constant value will fail, and the parser throws an error that it cannot statically determine with which thread to join (i.e. dynamic thread joining is not allowed).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Local.java
ThomasHaas: Yeah, will fix it

PR #480, opened 2023-07-02
Use p.getArch() for litmus when target option is not given

hernanponcedeleon:
The documentation says that the `target` option is optional for litmus tests. However, instead of inferring the arch from the test header, we were always using the default value `C11`. This PR fixes this by injecting the option in the configuration (unless the user has given a specific value).

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ You can also run Dartagnan from the console:
 ```
 java -jar dartagnan/target/dartagnan-3.1.1.jar <CAT file> [--target=<arch>] <program file> [options]
 ```
-For programs written in `.c` and `.bpl`, value `<arch>` specifies the programming language or architectures to which the program will be compiled. For program written in `.litmus` format, if the `--target` option is not given, dartagnan will automatically extract the `<arch>` from the litmus test header. `<arch>` must be one of the following: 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,7 +90,7 @@ You can also run Dartagnan from the console:
 ```
 java -jar dartagnan/target/dartagnan-3.1.1.jar <CAT file> [--target=<arch>] <program file> [options]
 ```
-For programs written in `.c` and `.bpl`, value `<arch>` specifies the programming language or architectures to which the program will be compiled. Program written in `.litmus` format do not require such option. `<arch>` must be one of the following: 
+For programs written in `.c` and `.bpl`, value `<arch>` specifies the programming language or architectures to which the program will be compiled. For program written in `.litmus` format, if the `--target` option is not given, dartagnan will automatically extract the `<arch>` from the litmus test header. `<arch>` must be one of the following:
```
ThomasHaas, line 93: Typo: "For program_s_ written in ..."
Also, I think we should capitalize "Dartagnan" since it is a proper name.

ThomasHaas (changes requested):

ThomasHaas:
Is the PR intended to target `master` directly? I don't mind, but I just wanna be sure.

hernanponcedeleon:
> Is the PR intended to target `master` directly? I don't mind, but I just wanna make sure.

This can cause Dartagnan to return wrong results so I think should immediately be merged into `master`. Once that is done I will merge `master` into `development`.

ThomasHaas (approved):
LGTM

PR #481, opened 2023-07-04
Add thread-creation-related C programs and tests

ThomasHaas:
This PR adds 4 new C programs that contain chaining of thread creations and inlining/outlining of pthread functions.
Only one of the programs has become a unit test, the other three are wrongly treated on either `development` or `development_rework`.
`development_rework` throws errors on those three tests, while `development` handles two of them correctly but gives a wrong result on the third one.


hernanponcedeleon:
If we know that our current implementation is broken, what is the point of adding the C files to the repo, but not a fix?
If it is for "documentation" purposes, i.e. have examples that show the implementation is broken, it would be better to document this as an issue and add the code to the repo once we have a fix.

ThomasHaas:
The point is that we need examples that show problems so that we can systematically fix them, i.e., we need to test our fixes somehow. For at least one of the examples, there is no way to "just add a fix" because it (likely) requires moving the whole thread creation/inlining out of the parser.
While I can keep the code in issues, it just makes it harder to work with. I would rather have the files directly (in`rework`, not in `development`), so that I and René can do proper testing and once a benchmark works as expected, we add it permanently to the unit tests so that we do not break it again.
Also I don't know if tracking issues of a (somewhat unstable) work-in-progress branch is really worthwhile.

hernanponcedeleon:
Fair enough. Feel free to merge then.

PR #483, opened 2023-07-05
Inlining & Threading (rework)

xeren:
Moves inlining of boogie code into a new `ProgramProcessor`.

EDIT by @ThomasHaas : 
This PR moves static memory initialization, inlining, and thread creation from the parser into processing passes.
All thse passes run immediatly after parsing, meaning that the overall behaviour should not have changed.
There is a minor improvement in handling of `pthread_join` allowing us to handle slightly more cases.
As a side effect, the 4 pthread events `Create`, `Start`, `Join` and `End` are also gone now.

Work to do in future PRs:
- Move thread creation to after unrolling. This requires a lot of changes in our passes to work on functions rather than threads. We (likely) also need a notion of `LocalId` in addition to `GlobalId`
- We need to understand what kind of history-tracking we want to do, e.g., do we need an `InlineId` now? More generally, we need to look into metadata tracking with our more intricate program processing. I think tracking of `originalId` is also broken for now.
- More refactoring/cleanup.


dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
```diff
@@ -47,17 +52,25 @@ public Function(String name, FunctionType type, List<String> parameterNames, int
         }
     }
 
+    @Override
+    public Type getType() {
+        return TypeFactory.getInstance().getArchType();
```
hernanponcedeleon, line 57: why do we fallback to the default arch type for the type of a function?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -89,6 +89,7 @@ private void run(Program program) {
         for (MemoryCoreEvent e : memEvents) {
             processLocs(e);
         }
+        //FIXME: Add handling for thread parameters (or get rid of this class)
```
hernanponcedeleon, line 92: @ThomasHaas you mentioned to me the other day some bug related to passing arguments and alias analysis ... how was this solved?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -0,0 +1,187 @@
...
+
+import static com.dat3m.dartagnan.configuration.OptionNames.RECURSION_BOUND;
+import static com.dat3m.dartagnan.program.event.EventFactory.*;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+@Options
+public class Inlining implements ProgramProcessor {
+
+    @Option(name = RECURSION_BOUND,
+            description = "Inlines each function call up to this many times.",
```
hernanponcedeleon, line 38: I assume it is not "up to", but exactly this many times

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Override
+    public void run(Program program) {
+        for (Function function : program.getFunctions()) {
+            inlineAllCalls(function);
+        }
+    }
+
+    private void inlineAllCalls(Function thread) {
```
hernanponcedeleon, line 57: is the parameter a thread or a function?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -0,0 +1,187 @@
...
+                expression.ifPresent(iExpr -> inlinedBody.add(newLocal(result, iExpr)));
+                inlinedBody.add(newGoto(exitLabel));
+            } else {
+                Event copy = functionEvent.getCopy();
+                inlinedBody.add(copy);
+                replacementMap.put(functionEvent, copy);
+            }
+        }
+
+        //
```
hernanponcedeleon, line 125: add a comment or remove

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -0,0 +1,376 @@
...
+        config.inject(creation);
+        return creation;
+    }
+
+    @Override
+    public void run(Program program) {
+        if (program.getFormat().equals(Program.SourceLanguage.LITMUS)) {
+            return;
+        }
+        // TODO: test code
```
hernanponcedeleon, line 94: Is this still test code?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -0,0 +1,376 @@
...
+            program.addThread(thread);
+
+            final Map<Expression, Expression> tid2ComAddrMap = new HashMap<>();
+            for (DirectFunctionCall call : thread.getEvents(DirectFunctionCall.class)) {
+                final List<Expression> arguments = call.getArguments();
+                switch (call.getCallTarget().getName()) {
+                    case "pthread_create" -> {
+                        assert arguments.size() == 4;
+                        final Expression pidResultAddress = arguments.get(0);
+                        //final Expression attributes = arguments.get(1);
```
hernanponcedeleon, line 126: do we need the attributes for anything?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/LlvmProcedures.java
```diff
@@ -107,6 +110,7 @@ public static void handleLlvmFunction(VisitorBoogie visitor, Call_cmdContext ctx
                             throw new UnsupportedOperationException("Operation " + params.get(3).getText() + " is not recognized.");
                 }
                 visitor.addEvent(Llvm.newRMW(reg, p0, p1, op, mo));
+                return true;
```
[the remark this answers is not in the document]
ThomasHaas: I will remove it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
ThomasHaas: This one is actually not the `FunctionType`, but the expression type. If we use a function as an expression, then we always mean the function's address, i.e., you could do `ptr r = myFunc; call r;`. For all addresses we use the arch type for now, but will eventually use a dedicated pointer type.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
ThomasHaas: Probably. I have not touched this class, and we actually have not tested recursive code at all.
I will try to come up with some simple recursion test.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
ThomasHaas: I could have sworn that I removed this line already... it's dead code now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
ThomasHaas: A function now. Will rename

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
```diff
@@ -36,108 +29,36 @@ public class PthreadsProcedures {
...
             case "pthread_create":
-                pthread_create(visitor, ctx);
-                break;
             case "__pthread_join":
             case "pthread_join":
-                // VisitorBoogie already took care of creating the join event
-                // when it parsed the previous load.
-                pthread_join(visitor, ctx);
-                break;
+                // These are handled by a dedicated pass
```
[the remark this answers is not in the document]
ThomasHaas: Yes, we can remove it.
The question is if you get any more clear by replacing
```
            case "pthread_create":
            case "__pthread_join":
            case "pthread_join":
                // These are handled by a dedicated pass
                return false;
```
with a single comment somewhere.
```
// pthread_create and pthread_join are handled by a dedicated pass
```
The former code makes it very clear that those `pthread` functions have been considered and were deliberately skipped.

Btw. we can also remove all the static lists like `LLVMPROCEDURES' now, but I've kept them around for documentation. If you want, I can delete them all (they are unused for the most part).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
ThomasHaas: Recursion handling is indeed broken. Not sure if we should fix it in this PR though.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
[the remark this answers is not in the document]
hernanponcedeleon: > Yes, we can remove it. The question is if you get any more clear by replacing
> 
> ```
>             case "pthread_create":
>             case "__pthread_join":
>             case "pthread_join":
>                 // These are handled by a dedicated pass
>                 return false;
> ```
> 
> with a single comment somewhere.
> 
> ```
> // pthread_create and pthread_join are handled by a dedicated pass
> ```
> 
> The former code makes it very clear that those `pthread` functions have been considered and were deliberately skipped.
> 

I would go with the single line comment

> Btw. we can also remove all the static lists like `LLVMPROCEDURES` now, but I've kept them around for documentation. If you want, I can delete them all (they are unused for the most part).

While having them at the top make it easier to see what is handled by the class, I would get rid of them.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
hernanponcedeleon: Can we make use of the same fix? i.e. is this an easy fix?

While we normally do not use this option, I would prefer to keep it in case we eventually need it (assuming maintaining the extra cost is not too much burden)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
hernanponcedeleon: Since this broken code does not affect the current status, we can fix in a different PR.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
ThomasHaas: At first glance, the code is quite different and I don't know how to fix it easily. Maybe René knows the code better and can tell, but from my perspective the fix is not trivial.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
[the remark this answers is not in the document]
ThomasHaas: Ok, then I remove them.

hernanponcedeleon:
LGTM

PR #485, opened 2023-07-11
Inlining of nondet_int

xeren:
In this PR, the family of `__VERIFIER_nondet_int` is not inlined by the parser, anymore.  Instead a program processor `IntrinsicsInlining` is issued after unrolling and should remove all remaining calls to functions defined by the verifier (currently just the `nondet` family).  The boogie parser, on the other hand, ignores definitions of said functions.  This happens so that the `Inlining` process does not get precedence over `IntrinsicsInlining`.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/IntrinsicsInlining.java
```diff
@@ -0,0 +1,117 @@
...
+    public void run(Program program) {
+        constantId = 0;
+        for (Thread thread : program.getThreads()) {
+            run(thread);
+        }
+    }
+
+    private void run(Thread thread) {
+        for (Event event : List.copyOf(thread.getEvents())) {
+            if (!(event instanceof DirectFunctionCall call)) {
```
hernanponcedeleon, line 47: Can we simply call `getEvents()` with parameter `DirectFunctionCall` to only get instances of that class?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/IntrinsicsInlining.java
```diff
@@ -0,0 +1,117 @@
...
+            }
+            assert !call.getCallTarget().hasBody();
+
+            List<Event> replacement = switch (call.getCallTarget().getName()) {
+                case "__VERIFIER_nondet_bool",
+                        "__VERIFIER_nondet_int", "__VERIFIER_nondet_uint", "__VERIFIER_nondet_unsigned_int",
+                        "__VERIFIER_nondet_short", "__VERIFIER_nondet_ushort", "__VERIFIER_nondet_unsigned_short",
+                        "__VERIFIER_nondet_long", "__VERIFIER_nondet_ulong",
+                        "__VERIFIER_nondet_char", "__VERIFIER_nondet_uchar" -> inlineNonDet(call);
+                default -> throw new UnsupportedOperationException(
```
hernanponcedeleon, line 58: I don't get this. Isn't this matching every function call (which we need!) that is not a `__VERIFIER_nondet` one?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/IntrinsicsInlining.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    private List<Event> inlineNonDet(DirectFunctionCall call) {
+        TypeFactory types = TypeFactory.getInstance();
+        ExpressionFactory expressions = ExpressionFactory.getInstance();
+        assert call instanceof DirectValueFunctionCall;
+        Register register = ((DirectValueFunctionCall) call).getResultRegister();
+        String name = call.getCallTarget().getName();
+        String prefix = name.startsWith("__V") ? "__VERIFIER_nondet_" : "__SMACK_nondet_";
```
hernanponcedeleon, line 73: in line 57, we only call `inlineNonDet` for `__VERIFIER` methods ... why do we need this conditional?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/IntrinsicsInlining.java
ThomasHaas: Btw. `thread.getEvents()` returns a fresh copy, so List.copyOf is not needed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/IntrinsicsInlining.java
ThomasHaas: It is matching every other call. This pass runs after regular inlining and thread creation, so that all calls except for intrinsic one's are resolved. The pass then tries to resolve all the intrinsics and if there is one that it cannot resolve, it will throw an exception, which is fine.
While I don't think that this single pass will be able to resolve all intrinsics in the future, i.e., we will likely get rid of the exception throwing, at least for now it is correct.

hernanponcedeleon (approved):
LGTM

PR #486, opened 2023-07-12
Add dotnet-5.0 to Dockerfile

xeren:
This PR attempts to bypass a removed dependency of SMACK, which currently causes the docker image build to fail.


ThomasHaas:
Can we close this?

xeren:
Yes, we can close this.  The dependency for `dotnet` has been removed from the target branch.

PR #487, opened 2023-07-13
Various fixes/cleanup

ThomasHaas:
This PR adds minor fixes and cleanup:
- `ExecutionStatus` is printed as `not_exec(...)` instead of `status(...)` to more clearly convey its semantics.
- `ExecutionStatus` now can assign to boolean registers. If possible, compilation schemes were updated to use boolean registers. Also, in many cases the `ExecutionStatus` was removed from compilation cause it served no purpose (it was dead code).
- Fixed an issue with `DirectFunctionCall.copy()` where metadata was not copied over. This caused `SourceLocation` information to get lost.
- More generally, the `Inlining` pass preserves more metadata when inlining code. The `FunCall` and `FunRet` code annotations have been renamed to `FunCallMarker` and `FunReturnMarker` to better distinguish them from actual calls/returns.
- `ThreadCreation` now assigns `OriginalId` (which was previously done after parsing).
-  The logging of `RelationAnalysis` was cleaned up. 
- Fixed an issue where `ExecutionModel` would complain about uninitialized registers if they were assigned by a `ThreadArgument`.
- Changed witness generation to include threading-related events.


dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
```diff
@@ -89,10 +87,9 @@ public void generateGraphOfExecutionModel(Writer writer, String graphName, Execu
     }
 
     private boolean ignore(EventData e) {
-        return !e.getEvent().hasMetadata(SourceLocation.class) && !e.isInit();
+        return false; // We ignore no events for now.
```
hernanponcedeleon, line 90: isn't this affection our current witness generation?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
```diff
@@ -14,11 +15,13 @@
 
 public final class Relation implements Dependent<Relation> {
 
+    private final Wmm wmm;
```
hernanponcedeleon, line 18: Why do we want to store the Wmm associated to a relation?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AbstractEvent.java
```diff
@@ -33,7 +34,6 @@ protected AbstractEvent() {
     protected AbstractEvent(AbstractEvent other) {
         copyAllMetadataFrom(other);
         this.tags = other.tags; // TODO: Dangerous code! A Copy-on-Write Set should be used (e.g. PersistentSet/Map)
-        this.function = other.function;
```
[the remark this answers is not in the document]
ThomasHaas: The function? Cause when you copy an event, it is nowhere in the control-flow until you insert the copy. 
So technically, the copy should not belong to any function.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
ThomasHaas: I used this to be able to move configuration data into the `Wmm`.
More precisely, `Acyclicity` had this `reduceAcyclicityEncodeSet` option, which I moved into `Wmm` cause (1) we probably never want to configure each `Acyclicity` axiom separately, and (2) we can log this setting once instead of per axiom.
But now `Acyclicity` needs to know which `Wmm` it belongs to in order to retrieve the configuration data. By attaching the `Wmm` to `Relation`, every `Constraint/Axiom` is able to access the `Wmm`.
It is similar to how `Event` knows which `Function/Program` it belongs to.

While I don't like configuring a data object like `Wmm`, it was the easiest solution for now. To somewhat avoid the issue of configurating the `Wmm` directly, I also added the inner `Config` class which holds the data.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/boogie/PthreadsProcedures.java
```diff
@@ -15,12 +15,15 @@ public static boolean handlePthreadsFunctions(VisitorBoogie visitor, Call_cmdCon
...
             case "pthread_cond_broadcast":
-            case "pthread_exit":
             case "pthread_mutex_destroy":
                 // TODO: These are skipped for now
                 VisitorBoogie.logger.warn(
                         "Skipped call to {} because the function is not supported right now.", funcName);
                 return true;
+            case "pthread_exit":
+                final Expression retVal = (Expression) ctx.call_params().exprs().expr(0).accept(visitor);
+                visitor.addEvent(EventFactory.newFunctionReturn(retVal));
```
hernanponcedeleon, line 25: is it true that a `pthread_exit` acts simply as a function return?

ThomasHaas: In principle it is reversed: a return acts as a `pthread_exit` I believe. Not that it really matters: both terminate the thread and produce a value that can be read by `pthread_join`.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/visualization/ExecutionGraphVisualizer.java
hernanponcedeleon: Some addresses get a `null` representation in graphviz witnesses, see e.g. ticketlock with `-DACQ2RX`

ThomasHaas: Hmm, the `NULL` entries should already have been there before.
It seems we simply do not construct init events for the members of the `lock` variable, cause Smack fails to init properly.
That problem should already exist on `development`. 

While debugging, I noticed that we now do create a lot of memory objects related to smack internals.
I will try to fix this.

ThomasHaas: I fixed both issues. We now assume that Smack generated correct static initialization code IFF it generated at least one init value. This might be wrong if there is the possibility that Smack generates some but not all initialization of a memory object (i.e., it generates partial initialization).

hernanponcedeleon:
LGTM

PR #492, opened 2023-07-27
Function passes & Thread creation after unrolling

ThomasHaas:
This PR adds `FunctionProcessor` (maybe rename to `FunctionPass`?) as a new processor type.
Many `ProgramProcessors` have been changed to `FunctionProcessors`. A few passes have remained as `ProgramProcessor` just to generate logging output for the whole program(*).

`ThreadCreation` has been moved past most other processing passes, in particular past `LoopUnrolling`. This allows us to handle `pthread_create` inside (dynamic) loops. However, `pthread_join` in such loops still fails due to improper handling of its semantics.

The PR targets `development_rework` rather than `development`, because there are regressions in logging and coverage reports. Coverage should be fixable, but for logging we need to discuss what we actually want to log.

(*) In order to combine logging with function-level passes, we might want to allow for metadata attachment to functions or have a general statistics collector object. Then passes can generate information per function (possibly as metadata) and dedicated logging passes can collect, aggregate, and log the information.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorBase.java
```diff
@@ -29,25 +29,26 @@
 
 class VisitorBase implements EventVisitor<List<Event>> {
 
-    protected boolean forceStart;
```
hernanponcedeleon, line 32: How is force start handled now?

ThomasHaas: It is in `ThreadCreation` as that pass is responsible for setting up all thread-related code.

benchmarks/c/miscellaneous/nondet_loop.c
```diff
@@ -1,9 +1,11 @@
 #include <stdint.h>
 #include <assert.h>
+#include "dat3m.h"
```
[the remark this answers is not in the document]
ThomasHaas: I will change it.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -57,6 +57,8 @@
 import static java.lang.Boolean.TRUE;
 import static java.lang.String.valueOf;
 
+;
```
hernanponcedeleon, line 60: remove

dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
```diff
@@ -123,12 +123,12 @@ public Collection<INonDet> getConstants() {
     /**
      * Iterates all events in this program.
      *
-     * @return {@code cId}-ordered complete sequence of all events in this program.
+     * @return {@code globalId}-ordered complete sequence of all events in this program.
      */
-    public List<Event> getEvents() {
+    public List<Event> getThreadEvents() {
```
hernanponcedeleon, line 128: The name suggest this is expected to be called after threads were built. What if the developer calls this before? If we intended to be used only after building threads, we just add some check to detect wrong usage

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
```diff
@@ -38,12 +38,14 @@ public void run(Program program) {
...
 
     private void processMallocs(Program program) {
-        for (Malloc malloc : program.getEvents(Malloc.class)) {
-            final MemoryObject allocatedObject = program.getMemory().allocate(getSize(malloc), false);
-            final Local local = EventFactory.newLocal(malloc.getResultRegister(), allocatedObject);
-            local.addTags(Tag.Std.MALLOC);
-            local.copyAllMetadataFrom(malloc);
-            malloc.replaceBy(local);
+        for (Thread thread : program.getThreads()) {
+            for (Malloc malloc : thread.getEvents(Malloc.class)) {
```
hernanponcedeleon, line 42: can you use `program.getThreadEvents`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -109,10 +110,13 @@ private ProcessingManager(Configuration config) throws InvalidConfigurationExcep
                 IntrinsicsInlining.fromConfig(config),
                 MemoryAllocation.newInstance(),
                 // --- Statistics + verification ---
-                EventIdReassignment.newInstance(), // Normalize used Ids (remove any gaps)
-                printAfterProcessing ? DebugPrint.withHeader("After processing") : null,
-                CoreCodeVerification.fromConfig(config),
-                LogProgramStatistics.newInstance()
+                IdReassignment.newInstance(), // Normalize used Ids (remove any gaps)
+                printAfterProcessing ? DebugPrint.withHeader("After processing", Printer.Mode.THREADS) : null,
```
hernanponcedeleon, line 114: Is this intended to replace #491 ?

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
```diff
@@ -34,17 +40,29 @@ public String print(Program program){
...
 
-        for (Function function : program.getFunctions()) {
-            if (function instanceof Thread) {
-                continue;
+        if (mode == Mode.FUNCTIONS || mode == Mode.ALL) {
+            for (Function function : program.getFunctions()) {
+                if (function instanceof Thread) {
+                    continue;
+                }
+                appendFunction(function);
```
hernanponcedeleon, line 64: Can't you simply iterate over all function and if mode is not function, skip those that are not threads?

dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
ThomasHaas: I can add a check. If someone calls it before, an empty list gets returned right now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
ThomasHaas: Actually, this code used to call `program.getThreadEvents` but I changed it in an attempt to get rid of that function. It turned out to be quite difficult to replace, so I might as well revert my changes on that one.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
ThomasHaas: Partly yes. We still might want to delete functions if they end up consuming too much memory, but just deleting them for the sake of better printing seemed a little overkill, so I made the `Printer` more smart.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
ThomasHaas: Actually, `program.getFunctions` does not return `Threads`. I wasn't yet sure if I want threads to be in that list or not, so I put a check into `Printer` so it works correctly in either case.
But even if regular functions and threads were in that list, it would be hard to print a replacement for the skipped init threads

hernanponcedeleon:
LGTM

PR #493, opened 2023-08-01
LLVM grammar

xeren:
Allows Dartagnan to parse basic features of LLVM IR.  This includes
- Functions, calls and returns (static)
- Aggregates, arrays (partial support, sufficient for cmpxchg)
- Jumps, Phi nodes (parsed into register-based form)
- Memory instructions, RMW, cmpxchg, mallocs
- Integer arithmetics and bit vector operations
- Basic type conversions

Not included:
- ~Metadata~
- ~Memory alignment, proper type sizes~
- Exception handling
- ~Virtual functions~

EDIT by @ThomasHaas: 
- Added partial support for metadata (enough to get source location information)
- Added proper type sizes with alignment considerations.
- Added simple support for virtual functions.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RegisterDecomposition.java
```diff
@@ -0,0 +1,83 @@
...
+                        //TODO recur if no leaf type
+                    }
+                    local.replaceBy(componentAssignments);
+                }
+            }
+            for (Event event : function.getEvents()) {
+                if (event instanceof RegReader reader) {
+                    reader.transformExpressions(transformer);
+                }
+            }
```
ThomasHaas, line 63: `function.getEvents(RegReader.class)`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RegisterDecomposition.java
```diff
@@ -0,0 +1,83 @@
...
+        return new RegisterDecomposition();
+    }
+
+    @Override
+    public void run(Program program) {
+        final var transformer = new ExtractSubstitutor();
+        for (Function function : program.getFunctions()) {
+            //TODO assume that no constructor occurs anywhere other than as the root operation of a local
+            for (Event event : function.getEvents()) {
+                if (event instanceof RegWriter writer &&
```
ThomasHaas, line 39: `function.getEvents(RegWriter.class)`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RegisterDecomposition.java
```diff
@@ -0,0 +1,83 @@
...
+                if (event instanceof RegReader reader) {
+                    reader.transformExpressions(transformer);
+                }
+            }
+        }
+    }
+
+    private record RegisterIndex(Register origin, int index) {}
+
+    // Visitor methods should return null if there was no structural
```
ThomasHaas, line 69: Incomplete comment

hernanponcedeleon:
I added a few more llvm files for which the current PR fails (mostly parsing, but also some problems with pthread_create/join)

ThomasHaas:
This branch has more errors:
- LLVM can generate "true"/"false" for `1/0 bv1`. The parser seems to parse these as `NULL` (happens mainly in `xor` instructions).
- LLVM type names can be used before their declaration:
   ```
   %struct.WorkStealQueue = type { %union.pthread_mutex_t, i32, i32, i32, i32, [16 x %struct.Obj*], i32 }
   %union.pthread_mutex_t = type { %struct.__pthread_mutex_s }
   ```
It seems the parser throws an error when parsing the first type, because it refers to an "undefined named type" (which is defined directly after).

dartagnan/src/test/java/com/dat3m/dartagnan/c/LLVMTest.java
```diff
@@ -76,11 +82,7 @@ protected long getTimeout() {
     }
 
     protected Provider<Integer> getBoundProvider() {
-        int bound = name == null ? 1 : switch (name) {
-            case LFDS_CHASELEV, LFDS_DGLM, LFDS_HASHTABLE, LFDS_MICHAELSCOTT, LFDS_SAFESTACK, LFDS_TREIBER, LFDS_WSQ -> 2;
-            default -> 1;
-        };
-        return Provider.fromSupplier(() -> bound);
+        return Provider.fromSupplier(() -> bound[0]);
```
ThomasHaas, line 85: The issue with your previous code was just that you put the computation of the bound outside of the provider, which caused it to get evaluated at the wrong time. The following would have worked:
```
return () -> {
     return name == null ? 1 : switch (name) {
            case LFDS_CHASELEV, LFDS_DGLM, LFDS_HASHTABLE, LFDS_MICHAELSCOTT, LFDS_SAFESTACK, LFDS_TREIBER, LFDS_WSQ -> 2;
            default -> 1;
        };
}
```

Also, in your newer version I don't think the array is needed. A simple `int` is enough.

hernanponcedeleon:
It seems we know print the result as for litmus tests even if we pass llvm files
```
> java -jar dartagnan/target/dartagnan-3.1.1.jar cat/imm.cat --target=imm output/paper-E3.1.ll 
...
[05.08.2023] 09:29:54 [INFO] RefinementSolver.run - Verification finished with result PASS
Condition forall (assert_0)
Ok
Total verification time(ms): 718
```

hernanponcedeleon:
We might need to add support for parsing dummy assembly like this
```
__asm__ __volatile__ (""   : : : "memory");
```

ThomasHaas:
> It seems we know print the result as for litmus tests even if we pass llvm files

Yes, that's because some code checks for `!= SourceLanguage.BOOGIE`, which also matches the new `SourceLanguage.LLVM`. I suggest to get rid of `BOOGIE` and just have `LLVM`, because even the boogie code is just meant represent LLVM in the first place.

> We might need to add support for parsing dummy assembly like this
```__asm__ __volatile__ (""   : : : "memory");```

Is this just a noop or a compiler barrier? Or what are we supposed to do with it?

**Comment by hernanponcedeleon** [comment]
The assembly indeed comes from a compiler barrier in libvsync. If the barrier is important for a language level MM, then we will have some header to handle this (as we currently do for Lkmm). If the assembly made it all the way to the llvm file, then this is simple a nop (the problem right now is that the visitor throws an exception)

**Comment by ThomasHaas** [comment]
I merged both #506 (with the change to 8-byte aligned funtion pointers) and #507 into this PR.

File: dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java [hunk 5]
```diff
@@ -39,6 +41,9 @@ public class Inlining implements FunctionProcessor {
             secure = true)
     private int bound = 1;
 
+    // This map is updated when a new program is passed to this instance.
+    private final Map<Function, CacheEntry> cache = new HashMap<>();
+
```
ThomasHaas, line 46: Is the cache necessary for correctness (maybe because you need to cache the original state of a function)?
If so, then maybe this pass should be changed back to a `ProgramProcessor` so it can properly traverse all functions before transformation.
Because as it stands, the cache has two flaws:
- Rerunning the pass on the same program can easily break if additional function calls get inserted afterwards (by some other pass).
- The cache can take up memory indefinitely (depending on whether references to `Inlining` survive). A potential fix would be to use a weak hash map (or some Guava cache with eviction policies), though it has the potential to evict entries during processing which would break the pass. Using a Guava cache with time-based eviction policy would be workable in practice though.

xeren: What was bugging me while testing was that `Inlining` was not parallelizable along the functions.  Maybe it should indeed be a `ProgramProcessor`.  For indirect recursion, the sequencing of `Inlining` means that the minimal recursion bound for finding a bug differs by 1 based on the definition order in the program.  Also, I could not find a better naming scheme for registers and labels.  `Inlining` names registers like `1:1:6:3:2:r8` to avoid name conflicts.
The version with caches helped me with debugging, but now I managed to fix the remaining bugs.  Along with the cache, I made use of an event tag to exclude writers to the result register of the call when renaming registers.  The new commit got rid of both again.

ThomasHaas:
I think you should run the usual llvm optimization passes, including `mem2reg` on the recursion llvm code you use for the unit test. Cause recursion + non-optimized code leads to many memory accesses making the processing/analysis expensive (the verification time is almost nothing).

benchmarks/c/miscellaneous/recursion.c
```diff
@@ -0,0 +1,70 @@
...
+    atomic_store_explicit(&x, a, memory_order_relaxed);
+    int result = atomic_load_explicit(&x, memory_order_relaxed);
+    pthread_join(t0, NULL);
+#ifdef ACKERMANN_LARGE
+    //ack(ack(1,1),ack(1,1)) needs a recursion bound of 64.
+    assert(result != 61);
+#else
+#ifdef ACKERMANN_SMALL
+    //ack(ack(0,0),ack(0,0) needs recursion bound of 4 and mem2reg after unrolling to be feasible.
+    assert(result != 3);
```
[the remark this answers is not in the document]
xeren: Yes, doing the tests for PASS would be better.
Since inlining happens in one phase and in-place, recursive functions behave complicatedly:  Each recursive call copies the current state of the function into itself.  This means that the resulting depth in recursive functions grows exponentially with the provided bound.  This particular test has the following call tree.  For `fibonacci(2)`, there is apparently already enough program logic to avoid all bound events.
```
fibonacci (depth = 0)
  fib (depth = 0)
    fibonacci (depth = 1)
      fib (already inlined)
        fibonacci (depth = 2, BOUND)
      fibonacci (depth = 2, BOUND)
  fibonacci (depth = 1)
    fib (already inlined)
      fibonacci (already inlined)
        fib (already inlined)
          BOUND
        BOUND
    fibonacci (depth = 2, BOUND)
```

xeren:
I created the recursion test with `clang -O0` and manually optimized it with `opt -mem2reg -sroa -early-cse -indvars -loop-unroll -simplifycfg -gvn -loop-simplify -instcombine`.  I hope that matches with the other tests.

benchmarks/c/miscellaneous/recursion.c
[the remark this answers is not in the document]
ThomasHaas: Because of this exponential blowup I thought you introduced the cache originally so that you just inline the original state of the function repeatedly (this of course requires this pass to be a program pass rather than a function pass). I think we should avoid exponential blowup, but it is fine for now cause it's still better than not supporting recursion at all.

ThomasHaas:
Yes, that's fine. We usually do not use `-instcombine` but you can use it. Btw. our readme states all the passes usually do:
`export ATOMIC_REPLACE_OPTS="-mem2reg -sroa -early-cse -indvars -loop-unroll -fix-irreducible -loop-simplify -simplifycfg -gvn"`

xeren:
Both lists of passes produce the same assembly for `recursion.ll`.  So the current version should already fit in with the other tests.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -124,12 +130,20 @@ public static Init newInit(MemoryObject base, int offset) {
...
 
-    public static DirectVoidFunctionCall newVoidFunctionCall(Function function, List<Expression> arguments) {
-        return new DirectVoidFunctionCall(function, arguments);
+    public static ValueFunctionCall newValueFunctionCall(Register resultRegister, FunctionType funcType,
+                                                         Expression funcPtr, List<Expression> arguments) {
+        return new ValueFunctionCall(resultRegister, funcType, funcPtr, arguments);
+    }
+
+    public static VoidFunctionCall newVoidFunctionCall(Function function, List<Expression> arguments) {
+        return new VoidFunctionCall(function.getFunctionType(), function, arguments);
```
hernanponcedeleon, line 143: Does this guarantee that the return type is actually `void`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
```diff
@@ -10,7 +11,7 @@ public class LKMMLock extends SingleAccessMemoryEvent {
     public LKMMLock(Expression lock) {
         // This event will be compiled to LKMMLockRead + LKMMLockWrite
         // and each of those will be assigned a proper memory ordering
-        super(lock, "");
+        super(lock, TypeFactory.getInstance().getIntegerType(32), "");
```
hernanponcedeleon, line 14: Are you sure spinlocks in the kernel are 32 bits?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -0,0 +1,784 @@
...
+            new Info("__VERIFIER_assert", false, false, false, false, this::inlineAssert),
+            new Info("__VERIFIER_nondet",
+                    List.of("__VERIFIER_nondet_bool",
+                            "__VERIFIER_nondet_int", "__VERIFIER_nondet_uint", "__VERIFIER_nondet_unsigned_int",
+                            "__VERIFIER_nondet_short", "__VERIFIER_nondet_ushort", "__VERIFIER_nondet_unsigned_short",
+                            "__VERIFIER_nondet_long", "__VERIFIER_nondet_ulong",
+                            "__VERIFIER_nondet_char", "__VERIFIER_nondet_uchar"),
+                    false, false, true, false, this::inlineNonDet),
+            // --------------------------- LLVM ---------------------------
+            new Info("llvm.*",
```
hernanponcedeleon, line 174: Why do we have `*` here as "matching everything" but not in other ones have several variants? e.g. nondet

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -0,0 +1,784 @@
...
+        } else if (name.startsWith("llvm.ctlz")) {
+            return inlineLLVMCtlz(valueCall);
+        } else if (name.startsWith("llvm.ctpop")) {
+            return inlineLLVMCtpop(valueCall);
+        } else if (name.contains("add.sat")) {
+            return inlineLLVMSaturatedAdd(valueCall);
+        } else if (name.contains("sub.sat")) {
+            return inlineLLVMSaturatedSub(valueCall);
+        } else if (name.startsWith("llvm.smax") || name.startsWith("llvm.smin")
+                || name.startsWith("llvm.umax") || name.startsWith("llvm.umin")) {
```
hernanponcedeleon, line 366: why `startsWith` and not `equal`? Do they have type postfixes?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -0,0 +1,784 @@
...
+            return inlineLLVMMinMax(valueCall);
+        } else {
+            final String error = String.format(
+                    "Call %s to LLVM intrinsic %s cannot be handled.", call, call.getCalledFunction());
+            throw new UnsupportedOperationException(error);
+        }
+    }
+
+    private List<Event> inlineLLVMCtlz(ValueFunctionCall call) {
```
hernanponcedeleon, line 375: don't we have an expression to deal with?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -0,0 +1,784 @@
...
+            ));
+        }
+        replacement.add(endCtlz);
+
+        return replacement;
+        /*final Expression result = expressions.makeCTLZ(input, (IntegerType) call.getResultRegister().getType());
+        return List.of(EventFactory.newLocal(call.getResultRegister(), result));*/
+    }
+
+    private List<Event> inlineLLVMCtpop(ValueFunctionCall call) {
```
hernanponcedeleon, line 407: Can you describe what the intrinsic is suppose to do?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -74,19 +74,26 @@ public class ProcessingManager implements ProgramProcessor {
...
-
+        final Intrinsics intrinsics = Intrinsics.newInstance();
+        final FunctionProcessor sccp = constantPropagation ? SparseConditionalConstantPropagation.fromConfig(config) : null;
         programProcessors.addAll(Arrays.asList(
-                printBeforeProcessing ? DebugPrint.withHeader("Before processing", Printer.Mode.ALL) : null,
+                intrinsics.markIntrinsicsPass(),
+                GEPToAddition.newInstance(),
+                RegisterDecomposition.newInstance(),
                 StaticMemoryInitializer.newInstance(),
+                printBeforeProcessing ? DebugPrint.withHeader("Before processing", Printer.Mode.ALL) : null,
```
hernanponcedeleon, line 84: shouldn't this be the very first pass?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -177,9 +180,15 @@ private void handlePthreadJoins(Thread thread, Map<IValue, Expression> tid2ComAd
...
-            if (!(targetName.equals("pthread_join") || targetName.equals("__pthread_join"))) {
+        for (FunctionCall call : thread.getEvents(FunctionCall.class)) {
+            if (!call.isDirectCall()) {
+                continue;
+            }
+            final String targetName = call.getCalledFunction().getName();
+            if (!(targetName.equals("pthread_join")
+                    // The following two calls can be generated on macOS
+                    || targetName.equals("__pthread_join")
+                    || targetName.equals("\"\\01_pthread_join\""))) {
```
hernanponcedeleon, line 191: can you somehow use the intrinsic Info that groups them all together?

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessBuilder.java
```diff
@@ -155,7 +156,8 @@ public WitnessGraph build() {
 				graph.addEdge(edge);
 
 				nextNode++;
-				if (e.hasTag(Tag.ASSERTION)) {
+				if (e instanceof Assert) {
+					// FIXME: Execution of an assertion is not a problem, if the assertion holds!
```
hernanponcedeleon, line 160: can you simply add `prover.getModel().evaluate(e.getExpression())`?

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
```diff
@@ -24,8 +25,15 @@ private GlobalSettings() {}
             secure = true)
     private int arch_precision = -1;
 
+    @Option(name = MIXED_TYPE,
+            description = "If false, all non-boolean, non-aggregate data types are unified into one common integer type determined by " + ARCH_PRECISION + ".",
+            secure = true)
+    private boolean mixedType = false;
```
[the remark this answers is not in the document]
ThomasHaas: We should definitely change the options. 
I would suggest to use `BV` by default with 64-bit precision for pointers and proper program types.
We can then have a single `useIntegers` option that makes everything integer.
We can get rid of `arch_precision` because everything apart from `64` is wrong for LLVM code (which often assumes pointers to be 64 bit) and the integer setting "`-1`" is better expressed by a separate option

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -56,6 +58,29 @@ public Expression makeBinary(Expression leftOperand, BOpBin operator, Expression
...
+        } else if (type == booleanType) {
+            return makeFalse();
+        } else {
+            throw new UnsupportedOperationException("Cannot create zero of type " + type);
+        }
+    }
+
     public IValue makeZero(IntegerType type) {
         return makeValue(BigInteger.ZERO, type);
     }
```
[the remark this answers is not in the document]
ThomasHaas: The return types are different. I think the specialization here is fine.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/TypeFactory.java
```diff
@@ -41,14 +43,68 @@ public IntegerType getIntegerType(int bitWidth) {
...
+    }
+
+    public AggregateType getAggregateType(List<Type> fields) {
+        checkNotNull(fields);
+        checkArgument(fields.stream().noneMatch(t -> t == voidType), "Void fields are not allowed");
+        return typeNormalizer.normalize(new AggregateType(fields));
+    }
+
+    public ArrayType getArrayType(Type element) {
+        return typeNormalizer.normalize(new ArrayType(element, -1));
```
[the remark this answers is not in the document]
ThomasHaas: Yeah, we should stick to one or the other. I think `-1` might be better cause that is definitely an invalid array size, whereas `0` could technically be valid.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -1,18 +1,1434 @@
...
+                    "dummy_" + (unnamedParameters++);
+            parameterNames.add(adjustRegisterName(parameterName));
+        }
+        final boolean isVarArgs = ctx.params().ellipsis != null;
+        final FunctionType functionType = types.getFunctionType(returnType, parameterTypes, isVarArgs);
+        //TODO there are more attributes in ctx
+        final var declaredFunction = new Function(name, functionType, parameterNames, functionCounter++, null);
+
+        program.addFunction(declaredFunction);
+        constantMap.put(name, declaredFunction);
```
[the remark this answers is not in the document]
ThomasHaas: It doesn't interfere at all. The map is used to replace references to function names by the concrete function.
If a call goes to a generic function pointer, than it will obviously not get replaced.
Some pass in the processing pipeline resolves calls to function pointers to direct calls, but that happens outside of the parser/visitor.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -1,18 +1,1434 @@
...
+        if (ctx.inlineAsm() != null) {
+            // FIXME: We ignore all inline assembly.
+            return null;
+        }
+        final Expression callTarget = checkPointerExpression(ctx.value());
+        if (callTarget == null) {
+            //FIXME ignores metadata functions, but also undeclared functions
+            return null;
+        }
+        //checkSupport(callTarget instanceof Function, "Indirect call in %s.", ctx);
```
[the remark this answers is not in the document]
ThomasHaas: The second comment? I think I forgot to remove it. That one used to check that there are no indirect calls (cause we didn't support them).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -1,18 +1,1434 @@
...
+
+    // ----------------------------------------------------------------------------------------------------------------
+    // Instructions producing a value
+
+    @Override
+    public Expression visitAllocaInst(AllocaInstContext ctx) {
+        // see https://llvm.org/docs/LangRef.html#alloca-instruction
+        final Register register = getOrNewCurrentRegister(pointerType);
+        //final var inalloca = ctx.inAllocaTok != null;
+        //final var swifterror = ctx.swiftError != null;
```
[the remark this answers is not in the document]
ThomasHaas: I haven't checked the context of those comments, but IIRC they are actually quite meaningful.
Those comments show that there is more information that we do not actually use.
Though I agree that we can remove thinks like `swiftError` cause we do not intend to support that anytime in the near future.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -1,18 +1,1434 @@
...
+        final Type elementType = parseType(ctx.type());
+        final Expression sizeExpression;
+        if (ctx.typeValue() == null) {
+            sizeExpression = expressions.makeOne(integerType);
+        } else {
+            final Type sizeType = parseType(ctx.typeValue().firstClassType());
+            sizeExpression = checkExpression(sizeType, ctx.typeValue().value());
+        }
+        //final int alignment = parseAlignment(ctx.align());
+        //final int addressSpace = parseAddressSpace(ctx.addrSpace());
```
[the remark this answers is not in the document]
ThomasHaas: This is probably about pointers, whose information (in particular the address space) we eventually need when dealing with GPUs.
I would keep those comments cause that information is actually exploitable, but we simply do not use it yet.
If you want, we could add a `TODO` on top of those comments.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -1,18 +1,1434 @@
...
+    // Conversions
+
+    @Override
+    public Expression visitTruncInst(TruncInstContext ctx) {
+        return conversionInstruction(ctx.typeValue(), ctx.type(), false);
+    }
+
+    @Override
+    public Expression visitZExtInst(ZExtInstContext ctx) {
+        return conversionInstruction(ctx.typeValue(), ctx.type(), false);
```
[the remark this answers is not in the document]
ThomasHaas: We use a single conversion expression for both extending and truncating. The difference is only the target type (if it is smaller, then we truncate, otherwise we extend).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
ThomasHaas: I think the constructor of `VoidFunctionCall` checks this.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
ThomasHaas: Didn't we have a discussion about this? Anyway, we mock the locks with our own header and that uses `int` internally. I also think that other sizes caused problems in some tests (not sure though)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: The meaning is different. `*` means that each listed variant is an incomplete name, because, e.g., all the llvm intrinsics will have a suffix like `.i32.i64`. These suffixes are not present in the `__VERIFIER_nondet` case, meaning that all listed names have to match exactly.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: Yes, see my reply above and why `*` was used.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: The expression might still be there, but I wanted to get rid of it since we can just implement it directly in our IR.
It would be strange to have `ctlz` as expression but not the other tons of llvm intrinsics. Similarly, I wouldn't want to add expressions for all the llvm intrinsics (if we don't need to)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: It counts the number of bits that are 1. Do you want to have explanations for all llvm intrinsics added as comments?
Cause we could also just reference the LLVM manual

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
ThomasHaas: Probably yes (minus potentially the `StaticMemoryInitializer`, but that one is only there for Boogie code).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
ThomasHaas: Yeah, I guess that should be doable.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessBuilder.java
ThomasHaas: Yes. But this code is true to the original code (which was already wrong) :)

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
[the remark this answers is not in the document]
ThomasHaas: We can just change our configuration for the unit tests to still use integers! No need to compromise the performance there.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: This class generates a long data dependency plus control flow which, as stated by a comment, it introduces dependencies that should not be there. The former might have performance impact wrt the one explicitly encoding `ctlz` semantics at the SMT level. The former impacts soundness of our analysis.

I think those two are sufficient reasons to treat this particular intrinsic differently than the rest.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: The other at least have the implementation of the intrinsic that I can use to check if the implementation is correct. This one has nothing. A reference to the LLVM manual would be enough.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessBuilder.java
hernanponcedeleon: the original code was only used with boogie which always moved the expression to the control flow and generated an `assert(false)`, thus the original code was (practically) sound while this one is not.

ThomasHaas: I don't think this is specific to boogie. LLVM generates the code such that it only has `assert(false)`. So it is still practically sound.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: What do you mean by the other one's have an implementation? All of them have an internal implementation.
You can also use `__builtin_ctpop` to generate the intrinsic in code and compare it against a manually implemented loop that counts the bits. I checked that this intrinsic works as expected on a few examples (I used the `ctlz.c` test and modified it).

hernanponcedeleon: I meant "an implementation as documentation" like [here](https://github.com/hernanponcedeleon/Dat3M/blob/9d1b472bc597e5f2a1a469e38fd006418d9e6593/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java#L445-L454)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: I agree with the ctrl-dep but that should get fixed by allowing jumps that do not produce dependencies.
The performance argument I only buy if you find a measurable difference in some code (even synthetic code would be fine).
Handling intrinsic `ctlz` expressions is not just a matter of adding some encoding in the backend, but you also need to provide handling in all processing, particular in CP. We get CP for free by giving an implementation.
Btw. ctpop suffers from the same long dependency chains.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: The implementation for the saturated operations is highly non-trivial, that's why I put it there. In particular, even if you know what saturated arithmetic does, it is not clear how to translate it properly. This is not the case for ctpop/ctlz, which are just straightforward bit-counting loops (i.e., once you know what they do, the implementation is trivial).
Anyway, I can write pseudo-code there, but then we should do this for all llvm intrinsics (at least for ctlz as well).

hernanponcedeleon: If the implementation is trivial, there is no need as long as we provide a link to the documentation. This way we have a reference of what the implementation should do

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: ok, but then the ctrl-dep problem should be fixed as part of this PR

ThomasHaas: Then I would suggest to use the `ctlz` expression for now (the code is still there, just commented out) and make the changes in a latter PR.
This PR is already big enough as it is.

hernanponcedeleon: We are still using the implementation (which introduces ctrl dependencies) rather than our own expression

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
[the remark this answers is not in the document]
hernanponcedeleon: These changes are not yet implemented

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -352,9 +376,7 @@ private List<Event> handleLLVMIntrinsic(FunctionCall call) {
         final ValueFunctionCall valueCall = (ValueFunctionCall) call;
         final String name = call.getCalledFunction().getName();
 
-        if (name.startsWith("llvm.assume")) {
-            return List.of(EventFactory.newAssume(call.getArguments().get(0)));
```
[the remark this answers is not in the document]
ThomasHaas: That one is handled by a different function. We might want to rename `handleLlvmIntrinsics` to `handleLlvmMathIntrinsics` to make it clear that not all of them are handled in this single function.

hernanponcedeleon:
There used to be some manual changes that were required when merging development (I think it was related to which parser to use, but maybe there was something else). We need to be sure we are not overriding some code

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -228,6 +229,7 @@ private void runInternal(SolverContext ctx, ProverEnvironment prover, Verificati
         logger.info("Checking target property.");
         prover.push();
         prover.addConstraint(propertyEncoder.encodeProperties(task.getProperty()));
+        prover.printAll(System.out);
```
ThomasHaas, line 232: You have committed debug code

benchmarks/c/miscellaneous/recursion.c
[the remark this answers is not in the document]
xeren: The current version now does inlining based on copies of the original functions.  This way, each call has a predictable replacement and does not grow exponentially anymore.  Recursion depth is already counted per-function, so non-recursive programs will still be completely expanded.  I set the default recursion bound to 1, so each function may appear at most once in the call stack of the checked program fragment.  A recursion bound of 0 would mean that every call is replaced with a bound marker.

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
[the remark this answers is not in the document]
xeren: They are implemented now.  A proper pointer type is planned for another PR.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -168,4 +175,99 @@ protected void saveFlaggedPairsOutput(Wmm wmm, WmmEncoder encoder, ProverEnviron
         }
     }
 
+    protected static class LoggingProver implements ProverEnvironment {
+
+        private final FormulaManager manager;
+        private final ProverEnvironment inner;
+        private final List<String> log = new ArrayList<>();
```
ThomasHaas, line 182: There should be a way to disable the logging, cause we generate huge formulas and having a full copy of the formula sitting in some logs seems to be wasteful. Indeed, I would prefer to have this disabled by default.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: Not anymore.  Calls to `llvm.ctlz` now produce our built-in `CTLZ` expression.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
xeren: The cache's lifetime is now completely inside `run(Program)`.  With this, `Inlining` should be reusable again.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
xeren: This was not intended for upstream.  I have removed it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: How? The code appears to be unchanged

.github/workflows/maven.yml
```diff
@@ -5,9 +5,9 @@ name: Build
 
 on:
   push:
-    branches: [ master, development ]
+    branches: [ development, development_rework ]
   pull_request:
-    branches: [ master, development ]
+    branches: [ development, development_rework ]
```
xeren, line 10: This looks outdated.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: This diff viewer appears to be outdated.  You can look at the [actual code of the branch](https://github.com/hernanponcedeleon/Dat3M/blob/b2d17f1eca6a0a88d5aa00369acb4bc78294572f/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java#L402).

xeren: I think this happened because the marked code of this conversation, the method header, has not changed.

.github/workflows/maven.yml
ThomasHaas: Yes, this and the prototype code will have to go eventually.

xeren:
> There used to be some manual changes that were required when merging development (I think it was related to which parser to use, but maybe there was something else). We need to be sure we are not overriding some code

I encountered numerous, but only minor conflicts when I merged `development` into this branch.  Mainly the test classes were affected.  Is there an issue with running any component (`svcomp`, `ui`) on your machines?

dartagnan/src/main/java/com/dat3m/dartagnan/prototype/expr/BinaryExpression.java
```diff
@@ -0,0 +1,8 @@
+package com.dat3m.dartagnan.prototype.expr;
+
+public interface BinaryExpression extends Expression {
+
+    default Expression getLeft() { return getOperands().get(0); }
+    default Expression getRight() { return getOperands().get(1); }
+
+}
```
xeren, line 8: The `com.dat3m.dartagnan.prototype` package is unused and all but a few of its features (`Free`. `PhiEvent`, `PointerType`) remain uncovered.  Can we remove this before merging?

ThomasHaas: You can remove the whole prototype package. It served its purpose.
And we can always recover the code if needed.

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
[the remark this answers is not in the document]
xeren: Looks like `AbstractCompilationTest` etc. still need mathematical integers for pointers.  Maybe I should revert the removal of `modeling.precision` for now and have it properly handled later.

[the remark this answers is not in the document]
xeren: The current patch has `lastValue` return 64-bit vectors if `encoding.integers` is false.

hernanponcedeleon:
Can it be that the tests still use the default BV Encoding and because of this safe stack times out?

ThomasHaas:
> Can it be that the tests still use the default BV Encoding and because of this safe stack times out?

Yes, the configuration of the tests was not updated to use integers.

hernanponcedeleon:
LGTM. Unless @ThomasHaas has further comments, I will merge

ThomasHaas:
You can merge.

PR #494, opened 2023-08-01
Thread cf & must edges

ThomasHaas:
This PR adds proper control-flow dependencies between spawned threads and their spawners/creators.

- Each thread's entry event is now a `ThreadStart` event that may have a `ThreadCreate` event as a `creator`. If `creator == null`, the thread is executed unconditionally (e.g, the `main` thread, init threads, and all threads in litmus tests).
- The control-flow encoding now allows the execution of `ThreadStart` only if its associated `creator` (if any) gets executed.
- `BranchEquivalence` has been reworked to respect this new control-flow and consider the control-flow relationship between a thread and its creator. 
- Synchronization events between spawning thread and spawned thread have been simplified.
- We now generate must-rf edges for synchronizing events.

Minor side changes:
- SCCP now properly recognizes `AbortIf (true)` for dead code detection
- SCCP now eliminates trivial casts that do not change types (e.g., `i64` to `i64` casts).
- `Thread.getEntry` and `Thread.getExit` now return `ThreadStart` resp. `Label`.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
@@ -112,7 +112,7 @@ public class ProgramEncoder implements Encoder {
 
     private BooleanFormula encodeThreadCF(Thread thread) {
         final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
-        final ThreadStart startEvent = (ThreadStart) thread.getEntry();
+        final ThreadStart startEvent = thread.getEntry();
         final List<BooleanFormula> enc = new ArrayList<>();
 
         final BooleanFormula cfStart = context.controlFlow(startEvent);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
@@ -10,6 +10,7 @@ import com.dat3m.dartagnan.program.analysis.ExecutionAnalysis;
 import com.dat3m.dartagnan.program.event.core.CondJump;
 import com.dat3m.dartagnan.program.event.core.Event;
 import com.dat3m.dartagnan.program.event.core.Label;
+import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.dat3m.dartagnan.program.event.core.utils.RegWriter;
 import com.dat3m.dartagnan.program.memory.Memory;
 import com.google.common.base.Preconditions;
@@ -21,7 +22,9 @@ import org.sosy_lab.common.configuration.Options;
 import org.sosy_lab.java_smt.api.*;
 import org.sosy_lab.java_smt.api.NumeralFormula.IntegerFormula;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 
 import static com.dat3m.dartagnan.GlobalSettings.getArchPrecision;
 import static com.dat3m.dartagnan.configuration.OptionNames.INITIALIZE_REGISTERS;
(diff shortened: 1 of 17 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/threading/ThreadStart.java
```diff
@@ -0,0 +1,71 @@
...
+    // May be NULL, if this thread is not spawned.
+    private ThreadCreate creator;
+    //TODO: It is probably better to make the ThreadCreate conditional rather than
+    // allow ThreadStart to fail spuriously.
+    private boolean mayFailSpuriously;
+
+    public ThreadStart(ThreadCreate creator) {
+        this.creator = creator;
+
+        mayFailSpuriously = (creator != null);
```
hernanponcedeleon, line 22: what happens if I use the CLI option to force thread creation? If that is the case do we call `setMayFailSpuriously()`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -796,6 +800,40 @@ public Knowledge visitReadFrom(Relation rel) {
                     }
                 }
             }
+
+            // Here we add must-rf edges between loads/stores that synchronize threads.
+            for (Thread thread : program.getThreads()) {
+                final ThreadStart start = (ThreadStart) thread.getEntry();
```
hernanponcedeleon, line 806: Remove cast if getEntry returns ThreadStart

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -796,6 +800,40 @@ public Knowledge visitReadFrom(Relation rel) {
...
+                while (!(cur instanceof Store endStore)) { cur = cur.getPredecessor(); }
+                cur = start.getCreator();
+                while (cur != null && !(cur instanceof Load endLoad && endLoad.getAddress().equals(endStore.getAddress()))) {
+                    cur = cur.getSuccessor();
+                }
+
+                if (cur instanceof Load endLoad) {
+                    must.add(new Tuple(endStore, endLoad));
+                }
+            }
```
hernanponcedeleon, line 835: can't we remove something from may as line 821?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -109,23 +112,30 @@ public BooleanFormula encodeControlFlow() {
 
     private BooleanFormula encodeThreadCF(Thread thread) {
         final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
-
-        Map<Label, Set<CondJump>> labelJumpMap = new HashMap<>();
-        Event pred = null;
+        final ThreadStart startEvent = (ThreadStart) thread.getEntry();
```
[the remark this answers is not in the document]
ThomasHaas: Probably yes. I wanted to do it, but then I noticed that `getEntry` is inherited from `Function` which does not have to start with that event... but that doesn't matter, cause I should be able to override the `getEntry` function and change its return type. Not sure why I didn't think of this immediately.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/threading/ThreadStart.java
ThomasHaas: Yes, `ThreadCreation` will spawn threads and set the option based on whether thread creation is forced to succeed.
At some point, I will probably get rid of this flag again and instead let `ThreadCreation` generate a non-deterministic choice between success and failure like so:
```
// compiling "int success = pthread_create(&t, NULL; worker, arg);" to
if (*) {
    ThreadCreate(worker, arg);
    store (t, TID);
    success = 1;
} else {
    // store (t, -1);
    success = 0;
}
```
Then `ThreadCreate/ThreadStart` can always be assumed to succeed if executed.

hernanponcedeleon:
LGTM

PR #495, opened 2023-08-03
Benchmarks

hernanponcedeleon:
This PR solves several compiler warning in the locks and lfds benchmarks. With these changes I should be able to regenerate the llvm files having problems in #493


ThomasHaas (approved):
LGTM

PR #496, opened 2023-08-07
Dynamic Pthread join

ThomasHaas:
This PR adds support for dynamic pthread_join calls.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
@@ -2,8 +2,10 @@ package com.dat3m.dartagnan.program.processing;
 
 import com.dat3m.dartagnan.configuration.Arch;
 import com.dat3m.dartagnan.exception.MalformedProgramException;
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -185,12 +159,90 @@ public void run(Program program) {
...
+                    assert resultRegister.getType() instanceof IntegerType;
+
+                    // We create a switch-case over all possible threads to join with
+                    final Register joinDummyReg = thread.getOrNewRegister("__join#" + joinCounter, types.getBooleanType());
+                    final Label joinEnd = EventFactory.newLabel("L__joinEnd#" + joinCounter);
+
+                    // Construct a case for each possible tid
+                    final Map<Expression, List<Event>> tid2joinCases = new LinkedHashMap<>();
+                    for (Expression tidCandidate : tid2ComAddrMap.keySet()) {
+                        final int tid = ((IValue)tidCandidate).getValueAsInt();
```
hernanponcedeleon, line 194: is this cast safe? If we know values are always `IValue`, why not make the map know about this type?

PR #499, opened 2023-08-14
Added simple pass that deletes unused memory objects

ThomasHaas:


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveUnusedMemory.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    public void run(Program program) {
+        final Memory memory = program.getMemory();
+        final MemoryObjectCollector collector = new MemoryObjectCollector();
+        program.getThreadEvents(RegReader.class).forEach(r -> r.transformExpressions(collector));
+        Sets.difference(memory.getObjects(), collector.memoryObjects).forEach(memory::deleteMemoryObject);
+    }
+
+    private static class MemoryObjectCollector implements ExpressionVisitor<Expression> {
```
hernanponcedeleon, line 32: With the exception of the visitor to the memory object (and maybe I miss some other), this pretty much calls the visitor of the children or return the object when it is a "leaf". With instead of returning `null` in the default visitor, we make the default implementation to call the visitor to the children / return when lead. I think this is the usual approach followed with visitors. I might allows us to get rid of a lot of code similar to this one in other visitors

ThomasHaas: The default visitor cannot be changed to simplify this code here. The only thing the default visitor can return is `null` or the value produced by one of its children (which is non-canonical if there a multiple children), but I need to return the original input expression.
The issue here is that I abuse the `RegReader.transformExpressions` method which is side-effect-full cause it replaces the Event's expressions. To avoid the side-effect, all the visitor methods need to return the original expression.

A cleaner way would be to have a method like `Event.visitExpressions` that only observes the expressions rather than modifying them. In that case, a default visitor could be used to traverse the expression tree.

PR #502, opened 2023-08-18
Proper memory access typing & updated compilation

ThomasHaas:
Updated memory events to have the correct access type (we defaulted to the arch type).
Updated the compilation schemes to never explicitly use the arch type when generating code.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
@@ -11,7 +11,7 @@ public class LKMMLock extends SingleAccessMemoryEvent {
     public LKMMLock(Expression lock) {
         // This event will be compiled to LKMMLockRead + LKMMLockWrite
         // and each of those will be assigned a proper memory ordering
-        super(lock, TypeFactory.getInstance().getArchType(), "");
+        super(lock, TypeFactory.getInstance().getIntegerType(32), "");
     }
 
     protected LKMMLock(LKMMLock other) {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
@@ -1,6 +1,7 @@
 package com.dat3m.dartagnan.program.event.lang.linux;
 
 import com.dat3m.dartagnan.expression.Expression;
+import com.dat3m.dartagnan.expression.type.TypeFactory;
 import com.dat3m.dartagnan.program.event.MemoryAccess;
 import com.dat3m.dartagnan.program.event.common.SingleAccessMemoryEvent;
 import com.dat3m.dartagnan.program.event.visitors.EventVisitor;
@@ -10,7 +11,7 @@ public class LKMMLock extends SingleAccessMemoryEvent {
     public LKMMLock(Expression lock) {
         // This event will be compiled to LKMMLockRead + LKMMLockWrite
         // and each of those will be assigned a proper memory ordering
-        super(lock, "");
+        super(lock, TypeFactory.getInstance().getArchType(), "");
     }
 
     protected LKMMLock(LKMMLock other) {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
@@ -57,7 +57,7 @@ class VisitorArm8 extends VisitorBase {
 
     @Override
     public List<Event> visitLock(Lock e) {
-        IntegerType type = types.getArchType();
+        IntegerType type = (IntegerType)e.getAccessType();
         Expression zero = expressions.makeZero(type);
         Expression one = expressions.makeOne(type);
         Register dummy = e.getFunction().newRegister(type);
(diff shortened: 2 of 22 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
```diff
@@ -10,7 +11,7 @@ public class LKMMLock extends SingleAccessMemoryEvent {
     public LKMMLock(Expression lock) {
         // This event will be compiled to LKMMLockRead + LKMMLockWrite
         // and each of those will be assigned a proper memory ordering
-        super(lock, "");
+        super(lock, TypeFactory.getInstance().getArchType(), "");
```
hernanponcedeleon, line 14: Not sure what the type will be used for, but can't we use `BooleanType`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -57,7 +57,7 @@ public List<Event> visitInitLock(InitLock e) {
 
     @Override
     public List<Event> visitLock(Lock e) {
-        IntegerType type = types.getArchType();
+        IntegerType type = (IntegerType)e.getAccessType();
```
hernanponcedeleon, line 60: can we use `BooleanType`? ... the same comment applies to all lock events touched in this PR

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMLock.java
ThomasHaas: I think there may be an issue with the init value of the lock's address.

hernanponcedeleon: The lock expression is a pointer to spinlocks with `%struct.spinlock = type { i32 }`. Even if we simply treat this as a boolean, probably the verification speed up we get is not worth the struggle.

Feel free to merge.

ThomasHaas: If the type is `i32`, then shouldn't the arch type also be wrong?

hernanponcedeleon: Probably, but then I don't get why we don't get a type mismatch error in Code like [this](https://github.com/hernanponcedeleon/Dat3M/blob/llvm-grammar0/dartagnan/src/test/resources/lkmm/2%2B2W%2Bonces%2Blocked.ll) 
```
%struct.spinlock = type { i32 }

@lock_x = dso_local global %struct.spinlock zeroinitializer, align 4, !dbg !0
...
%2 = call i32 @__LKMM_SPIN_LOCK(%struct.spinlock* noundef nonnull @lock_x) #5, !dbg !52
```

ThomasHaas: Have you compiled with BV and mixedType=true?

hernanponcedeleon: Running this (I also tried other archs) on the llvm branch (which is the one having the mixType option) does not raise any error
```
CFLAGS="-I./include" java -jar dartagnan/target/dartagnan-3.1.1.jar cat/linux-kernel.cat --target=lkmm benchmarks/lkmm/2+2W+onces+locked.c --modeling.precision=64 --modeling.mixedType=true
```

ThomasHaas: Probably because you ran this on the C code. The boogie parser will run instead and produce the uniform types, i.e., ignore the mixed size option.

hernanponcedeleon: You are right. If I use the llvm file directly I get a mismatch.

I cherry pick this commit into that branch and update the types to be 32 and then the error is gone. So we should be using `TypeFactory.getInstance().getIntegerType(32)` here

ThomasHaas: But then it will fail if you do not have mixed types enabled, e.g., when you do integer encoding

hernanponcedeleon: With the changes I mentioned above I get these
````
> CFLAGS="-I./include" java -jar -Dlog4j2.configurationFile=file:$DAT3M_HOME/dartagnan/src/main/resources/log4j2.xml dartagnan/target/dartagnan-3.1.1.jar cat/aarch64.cat --target=arm8 dartagnan/src/test/resources/lkmm/2+2W+onces+locked.ll --modeling.precision=64 --modeling.mixedType=true
PASS
Total verification time(ms): 426

> CFLAGS="-I./include" java -jar -Dlog4j2.configurationFile=file:$DAT3M_HOME/dartagnan/src/main/resources/log4j2.xml dartagnan/target/dartagnan-3.1.1.jar cat/aarch64.cat --target=arm8 dartagnan/src/test/resources/lkmm/2+2W+onces+locked.ll --modeling.precision=64 --modeling.mixedType=false
PASS
Total verification time(ms): 418

> CFLAGS="-I./include" java -jar -Dlog4j2.configurationFile=file:$DAT3M_HOME/dartagnan/src/main/resources/log4j2.xml dartagnan/target/dartagnan-3.1.1.jar cat/aarch64.cat --target=arm8 dartagnan/src/test/resources/lkmm/2+2W+onces+locked.ll
```

ThomasHaas: Your code shouldn't work in principle, but it seems our equality encoding automatically does type conversions.
I haven't yet checked why we sometimes get the error although we do these conversions.

PR #504, opened 2023-08-21
Remove version number from jar files

hernanponcedeleon:


ThomasHaas:
Keep in mind that the CNA repo might also need an update to avoid the version number (if you want to keep that repo alive).

hernanponcedeleon:
The CNA scrips checkout a concrete commit of dartagnan, so this should not cause any problem

dartagnan/pom.xml
```diff
@@ -653,6 +653,7 @@
   </dependencies>
 
   <build>
+    <finalName>${project.artifactId}</finalName>
```
ThomasHaas, line 656: What's the use of this? Why didn't we need it before?

hernanponcedeleon: By default the name of the generated jar is `${project.artifactId}-${version}` thus I override the file name to remove the version

ThomasHaas (approved):
LGTM

PR #506, opened 2023-08-24
Function pointers in initial memory

ThomasHaas:
This PR generalizes `MemoryObject`'s initial values to contain any expression rather than just `IConst`. This is currently needed to hold other types of constants like `Function`. 

The LLVM parser can now parse functions referenced in initializers.

Devirtualisation has been extended to include function pointers referenced in initial memory.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
@@ -2,6 +2,7 @@ package com.dat3m.dartagnan.parsers.program.utils;
 
 import com.dat3m.dartagnan.configuration.Arch;
 import com.dat3m.dartagnan.exception.MalformedProgramException;
+import com.dat3m.dartagnan.expression.Expression;
 import com.dat3m.dartagnan.expression.ExpressionFactory;
 import com.dat3m.dartagnan.expression.IConst;
 import com.dat3m.dartagnan.expression.INonDet;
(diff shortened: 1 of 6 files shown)


hernanponcedeleon (commented):
do we loose anything by fully moving from `IConst` to `Expression` instead of having a `Const` interface implemented both by `IConst` and `FConst`?

General comments about `NaiveDevirtualisation.java` (I cannot comment directly on those lines that are not changed by the PR)

- can `nextAvailableFuncAddress = 1;` cause problem due to the fact addresses must be even? (see #346)?
-

ThomasHaas:
> do we loose anything by fully moving from `IConst` to `Expression` instead of having a `Const` interface implemented both by `IConst` and `FConst`?

No, not really. I still want to have a `Constant` interface, but `IConst` is simply not right I think.

> General comments about `NaiveDevirtualisation.java` (I cannot comment directly on those lines that are not changed by the PR)
>     * can `nextAvailableFuncAddress = 1;` cause problem due to the fact addresses must be even? (see [Addresses must be even #346](https://github.com/hernanponcedeleon/Dat3M/pull/346))?

Hmm, I guess if people really wanted to to bit fiddling on function pointers, then maybe. But I also do not make sure that the function pointers get different addresses than other memory, assuming that people do not do weird stuff with them (e.g. loading from a function pointer). It's a "naive" implementation after all, but probably good enough in practice in terms of soundness. but not terms of performance.
That being said, making the addresses 8 byte aligned would be an easy change.

ThomasHaas:
Merged into #493 with the 8-byte alignment change.

PR #507, opened 2023-08-24
Dedicated Assert event

ThomasHaas:
This PR adds a dedicated assert event rather than exploiting locals.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -10,10 +10,10 @@ import com.dat3m.dartagnan.program.Program;
 import com.dat3m.dartagnan.program.Program.SourceLanguage;
 import com.dat3m.dartagnan.program.analysis.SyntacticContextAnalysis;
 import com.dat3m.dartagnan.program.event.Tag;
+import com.dat3m.dartagnan.program.event.core.Assert;
 import com.dat3m.dartagnan.program.event.core.CondJump;
 import com.dat3m.dartagnan.program.event.core.Event;
 import com.dat3m.dartagnan.program.event.core.Load;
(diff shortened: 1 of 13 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -248,15 +248,16 @@ public static String generateResultSummary(VerificationTask task, ProverEnvironm
                 printWarningIfThreadStartFailed(p, encCtx, prover);
                 if (props.contains(PROGRAM_SPEC) && FALSE.equals(model.evaluate(PROGRAM_SPEC.getSMTVariable(encCtx)))) {
                     summary.append("===== Program specification violation found =====\n");
-                    for(Event e : p.getThreadEvents(Local.class)) {
-                        if(e.hasTag(Tag.ASSERTION) && TRUE.equals(model.evaluate(encCtx.execution(e)))) {
+                    for(Assert ass: p.getThreadEvents(Assert.class)) {
+                        if(FALSE.equals(model.evaluate(encCtx.encodeExpressionAsBooleanAt(ass.getExpression(), ass)))) {
```
hernanponcedeleon, line 252: Why false? is the condition in the new assert event reverted wrt locals?

ThomasHaas: The original code checked for the assertion being executed not for its condition. So the old code implicitly assumed that we only ever have `assert(false)` (I think some pieces of code still have this assumption).

hernanponcedeleon:
LGTM

ThomasHaas:
I would still like to get the CI running before merging. I guess we can just retarget this PR to `development`, let the CI run, and then retarget it back.

hernanponcedeleon:
That seems like more work than running the unit tests locally on you machine 😝

ThomasHaas:
Merged into #493

PR #510, opened 2023-09-06
Cleanup of pom files and fix windows build

natgavrilenko:


dartagnan/src/test/java/com/dat3m/dartagnan/compilation/C11ToPPCTest.java
```diff
@@ -49,9 +50,12 @@ protected Provider<Arch> getTargetProvider() {
 	protected List<String> getCompilationBreakers() {
 		return Stream.of(
 				"manual/IRIW-sc-sc-acq-sc-acq-sc",
-				"manual/RWC-sc-acq-sc-sc-sc").map(p -> ResourceHelper.LITMUS_RESOURCE_PATH + "litmus/C11/" + p + ".litmus").collect(Collectors.toList());
+				"manual/RWC-sc-acq-sc-sc-sc")
+				.map(p -> ResourceHelper.LITMUS_RESOURCE_PATH + "litmus/C11/" + p + ".litmus")
+				.map(p -> Path.of(p).toString())
```
natgavrilenko, line 55: This is a quick fix for windows, but I think it needs a proper refactoring. It is not a good practice to hardcode test case names this way,

pom.xml
```diff
@@ -1,89 +1,636 @@
...
+                            </goals>
+                            <configuration>
+                                <outputDirectory>${project.build.directory}/${dat3m.library.path}</outputDirectory>
+                                <includeTypes>jar</includeTypes>
+                                <includeScope>compile</includeScope>
+                                <excludeGroupIds>com.dat3m</excludeGroupIds>
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <id>copy-binaries</id>
```
natgavrilenko, line 91: These are the dependencies from the project itself (i.e., dartagnan.jar), and they cannot be copied before the packaging phase. For example, "mvn test -Dtest="BuildWitnessTest" -pl dartagnan" would fail if this were executed before packaging.

pom.xml
```diff
@@ -1,89 +1,636 @@
...
+                    <artifactId>maven-site-plugin</artifactId>
+                    <version>${maven-site-plugin.version}</version>
+                </plugin>
+                <plugin>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <version>${maven-surefire-plugin.version}</version>
+                    <configuration>
+                        <reuseForks>false</reuseForks>
+                        <systemPropertyVariables>
+                            <LOGNAME>surefire</LOGNAME>
```
natgavrilenko, line 135: This is quick fix to prevent logger crashing on windows when running unit tests. On the long run, we should consider a proper fix and refactoring the logging configuration.

pom.xml
```diff
@@ -1,89 +1,636 @@
...
+                        </systemPropertyVariables>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
+
+    <profiles>
+        <!-- TODO: Why does the arch matter for osx but not for linux and windows? -->
```
natgavrilenko, line 144: I've tested it with Windows, Linux, and OSX, but I don't have any ARM machine profiles for the ARM arch.

hernanponcedeleon:
I'll take a closer look to this one once I'm back from holidays.

In the meantime, @xeren since you have a windows machine, can you please test that all works correctly? i.e. run the unit tests and try building, running dartagnan from the console.

natgavrilenko:
Running dartagnan from console will still have a problem on windows because of the logger (see below). But now it builds and passes the tests, which wasn't the case before.

C:\Users\n00857603\Desktop\code\Dat3M>java -jar dartagnan\target\dartagnan.jar
ERROR StatusConsoleListener FileManager (${env:DAT3M_OUTPUT}/logs/${sys:LOGNAME}.log) java.io.IOException: The filename, directory name, or volume label syntax is incorrect

I would rather fix the logger problem in a separate commit, because there are other logger issues that need to be fixed and reflected in the shell scripts. With small commits there will be less conflicts with the other pending merge requests.

ThomasHaas:
I cannot even run the unit tests from console on my Mac (different issue though).
The file name issue is also present on my MacOS, but rather than complaining it simply generates a file named "${sys:LOGNAME}.log".
Anyway, I'm happy if you can fix the logger problems in multiple small commits, though I don't think you will cause conflicts with pending PRs, because none of them change log configs or pom configs in any way. So it's up to you how fine-grained you wanna do it.

natgavrilenko:
> I cannot even run the unit tests from console on my Mac (different issue though). The file name issue is also present on my MacOS, but rather than complaining it simply generates a file named "${sys:LOGNAME}.log". Anyway, I'm happy if you can fix the logger problems in multiple small commits, though I don't think you will cause conflicts with pending PRs, because none of them change log configs or pom configs in any way. So it's up to you how fine-grained you wanna do it.

As far as I see, the logging should work exactly the same way as in the development branch, i.e. no new problems have been introduced. Please correct me if I am wrong. If this is correct, I would prefer to fix logging in a separate merge request after this one is merged. A proper fix for logging will be quite big.

ThomasHaas:
I should have been more clear: the issues I mentioned are not specific to this branch, they exist on the other branches as well and probably for years now (I never ran the unit tests via console).
I'm fine with committing this branch, but it seems @hernanponcedeleon wants to take a look first?
I don't know much about maven, pom, and the logging stuff, so from my point of view you can take full control and fix whatever you think is a problem.

natgavrilenko:
Yeah, let's wait for Hernan, no hurry with it. I have some ideas how I would fix the logger, but I want to make sure with Hernan that he is OK with my plan, as it is not a trivial small change.

pom.xml
hernanponcedeleon, line 144: I think @ThomasHaas introduced the different osx profiles when he switched to an M1 MacBook because [he encountered some problems](https://github.com/hernanponcedeleon/Dat3M#troubleshooting) with the libraries of the smt solvers. We never tried Linux or Windows on arm processors because none of us has that hardware + OS combination, but I would not be surprised if the same problem to arise there.

hernanponcedeleon (commented):
Please put back the changes that avoid the jar having the version number. As we discussed today, we might want to re-introduced the version number, but we need to find a solution for the other problems I mentioned, thus we should tackle that in a different PR.

ThomasHaas:
I'm curious about why you want to have the version number back. What is the problem if you don't have one?

pom.xml
natgavrilenko: Ok, I removed this comment for now

natgavrilenko:
> Please put back the changes that avoid the jar having the version number. As we discussed today, we might want to re-introduced the version number, but we need to find a solution for the other problems I mentioned, thus we should tackle that in a different PR.

Put it back, no more version numbers

natgavrilenko:
> I'm curious about why you want to have the version number back. What is the problem if you don't have one?

The common practice is to have version numbers in jars (check, for example, the dependencies). Also, missing version numbers can cause problems if we need to publish Dartagnan jars to a (public or private) artifactory. But we are far from there anyway.

P.S. I didn't remove them intentionally, just noticed that the version number are back after I resolved conflicts with the development branch.

PR #511, opened 2023-09-15
Platform-independent path in unit tests

natgavrilenko:


dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
```diff
@@ -53,4 +57,30 @@ public static void LogGlobalSettings() {
...
+        String path = getOutputDirectory();
+        Files.createDirectories(Paths.get(path));
+        return path;
+    }
+
+    public static String getOutputDirectory() {
+        if (isJUnitTest()) {
+            return "target/output";
+        }
+        String env = System.getenv("DAT3M_OUTPUT");
```
natgavrilenko, line 71: In my opinion, it would be cleaner to use a config file instead of an environment variable to read output directory. But this would require changes in many places (e.g., vsyncer), so I will skip it for now.

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/CSVLogger.java
```diff
@@ -10,36 +10,16 @@
...
-             ( See ResourceHelper.getCSVFileName for details)
-           - The <nameSupplier> is used to name the entry inside the .csv file
-
-    NOTE: To properly work with a Timeout Rule, the CSVLogger needs to be put before the Timeout Rule.
-          As a Timeout Rule runs the test in a separate thread, this may force some initialization code
-          to be moved AFTER the Timeout rule (cause some resources can only be used by the thread that created
-          it, which is, e.g., the case for Z3's Context and Solver objects).
-          Overall, this forces some setup code to be measured by the Logger (which is usually not that crucial)
- */
 public class CSVLogger extends TestWatcher {
```
ThomasHaas, line 23: Is there are reason you deleted the comments on this class? I think those are kinda crucial to understand the class or rather what one has to be careful about when using it. I would either keep the comments or just delete the whole class (we are not using it anyway).

ThomasHaas: What it does might be obvious, how it works is a little less obvious, and how it interacts with the timeout rule is not at all obvious IMO.
Anyway, you can just delete the class. I think it is never used

natgavrilenko: Deleted

PR #512, opened 2023-09-15
Add labels and jumps to PTX

tonghaining:
- Add labels, direct and conditional jumps to PTX
- `toLinuxName()` -> `toTextualName()` since this is not anymore LKMM specific
- Fix small issues in `defaultString()` for PTX events
- Move original PTX tests to `litmus/PTX/Nvidia`
- Add ticketlock litmus test with variants


hernanponcedeleon:
@ThomasHaas would you mind taking a look?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
```diff
@@ -39,7 +39,7 @@ public String toString() {
         }
     }
 
-    public String toLinuxName(){
+    public String toTextualName(){
```
ThomasHaas, line 42: The naming is odd: What would be a "non-textual" name?
I think we should just call it `getName` and return the name of the enum value in lower case.
And we should rename the enum values `PLUS` to `ADD`, `MINUS` to `SUB` and `MULT` to `MUL`, and remove the underscores from `L_SHIFT` and the other enum names.
Then you could just do `return this.name().toLowerCase()` and it works for all op codes (it's weird to have undefined names for some enum values in the first place).

On a side note: I think it would be most natural to have `toString` return the name and add a `getSymbol` method to get the operator symbol. But this is a bigger change since all the callers of `op.toString` would need to get updated, so it should be done in a separate PR.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusPTX.java
```diff
@@ -210,7 +215,7 @@ public Object visitAtomConstant(LitmusPTXParser.AtomConstantContext ctx) {
         IOpBin op = ctx.operation().op;
         String mo = ctx.mo().content;
         String scope;
-        if (mo.equals(Tag.PTX.ACQ_REL) || mo.equals(Tag.PTX.RLX)) {
+        if (mo.equals(Tag.PTX.ACQ) || mo.equals(Tag.PTX.REL) || mo.equals(Tag.PTX.ACQ_REL) || mo.equals(Tag.PTX.RLX)) {
             scope = ctx.scope().content;
         } else {
             throw new ParsingException("Atom instruction doesn't support mo: " + mo);
```
ThomasHaas, line 221: Why only read the scope if the mo is of a certain shape? If `ctx.scope().content` cannot throw a `NullPointerException`, then just do it before doing the mo-check. If it can throw the exception, just do it after the mo-check.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusPTX.java
```diff
@@ -222,13 +227,13 @@ public Object visitAtomConstant(LitmusPTXParser.AtomConstantContext ctx) {
...
         MemoryObject object = programBuilder.getOrNewMemoryObject(ctx.location().getText());
-        Register register_operand = programBuilder.getOrNewRegister(mainThread, ctx.register().get(1).getText(), TypeFactory.getInstance().getArchType());
+        Register register_operand = programBuilder.getOrNewRegister(mainThread, ctx.register().get(1).getText(), archType);
         IOpBin op = ctx.operation().op;
         String mo = ctx.mo().content;
         String scope;
-        if (mo.equals(Tag.PTX.ACQ_REL) || mo.equals(Tag.PTX.RLX)) {
+        if (mo.equals(Tag.PTX.ACQ) || mo.equals(Tag.PTX.REL) || mo.equals(Tag.PTX.ACQ_REL) || mo.equals(Tag.PTX.RLX)) {
             scope = ctx.scope().content;
         } else {
```
ThomasHaas, line 238: Same as above. The pattern happens quite often, so please fix it in the remaining code as well.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/IOpBin.java
ThomasHaas: You didn't rename `PLUS`/`MINUS`. This will cause some events to be named like `atomic_fetch_plus` rather than `atomic_fetch_add`.

ThomasHaas (approved):
LGTM

PR #513, opened 2023-09-18
Introduce TaggedEvent

hernanponcedeleon:
This PR introduces a general `TaggedEvent` whose semantics are purely defined by the tags it has. 
Currently this is only used to implement fences, but there is another use case in the pipeline related to vulkan memory model.

The `Fence` class has almost no code now. I considered completely removing it, but since fences are such a basic and used concept in WMM, I decided to keep it.

I also refactor `FenceBase` to avoid repeated code between this class and `Fence`.


dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -361,7 +361,7 @@ private void addEvent(Event e, int globalId, int localId) {
 
         } else if (data.isFence()) {
             // ===== Fences =====
-            String name = ((Fence)data.getEvent()).getName();
+            String name = ((TaggedEvent)data.getEvent()).getName();
             fenceMap.computeIfAbsent(name, key -> new HashSet<>()).add(data);
```
ThomasHaas, line 365: This code certainly violates the fact that `TaggedEvents` should be identified by their tags only (semantics-wise).
IIRC, this is done like this right now because we treat `fencerel(X)` as a base relation rather than a short-cut for `po;[X];po`.
I would put a `FIXME` there so we fix this eventually. You should double check what `data.isFence()` does and if it still applies for general `TaggedEvents`!

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/TaggedEvent.java
```diff
@@ -0,0 +1,39 @@
...
+import com.dat3m.dartagnan.program.event.Tag;
+import com.dat3m.dartagnan.program.event.visitors.EventVisitor;
+
+public class TaggedEvent extends AbstractEvent {
+
+    protected final String name;
+
+    public TaggedEvent(String name, String ... tags) {
+        this.name = name;
+        this.addTags(Tag.VISIBLE, name);
```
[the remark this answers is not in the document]
hernanponcedeleon: This is needed for many WMM that use the name in the cat file, e.g. `MFENCE`. Do you want to have a parameter to the constructor to control if the name should be added or not?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/TaggedEvent.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public String defaultString() {
+        return name;
+    }
```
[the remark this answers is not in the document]
hernanponcedeleon: I don't think we should do that because that is mixing two things. Whatever is chosen for a name should be informative enough to understand the event semantics.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/EventVisitor.java
```diff
@@ -34,7 +34,8 @@ public interface EventVisitor<T> {
     default T visitLocal(Local e) { return visitEvent(e); }
     default T visitSkip(Skip e) { return visitEvent(e); }
 
-    default T visitFence(Fence e) { return visitEvent(e); }
+    default T visitTaggedEvent(TaggedEvent e) { return visitEvent(e); }
+    default T visitFence(Fence e) { return visitTaggedEvent(e); }
```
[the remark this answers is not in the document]
hernanponcedeleon: I don't get what you mean. `visitFence` is gone because I removed `Fence` (push pending). But it seems you also want to remove `visitTaggedEvent` ...

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
hernanponcedeleon: I don't fully understand what your concern is.
```
public boolean isFence() { return event.hasTag(FENCE); }
```
thus the event is being identified by its tag (and if that invariant is satisfied, I don't see the problem with the cast)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/visitors/EventVisitor.java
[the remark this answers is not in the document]
ThomasHaas: We can keep `visitTaggedEvent` but I'm not sure if that is really a good idea, because the whole point is that you never need to know the class but only its tags (i.e., any check like `instanceof TaggedEvent` or a cast to `TaggedEvent` is HIGHLY suspicious).
But I think it is fine to keep `visitTaggedEvent` because it does have one parameter that is not just a tag, namely its name. That might be enough to justify the existence of dedicated visitor methods for it.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
ThomasHaas: The problem is that the code assumes not only that the event has a fence tag but also that it is a `TaggedEvent`.
What if you had something like a `FencedLoad` that acts as a fence and a load simultaneously? In fact, acquire/release memory accesses fit into that picture (they just don't have the `F` tag).

hernanponcedeleon: ok, now I got it. I agree, but the old code had the same issue. For now I will simply add a FIXME

ThomasHaas:
I need to think about what the best way is to handle `TaggedEvent`. Just from seeing the resulting code, I really dislike its usage.
That being said, I also think that our old usage of `Fence` was only fine as long as one could assume the invariant that fences are disjoint from all other visible events. I am not sure how well this invariant applies to `TaggedEvent`...

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/TaggedEvent.java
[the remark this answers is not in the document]
ThomasHaas: I think the `EventFactory` methods can just put the name (e.g., `MFENCE`) twice, once as the display name and once as a tag. No need to have an extra flag.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -106,12 +106,12 @@ public static Store newStoreWithMo(Expression address, Expression value, String
         return store;
     }
 
-    public static Fence newFence(String name) {
-        return new Fence(name);
+    public static GenericVisibleEvent newFence(String name) {
+        return new GenericVisibleEvent(name, Tag.FENCE);
```
ThomasHaas, line 110: I would return `GenericVisibleEvent(name, name, Tag.FENCE)` to make it semantically equivalent with the previous version. Then you do not need to change any of the other fence factory methods.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -447,41 +447,59 @@ public static class DMB {
...
+                return fence;
             }
 
-            public static Fence newISHBarrier() {
-                return new Fence("DMB.ISH");
+            public static GenericVisibleEvent newISHBarrier() {
+                String name = "DMB.ISH";
+                GenericVisibleEvent fence = newFence(name);
+                fence.addTags(name);
+                return fence;
```
ThomasHaas, line 465: If you fix the `newFence` method, you can reduce this back to `return newFence("DMB.ISH")`.
Same for all other fence methods

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/FenceBase.java
```diff
@@ -2,20 +2,19 @@
...
 import com.dat3m.dartagnan.program.event.Tag;
 import com.dat3m.dartagnan.program.event.core.AbstractEvent;
+import com.google.common.base.Preconditions;
 
 @NoInterface
 public abstract class FenceBase extends AbstractEvent {
 
     protected final String name;
-    protected final String mo; // May be NULL or empty for fences that do not support a mo.
+    protected final String mo; // For fences that do not support a mo, use Fence instead.
```
ThomasHaas, line 11: The comment is outdated.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/FenceBase.java
```diff
@@ -2,20 +2,19 @@
...
 
     public FenceBase(String name, String mo) {
+        Preconditions.checkArgument(!mo.isEmpty());
         this.name = name;
         this.mo = mo;
-        this.addTags(Tag.VISIBLE, Tag.FENCE);
-        if (mo != null && !mo.isEmpty()) {
-            addTags(mo);
-        }
+        addTags(Tag.VISIBLE, Tag.FENCE, mo);
```
ThomasHaas, line 17: Why do we require a memory order now? I think fences without a `mo` are reasonable.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LogThreadStatistics.java
```diff
@@ -38,7 +39,7 @@ public void run(Program program) {
                     storeCount++;
                 } else if (e instanceof Load) {
                     loadCount++;
-                } else if (e instanceof Fence) {
+                } else if (e.hasTag(Tag.FENCE)) {
```
ThomasHaas, line 42: Maybe we should change this to `e instance GenericVisibleEvent`, cause the intention of the program statistics is to give a rough understanding of all the visible events to gauge the size of the execution graphs.
Like this, you will miss visible non-fences (like the ones in Vulkan).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -171,7 +171,7 @@ public List<Event> visitLlvmCmpXchg(LlvmCmpXchg e) {
     @Override
     public List<Event> visitLlvmFence(LlvmFence e) {
         String mo = e.getMo();
-        Fence fence = mo.equals(C11.MO_RELEASE) || mo.equals(C11.MO_ACQUIRE_RELEASE) || mo.equals(C11.MO_SC) ?
+        GenericVisibleEvent fence = mo.equals(C11.MO_RELEASE) || mo.equals(C11.MO_ACQUIRE_RELEASE) || mo.equals(C11.MO_SC) ?
```
ThomasHaas, line 174: I would rather use `Event` as variable type. The code doesn't really need to know about what type of event it is. I think we can do that for pretty much all cases where `GenericVisibleEvent` is used right now.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusRISCV.java
```diff
@@ -229,12 +230,50 @@ public Object visitBranchCond(LitmusRISCVParser.BranchCondContext ctx) {
 
 	@Override
 	public Object visitFence(LitmusRISCVParser.FenceContext ctx) {
-		return programBuilder.addChild(mainThread, EventFactory.newFence("Fence." + ctx.fenceMode().mode));
+        GenericVisibleEvent fence;
+        switch(ctx.fenceMode().mode) {
+            case "r.r": 
+                fence = EventFactory.RISCV.newRRFence();
+                break;
+            case "r.w":
```
ThomasHaas, line 238: you can use a switch expression:
```
Event fence = switch (...) {
  case "r.r" -> EventFactory.RISCV.newRRFence();
  case "r.w" -> ....
}
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
hernanponcedeleon: That would add `name` to the tag to every single event which is exactly what you initially wrote we should not do. I am having doubts about what would be the best approach. On the one side, I think it is better not to add the name by default and do it only if needed by creating a specific factory method for the special instruction and add the tag there (in like we are doing now in e.g., RISCV). However, except for PTX (and the soon coming vulkan), it seems pretty much all our memory models indeed require having the name as a tag.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/FenceBase.java
hernanponcedeleon: No ... these are language level fences and I argue that for those `mo` is mandatory (which is aligned with the fact we have `Preconditions.checkArgument(!mo.isEmpty());` in the constructor)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
hernanponcedeleon: What about the factory? There we always return the most precise type so for consistency I would still use `GenericVisibleEvent`

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/FenceBase.java
ThomasHaas: `Fence` does not exist anymore, hence it is outdated. Strictly speaking, this class is also not just for language-level fences, it is just a common(!) implementation that can be reused for whatever purposes.
Whether all fences have a `mo`or not is likely a matter of convention. Intuitively, I would have said a compiler barrier has no memory ordering, but in our `linux.h` we do model compiler barriers as a fence with `barrier` memory ordering. I'm fine with that, though I don't necessarily see why `FenceBase` should enforce that.
After all, this class is abstract and any subclass is free to decide if it needs a memory ordering or not.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
ThomasHaas: I said that it would not be a good idea to do it in the constructor of `GenericVisibleEvent` and that the `EventFactory` could do it, if necessary. Not doing it in the constructor allows you to easily support `EventFactory.newFence` which automatically adds name + F-tag, while also supporting something like `EventFactory.newVisibleEvent` which does not add anything at all. Honestly, we might even go a step further and just have `GenericEvent` that does not even need to be visible and let the `EventFactory` add the visible tag.
That being said, I understood that `PTX` and `Vulkan` just do not use the F-tag, so they have non-fence events.
So you wouldn't call `EventFactory.newFence` in the first place, whether it adds the name or not.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
ThomasHaas: In the factory you can use `GenericVisibleEvent`. I would see it similar to how we use `Set<T>` as a type even if we instantiate it with `HashMap<T>` most of the time. If the concrete type is not so important at the call site, I would stick with the more generic type there.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
hernanponcedeleon: I am not referring to the new instructions from vulkan that gave raise to the change in this PR, but rather to PTX or vulkan **fences**. E.g. PTX has [this kind](https://github.com/hernanponcedeleon/Dat3M/blob/027b4024dd3b9894473233bc5f197912a7cf7421/litmus/PTX/Proxy-Alias-AliasFence.litmus#L10) of instructions which in the memory model are simply referred as `F` (different to e.g. `tso` which expliclty uses `mfence`)

ThomasHaas: I see. Well, that is also the case for all C11 fences: the cat model always refers to `F & mo` never to the actual fence name which would be `atomic_thread_fence` I suppose. So there a few cases where the name of the fence is irrelevant.

hernanponcedeleon: Exactly and this is why I don't want `newFence` to add it. Instead, we should call e.g., `newSYBarrier` and that one would add the needed tag.

ThomasHaas: Ok, but you could still add a method like `newFenceWithoutNameTag` (you can even make it private if you want) to avoid the code repetition. I mean, we can easily provide different options with different factory methods.
Though I guess in most cases it does not really hurt to add the name as a tag as well, but I agree that we can avoid it entirely.

hernanponcedeleon: It seems pretty much all our MMs were relying on the name being add as a tag, either because the cat file refers to it (e.g., `mfence`) or because we about this and use it to pass the memory ordering (e.g., compilation of C11). Thus, I decided to keep the functionality as closed as possible to before.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -638,17 +638,17 @@ public Void visitSyncBarrier(Relation rel) {
         @Override
         public Void visitSyncFence(Relation syncFence) {
             boolean idl = !context.useSATEncoding;
-            List<Fence> allFenceSC = program.getThreadEvents(Fence.class).stream()
-                    .filter(e -> e.hasTag(Tag.PTX.SC))
+            List<GenericVisibleEvent> allFenceSC = program.getThreadEvents(GenericVisibleEvent.class).stream()
+                    .filter(e -> e.hasTag(Tag.FENCE) && e.hasTag(Tag.PTX.SC))
```
ThomasHaas, line 642: Can we use `List<Event>` and go over all(!) events and filter for those that have `Visible`, `Fence` and `PTX.SC` tag?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusRISCV.java
```diff
@@ -229,12 +230,27 @@ public Object visitBranchCond(LitmusRISCVParser.BranchCondContext ctx) {
...
+            case "r.rw" -> EventFactory.RISCV.newRRWFence();
+            case "w.r" -> EventFactory.RISCV.newWRFence();
+            case "w.w" -> EventFactory.RISCV.newWWFence();
+            case "w.rw" -> EventFactory.RISCV.newWRWFence();
+            case "rw.r" -> EventFactory.RISCV.newRWRFence();
+            case "rw.w" -> EventFactory.RISCV.newRWWFence();
+            case "rw.rw" -> EventFactory.RISCV.newRWRWFence();
+            case "tso" -> EventFactory.RISCV.newTsoFence();
+            case "i" -> EventFactory.RISCV.newSynchronizeFence();
+            default -> throw new ParsingException("No support fence mode");
```
ThomasHaas, line 245: Add the wrong fence mode as part of the error message, so it is easy to see what is wrong.
I would also say "Invalid fence mode" cause "no support/unsupported" sounds like we have an incomplete implementation (do we?).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -976,7 +976,8 @@ public Knowledge visitSameScope(Relation rel, String specificScope) {
             List<Event> events = new ArrayList<>();
             events.addAll(program.getThreadEvents(Load.class));
             events.addAll(program.getThreadEvents(Store.class));
-            events.addAll(program.getThreadEvents(Fence.class));
+            events.addAll(program.getThreadEvents(GenericVisibleEvent.class));
+            events.removeIf(e -> e instanceof GenericVisibleEvent && !e.hasTag(Tag.FENCE));
```
ThomasHaas, line 980: Here I would also go over all events and take those that have `Visible` and `Fence` tag.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -1021,9 +1022,9 @@ public Knowledge visitSyncBarrier(Relation sync_bar) {
         @Override
         public Knowledge visitSyncFence(Relation sync_fen) {
             Set<Tuple> may = new HashSet<>();
-            List<Fence> fenceEvents = program.getThreadEvents(Fence.class);
-            for (Fence e1 : fenceEvents) {
-                for (Fence e2 : fenceEvents) {
+            List<GenericVisibleEvent> fenceEvents = program.getThreadEvents(GenericVisibleEvent.class).stream().filter(e -> e.hasTag(Tag.FENCE)).toList();
+            for (GenericVisibleEvent e1 : fenceEvents) {
+                for (GenericVisibleEvent e2 : fenceEvents) {
                     if (e1.hasTag(Tag.PTX.SC) && e2.hasTag(Tag.PTX.SC) && !exec.areMutuallyExclusive(e1, e2)) {
```
ThomasHaas, line 1028: Same as above. But you can additionally filter for `Tag.PTX.SC`, because the loop only ever cares about those events.

ThomasHaas (commented):
Apart from the remaining few and minor issues, the PR looks good to me. I will approve once the changes are there

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusRISCV.java
hernanponcedeleon: We are complete. We even support the synchronization fence `.i` which is used in several litmus tests, but it is really not part of the cat file (I noticed this the other day). I double checked both in herd's repo and RISCV documentation, and indeed this fence is not part of the formal memory model.

ThomasHaas (approved):
LGTM (I pushed some minor changes so you might want to double check before merging).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -636,11 +635,10 @@ public Void visitSyncBarrier(Relation rel) {
...
         public Void visitSyncFence(Relation syncFence) {
-            boolean idl = !context.useSATEncoding;
-            List<Event> allFenceSC = program.getThreadEvents(Event.class).stream()
-                    .filter(e -> e.hasTag(Tag.VISIBLE) && e.hasTag(Tag.FENCE) && e.hasTag(Tag.PTX.SC))
-                    .sorted(Comparator.comparingInt(Event::getGlobalId))
-                    .toList();
+            final boolean idl = !context.useSATEncoding;
+            final String relName = syncFence.getName().get(); // syncFence is base, it always has a name
+            List<Event> allFenceSC = program.getThreadEventsWithAllTags(VISIBLE, FENCE, PTX.SC);
+            allFenceSC.sort(Comparator.comparingInt(Event::getGlobalId));
```
hernanponcedeleon, line 641: why do we need the sort?

ThomasHaas: The original code used a sort, so I kept it. I think it is not necessary for correctness, but it makes sense to go over the events in that order.
That being said, the call to `program.getThreadEventsWithAllTags` should already return a sorted list, so the sort will do nothing (IIRC, java's sort is optimized to run in linear time on already sorted lists).
You can remove the call, if you want.

hernanponcedeleon: Yes, the whole things boils down to call `e.getSuccessors()` so the ordering should be guaranteed (and even if not, it should not affect correctness). I will remove

PR #515, opened 2023-09-21
Remove id from recomputed co

hernanponcedeleon:
In PTX, coherence is [a partial transitive order](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#coherence-order) which clearly does not include `id`.

@natgavrilenko noticed that due to how we recompute coherence to guarantee transitivity, the `id`-pairs end up in the may set of coherence. This does not affect correctness because of some axioms, e.g. "SC-per-Location". However, we should not rely on axioms to guarantee the semantics of base relations.

### DIFF
--- a/cat/ptx-v6.0.cat
+++ b/cat/ptx-v6.0.cat
@@ -17,7 +17,7 @@ PTX
 (*******************)
 
 // Explicitly add transitivity for coherence since the co in PTX is not total
-let co = co+
+let co = co+ \ id
 
 (* Events *)
 let strong-write = W & (RLX | REL)
--- a/cat/ptx-v7.5.cat
+++ b/cat/ptx-v7.5.cat
@@ -31,7 +31,7 @@ PTX
 (*******************)
 
 // Explicitly add transitivity for coherence since the co in PTX is not total
-let co = co+
+let co = co+ \ id
 
 (* Events *)
 let strong-write = W & (RLX | REL)


ThomasHaas:
How would `id` end up in the coherence relation? Are you saying that `co` can be cyclic? In that case, it is wrong with and without `id`.

hernanponcedeleon:
Assume we have
```
e1: x=1
e2: x=2
```
Our base `PTX-co` would then have `(e1,e2) \in PTX-co.may` and ``(e2,e1) \in PTX-co.may``. Since the cat file defines `co = PTX-co+`, we end up with `(e1,e1) \in co.may`. The solver will take care (dynamically) of guaranteeing that coherence is not acyclic (using e.g., "SC-per-Location"), but with the change in this PR we also remove those pairs from the may set.

ThomasHaas:
Ok, you are talking about the may-set and not about the actual relation. `co` is always acyclic and will never contain `id` even without any axioms. I find it rather odd to add `\ id` as a micro-optimization into the cat model just so that our relation analysis is a little bit more precise in the very odd case where there are no axioms in the model?!
I mean, you could also add `acyclic(co)` and you would get even more information (assuming XRA is running). But that is obviously redundant with `SC-per-loc`.

Overall, I think this is a confusing change that does not affect semantics and neither improves analysis in practice. It is also not as truthful as, e.g., `acyclic(co)`.

hernanponcedeleon:
Ok. I was not convinced that PTX-co was acyclic, but you are right that the use of the clock variables (if using IDL) will guarantee this.

PR #517, opened 2023-09-22
Performance of relation analysis propagator

natgavrilenko:
The first part of Relation Analysis performance improvements.

Postpones the creation of tuple objects until all combinations have been computed. Reduced execution time about 4-5 times for a transitive closure with ~1.3M tuples and about 2 times for large composition relations.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
@@ -1071,9 +1071,34 @@ public class RelationAnalysis {
     }
 
     public final class Propagator implements Definition.Visitor<Delta> {
-        public Relation source;
-        public Set<Tuple> may;
-        public Set<Tuple> must;
+
+        private Relation source;
+        private Set<Tuple> may;
+        private Set<Tuple> must;
+
+        public Relation getSource() {
+            return source;
+        }
+
+        public void setSource(Relation source) {
+            this.source = source;
+        }
+
+        public Set<Tuple> getMay() {
+            return may;
+        }
+
+        public void setMay(Set<Tuple> may) {
+            this.may = may;
+        }
+
+        public Set<Tuple> getMust() {
+            return must;
+        }
+
+        public void setMust(Set<Tuple> must) {
+            this.must = must;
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -1196,48 +1208,52 @@ public Delta visitInverse(Relation rel, Relation r1) {
...
+                    Event e1 = entry.getKey();
+                    Set<Event> update = new HashSet<>();
+                    for (Event e2 : entry.getValue()) {
+                        if (isMay) {
+                            update.addAll(outer.getOrDefault(e2, Set.of()));
+                        } else {
+                            boolean implies = exec.isImplied(e1, e2);
+                            update.addAll(outer.getOrDefault(e2, Set.of()).stream()
+                                    .filter(e -> implies || exec.isImplied(e, e2))
+                                    .collect(toSet()));
```
ThomasHaas, line 1237: Not sure how much impact this has, but you could either use `.collect(toList())` here (cheaper than set construction, faster iteration) or directly do `outer.getOrDefault(...).stream().filter().foreach(update::add)` which would skip the construction of an intermediate collection altogether.
Since the must-case is usually not the bottleneck, I don't expect much gain so you are free to keep it as is if you want.

ThomasHaas (approved):
LGTM overall, but I haven't checked the details.

I'm not sure if this issue was already in previous version, but the most precise must-set R+ of a transitive closure is not obtained by a linear fixed point iteration `R+ = R + R;R+` but rather by iterating the non-linear equation `R+ = R + R+;R+`.
Since the latter is more expensive and there is not a big difference (if any) in practice, using the former computation scheme is fine. I just wanted to let you know there is a theoretical difference between the transitive closures of the may and must sets, because the composition of must sets is not associative.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
natgavrilenko: Changed to a list, though don't see a difference on "my" benchmarks.

PR #520, opened 2023-09-25
Performance of RA ExtendedPropagator

natgavrilenko:


ThomasHaas:
Do you have any statistics on how much this improves performance?

natgavrilenko:
> Do you have any statistics on how much this improves performance?

On the (non-opensource) benchmark that I have been analyzing most detailed:
Regular RA - before 806 sec, after 352 sec
Regular+Extended RA - before 2264 sec, after 1304 sec.
Both cases were ran at the same time to ensure equal load.

The boilerplate parts: 
- composition and transitive closure in regular RA (after my changes, transitive closure is more of less OK, but composition is still a problem),
- composition in extended RA (propagation of disabled set from child to parent).

There are other suboptimal parts, I don't see a point to invest time into optimizing them now (for these benchmarks), because we cannot get reasonable performance for composition.

ThomasHaas:
This PR targets only the extended RA, right? I assume your statistics include the changes from this PR and from #517, 
but I'm wondering how much effect this particular PR has relative to #517 (a ballpark estimate is fine). 

I find it really interesting that the extended RA (we also call it XRA for short) takes quite a bit more time than the regular one, because that didn't really happen on all my smaller benchmarks. 

I think you will quickly hit a wall with these "simple" optimizations, way before XRA gets to reasonable performance on those large benchmarks. I think we need a properly engineered solution in the long run.
As a case in point, child-to-parent propagation of disabled sets can be heavily reduced (almost eliminated?) by interleaving XRA with RA.
I don't know how much time you have/are willing to commit to improving the relation analysis though :)

natgavrilenko:
> This PR targets only the extended RA, right? I assume your statistics include the changes from this PR and from #517, but I'm wondering how much effect this particular PR has relative to #517 (a ballpark estimate is fine).
> 
> I find it really interesting that the extended RA (we also call it XRA for short) takes quite a bit more time than the regular one, because that didn't really happen on all my smaller benchmarks.
> 
> I think you will quickly hit a wall with these "simple" optimizations, way before XRA gets to reasonable performance on those large benchmarks. I think we need a properly engineered solution in the long run. As a case in point, child-to-parent propagation of disabled sets can be heavily reduced (almost eliminated?) by interleaving XRA with RA. I don't know how much time you have/are willing to commit to improving the relation analysis though :)

Yes, the numbers are for both PRs together. 

And yes, these simple optimizations are not enough. Interleaving RA with XRA might work, but XRA is not my main concern for now, so I am not planning to try optimizing it further in near future.

PR #522, opened 2023-09-29
Add Mem2reg

xeren:
Adds a program processing step called `MemToReg`, which replaces load and store operations on thread-local memory objects.  This targets especially LLVM assembly.  Preferably, it should be as strong as `opt -mem2reg` and also applicable after unrolling.  A memory object is considered promotable, if
- it is statically sized, and
- its address is neither used in a complex expression (boolean, integer multiplication, etc.), passed to another function, nor returned by the function.

An inter-procedural approach would be unnecessary, considering that inlining should have already taken place before.  When positioned before `SimpleSpinLoopDetection`, the process can help recognizing side-effect-free loops.  When positioned after unrolling, it can make further use of the finiteness of executions, promoting even more allocations and simplifying the SMT encoding.

Issues:
- The underlying memory model has to guarantee that every process perceives its own memory operations in program order.  Before merging, this has to be accounted for in some form.
- Common memory operations like `free`, `memset`, `memcpy` and `memcmp` need to be supported.


ThomasHaas:
I think the first issue is resolved by just making `mem2reg` optional. It would even be fine to have it enabled by default and just allow the user to disable it. We might also have to disable it for GPU code where `SC-per-loc` does not necessarily hold (if it holds, then a thread has to perceive its own code in program order).
That being said, technically speaking it is not sufficient to assume that a thread only observes its most recent operations, because it could still force more (stupid) constraints like `empty (rfi)` which are able to distuinguish an internal read from a register access.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
```diff
@@ -0,0 +1,353 @@
...
+                assert entry.getKey().getUsers().isEmpty();
+                entry.getKey().tryDelete();
+            }
+        }
+        if (loadCount + storeCount > 0) {
+            logger.debug("Removed {} loads and {} stores in function \"{}\".", loadCount, storeCount, function.getName());
+        }
+    }
+
+    private List<Type> getReplacingRegisters(Alloc allocation) {
```
ThomasHaas, line 123: I think the name is not fitting because this method does not return registers. 
`getReplacementTypes` might be better.
If the method is intended to only return primitive/simple types (non-aggregates), you can also call it `getPrimitiveReplacementTypes`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
```diff
@@ -0,0 +1,353 @@
...
+        } else {
+            //TODO aggregate types
+            return null;
+        }
+        //TODO check for mixed-size accesses
+        return replacementTypes;
+    }
+
+    // Invariant: base != null
+    private record AddressOffset(RegWriter base, int offset) {
```
ThomasHaas, line 146: You could also state the invariant with `@Nonnull` annotations on the record component.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
```diff
@@ -0,0 +1,353 @@
...
+        // Keeps track of pointer operations.  Maps assignments to the address value they must return.
+        private final Map<Local, AddressOffset> assignments = new HashMap<>();
+        private boolean dead;
+
+        @Override
+        public Label visitEvent(Event e) {
+            if (dead) {
+                return null;
+            }
+            // Publish all values passed as arguments.
```
ThomasHaas, line 177: This code seems to assume that the parameter is a Function call (according to the comment).
If so, this assumption should be made more explicit (a comment would be sufficient).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
```diff
@@ -0,0 +1,353 @@
...
+    // Processes events in program order.
+    // Returns a label, if it is program-ordered before the current event and its symbolic state was updated.
+    private static final class Matcher implements EventVisitor<Label> {
+
+        // Current local symbolic state.  Missing values mean irreplaceable contents.
+        private final Map<Object, AddressOffset> state = new HashMap<>();
+        // Maps labels and jumps to symbolic state information.
+        private final Map<Label, Map<Object, AddressOffset>> jumps = new HashMap<>();
+        // Since it is not allowed to get the address of a register, once an address is
+        // Join over all memory information.  Initialized to all empty.  Missing entries mean irreplaceable address.
```
ThomasHaas, line 164: Incomplete comment.

hernanponcedeleon:
I don't get why this works only if "process perceives its own memory operations in program order".

Is the second point really an "issue"? The comment in the class says that they are treated as calls to unknown and I would assume this simply results in a more imprecise, but still sound, analysis.

ThomasHaas:
> I don't get why this works only if "process perceives its own memory operations in program order".

He means that it is only sound if the memory model is "nice enough". `Mem2reg` is in general unsound since it eliminates visible operations which may affect consistency. Of course, no reasonable model will disallow this optimization.

hernanponcedeleon:
> `Mem2reg` is in general unsound since it eliminates visible operations which may affect consistency. 

If it only eliminates events accessing thread-local memory, those events should not be visible from the point of view of the memory model (implementation-wise they are visible because we mark every Load/Store as visible, but conceptually they should not be).

I also did not get your comment about why we might need to disable this pass for GPU memory models.

ThomasHaas:
I'm not sure how it is decided what is and what is not supposed to be visible. If you say that accesses to thread-local addresses are conceptually never visible, then we have the opposite problem: if we do not run `mem2reg` our execution graphs are wrong.
I don't think this is true though, because I interpret "visible" as being visible to the memory subsystem rather than being visible to other threads. That makes accesses to memory-allocated storage visible by default. In fact, allocated memory could always be accessed by any other thread that guesses the correct address (it would be UB for C-code, but not for hardware code).

Regarding GPUs: Currently there cannot be any issue since we have no allocation mechanism in our litmus code. But hypothetically speaking, a thread could allocate local memory, create an virtual alias to that memory, store to that virtual alias, and then load from the physical address directly failing to observe its own store.
In practice though, this won't be a problem because we will likely never apply `mem2reg` to any objects that have virtual aliases.

ThomasHaas (approved):
LGTM

PR #524, opened 2023-10-02
Optimize memory consumption for RA sets

natgavrilenko:


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/Tuple.java
```diff
@@ -2,25 +2,7 @@
...
-    }
-
-    public Event getFirst(){
-        return first;
-    }
-
-    public Event getSecond(){
-        return second;
-    }
+public record Tuple(Event first, Event second) implements Comparable<Tuple> {
```
natgavrilenko, line 5: Still used in SymmetryEncoder and CoSymmetryBreaking. It would be nice to drop it completely.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -262,6 +260,15 @@ private void run() {
                 if (delta.may.isEmpty() && delta.must.isEmpty()) {
                     continue;
                 }
+
+                // TODO: Quick fix for https://github.com/hernanponcedeleon/Dat3M/issues/523
```
natgavrilenko, line 264: Quick fix for the old issue to make GPU unit tests pass. A proper fix is needed.

ThomasHaas:
The return of the `TupleSet` :). 
I feel like we should be more general here. Your `TupleSet` class really doesn't care about the fact that it contains events, it might as well be a `TupleSet<T>` with a generic parameter. On top of that, I think rather than being a set, this class represents more of a directed graph in the form of adjacency sets. And I think the fact that we deal with a graph is the most important aspect for future development.
Ideally, I imagine an interface like `DirectedGraph<T>` with a straightforward implementation (your `TupleSet` but maybe with a better name), but also with the possibility to implement differently (e.g., implicitly) represented graphs. This would align nicely with #437 where the goal was to get rid of the explicit representation of some of the large base relations like `ext`, which can be represented in constant space instead of quadratic space. Changing the code to now use an explicit implementation everywhere clashes with that idea.

ThomasHaas:
Also, do you have any estimates on how much memory this saves on your larger benchmarks? I assume it also comes with performance improvements due to the smaller memory footprint.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -252,6 +249,7 @@ private void run() {
                 List<Delta> d = qGlobal.remove(r);
                 if (d != null) {
                     qLocal.put(r, d);
+
```
ThomasHaas, line 252: Why the added whitespace?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/TupleSet.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public int size() {
+        return data.values().stream()
+                .map(Set::size)
+                .reduce(0, Integer::sum);
+    }
+
+    public boolean contains(Event e1, Event e2) {
+        return data.containsKey(e1) && data.get(e1).contains(e2);
+    }
```
ThomasHaas, line 39: This code makes one redundant lookup. I don't think the code is used in a hot path, so feel free to keep it as is.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
natgavrilenko: removed

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/TupleSet.java
natgavrilenko: Fixed. These is still a room for optimizations, both in the TupleSet class and in the classes using it, but I don't want to spend time on this now as it is not the bottleneck.

natgavrilenko:
> The return of the `TupleSet` :). I feel like we should be more general here. Your `TupleSet` class really doesn't care about the fact that it contains events, it might as well be a `TupleSet<T>` with a generic parameter. On top of that, I think rather than being a set, this class represents more of a directed graph in the form of adjacency sets. And I think the fact that we deal with a graph is the most important aspect for future development. Ideally, I imagine an interface like `DirectedGraph<T>` with a straightforward implementation (your `TupleSet` but maybe with a better name), but also with the possibility to implement differently (e.g., implicitly) represented graphs. This would align nicely with #437 where the goal was to get rid of the explicit representation of some of the large base relations like `ext`, which can be represented in constant space instead of quadratic space. Changing the code to now use an explicit implementation everywhere clashes with that idea.

We don't use this TupleSet / DirectedGraph for anything other that Event, so no need to add generics here. But it should be easy to change when / if needed at some point. 

Regarding the interface, yes, it will be needed for the complement representation of sets (like you suggest for "ext"). Maybe also some implementation with sorted entries for SymmetryEncoder.

For the name, I am fine with DirectedGraph. Do you want me to rename it?

ThomasHaas:
The concept of directed graphs is universally useful, so having a generic interface floating around might encourage more usages in the future. This only applies to the interface; you could have a generic interface `DirectedGraph<T>` with a concrete implementation `EventGraph implements DirectedGraph<Event>` (~TupleSet).
That being said, for this PR I would be fine with (i) having some interface (generic or just for events) for (event) graphs and (ii) renaming `TupleSet` (and `EmptyTupleSet`) to something graph-related and implementing the aforementioned interface
It would probably be fine to have your IDE automatically extract an interface from `TupleSet`.

Some (incomplete) naming suggestions:
(1) `interface (Directed)EventGraph`, `class ExplicitEventGraph implements (Directed)EventGraph` or `class SimpleEventGraph implemens Directed)EventGraph` (or something similar).
(2) `interface DirectedGraph<T>`, `class EventGraph implements DirectedGraph<Event>`

natgavrilenko:
> Also, do you have any estimates on how much memory this saves on your larger benchmarks? I assume it also comes with performance improvements due to the smaller memory footprint.

Here are the numbers (for the same benchmark as before):
Regular RA: original 806 sec, optimizations 352 sec, optimizations+TupleSet 274 sec.
Extended+Regular RA: original 2264 sec, optimizations 1304 sec, optimizations+TupleSet 902 sec.
Memory usage: original/optimizations 7.4GB, optimizations+TupleSet 3,2GB.

The main motivation for the TupleSet was the memory consumption. In the earlier version, I wasn't able to run the a bigger benchmark (the same test with unrolling bound 2), because RA was crashing with OOM, given 12GB of memory. This was only for the initial RA (the mandatory part for the base relations) crashing before the middle. Now Dartagnan passes the initial RA in 108 seconds (though not able to solve it). Regarding the solving times, this is just a preliminary info, I will need to run more variations and investigate this problem further later on.

Overall, based on memory logs for individual relations, the "tupleset" version requires about 2.5 times less memory compared to the older version (not that I did comparison with the older version as-is, with partially reused tuple objects).

ThomasHaas:
The improvement sounds good.

I see that you did not add an interface, but I guess that is fine for now. We will do it when needed.

natgavrilenko:
> The concept of directed graphs is universally useful, so having a generic interface floating around might encourage more usages in the future. This only applies to the interface; you could have a generic interface `DirectedGraph<T>` with a concrete implementation `EventGraph implements DirectedGraph<Event>` (~TupleSet). That being said, for this PR I would be fine with (i) having some interface (generic or just for events) for (event) graphs and (ii) renaming `TupleSet` (and `EmptyTupleSet`) to something graph-related and implementing the aforementioned interface It would probably be fine to have your IDE automatically extract an interface from `TupleSet`.
> 
> Some (incomplete) naming suggestions: (1) `interface (Directed)EventGraph`, `class ExplicitEventGraph implements (Directed)EventGraph` or `class SimpleEventGraph implemens (Directed)EventGraph` (or something similar). (2) `interface DirectedGraph<T>`, `class EventGraph implements DirectedGraph<Event>`

Renamed to EventGraph and EmptyEventGraph, but let's keep them class + child class until we really need an interface. The same for generics, it can be easily added when really needed.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -959,60 +915,55 @@ private Knowledge computeInternalDependencies(Set<UsageType> usageTypes) {
...
                 for (Event e2 : events) {
-                    if (exec.areMutuallyExclusive(e1, e2)) {
-                        continue;
-                    }
                     Thread thread1 = e1.getThread();
                     Thread thread2 = e2.getThread();
                     if (specificScope != null) { // scope specified
-                        if (thread1.getScopeHierarchy().sameAtHigherScope(thread2.getScopeHierarchy(), specificScope)) {
-                            must.add(new Tuple(e1, e2));
+                        if (thread1.getScopeHierarchy().sameAtHigherScope(thread2.getScopeHierarchy(), specificScope) && !exec.areMutuallyExclusive(e1, e2)) {
```
ThomasHaas, line 938: You have overwritten some changes recently done by @tonghaining in #516. This mutual exclusion check was refactored together with the one further below.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -1021,38 +972,37 @@ public Knowledge visitSyncBarrier(Relation sync_bar) {
...
         public Knowledge visitSyncFence(Relation sync_fen) {
-            Set<Tuple> may = new HashSet<>();
+            EventGraph may = new EventGraph();
             List<Event> fenceEventsSC = program.getThreadEventsWithAllTags(VISIBLE, FENCE, PTX.SC);
             for (Event e1 : fenceEventsSC) {
                 for (Event e2 : fenceEventsSC) {
-                    if (!exec.areMutuallyExclusive(e1, e2)) {
-                        may.add(new Tuple(e1, e2));
+                    if (e1.hasTag(Tag.PTX.SC) && e2.hasTag(Tag.PTX.SC) && !exec.areMutuallyExclusive(e1, e2)) {
+                        may.add(e1, e2);
```
ThomasHaas, line 980: Same as above: you undid changes of #516. Please make sure that no changes get undone.

natgavrilenko: Oops, good that you noticed it. Some weird automatic merging effect that I didn't catch. Fixed.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
natgavrilenko: It is semantically the same as before: the mutually exclusive check was in the beginning of the "for" loop. But the tonghaining's version is better, so changed to this to his version.

ThomasHaas: Yes, semantically they are identical. I think none of the changes in #516 affected semantics.

ThomasHaas (approved):
LGTM.

PR #525, opened 2023-10-03
PTX examples

hernanponcedeleon:
This PR adds a few more PTX examples and support for some new required PTX instructions.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPTX.java
```diff
@@ -33,6 +35,30 @@ public List<Event> visitPtxAtomOp(PTXAtomOp e) {
         );
     }
 
+    // PTX CAS semantics from
+    // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-atom
+    // atomic {
+    //   d = *a;
+    //   *a = (*a == b) ? c : *a
+    // }
```
ThomasHaas, line 43: `d` is not used. Is that a typo or is `d` supposed to be the return value of the CAS (if so, make that more clear)?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPTX.java
```diff
@@ -33,6 +35,30 @@ public List<Event> visitPtxAtomOp(PTXAtomOp e) {
...
+    @Override
+    public List<Event> visitPtxAtomCAS(PTXAtomCAS e) {
+        Register resultRegister = e.getResultRegister();
+        String mo = e.getMo();
+        Expression address = e.getAddress();
+        Expression expected = e.getExpectedValue();
+        Expression newValue = e.getStoreValue();
+        Expression storeValue = expressions.makeConditional(expressions.makeBinary(resultRegister, COpBin.EQ, expected),
+                newValue, resultRegister);
```
ThomasHaas, line 52: You can use the `makeEQ` method (not sure about the name). Also, this code does not align with the pseudo code, but from looking in the PTX docu, I think their pseudo code is misleading.
The pseudo code suggests that there is a fresh load `*a` happening instead of reusing the value in `d`. But it surely makes more sense to reuse the already loaded value.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPTX.java
hernanponcedeleon: `d` is the destination/return register. I can add a comment.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPTX.java
hernanponcedeleon: I don't think we should precisely follow the pseudo code. I just added it because the PTX CAS is slightly different than the other ones we have in the sense that it always performs the Store (it the comparison fails, it simple stores the old value).

ThomasHaas: Are you sure that the ternary operator is meant as a `select/cmov` or as a proper branching? This makes a difference in dependencies.

hernanponcedeleon: The documentation of atomic operations (which includes `cas`) states
```
Atomically loads the original value at location a into destination register d, performs a reduction operation with operand b and the value in location a, and stores the result of the specified operation at location a, overwriting the original value.
```
As I understand this, the store **is always executed**. 
If that is the case, branching makes nonsense to me.

ThomasHaas:
```
d = *a;
if (d == b)
  *a = c;
else 
  *a = d;
```
This is still feasible code that always stores. Though I agree that the code is most likely referring to a `cmov` (I just wanted to be sure).

dartagnan/src/main/antlr4/LitmusPTX.g4
```diff
@@ -108,13 +108,33 @@ storeRegister
...
+localAdd
+    :   Add register Comma register Comma constant
+    ;
+
+localSub
+    :   Sub register Comma register Comma constant
+    ;
+
+localMul
+    :   Mul register Comma register Comma constant
```
tonghaining, line 131: Should it be `Mult`?

hernanponcedeleon: The instruction is called `mul` in PTX ISA. Also, the only other grammar we have with a similar rule (LLVM) also uses `Mul`

tonghaining: Okay, but I think the [token](https://github.com/hernanponcedeleon/Dat3M/blob/ptx-examples/dartagnan/src/main/antlr4/LitmusPTX.g4#L336) used here was named as `Mult` in this file?

hernanponcedeleon: There is definitely a mismatch between our syntax and the PTX one (not only on mult, but also plus and minus). The alloy tests seems to use the correct `add, dec ...` syntax. 

@tonghaining do you remember why you opted for using `plus, minus, mult ...` in our grammar?

hernanponcedeleon: I updated the grammar and the tests

tonghaining: > There is definitely a mismatch between our syntax and the PTX one (not only on mult, but also plus and minus). The alloy tests seems to use the correct `add, dec ...` syntax.
> 
> @tonghaining do you remember why you opted for using `plus, minus, mult ...` in our grammar?

No I don't remember, I also think we should update them

hernanponcedeleon:
Anybody has further comments, or can I merge?

ThomasHaas (approved):
LGTM

PR #526, opened 2023-10-04
Update README (optimizations flags + move Natalia to active developer), remove atomic-replace code

hernanponcedeleon:
- #521 removed the references in the README about the `ATOMIC_REPLACE_OPTS` which we still use. Put it back and renamed variable.
- Removed the unneeded atomic-replace code.
- Moved Natalia to active developers list.


natgavrilenko (commented):
Could you use a more meaningful commit name, please? "Updates" doesn't tell much.

PR #528, opened 2023-10-06
Coherence is not total in Vulkan

hernanponcedeleon:
By reading [Vulkan documentation](https://registry.khronos.org/vulkan/site/spec/latest/appendices/memorymodel.html#memory-model-memory-semantics), I noticed coherence is also not total there.

> Non-atomic memory operations, by default, may be observed by one agent in a different order than they were written by another agent.

I confirmed this with their alloy model using the following test
```
NEWWG
NEWSG
NEWTHREAD
st.sc0 x = 1
st.sc0 x = 2
NEWWG
NEWSG
NEWTHREAD
ld.sc0 x = 2
ld.sc0 x = 1
SATISFIABLE consistent[X]
```
I added Vulkan to our list of models where `co` is not total and added the test above (in our format) for regression.

I also noticed that we need to recompute `fr` based on the new definition of `co`, also in PTX.


ThomasHaas:
The Vulkan model already defined `asmo` in place of coherence, which is only total on the atomic stores.
The only pure usage of `co` is in the atomicity axiom where it might very well be intended to make RMWs atomic w.r.t. to non-atomics (not sure if this is true in Vulkan). In other words, the coherence on non-atomic stores is mostly ignored by the model!
The only pure usages of `fr` are in the atomicity axiom and the SC-per-loc (concistency-cycle) axiom.
Again, the first usage might very well be correct, but the second one is actually what breaks your test.
I can imagine that one needs to have something like `afre = rf^-1;asmo` be used there to get a `fr` relation only concerning atomics stores.

ThomasHaas:
I think the formulation of the scoped coherence in Vulkan is quite nice since it does not require updating the underlying total coherence order. I think the same can be done for PTX by just defining `co = ((co & morally-strong) | ([W];cause;[W] & loc))+`.

ThomasHaas:
PTX mentions that `co` is non-total but that does not collide with my suggestion. We can still start from a total `co` and then derive a partial order from that one by selectively ignoring certain orderings. Right now we also start from a wrong coherence order (containing too few edges and even failing to be transitive) and then fix it on the CAT side, so it is not morally any different.
In fact, I believe we could avoid changing our anarchic semantics if we just define ALL properties in CAT using a "free relation" (i.e., one that can be interpreted arbitrarily):
```
let co = ([W];free;[W] & loc)+ // co is ANY transitive relation that relates same-address writes
acyclic (co) // co is acyclic
empty (([W];morally-strong;[W] & loc) \ (co | co^-1)) // co is total on morally-strong writes
empty (([W];cause;[W] & loc) \ co) // co relates causally-ordered writes (even if they are not morally strong)
```
This is a lot better than what we have right now! So my suggestion is to either take the proper total `co` and build a partial order from that one (like Vulkan) or just start from a totally free relation and impose axioms to get the partial order.
The last formulation is even more truthful to the alloy model than what we have right now.

cat/spirv.cat
```diff
@@ -57,10 +57,11 @@ let avvisinc = ((avvisSem | avvisSem^-1) \ id) |
...
 let posemtosc = ((SEMSC0 * SC0) | (SEMSC1 * SC1)) & po
 
-let asmo = co & (ATOM * ATOM)
+let asmo = co & ((ATOM | IW) * ATOM)
 // From-read is defined based on scoped modification order
-// instead of coherence
-let fr = rf^-1; asmo
+// instead of coherence. For reads that read-from init, they
+// from-read all writes at the same location.
+let fr = (rf^-1; asmo) | (([IW]; rf)^-1; co)
```
ThomasHaas, line 64: Can't you define `let asmo = co & (ATOM * ATOM | IW * W)`? Then you don't need to change `fr`.

hernanponcedeleon: I don't think that matches the specification

> For a given atomic write A, all atomic writes that are mutually-ordered with A occur in an order known as A’s scoped modification order. A’s scoped modification order relates no other operations.

Precisely, the "relates no other operations" part.

Even if it were, I still need to redefine `fr`. Also, this way [is closer to the alloy model.](https://github.com/KhronosGroup/Vulkan-MemoryModel/blob/66c31d9cfcc03b1f5231779fbf3c42dc983b62e2/alloy/spirv.als#L308-L312)

ThomasHaas: By "not change `fr`" I meant that you can keep `fr = rf^-1;asmo` not the original one of course.
However, I agree with not having non-atomic writes in `asmo`. Do you still think that `IW * ATOM` should be there, i.e., should initial writes be considered atomic or not? They probably are, but then we should maybe just define `let ATOM = ATOM | IW` to make init writes appear atomic everywhere?

Lastly, why don't we define `let fr = (rf^-1; asmo) | (([IW]; rf)^-1; loc;[W] \ id)` to really match the alloy model? It also avoids this awkward usage of `co`.
It could also be refactored to `let fr = rf^-1;(asmo | ([IW];loc;[W] \ id))` which I think is nicer, but the above version is a little closer to the original alloy model.

hernanponcedeleon: > They probably are, but then we should maybe just define let ATOM = ATOM | IW to make init writes appear atomic everywhere?

I think they probably are there, but only for those variables having at least once atomic access. If we use `ATOM = ATOM | IW`, we would be even adding initial stores for variables that are only accessed as non-atomic. Using `asmo = co & ((ATOM | IW) * ATOM)` we don't have this problem.

> Lastly, why don't we define let fr = (rf^-1; asmo) | (([IW]; rf)^-1; loc;[W] \ id) to really match the alloy model?

That might be the best. I wrote the other version manually simply following the "For reads that read-from init, they from-read all writes at the same location."

hernanponcedeleon:
> co = ((co & morally-strong) | ([W];cause;[W] & loc))+

I am still giving this one a though. However, I wanted to test it first and I cannot even make it pass the tests.

ThomasHaas:
have you made sure that the co is total though? If you still use the partial order co, it won't work.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
I will merge this PR as it is and try to think about a solution for the queries referring to locations problems. The alloy-based tool does not allow to use locations in the final conditions, but it might be exactly for this reason.

PR #531, opened 2023-10-09
Update Ptx constant-proxy rule

tonghaining:
This PR contains 2 updates:
1. As verified in PTX [document](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-ld), `weak` instructions don't hold the scope tag. It is removed from the visitors now.
2. We found the addition of the constant-proxy rule will no longer be necessary following the modification that explicitly defines locations as Constant virtually aliasing to Generic locations. There is a corresponding [pull request](https://github.com/NVlabs/mixedproxy/pull/7) toward the original PTX alloy approach. We could await confirmation on whether it is appropriate to proceed with this change.


hernanponcedeleon:
There has not been any answer regarding the PR to the repository with the alloy model. However, the model in this PR is more faithful to their model, so I will merge.

PR #536, opened 2023-10-18
Support for Llvm intrinsics: lifetime and expect

ThomasHaas:
This PR adds support for `llvm.lifetime` intrinsics (we just ignore them) and `llvm.expect(a, b)`.
The latter just returns its first parameter `a` but tells the optimizer that it likely evaluates to `b`. So we just replace the call by its first parameter. This call can be generated from `assert(x == val)`, because the compiler will be able to expect that `x == val` holds in the subsequent code (it's like a soft form of `assume`?).


hernanponcedeleon:
For verification purposes, I don't think we can make use of `assumes`. If this intrinsic is generated after an `assert` and we add an `assume`, we would filter executions where the assertion does not hold.

Other than the comment, LGTM.

ThomasHaas:
The assertion failure terminates the thread (usually), so the assume would not get executed. Anyway, my comment was about the LLVM compiler and the reason why it generates `expect` statements. We do not translate those to assumes in Dartagnan.

PR #540, opened 2023-10-23
Clean ptx vulkan tests

tonghaining:
This PR organizes the tests files and moves the manual tests out of the original folder.
Some missing tests of the PTX model are also added.


hernanponcedeleon:
We don't really need the `DR/Manual` folder.

For the tests coming from the vulkan repo, we needed to divide between `KronosGroup` and `DR` because we were using their expected results to know what to use in data race detection. Their expected results were based on this `filter` semantics (i.e., they only checked if the filtered executions contained a race, but it could be the case that they say "no, there is no race", but in reality, there is a race in the non-filter ones).

For the tests we manually wrote, we care about checking races in every execution, thus we can simply use the litmus with the `exists` query.

dartagnan/src/test/resources/VULKAN-DR-expected.csv
```diff
@@ -68,14 +68,3 @@ litmus/VULKAN/DR/test6-filter.litmus,1
...
-litmus/VULKAN/DR/Manual/ticketlock-diff-wg.litmus,0
-litmus/VULKAN/DR/Manual/ticketlock-acq2rlx-1.litmus,1
-litmus/VULKAN/DR/Manual/ticketlock-acq2rlx-2.litmus,0
-litmus/VULKAN/DR/Manual/ticketlock-rel2rlx.litmus,0
-litmus/VULKAN/DR/Manual/MP-avvis.litmus,1
-litmus/VULKAN/DR/Manual/MP-no-avvis.litmus,0
-litmus/VULKAN/DR/Manual/MP-mesa.litmus,1
-litmus/VULKAN/DR/Manual/MP-mesa-load-acq.litmus,1
-litmus/VULKAN/DR/Manual/MP-mesa-fence-loop.litmus,1
-litmus/VULKAN/DR/Manual/MP-mesa-optimized.litmus,1
```
hernanponcedeleon: We still want to test those benchmarks. It is ok you removed the files, but in the expected file you need to use the `VULKAN/Manual` tests

hernanponcedeleon:
@tonghaining can you please resolve the conflicts so the CI starts?

hernanponcedeleon:
@tonghaining  I don't understand your last commit. Why did you remove certain tests?

tonghaining:
> @tonghaining I don't understand your last commit. Why did you remove certain tests?

Those test files don't exist, for example `SB+bar-dyn-reg-const.litmus` is locates under the Manual/. I forget where did those lines come.

PR #542, opened 2023-10-25
Add whiles to C litmus parser

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -225,6 +226,26 @@ public Object visitIfExpression(LitmusCParser.IfExpressionContext ctx) {
         return null;
     }
 
+	@Override
+    public Object visitWhileExpression(LitmusCParser.WhileExpressionContext ctx) {
+        Expression expr = (Expression) ctx.re().accept(this);
+
+        whileId++;
+        Label headL = programBuilder.getOrCreateLabel(currentThread,"head_" + whileId);
+        Label endL = programBuilder.getOrCreateLabel(currentThread,"end_" + whileId);
```
ThomasHaas, line 235: Do we have a `.createLabel/.newLabel` method? If so, we should prefer that over `getOrCreate` because the call should never return an existing label.

hernanponcedeleon: We don't have such utility in the program builder. However, we will never get an existing Label because we increase the `whileId` at every call

ThomasHaas: I understand that we never get a duplicate label, which is why I wanted to make that explicit in the choice of function call. But if we do not have such a function `ProgramBuilder`, then that's fine.

ThomasHaas (approved):
LGTM

PR #543, opened 2023-10-26
Fix error in C litmus visitor for while rule

hernanponcedeleon:
There was an error in #542: the while condition can generate events and thus needs to be inside the body. Also, the back jumping to head was missing.


ThomasHaas (approved):
LGTM

PR #546, opened 2023-10-30
More tests about coherence for both PTX and Vulkan

hernanponcedeleon:


litmus/PTX/Manual/Coherence-weak.litmus
```diff
@@ -0,0 +1,12 @@
+PTX Coherence-weak
+"Test that threads 3 and 4 can't read updates to x in conflicting order"
```
tonghaining, line 2: should it be `Test that threads 3 and 4 CAN read updates to x in conflicting order`?

hernanponcedeleon: Good catch. I copied the `asmo.litmus` file and simply update the instructions. However I like to have the same comment for similar tests which use weaker or stronger instructions, this I will change the comment in all files to "Test **IF** threads 3 and 4 can't read updates to x in conflicting order"

PR #550, opened 2023-11-03
Memory Datatype and Pointer type

xeren:
Adds datatypes to memory objects, as well as the opaque pointer type to the IR.  Removes the preprocessing step that replaces GEP expressions with simple additions.
To infer the datatype of dynamically-allocated objects, the program is scanned for occurrences of that address.  If those occurrences do not agree on one type, it defaults to byte.  The current implementation is unsound, as it ignores registers and the base address being stored in memory, where another reader may assume another type.  The datatype currently controls initializations.
The pointer type will allow the encoder to choose a different theory for address expressions.  The IR will be closer to the LLVM IR and should become more readable.

More:
- Since memory objects and functions cannot keep being implementations of `IConst` anymore, the interface is removed entirely.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/SparseConditionalConstantPropagation.java
```diff
@@ -67,8 +70,10 @@ public static SparseConditionalConstantPropagation fromConfig(Configuration conf
     @Override
     public void run(Function func) {
         final Predicate<Expression> checkDoPropagate = propagateCopyAssignments
-                ? (expr -> expr instanceof IConst || expr instanceof BConst || expr instanceof Register)
-                : (expr -> expr instanceof IConst || expr instanceof BConst);
+                ? (expr -> expr instanceof IValue || expr instanceof MemoryObject || expr instanceof Function ||
+                        expr instanceof BConst || expr instanceof Register)
+                : (expr -> expr instanceof IValue || expr instanceof MemoryObject || expr instanceof Function ||
+                        expr instanceof BConst);
```
ThomasHaas, line 76: I think we can refactor the common check into a `isConstant` method.

xeren:
With how big this has gotten again, and the number of issues, I will close this PR.

PR #552, opened 2023-11-04
Fix relation analysis for virtual location - fixes #551

### LINKED ISSUE #551: Must set of virtual location
I think the relation analysis for virtual location is wrong due to the use of `alias.mayAlias`. This has not cause problems so far because for those litmus tests, may and must alias probably return the same.
I think the code should be
```
        @Override
        public Knowledge visitVirtualLocation(Relation rel) {
            EventGraph must = new EventGraph();
            EventGraph may = new EventGraph();
            List<MemoryCoreEvent> events = program.getThreadEvents(MemoryCoreEvent.class);
            for (MemoryCoreEvent e1 : events) {
                for (MemoryCoreEvent e2 : events) {
                    if (sameGenericAddress(e1, e2) && !exec.areMutuallyExclusive(e1, e2)) {
                        if (alias.mustAlias(e1, e2)) {
                            must.add(e1, e2);
                        }
                        if (alias.mayAlias(e1, e2)) {
                            may.add(e1, e2);
                        }
                    }
                }
            }
            return new Knowledge(must, may);
        }
```

hernanponcedeleon:


ThomasHaas (approved):
LGTM

PR #554, opened 2023-11-07
Add 3 tests from paper "heterogeneous race free memory models"

tonghaining:


litmus/PTX/Manual/MP-3-transitive.litmus
```diff
@@ -0,0 +1,11 @@
+PTX MP-3-transitive
+"Adapted from Figure 6 of paper: Heterogeneous-race-free Memory Models"
+{
+x=0; y=0; z=0;
+P2:r0=0; P2:r1=0; P2:r2=0;
```
hernanponcedeleon, line 5: `P1:r0=0`

litmus/PTX/Manual/MP-3-transitive.litmus
```diff
@@ -0,0 +1,11 @@
...
+}
+ P0@cta 0,gpu 0         | P1@cta 0,gpu 0         | P2@cta 0,gpu 1         ;
+ st.weak x, 1           | ld.acquire.cta r0, y   | ld.acquire.gpu r1, z   ;
+ st.release.cta y, 1    | st.release.gpu z, 1    | ld.weak r2, x          ;
+exists
+(P1:r0 == 1 /\ P2:r1 == 1 /\ P2:r2 == 1)
```
hernanponcedeleon, line 11: Shouldn't it be `P2:r2 != 1`?

litmus/PTX/Manual/MP-3-transitive.litmus
```diff
@@ -0,0 +1,11 @@
+PTX MP-3-transitive
+"Adapted from Figure 6 of paper: Heterogeneous-race-free Memory Models"
+{
+x=0; y=0; z=0;
+P2:r0=0; P2:r1=0; P2:r2=0;
+}
+ P0@cta 0,gpu 0         | P1@cta 0,gpu 0         | P2@cta 0,gpu 1         ;
```
hernanponcedeleon, line 7: I think it should be `P2@cta 1,gpu 0`.
The release-acquire sequence between `P1` and `P2` has `gpu` scope so both threads must be in the same `gpu 0` (and we need to palce them in different `cta` otherwise the `gpu` scope would not be necessary)

litmus/PTX/Manual/MP-3-transitive.litmus
tonghaining: Updated

litmus/PTX/Manual/MP-3-transitive.litmus
tonghaining: Then the assertion should be `~exists (P1:r0 == 1 /\ P2:r1 == 1 /\ P2:r2 != 1)`? to be consistent with the paper's assertion "R3 will get value 1"

litmus/PTX/Manual/MP-3-transitive.litmus
tonghaining: Yes, updated.

litmus/PTX/Manual/MP-3-transitive.litmus
hernanponcedeleon: In my head I find it easier to think about `exists` conditions than `~exists` ... but at the end of the day, we can use one or the other and have 1 or 0 in the expected file.

hernanponcedeleon (approved):

PR #556, opened 2023-11-08
Preliminary support for llvm.objectsize & __memset_chk

ThomasHaas:
This PR adds very rudimentary support for the intrinsics `llvm.objectsize` and `__memset_chk`.

`llvm.objectsize(ptr, ...)` returns the size of an object pointed to by `ptr` but may return 0 or -1 if the size cannot be statically determined. For now, I just always assumed the second case, so `llvm.objectsize` gets replaced by 0/-1 no matter what.
There is a dedicated pass (rather than `Inlining`) that makes this replacement, because in principle this replacement would require a global analysis to determine the object sizes (*).

`__memset_chk` is just a regular memset that additionally checks if the range of the memset is within a given bound.
If it is not, `__memset_chk` is supposed to abort the program.
However, the bound parameter is given by `llvm.objectsize` which for now always returns 0/-1 and hence the memset would always be out-of-bounds (**).
So what we do for now is to just ignore the bounds check and treat `__memset_chk` as identical to `memset`.

(*) I originally implemented an analysis that traverses use-def-chains to find the allocation site of the pointer. However, a correct implementation also has to track pointer offsets on the way, and this makes the analysis quite complicated especially when the array indexing is involved.

(**) I do not understand what llvm does if it fails to resolve `llvm.objectsize` because then  `__memset_chk` cannot possibly work. Unfortunately, the documentation of `llvm.objectsize` and its interaction with `__memset_chk` is not well-documented.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveLLVMObjectSizeCalls.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveLLVMObjectSizeCalls.java
@@ -1,0 +1,61 @@
+package com.dat3m.dartagnan.program.processing;
+
+import com.dat3m.dartagnan.expression.ExpressionFactory;
+import com.dat3m.dartagnan.expression.IValue;
+import com.dat3m.dartagnan.expression.type.IntegerType;
+import com.dat3m.dartagnan.program.Function;
+import com.dat3m.dartagnan.program.event.EventFactory;
+import com.dat3m.dartagnan.program.event.core.Event;
+import com.dat3m.dartagnan.program.event.functions.ValueFunctionCall;
+import org.sosy_lab.common.configuration.Configuration;
+
+import java.math.BigInteger;
+
+/*
+    https://llvm.org/docs/LangRef.html#llvm-objectsize-intrinsic
+    A call to llvm.objectsize(ptr, ...) returns the size of the (accessible) object pointed to by <ptr>.
+    For example, if ptr_base is returned by alloc(100), then llvm.objectsize(ptr_base) == 100,
+    but llvm.objectsize(ptr_base + 80) == 20 (~ there are only 20 accessible bytes at that offset).
+    If the size cannot be determined statically, the call is replaced by either 0 or -1.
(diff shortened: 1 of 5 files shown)


hernanponcedeleon:
I guess that a case where the `memset` parameter is out-of-bound is UB, right? So the current implementation is sound only if the program has not UB (which is an assumption we made in many other places).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveLLVMObjectSizeCalls.java
```diff
@@ -0,0 +1,61 @@
...
+
+    NOTE: For simplicity, we will always assume the object size to be statically unknown and return -1/0.
+
+    FIXME: A correct implementation needs to traverse the use-def chain from a llvm.objectsize call back to
+     the allocation site while tracking offset computations.
+     Unfortunately, this is quite hard to do in the presence of array indexing which can result in dynamic offsets.
+     A way around this is to rely on unrolling + CP to resolve dynamic offsets, then run this pass, and finally
+     run a second CP to propagate the resolved object sizes.
+ */
+public class ResolveLLVMObjectSizeCalls implements FunctionProcessor {
```
hernanponcedeleon, line 31: We should at least log a warning

ThomasHaas:
With `__memset_chk` it would not be UB, because the program would terminate before actually doing the UB operation.
On the other hand, the source language does use the unchecked `memset`, so the while the LLVM would be "fine" the source code would have UB I think.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveLLVMObjectSizeCalls.java
ThomasHaas: I don't think a warning here would be helpful because there is not a single case (for now) where this handling is unsound. After all, it is valid for LLVM to reduce the call to 0/-1 according to the specification.
I would rather add a warning to `__memset_chk` that shows we skipped a bound check.

hernanponcedeleon: You are right ... The warning there makes more sense

PR #557, opened 2023-11-08
Fix devirtualization

ThomasHaas:
`NaiveDevirtualization` used to throw an exception if it could not find any valid call target for a dynamic call.
However, this is no problem if the dynamic call is unreachable anyway (i.e., it is dead code). This happens commonly in bigger code bases where the client-code only exercises a small subset of the functions, and never those that make use of function pointers.
Therefore, we just generate a warning rather than an exception.

Also, `NaiveDevirtualization` now places a proper assertion in the code for the case where the provided function pointer is invalid.

Lastly, I added `#Annotations` to the program statistics output, because those inflate the number of events


hernanponcedeleon:
If for reachable code, we cannot find any valid call target for a dynamic call, we will return FAIL as the result of the verification, am I right?

ThomasHaas:
Yes, if we are checking for assertion violations. The error message will be "invalid function pointer" rather than "user assertion", so the user can see that the error is not due to his assertions.

PR #565, opened 2023-11-12
Parse poison values as non-deterministic values

hernanponcedeleon:
The [LLVM documentation](https://llvm.org/docs/LangRef.html#poison-values) states that "It is correct to replace a poison value with an undef value or any value of the type.", thus we parse poision values as non-deterministic values.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
@@ -868,9 +868,7 @@ public class VisitorLlvm extends LLVMIRBaseVisitor<Expression> {
     @Override
     public Expression visitPoisonConst(PoisonConstContext ctx) {
         // It is correct to replace a poison value with an undef value or any value of the type.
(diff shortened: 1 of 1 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -859,6 +859,14 @@ public Expression visitArrayConst(ArrayConstContext ctx) {
         return expressions.makeArray(elementType, arrayValues, true);
     }
 
+    @Override
+    public Expression visitPoisonConst(PoisonConstContext ctx) {
+        // It is correct to replace a poison value with an undef value or any value of the type.
+        BooleanType booleanType = types.getBooleanType();
+        var nondeterministicExpression = new BNonDet(booleanType);
+        return expressions.makeCast(nondeterministicExpression, expectedType);
```
ThomasHaas, line 867: Why do you always use a boolean type? I think you should use the correct type. 
Btw. I think it would also be fine to use a constant value like `0` (we do this for `undef`).

ThomasHaas (approved):
LGTM

PR #566, opened 2023-11-12
Add support for sleep intrinsic as a nop

hernanponcedeleon:
@ThomasHaas can you double check that the `alwaysReturn:true` parameter makes sense?


ThomasHaas (approved):
LGTM.
I think it is fair to say that `sleep` always returns (btw. we do not make use of that information right now).

PR #568, opened 2023-11-13
Support for GEP in init values & pointer subtraction in alias analysis

ThomasHaas:
This PR fixes the following two problems.

(1) #567 added support for parsing GEPs in initializers of globals, however, our `GEPToAddition` didn't lower those GEPs to additions which caused exceptions during encoding (cause GEPs were not encodable).

(2) When dealing with code that uses `container_of`-magic, the IR ends up having pointer subtractions (e.g. `ptr - 12`).
Our alias analysis wrongly handled those (it was not just imprecise but also incorrect!).
I added handling for subtraction. @xeren Can you check that my simple fix makes sense?


xeren (approved):
LGTM

PR #569, opened 2023-11-13
Update morally-strong to follow HRF-indirect and add test

tonghaining:


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -606,10 +606,11 @@ public Void visitSyncFence(Relation syncFence) {
                 for (Event z : allFenceSC.subList(i + 1, allFenceSC.size())) {
                     String scope1 = Tag.getScopeTag(x, program.getArch());
                     String scope2 = Tag.getScopeTag(z, program.getArch());
-                    if (!scope1.equals(scope2) || scope1.isEmpty()) {
+                    if (scope1.isEmpty() || scope2.isEmpty()) {
                         continue;
                     }
-                    if (!x.getThread().getScopeHierarchy().sameAtHigherScope((z.getThread().getScopeHierarchy()), scope1)) {
+                    if (!x.getThread().getScopeHierarchy().sameAtHigherScope((z.getThread().getScopeHierarchy()), scope1) ||
+                            !z.getThread().getScopeHierarchy().sameAtHigherScope((x.getThread().getScopeHierarchy()), scope2)) {
```
hernanponcedeleon: Do we really need both checks? Isn't the check symmetric?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -938,13 +938,14 @@ public Knowledge visitSameScope(Relation rel, String specificScope) {
...
-                        if (thread1.getScopeHierarchy().sameAtHigherScope(thread2.getScopeHierarchy(), specificScope)) {
+                        if (thread1.getScopeHierarchy().canSyncAtScope(thread2.getScopeHierarchy(), specificScope)) {
                             must.add(e1, e2);
                         }
                     } else {
                         String scope1 = Tag.getScopeTag(e1, program.getArch());
                         String scope2 = Tag.getScopeTag(e2, program.getArch());
-                        if (scope1.equals(scope2) && !scope1.isEmpty() && thread1.getScopeHierarchy().sameAtHigherScope(thread2.getScopeHierarchy(), scope1)) {
+                        if (!scope1.isEmpty() && !scope2.isEmpty() && thread1.getScopeHierarchy().canSyncAtScope(thread2.getScopeHierarchy(), scope1)
+                                && thread2.getScopeHierarchy().canSyncAtScope(thread1.getScopeHierarchy(), scope2)) {
```
hernanponcedeleon, line 948: Do we really need both checks? Isn't the check symmetric?

litmus/PTX/Manual/MP-cta1.litmus
```diff
@@ -0,0 +1,13 @@
+PTX MP-cta1
```
hernanponcedeleon, line 1: `MP-cta-gpu` is probably a better name

hernanponcedeleon:
I did some cleanup to the `ScopeHierarchy` class. @tonghaining can you take a look that everything makes sense?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
tonghaining: I think so, because a morally strong pair requires that `each operation is strong and specifies a scope that includes the thread executing the other operation`. `thread1 \in scope2` doesn't guarantee `thread2 \in scope1`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopeHierarchy.java
```diff
@@ -40,27 +39,27 @@ public int getScopeId(String scope) {
         return scopeIds.getOrDefault(scope, -1);
     }
 
-    // For each scope S higher than flag, we check both events are in the same scope S
-    public boolean sameAtHigherScope(ScopeHierarchy thread, String flag) {
-        if (!this.getClass().equals(thread.getClass()) || !this.getScopes().contains(flag)) {
+    // For any scope higher than the given one, we check both threads have the same scope id.
```
tonghaining, line 42: Maybe "// For any scope higher than the given one, we check both **ScopeHierarchy** have the same scope id."? Because the Thread class is not involved here.

litmus/PTX/Manual/MP-cta1.litmus
tonghaining: changed

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopeHierarchy.java
hernanponcedeleon: I think "thread" is a better explanation of what if going on at the high level. The scope hierarchy is an artifact of our implementation, but at the end of the day, what we really care is to know which is the scope id for a given thread.

tonghaining: Okay, yes "thread" makes more sense at the high level.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
hernanponcedeleon: Ok, I missed the fact that the second parameter is different, i.e. `scope1` vs `scope2`

hernanponcedeleon (approved):

PR #570, opened 2023-11-14
SVCOMP qualification

hernanponcedeleon:
SVCOMP qualification rules requires to "start the verification" for at least half of the tasks.
Thus, we should return UNKNOWN instead of throwing an exception for SVCOMP unsupported properties.

### DIFF
--- a/svcomp/src/main/java/com/dat3m/svcomp/SVCOMPRunner.java
+++ b/svcomp/src/main/java/com/dat3m/svcomp/SVCOMPRunner.java
@@ -29,47 +29,47 @@ import static java.lang.Integer.parseInt;
 @Options
 public class SVCOMPRunner extends BaseOptions {
 
-	private Property property;
-
-	@Option(
-		name= PROPERTYPATH,
-		required=true,
-		description="The path to the property to be checked.")
-	private void property(String p) {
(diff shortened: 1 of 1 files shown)


svcomp/src/main/java/com/dat3m/svcomp/SVCOMPRunner.java
```diff
@@ -76,30 +78,34 @@ public static void main(String[] args) throws Exception {
...
+            .filter(s->s.startsWith("-"))
+            .toArray(String[]::new);
+        Configuration config = Configuration.fromCmdLineArguments(argKeyword);
+        SVCOMPRunner r = new SVCOMPRunner();
+        config.recursiveInject(r);
+
+        if(r.property == null) {
+            System.out.println("UNKNOWN");
+            return;
+        }
```
ThomasHaas, line 101: Can we also log a warning that the input property is not supported so we can better distinguish between UNKNOWN due to unrolling and UNKNOWN due to unsupported property?

hernanponcedeleon: Done

ThomasHaas (approved):
LGTM

PR #571, opened 2023-11-15
Added MultiRegWriter interface

ThomasHaas:
`LlvmCmpXchg` is now a `MultiRegWriter`
Code that special-handled `LlvmCmpXchg` now special handles the more generic `MultiRegWriter` events.


ThomasHaas:
As it turned out that we do not need `MultiRegWriters` for now, I will close this.

PR #572, opened 2023-11-15
Add IntegerType.isPointer()

xeren:
This PR adds means to differentiate the types `ptr` and `bv64`.
This is a lightweight alternative to #550.  It does not forbid multiplications or bit operations on pointers, and does not require pointer expressions as addresses of memory events.  Assignments, however, raise an exception on type mismatch.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
@@ -98,11 +98,15 @@ public final class ExpressionFactory {
     }
 
     public Expression makeCast(Expression expression, Type type) {
+        return makeCast(expression, type, false);
+    }
+
+    public Expression makeCast(Expression expression, Type type, boolean signed) {
         if (type instanceof BooleanType) {
             return makeBooleanCast(expression);
         }
         if (type instanceof IntegerType integerType) {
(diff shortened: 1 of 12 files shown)


ThomasHaas:
After thinking about this for some time, I don't think we should start to flag integers as being pointer/no pointers.
One reason for a dedicated pointer type is the idea of address spaces, which might be necessary in the near future when dealing with OpenCL or CUDA, or maybe even just ASL.
I can also imagine that a distinction between physical and virtual pointers could be useful.
All these things should not be implemented on top of integers with extra attributes I believe.

I think one of the ugly parts about supporting pointer types is the GEP expression.
Maybe adding pointer types is not so bad if we avoid `GEP` for now and instead allow pointer-typed left operands in all binary expressions. So `ptr + 8` would still be valid and it would be of pointer type (the type of the binary expressions matches the left operand).
Alternatively, we can have something like `GEP` possibly without the type parameter (i.e., it we always use `byte` offsets), but still use the standard `makeAdd/makeSub/etc.` factory methods which could automagically generate the `GEP` if a pointer was provided instead of an int.

Either way, I think we need proper pointer types (eventually). To eaze in the implementation, I think it would be fine to still allow arbitrary integer expressions as addresses, so that the backend doesn't need to change immediately. We can then step-by-step introduce pointer types into use, temporarily allowing for both pointers and integers serving as addresses.

PR #575, opened 2023-11-21
Fix error of ExecutionModel when reading nondet variables

ThomasHaas:
Fixed `ExecutionModel` failing to read values of nondet variables if they were optimized away.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
@@ -23,6 +23,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.sosy_lab.common.configuration.InvalidConfigurationException;
 import org.sosy_lab.java_smt.api.BooleanFormula;
+import org.sosy_lab.java_smt.api.Formula;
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModel.java
```diff
@@ -337,9 +338,13 @@ private void addEvent(Event e, int globalId, int localId) {
...
 
             if (data.isRead() || data.isWrite()) {
-                String valueString = String.valueOf(model.evaluate(encodingContext.value((MemoryEvent) e)));
+                Formula valueFormula = encodingContext.value((MemoryCoreEvent)e);
+                assert valueFormula != null;
+                String valueString = String.valueOf(model.evaluate(valueFormula));
                 BigInteger value = switch(valueString) {
-                    case "false" -> BigInteger.ZERO;
+                    // NULL case can happen if the solver optimized away a variable.
+                    // This should only happen if the value is irrelevant, so we will just pick 0.
```
hernanponcedeleon, line 346: Can we log a warning so we can spot when this happen and check that indeed the variable is irrelevant?

ThomasHaas: Sure. I can only log the event though, so it might be hard to connect that back to the source code.

PR #577, opened 2023-11-25
Add pthread_self & __VERIFIER_tid intrinsics

ThomasHaas:
Added support for `pthread_t pthread_self(void)` as well as `unsigned int __VERIFIER_tid(void)` (the semantics are identical, but they live in different headers and have a different return type).
The id assigned by `pthread_create(&t, ...)` (to `t`) matches with the return value of `pthread_self()` in the child thread.

Added a unit test to check that the intrinsics work correctly.

### DIFF
--- a/benchmarks/c/miscellaneous/thread_id.c
+++ b/benchmarks/c/miscellaneous/thread_id.c
@@ -1,0 +1,29 @@
+#include <pthread.h>
+#include <stdatomic.h>
+#include <assert.h>
+#include <dat3m.h>
+
+/*
+    The test shows that pthread_self and __VERIFIER_tid return the same value.
+    Furthermore, it shows that the id written by pthread_create(&t, ...) into "t" matches the
+    the id returned by pthread_self within the spawned child thread.
+    Expected result: PASS
+*/
+
+pthread_t childTid;
+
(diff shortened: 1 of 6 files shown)


benchmarks/c/miscellaneous/thread_id.c
```diff
@@ -0,0 +1,29 @@
...
+}
+
+int main()
+{
+    pthread_t t1;
+    pthread_create(&t1, NULL, thread1, NULL);
+    pthread_join(t1, NULL);
+
+    assert(childTid == t1);
+    assert(__VERIFIER_tid() == pthread_self());
```
hernanponcedeleon, line 28: How does this equality hold if both functions have different return type?

ThomasHaas: I don't exactly know why the compiler doesn't complain here. Maybe on my system `pthread_t` is compatible with `unsigned int`.
As long as it compiles, the equality holds because internally we treat `pthread_t` as an integer.

hernanponcedeleon: I just checked and on Linux, the compiler does not complain..

PR #579, opened 2023-11-28
More control barrier tests

hernanponcedeleon:
@tonghaining can you take a look?


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -961,7 +961,9 @@ public Knowledge visitSyncBarrier(Relation sync_bar) {
             List<FenceWithId> fenceEvents = program.getThreadEvents(FenceWithId.class);
             for (FenceWithId e1 : fenceEvents) {
                 for (FenceWithId e2 : fenceEvents) {
-                    if(exec.areMutuallyExclusive(e1, e2)) {
+                    // “A bar.sync or bar.red or bar.arrive operation synchronizes with a bar.sync
```
tonghaining, line 964: what is `bar.red`?

tonghaining: Okay, I think the current implementation support (`bar.sync`, `bar.sync`) and (`bar.arrive`, `bar.sync`).
Shall we also update the comments on two ptx cat files about the `sync_barrier` relation?

hernanponcedeleon: I don't think we need that level of detail in the cat file. E.g. we do not specify either all the cases covered by `sr`.

tonghaining:
LGTM

PR #580, opened 2023-12-01
Improve lkmm.h

hernanponcedeleon:
The changes were actually done by @ThomasHaas.
With this header file, qspinlock compiles with no warnings.


ThomasHaas:
The changed header does have semantic differences: it is not a straight improvement over the previous one.
For example, the following code will generate a `bv64` load that gets truncated to `bv32` rather than directly generating the intended `bv32` load.
```
int value;
// ...
READ_ONCE(value);
```

Issue #585, opened 2023-12-02
Missing intrinsics


hernanponcedeleon:
- [ ] fflush
- [ ] fprintf
- [x] llvm.smul.with.overflow.i32, llvm.smul.with.overflow.i64
- [ ] strerror
- [ ] time
- [ ] __isoc99_scanf
- [ ] ffs
- [ ] sem_init, sem_post, sem_wait
- [ ] pthread_detach
- [x] memcpy_s
- [ ] pthread_barrier_init

hernanponcedeleon:
Most of the missing intrinsics are not called (as shown by a run after #871). Thus we can close.

PR #589, opened 2023-12-09
Properties

hernanponcedeleon:
This PR cleans up the support for checking overflows and memory-safety we added for svcomp 2024.

Instead of adding new properties as in the `svcomp-2024` branch, I treat all 3 properties uniformly by what they are: a reachability problem.

If we want to allow checking only one property at a time (as needed in svcomp), we need a way to disable certain type of assert events. Enabling / disabling those related to overflow / memsafety can be done by externally setting the corresponding instrumentation flag for the compiler. However, there was currently no way to disable checking from user assertions. I added such an option.

EDIT: I also added support for the multiplication overflow intrinsic which was missing in #585.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -368,6 +375,23 @@ private List<Event> inlineAssert(FunctionCall call) {
         return List.of(assertion, abort);
     }
 
+    private List<Event> integerOverflow(FunctionCall call) {
+        ExpressionFactory expressions = ExpressionFactory.getInstance();
+        final Expression condition = expressions.makeFalse();
+        final Event assertion = EventFactory.newAssert(condition, "integer overflow");
+        final Event abort = EventFactory.newAbortIf(expressions.makeTrue());
+        abort.addTags(Tag.EARLYTERMINATION, Tag.OVERFLOW);
+        return List.of(assertion, abort);
```
ThomasHaas, line 384: You are adding the tag to the abort, not to the assertion. I don't think that is intended.

Also, is the only purpose of the tag to distinguish this from user assertions so that the user can disable normal assertions while still having these?
That makes `Tag.NULLDEREF` and `Tag.OVERFLOW` functionally identical (for now).
Wouldn't it be easier and more general to allow the user to specify a set of strings to disable any property he wants?
Something like `properties.disable=user_assertion, invalid_deref, overflow`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -368,6 +375,23 @@ private List<Event> inlineAssert(FunctionCall call) {
...
+    }
+
+    private List<Event> nullDereference(FunctionCall call) {
+        final Expression condition = expressions.makeFalse();
+        final Event assertion = EventFactory.newAssert(condition, "invalid dereference");
+        final Event abort = EventFactory.newAbortIf(expressions.makeTrue());
+        abort.addTags(Tag.EARLYTERMINATION, Tag.NULLDEREF);
```
ThomasHaas, line 391: It should be `Tag.INVALIDDEREF` or similar (UBSAN checks for more than just null deref).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -543,6 +573,46 @@ private List<Event> inlineLLVMSaturatedAdd(ValueFunctionCall call) {
...
+
+    private List<Event> inlineLLVMSOpWithOverflow(ValueFunctionCall call, IOpBin op) {
+        final Register resultReg = call.getResultRegister();
+        final List<Expression> arguments = call.getArguments();
+        final Expression x = arguments.get(0);
+        final Expression y = arguments.get(1);
+        assert x.getType() == y.getType();
+
+        final IntegerType iType = (IntegerType) x.getType();
+        final Expression sum = expressions.makeBinary(x, op, y);
```
ThomasHaas, line 596: Don't call this `sum`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -543,6 +573,46 @@ private List<Event> inlineLLVMSaturatedAdd(ValueFunctionCall call) {
...
+    }
+
+    private Expression checkIfValueInRangeOfType(Expression value, IntegerType integerType, boolean signed) {
+        final Expression minValue = expressions.makeValue(integerType.getMinimumValue(signed), integerType);
+        final Expression maxValue = expressions.makeValue(integerType.getMaximumValue(signed), integerType);
+        return expressions.makeAnd(
+                expressions.makeLTE(minValue, value, true),
+                expressions.makeLTE(value, maxValue, true)
+        );
+    }
```
ThomasHaas, line 614: Now here is the biggest problem of why I didn't wanna merge the overflow checking right away.
This code does only work when doing integer encoding, meaning that with the standard BV encoding we won't detect overflows.
A correct implementation would extend the bit-width, perform the arithmetic, check for overflows, and then truncate.

From LLVM's language manual:
```
An operation overflows if, for any values of its operands A and B and for any N larger than
the operands’ width, ext(A op B) to iN is not equal to (ext(A) to iN) op (ext(B) to iN) where 
ext is sext for signed overflow and zext for unsigned overflow, and op is the 
underlying arithmetic operation.
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/specification/AssertInline.java
```diff
@@ -20,8 +21,10 @@ public AssertInline(Assert assertion){
     @Override
     public BooleanFormula encode(EncodingContext ctx) {
         final BooleanFormulaManager bmgr = ctx.getBooleanFormulaManager();
-        return bmgr.implication(ctx.execution(assertion),
-                ctx.encodeExpressionAsBooleanAt(assertion.getExpression(), assertion));
+        return (ctx.encodeUserAssertions() || assertion.hasTag(Tag.OVERFLOW) || assertion.hasTag(Tag.NULLDEREF))
+                ? bmgr.implication(ctx.execution(assertion),
+                        ctx.encodeExpressionAsBooleanAt(assertion.getExpression(), assertion))
+                : bmgr.makeTrue();
```
ThomasHaas, line 27: I think just not encoding the assertions is not right, because we still terminate early and just not report an error. Indeed, I think we could erroneously return PASS in edge-cases where the error we do not want to check for is always triggered.

Rather than changing the encoding, I think it might be better to have a pass that deletes the assertions + the abort, if they do not match with what the user wants to check for. Alternatively, the `Inlining` pass could just never generate the assertions in the first place.
This might be even better because it makes sure that no other pass ever sees the aborts and optimizes around them.

That being said, for user-placed assertions, LLVM generates an `unreachable` instruction which still causes an abort, even if the assertion itself gets deleted before inlining.
On the other hand, we should probably not use the standard assertions provided by `assert.h` and instead define `#define assert(cond, ...) __VERIFIER_assert(cond)` with a custom assert function.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -116,6 +116,7 @@ public enum Info {
         // --------------------------- LLVM ---------------------------
         LLVM(List.of("llvm.smax", "llvm.umax", "llvm.smin", "llvm.umin",
                 "llvm.ssub.sat", "llvm.usub.sat", "llvm.sadd.sat", "llvm.uadd.sat", // TODO: saturated shifts
+                "llvm.sadd.with.overflow", "llvm.ssub.with.overflow", "llvm.smul.with.overflow",
```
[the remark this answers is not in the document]
hernanponcedeleon: No, [there is not](https://llvm.org/docs/LangRef.html#arithmetic-with-overflow-intrinsics)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: I kinda dislike that we still have this weird range check, but I see that we need it.
It's like this: we generate encoding `A & B` and when doing integer encoding, only `A` is relevant and `B` is trivial, but when doing BV encoding `A` is trivial and `B` is relevant. Yet we encode both at all times.
Maybe the SMT solver detects this and removes the trivial part in either case.

Well, we can keep it like this, but I think we should document that this range-check is only relevant when doing integer-based reasoning/encoding and pointless for BVs.

dartagnan/src/main/java/com/dat3m/dartagnan/program/specification/AssertInline.java
```diff
@@ -2,6 +2,7 @@
 
 import com.dat3m.dartagnan.encoding.EncodingContext;
 import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.event.Tag;
```
ThomasHaas, line 5: This include is dead again, right?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -141,11 +164,17 @@ public enum Info {
...
         STD_FREE("free", true, false, true, true, Intrinsics::inlineAsZero),//TODO support free
-        STD_ASSERT(List.of("__assert_fail", "__assert_rtn"), false, false, false, true, Intrinsics::inlineAssert),
+        STD_ASSERT(List.of("__assert_fail", "__assert_rtn"), false, false, false, true, Intrinsics::inlineUserAssert),
         STD_EXIT("exit", false, false, false, true, Intrinsics::inlineExit),
         STD_ABORT("abort", false, false, false, true, Intrinsics::inlineExit),
         STD_IO(List.of("puts", "putchar", "printf"), false, false, true, true, Intrinsics::inlineAsZero),
         STD_SLEEP("sleep", false, false, true, true, Intrinsics::inlineAsZero),
+        // --------------------------- UBSAN ---------------------------
+        UBSAN_OVERFLOW(List.of("__ubsan_handle_add_overflow", "__ubsan_handle_sub_overflow", 
+                "__ubsan_handle_divrem_overflow", "__ubsan_handle_mul_overflow", "__ubsan_handle_negate_overflow"),
```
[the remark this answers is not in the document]
hernanponcedeleon: I just checked in the logs and I found this
```
./ring_2w1r-1.log:java.lang.UnsupportedOperationException: Unknown intrinsic function __ubsan_handle_divrem_overflow
```
However, the intrinsic is defined, but never called.

ThomasHaas (approved):
LGTM

PR #590, opened 2023-12-10
Non-recursive implementation of DependencyGraph to avoid stack overflow

ThomasHaas:


hernanponcedeleon:
@ThomasHaas do you remember any benchmark having this problem? I thought we had some in svcomp, but I cannot find any.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/dependable/DependencyGraph.java
```diff
@@ -118,47 +121,87 @@ private void tarjan() {
...
+                v.index = index;
+                v.lowlink = index;
+                stack.push(v);
+                v.isOnStack = true;
+                index++;
+            } else if (lastVisited != null) {
+                // Recursion in progress (we returned from a deeper recursion level)
+                v.lowlink = Math.min(v.lowlink, lastVisited.lowlink);
+            } else {
+                assert false;
```
hernanponcedeleon, line 156: I guess this point should never be reached. Maybe we should add an error msg to this sanity check.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/dependable/DependencyGraph.java
```diff
@@ -225,4 +257,42 @@ public String toString() {
             return content.toString();
         }
     }
+
+    // -------------------------------------------------------------------
+    // -------- Original (recursive) strongConnect implementation --------
```
hernanponcedeleon, line 262: We can get rid of this (and update the comment above referring to the recursive version)

dartagnan/src/main/java/com/dat3m/dartagnan/utils/dependable/DependencyGraph.java
ThomasHaas: I don't think an error message is reasonable for these kinds of invariants. But I will change the code to just have a `else { assert (cond); ...}` rather than `else if (cond) { ...} else { assert false; }`

dartagnan/src/main/java/com/dat3m/dartagnan/utils/dependable/DependencyGraph.java
ThomasHaas: What do you mean by update the comment above? In what way? 
I left the recursive algorithm as "specification", cause it is hard to understand the non-recursive version on its own.

hernanponcedeleon: I did not know you wanted to keep this comment as the specification of the code, and if this was removed, the comment referring to this had to be gone.

I am fine with keeping the commented code as specification, because I agree that the recursion-free version is not trivial.

ThomasHaas: If I keep the original code with special comments that reference it, you can bet that this is intentional :P.
The reason I want to keep this is that in the off-chance that I need to change something in the future, I won't be able to understand the non-recursive version easily. To be fair, I don't even fully understand why the recursive version works, but at least I understand the recursive structure itself (i.e., that it is a DFS) :).

dartagnan/src/main/java/com/dat3m/dartagnan/utils/dependable/DependencyGraph.java
hernanponcedeleon: Once you do this change, I can merge

PR #591, opened 2023-12-16
Devirtualise pthread_create

ThomasHaas:
This PR extends devirtualization to also handle `pthread_create`.
I think there are more pthread functions like `pthread_key_create` that might require devirtualization in the future, however, for now I only added `pthread_create` to make the unit tests in #582 run.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
@@ -64,7 +64,7 @@ public class NaiveDevirtualisation implements ProgramProcessor {
         }
 
         for (Function func : Iterables.concat(program.getThreads(), program.getFunctions())) {
-            devirtualiseNew(func, toAddressTransformer.func2AddressMap);
+            devirtualise(func, toAddressTransformer.func2AddressMap);
         }
     }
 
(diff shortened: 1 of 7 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
```diff
@@ -131,61 +136,108 @@ private void devirtualise(Function function, Map<Function, IValue> func2AddressM
...
+        } else if (call.getCalledFunction().getIntrinsicInfo() == Intrinsics.Info.P_THREAD_CREATE) {
+            final TypeFactory types = TypeFactory.getInstance();
+            final Type ptrType = types.getPointerType();
+            final Type threadType = types.getFunctionType(ptrType, List.of(ptrType));
+            possibleTargets = func2AddressMap.keySet().stream()
+                    .filter(f -> f.getFunctionType() == threadType).collect(Collectors.toList());
+        } else {
+            possibleTargets = List.of();
+            assert false;
+        }
```
hernanponcedeleon, line 211: I don't like this kind of asserts: if they fail, this tells nothing to the user. If we expect that after passing the first `if` we will certainly encounter some given functions (now only pthread_create, but possible more in the future), this assumption should be the one being made in the assert.

Also, the empty list of possible target is useless since the assert would terminate the execution

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
```diff
@@ -131,61 +136,108 @@ private void devirtualise(Function function, Map<Function, IValue> func2AddressM
...
+    }
+
+    private Expression getFunctionPointer(FunctionCall call) {
+        if (!call.isDirectCall()) {
+            return call.getCallTarget();
+        } else if (call.getCalledFunction().getIntrinsicInfo() == Intrinsics.Info.P_THREAD_CREATE) {
+            return call.getArguments().get(2);
+        }
+        assert false;
+        return null;
```
hernanponcedeleon, line 223: Same as comment above.

The fact that you need to add a `return null` to make the compiler not complain about a missing return type already says we should not be using `assert false`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
```diff
@@ -131,61 +136,108 @@ private void devirtualise(Function function, Map<Function, IValue> func2AddressM
...
+            } else {
+                return EventFactory.newVoidFunctionCall(devirtCallTarget, virtCall.getArguments());
             }
+        } else if (virtCall.getCalledFunction().getIntrinsicInfo() == Intrinsics.Info.P_THREAD_CREATE) {
+            final FunctionCall devirtCall = virtCall.getCopy();
+            devirtCall.setArgument(2, devirtCallTarget);
+            return devirtCall;
         }
+        assert false;
+        return null;
```
hernanponcedeleon, line 240: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
ThomasHaas: Unfortuntely, I have to assign `possibleTargets` for otherwise the compiler complaints.

I disagree with having an `assert xyz` instead of the last else-case. A reason is that the assert has to change whenever a new function is added, i.e., it is badly extensible. And this code is expected to get extended later on.

The code should not be reachable in principle, no matter what code it runs on.
More precisely, if that location is ever reached, it does not mean we have some bad input or some not-yet supported input, but rather the logic of this class is broken.
I can change it to `Verify.verify` which is the closest to Java's assertions, though it's a little less fitting.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
ThomasHaas: It sucks that `assert false` still requires a return afterwards. I guess this is because `asserts` can get compiled away, leading to ill-formed code.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
hernanponcedeleon: The branches in `getPossibleTargets` should cover the same cases as `needsDevirtualization`, right? If so, I would rather use `Precondition.checkArguments` than having the assertion

ThomasHaas: Yes, in all cases where `assert false` is placed, the assumption is that the checks cover all cases of `needsDevirtualization` (if they don't, there should be an error).
Not sure how you wanna use `Precondition.checkArguments` here, cause if you think about `Precondition.checkArguments(needsDevirtualiatuon(...))` then this is precisely what is not smart to do, because it passes whenever `needsDevirtualization` gets extended.

The only right thing to do is to throw an error in the else-case I think.

hernanponcedeleon (approved):
@ThomasHaas can you please rebase so I merge?

PR #594, opened 2023-12-20
Bug in morally strong definition in PXTv7.5

hernanponcedeleon:
There is currently a problem with the PTXv7.5 model. The model modified the notion of morally strong from the previous version by forcing the operations to be over the same proxy. The specification is not very precise about how proxies interact with fences, however [it claims](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#causality-axiom) that the fences in this test are morally strong.
```
PTX MP-sys-fence
"Message passing, sync because fences are morally strong and release/acquire pattern is formed"
{
x=0;
y=0;
P1:r1=0;
P1:r2=0;
}
 P0@cta 0,gpu 0         | P1@cta 0,gpu 0         ;
 st.weak x, 1           | ld.relaxed.sys r1, y   ;
 fence.acq_rel.sys      | fence.acq_rel.sys      ;
 st.relaxed.sys y, 1    | ld.weak r2, x          ;
~exists
(P1:r1 == 1 /\ P1:r2 != 1)
```
With the current implementation, they are not. To fix the problem, I now tag every non-proxy fence as `GEN`.

Doing this revealed another problem: the expected result in `Release-acquire-pattern.litmus` for PTXv7.5 was clearly wrong. In fact, in the current implementation we have a different result for PTXv6.0 when the claim is that both models are equivalent for benchmarks not using proxies.

To be sure we have no more mismatched between both PTX versions, I moved all proxy tests to specific folders and forced the remaining results in the expected files to be exactly the same.

Note: it might be easier to review the changes by checking each commit individually.


tonghaining:
LGTM

PR #597, opened 2023-12-28
Generate knowledge from SVCOMP witness

hernanponcedeleon:
This PR adds must knowledge to XRA from SVCOMP witnesses. It simply iterates over the witness edges and if it finds a store-load (resp. store-store) edge from different threads, it adds the pair to the must set of `rf` (resp. `co`). The definition of SC + the semantics of the witness guarantee that this knowledge is correct.

There is no point in trying to be smart and extract e.g., `fr` knowledge or edges within the same thread. XRA is supposed to take care of propagating information and deriving those (keep in mind we assume the memory model is locally consistent). 

In the few examples I tried, the number of encoded edges (the number logged as `Acyclic.getEncodeGraph - reduced encodeGraph size` is reduced by ~10%)

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
@@ -167,7 +167,7 @@ public class WitnessGraph extends ElemWithAttributes {
             currents = getEventsFromEdge(program, e);
             current = currents.size() == 1 ? currents.get(0) : null;
             // If a graph edge implies a hb-relation, inter-thread communication guarantees
-            // same address and thus rf
+            // same address and thus rf.
             if (last != null && current != null && last instanceof Store && current instanceof Load
                     && ((graphEdgeImpliesHbEdge() && !last.getThread().equals(current.getThread()))
                             || alias.mustAlias(last, current))) {
@@ -191,14 +191,16 @@ public class WitnessGraph extends ElemWithAttributes {
                 last = null;
                 continue;
             }
-            // If a graph edge implies a hb-relation, inter-thread communication guarantees same address and thus co
+            // If a graph edge implies a hb-relation, inter-thread communication guarantees same address and thus co.
             if (last != null && last instanceof Store && graphEdgeImpliesHbEdge()
                     && !last.getThread().equals(current.getThread())) {
                 k.add(last, current);
             }
-            // Previous stores to the same address are guaranteed to be in co
+            // Previous stores to the same address are guaranteed to be in co.
+            // Some tools create two edges for the same store (e.g. CPAChecker for pthread_create)
+            // We avoid adding self loops.
             for (MemoryCoreEvent s : lasts) {
-                if (alias.mustAlias(s, current)) {
+                if (alias.mustAlias(s, current) && s != current) {
                     k.add(s, current);
                 }
             }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
@@ -14,8 +14,6 @@ import org.sosy_lab.java_smt.api.*;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.*;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static com.dat3m.dartagnan.GlobalSettings.getOrCreateOutputDirectory;
 import static com.dat3m.dartagnan.witness.GraphAttributes.*;
@@ -71,14 +69,13 @@ public class WitnessGraph extends ElemWithAttributes {
         List<Edge> ret = new ArrayList<>();
         Node cur = getViolationNode();
         while(cur != getEntryNode()) {
-            for(Edge e : getEdges()) {
-                if(e.getTarget().equals(cur)) {
-                    ret.add(e);
-                    cur = e.getSource();
-                    break;
-                }
-            }
+            // Local variables in lambdas have to be final, thus the copy
+            final Node copy = cur;
+            Edge nextEdge = getEdges().stream().filter(e -> e.getTarget().equals(copy)).findAny().get();
+            ret.add(nextEdge);
+            cur = nextEdge.getSource();
         }
+        // Return path from entry to violation
         Collections.reverse(ret);
         return ret;
     }
@@ -103,6 +100,15 @@ public class WitnessGraph extends ElemWithAttributes {
         return str.toString();
     }
 
+    private List<MemoryCoreEvent> getEventsFromEdge(Program program, Edge edge) {
+        return edge.hasCline() ?
+            program.getThreadEvents(MemoryCoreEvent.class).stream()
+                .filter(e -> e.hasMetadata(SourceLocation.class))
+                .filter(e -> e.getMetadata(SourceLocation.class).lineNumber() == edge.getCline())
+                .toList() :
+            Collections.emptyList();
+    }
+
     public BooleanFormula encode(EncodingContext context) {
         Program program = context.getTask().getProgram();
         BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
@@ -122,16 +128,6 @@ public class WitnessGraph extends ElemWithAttributes {
 
     private boolean graphEdgeImpliesHbEdge() {
         return hasAttributed(PRODUCER.toString()) && getAttributed(PRODUCER.toString()).equals("Dartagnan");
-    }
-
-    private List<MemoryCoreEvent> getEventsFromEdge(Program program, Edge edge) {
-        Stream<MemoryCoreEvent> res = Stream.empty();
-        if (edge.hasCline()) {
-            res = program.getThreadEvents(MemoryCoreEvent.class).stream()
-                    .filter(e -> e.hasMetadata(SourceLocation.class))
-                    .filter(e -> e.getMetadata(SourceLocation.class).lineNumber() == edge.getCline());
-        }
-        return res.collect(Collectors.toList());
     }
 
     public EventGraph getReadFromKnowledge(Program program, AliasAnalysis alias) {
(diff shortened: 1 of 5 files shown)


ThomasHaas (commented):
Do we ever exploit control-flow information from the witness?
I think there are a few opportunities (not needed for this PR):
- Encode the control-flow information directly (might already be done?)
- Have an `ExecutionAnalysis` wrapper that uses witness information to give cf-information. This, in particular, improves XRA.
- Have a pass that resolves known branching/eliminates non-executed code. This automatically improves all other analyses.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
```diff
@@ -7,151 +7,195 @@
...
+        List<MemoryEvent> currents;
+        for (Edge e : getEdges()) {
+            currents = getEventsFromEdge(program, e);
+            current = currents.size() == 1 ? currents.get(0) : null;
+            if (current != null && last != null && c1.isInstance(last) && c2.isInstance(current)
+                    && !last.getThread().equals(current.getThread())) {
+                k.add(last, current);
+            }
+            last = current;
+        }
```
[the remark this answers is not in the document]
hernanponcedeleon: > I think there are various cases where this is inaccurate.

You are right that this PR does not exploit all information from the witness (I never claimed this). I simply prototype a solution making use of the easiest to access information (`rf` and `co` edges easily (*) extracted from communication between threads). 

> 
> (1) `rfi`-edges are never made must-edges. While `coi` is known by local consistency, this is not true for `rfi`. It is also not derivable via XRA. Generally, I don't see why you deliberately avoid same-thread edges. Removing the check makes the code easier/shorter and more accurate without implicit assumptions about local consistency.

(*) This is not as trivial as you make it sound. Due to the definition of SC, we know intra-thread witness edges come from `rf U co U fr` (this PR ignores `fr`) and this guarantees that the events access the same location. Deriving `rfi` requires not only knowing that the corresponding events **must** alias, but also that there are no other writes in between which **may** alias with the load. This certainly requires much more code than the one in this PR.

> 
> (2) It misses all must-coherences that are interrupted by a read.
> 
> Indeed, take the worst cast example showing (1) and (2):
> 
> ```
> x = 1;   |   x = 2; | ... | x = n
> r1 = x;  |  r2 = x; | ... | rn = x;
> ```
> 
> Assume the witness orders `x=1; r1=x; x=2; r2=x; ...; x=n; rn=x` (i.e., all threads execute after one another), then you do not derive a single(!) edge. All `rfi`'s are ignored and no two writes get co-ordered because they are not adjacent.
> 

Same problem as above. However, this case might be easier, because as soon as we know a store must alias with a previous store, the co edge is definitely in the must set no matter what happens in between.

> (3) It misses all transitive coherence edges. These are not derivable by XRA due to missing cf-information. Consider the coherence `w1->w2->w3->....->wn` of which you only derive adjacent must-edges. There could be an executuon that does not execute every second store, so you only have `w1, w3, w5, ...` without any known coherences. The reason is that XRA does not know that the witness also guarantees that all those intermediate events do get executed!
> 

Same as (2).

> (4) For the same lack of cf-information, XRA cannot derive from a must-edge `rf(w, r)` that all other rf-edges `rf(w', r)` become impossible. So this could also be derived explicitly from the witness.

I thought about this, and I actually asked myself why we do not currently remove from the may set the pairs where the load is part of a pair in the must set. I did not realize that the lack of execution information was the reason. I can easily add this one.

hernanponcedeleon:
I did quite a few changes
1. Added a method for the assumption that graph edge implies hb edge. For now, only witnesses generated by dartagnan would make use of this. However, instead of checking the producer, we could have a graph attribute representing this, adding it for witnesses generated by dartagnan, and encouraging other tool developers to add this to their witnesses if the assumption holds in their case.
2. When the above does not hold, we make use of must alias information.
3. Got rid of the event ids in the witness. This was fragile, and for what I have tested in our own witnesses, the speed up that we get with XRA is around the same that we were having by encoding the precise values read/stored. 
4. Removed pairs from rf-may-set when we something in the must-set.
5. While most tools generate witnesses representing a single path, this is not required (e.g. CPAChecker has a few branching edges). We know do some pre-processing to only reason about a single path leading to a violation.

I expected that 2. and 3. would result in XRA almost not being applied since it is hard to get the precise event (and only in this case we improve XRA) from the source code line. In the few cases I checked, we never generate `rf`. However, we generate a few more `co` and these seem to propagate quite a bit. Here is an example
```
mix006_tso.oepc generated by GemCutter

SVCOMP-2024: 
	#Relations: 16
	#Axioms: 1
	#may-edges removed (extended): 316
	total #must|may|exclusive edges: 240781|274828|0
	#must|may rf edges: 3|663
	#must|may co edges: 166|422
21:12:34 [INFO ] Acyclic.getEncodeGraph - reduced encodeGraph size: 14832
21:12:35 [INFO ] PropertyEncoder.encodeProgramSpecification - Encoding program specification
21:12:35 [INFO ] AssumeSolver.run - Starting first solver.check()
21:13:46 [INFO ] AssumeSolver.run - Verification finished with result FAIL

Now:
        #Relations: 16
        #Axioms: 1
        #may-edges removed (extended): 7624
        #must-edges added (extended): 1219
        total #must|may|exclusive edges: 243454|267520|0
        #must|may rf edges: 3|592
        #must|may co edges: 215|373
==========================================
[30.12.2023] 23:05:40 [INFO] Acyclic.getEncodeGraph - reduced encodeGraph size: 8270
[30.12.2023] 23:05:41 [INFO] PropertyEncoder.encodeProgramSpecification - Encoding program specification
[30.12.2023] 23:05:41 [INFO] AssumeSolver.run - Starting first solver.check()
[30.12.2023] 23:05:49 [INFO] AssumeSolver.run - Verification finished with result FAIL
```
It might be possible to increase the chances of getting the precise event from a graph edge. For example, right now we don't use the edge attribute saying from which thread the edge come from (we need a way to match witness thread id with the ids of our IR).

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
```diff
@@ -2,156 +2,207 @@
...
-					.filter(e -> e.getMetadata(SourceLocation.class).lineNumber() == edge.getCline())
-					.collect(Collectors.toList());
-			if (!previous.isEmpty() && !events.isEmpty()) {
-				enc.add(bmgr.or(Lists.cartesianProduct(previous, events).stream()
-						.map(p -> context.edgeVariable("hb", p.get(0), p.get(1)))
-						.toArray(BooleanFormula[]::new)));
-			}
-			if (!events.isEmpty()) {
-				previous = events;
-			}
```
hernanponcedeleon, line 94: This was the reason why we were not able to validate some of our own witnesses.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
```diff
@@ -2,156 +2,207 @@
...
+                    && ((graphEdgeImpliesHbEdge() && !last.getThread().equals(current.getThread()))
+                            || alias.mustAlias(last, current))) {
+                k.add(last, current);
+            }
+            last = current;
+        }
+        return k;
+    }
+
+    public EventGraph getCoherenceKnowledge(Program program, AliasAnalysis alias) {
```
ThomasHaas, line 153: I think this method can be written a lot cleaner by first iterating over all edges to extract the stores into a list. This list will only contain stores and no `null`-entries.
Now you can do something like this:
```Java
for (int i = 0; i < writes.size() - 1; i++) {
   final Store w1 = writes.get(i);
   // Add co-edges to all later same-address writes.
   writes.sublist(i + 1, writes.size()).stream()
        .filter(w2 -> alias.mustAlias(w1, w2))
        .forEach(w2 -> must.add(w1, w2));
   // Special case for cross-thread edges.
   final Store immSucc = writes.get(i+1);
   if (graphEdgeImpliesHbEdge() && !w1.getThread().equals(w2.getThread()) {
      must.add(w1, immSucc);
  }
}
```

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
```diff
@@ -2,156 +2,198 @@
...
+
+    public Node getEntryNode() {
+        assert(nodes.stream().anyMatch(n -> n.hasAttributed(ENTRY.toString())));
+        return nodes.stream().filter(n -> n.hasAttributed(ENTRY.toString())).findFirst().get();
+    }
+
+    public Node getViolationNode() {
+        assert(nodes.stream().anyMatch(n -> n.hasAttributed(VIOLATION.toString())));
+        return nodes.stream().filter(n -> n.hasAttributed(VIOLATION.toString())).findAny().get();
+    }
```
ThomasHaas, line 52: You haven't touched this code in the PR, but I think this should still be fixed.
Instead of asserting the existence, you can just use `.findAny/First().orElseThrow(...)`.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessGraph.java
```diff
@@ -2,156 +2,198 @@
...
+        BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
+        List<BooleanFormula> enc = new ArrayList<>();
+        List<MemoryCoreEvent> previous = new ArrayList<>();
+        for (Edge edge : edges.stream().filter(Edge::hasCline).toList()) {
+            List<MemoryCoreEvent> events = getEventsFromEdge(program, edge);
+            if (!previous.isEmpty() && !events.isEmpty()) {
+                enc.add(bmgr.or(Lists.cartesianProduct(previous, events).stream()
+                        .map(p -> context.edgeVariable("hb", p.get(0), p.get(1)))
+                        .toArray(BooleanFormula[]::new)));
+            }
```
ThomasHaas, line 123: This code is kinda sketchy because it generates hb-constraints between events that cannot be hb-related since checks against the may-set are omitted. 
I understand that you need to do that to get the desired ordering between non-neighboring events but I think this should be documented somehow.

ThomasHaas (approved):
LGTM

PR #553, opened 2023-11-05
Benchmarks coming from GPU concurrency: Weak behaviours and programming assumptions

hernanponcedeleon:
I found two issues in the litmus tests coming from the "GPU concurrency: Weak behaviours and programming assumptions" paper.
- in the paper, loads and stores have (an old?) annotation called `.cg` which is supposed to invalidate the L2 cache. Since the L2 is shared between all threads in the GPU, it makes sense to me that this is actually related to the `.gpu` scope
- `st.cg` and `ld.cg` are relaxed stores and loads (volatile are the weak ones)

@tonghaining can you please check this makes sense?

PR #600, opened 2024-01-05
add memalloy-ptx tests

tonghaining:
For the IRIW_gl_cta test, the [litmus](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.litmus) seems not matching with the [output](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.txt). Since this IRIW_gl_cta litmus is identical to the [IRRWIW_uniproc](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRRWIW_uniproc/IRRWIW_uniproc.litmus) test, the IRIW_gl_cta in this PR is written based on the output file.

### DIFF
--- a/litmus/PTX/Memalloy/IRIW_gl_cta.litmus
+++ b/litmus/PTX/Memalloy/IRIW_gl_cta.litmus
@@ -1,0 +1,18 @@
+PTX IRIW_gl_cta
+"CoRR"
+"https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.txt"
+{
+P0:r0=0;
+P1:r0=0;
+P1:r2=0;
+P2:r0=0;
+P3:r0=0;
+P3:r2=0;
+x=0;
+}
+ P0@cta 0,gpu 0           | P1@cta 1,gpu 0           | P2@cta 1,gpu 0           | P3@cta 1,gpu 0           ;
+ ld.weak r0, 1            | ld.weak r0, x            | ld.weak r0, 2            | ld.weak r0, x            ;
+ st.weak x, r0            | fence.acq_rel.gpu        | st.weak x, r0            | fence.acq_rel.cta        ;
+                          | ld.weak r2, x            |                          | ld.weak r2, x            ;
+exists
+(P1:r0 == 1 /\ P1:r2 == 0 /\ P3:r0 == 1 /\ P3:r2 == 0)
--- a/litmus/PTX/Memalloy/IRIW_gl_cta.litmus
+++ b/litmus/PTX/Memalloy/IRIW_gl_cta.litmus
@@ -12,7 +12,7 @@ x=0;
 }
  P0@cta 0,gpu 0           | P1@cta 1,gpu 0           | P2@cta 1,gpu 0           | P3@cta 1,gpu 0           ;
  ld.weak r0, 1            | ld.weak r0, x            | ld.weak r0, 2            | ld.weak r0, x            ;
- st.weak x, r0            | fence.acq_rel.gpu        | st.weak x, r0            | fence.acq_rel.cta        ;
+ st.weak x, r0            | fence.sc.gpu             | st.weak x, r0            | fence.sc.cta             ;
                           | ld.weak r2, x            |                          | ld.weak r2, x            ;
 exists
 (P1:r0 == 1 /\ P1:r2 == 0 /\ P3:r0 == 1 /\ P3:r2 == 0)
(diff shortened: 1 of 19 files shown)


hernanponcedeleon:
1) Where are the expected results coming from? Did you double check with the alloy model/tool?
2) IIRC from one of the PTX papers, `membar` is supposed to be equivalent (from the point of view of the memory model) to some fence. But was is `rel.acq` or `sc`?
3) According to #553 
> st.cg and ld.cg are relaxed stores and loads (volatile are the weak ones)
4) You have a link to some `.txt` file for each test. In the same folder there is a `.litmus` one (with a slightly different syntax than ours). Why not link to those directly?

tonghaining:
1. The results are from the [output txt files](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.txt#L40-L41). I haven't run the alloy tool, I can double check it after I setting up the alloy repo.
2. It is not explicitly defined in the [generate assembler](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/WRR%2B2W_uniproc/WRR%2B2W_uniproc.txt#L30) of the output txt files. I think we should use `.acq_rel` here since the [document](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-membar) said "If the optional `.sem` qualifier is absent, `.acq_rel` is assumed by default."
3. I checked the `.cg` again recently and I think it might should be used for weak operations. For example, the cop(cache-operation) of [load](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-ld) is only needed by weak operations.
4. I linked all tests to the `.litmus` one except the "IRIW_gl_cta" because its [litmus](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.txt#L10-L12) is not consistent with its [txt](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.txt#L10-L12). I suspect they somehow mistakenly pasted the litmus from [IRRWIW_uniproc](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRRWIW_uniproc/IRRWIW_uniproc.litmus) to IRIW_gl_cta.

hernanponcedeleon:
1. I don't think you can use those logs to define the expected behavior from the point of view of the memory model. Those logs shows if the behavior was observed in the real hardware. In the example you linked, it just shows that the behavior [was not observed](https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRIW_gl_cta/IRIW_gl_cta.txt#L58), but this does not guarantee the memory model does not allows it.
2. The sentence from the documentation refers to `fence`. not `membar`. According to Figure 3 (c) from the PTX paper
> fence instruction. membar is a synonym for fence.sc
3. It seems you are right. We need to update all other tests from #553 then.
4. Ok

tonghaining:
1. checked at https://github.com/tonghaining/mixedproxy/tree/memalloy/tests/memalloy
2. fixed.
3. updated

litmus/PTX/Memalloy/IRIW_gl_cta.litmus
```diff
@@ -0,0 +1,18 @@
...
+P0:r0=0;
+P1:r0=0;
+P1:r2=0;
+P2:r0=0;
+P3:r0=0;
+P3:r2=0;
+x=0;
+}
+ P0@cta 0,gpu 0           | P1@cta 1,gpu 0           | P2@cta 1,gpu 0           | P3@cta 1,gpu 0           ;
+ ld.weak r0, 1            | ld.weak r0, x            | ld.weak r0, 2            | ld.weak r0, x            ;
```
hernanponcedeleon, line 14: `ld.weak r0, 2` should be `ld.weak r0, 1`

litmus/PTX/Memalloy/IRWIW_gl_cta.litmus
```diff
@@ -0,0 +1,20 @@
+PTX IRWIW_gl_cta
+"CoRR"
+"https://github.com/johnwickerson/memalloy/blob/master/ptx_testing/IRRWIW_uniproc/IRRWIW_uniproc.litmus"
```
hernanponcedeleon, line 3: This link seems wrong

hernanponcedeleon (commented):
I don't think we need `weak` for those `ld` that get a constant. I expect those to be parsed as `Local` events and not `Load` ones, and the weak tag that not make sense for locals

litmus/PTX/Memalloy/IRWIW_gl_cta.litmus
tonghaining: updated

litmus/PTX/Memalloy/IRIW_gl_cta.litmus
tonghaining: fixed

hernanponcedeleon:
@tonghaining can you please rebase so I merge?

PR #601, opened 2024-01-12
Spir-V parser

natgavrilenko:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorSpirv.java
```diff
@@ -0,0 +1,15 @@
...
+
+import com.dat3m.dartagnan.configuration.Arch;
+import com.dat3m.dartagnan.expression.Expression;
+import com.dat3m.dartagnan.parsers.SpirvBaseVisitor;
+import com.dat3m.dartagnan.parsers.program.utils.ProgramBuilder;
+import com.dat3m.dartagnan.program.Program;
+
+public class VisitorSpirv extends SpirvBaseVisitor<Expression> {
+
+    private final ProgramBuilder programBuilder = ProgramBuilder.forArch(Program.SourceLanguage.LITMUS, Arch.TSO);
```
hernanponcedeleon, line 11: This does not make sense.

We should probably add a new entry to `Program.SourceLanguage` and set the target arch to `Arch.VULKAN`

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/ParserSpirvTest.java
```diff
@@ -0,0 +1,54 @@
...
+        return Stream.of(files)
+                .filter(file -> !file.isDirectory())
+                .map(f -> new Object[]{f.getName()})
+                .toList();
+    }
+
+    @Test
+    public void testParsingFile() throws IOException {
+        Program program;
+        Path path = Paths.get(getTestResourcePath("parsers/program/spirv/" + filename));
```
hernanponcedeleon, line 46: You can use `PATH` variable

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
```diff
@@ -71,12 +71,14 @@ public Program parse(String raw, String path, String format, String cflags) thro
...
-        String format = name.substring(name.lastIndexOf(".") + 1);
-        switch (format) {
-            case "ll":
-                return new ParserLlvm();
-            case "litmus":
-                return getConcreteLitmusParser(readFirstLine(file).toUpperCase());
+        if (name.endsWith(".ll")) {
+            return new ParserLlvm();
+        }
+        if (name.endsWith(".spv.dis")) {
```
hernanponcedeleon, line 77: This is just a name I came up with to differentiate between the binary file (`*.spv`) and the assembly one.

AFAIK there is no "standard" extension for spirv assembler so I am ok with keeping this one. However, when we make the support of spriv official, we need to document what is the expected input extension.

BTW ... you also need to add this format to `Dartagnan.supportedFormats` to be able to parse the files from the console

natgavrilenko: I would keep it hidden until we have the actual support, including the verification of spirv files

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/ParserSpirvTest.java
natgavrilenko: I think "getTestResourcePart" is the standard way to get a path to resource file. Relying on PATH or other env variables is 1) not a standard practice 2) is fragile and can break with different OS or configurations.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorSpirv.java
natgavrilenko: Oops, fixed

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/ParserSpirvTest.java
hernanponcedeleon: I was not referring to replace `getTestResourcePart`, but rather the string that is passed as the first part of its argument. I.e., similar to line 33

natgavrilenko: Ah, then you are right. Done.

hernanponcedeleon (approved):

PR #603, opened 2024-01-16
Make parsers fail on error in lexer

natgavrilenko:


hernanponcedeleon:
@natgavrilenko functionally speaking, what are the changes of this PR doing?

natgavrilenko:
> @natgavrilenko functionally speaking, what are the changes of this PR doing?

Functionally speaking, our parsers didn't fail on parsing errors if these errors come from a lexer (e.g., "extra" input that doesn't match any token):
- If an error is in a parser (matched some token but the token doesn't match the grammar), then Datagnan was failing (as it should),
- If an error is in a lexer (cannot match input to a token), then Dartagnan was showing a warning in the console, but didn't fail (despite it should).

This is because the lexers were using default ConsoleErrorListener, while parsers were using our custom ParserErrorListeneter (forces failing) or BailedErrorStrategy.

I noticed this problem while testing whether my Spir-V tests indeed fail if I add some crap to the input files. And noticed that some special kind of crap doesn't fail the tests. 

I added error listeners to the lexers and replaced BailedErrorStrategy with an ErrorListener, because the latter produces a more informative error message.

Note that one DiagnosticErrorListener is missing in ParserLlmv  and ParserLitmusC. In these parsers, some tests started failing if I enable ambiguity check DiagnosticErrorListener. I don't think that it is a major problem, but maybe we can create an issue and take a look at this some day.

hernanponcedeleon (approved):

PR #606, opened 2024-01-19
Refactor expression package & remove IConst interface

ThomasHaas:
This PR makes the following changes:
- Most expression classes were renamed (e.g. `IValue` -> `IntLiteral`, `BConst` -> `BoolLiteral`, `IOpBin` -> `IntBinaryOp`). Generally, the names became a little more verbose but more readable and streamlined IMO.
- The `IConst` interface was removed and `Expression.reduce` returns an `IntLiteral` now. The interface was implemented only by `IValue` and `MemoryObject` but for the latter we never really made use of it.
That being said, I'm not happy with the `Expression.reduce` method returning `IntLiteral`. I might reintroduce a general `Constant` interface for all types of constants: `IntLiteral`, `BoolLiteral`, `MemoryObject`, and `NonDet...`.
- I replaced some lengthy switch-cases with the new switch-expressions.


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -156,7 +156,7 @@ public BooleanFormula encodeMemory() {
         // For all objects, their 'final' value fetched here represents their constant value.
         final var addrExprs = new ArrayList<BooleanFormula>();
         for (final MemoryObject object : memory.getObjects()) {
-            final BigInteger addressInteger = object.getValue();
```
hernanponcedeleon, line 159: How was this never causing problems in the past?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -115,8 +115,8 @@ public Expression makeCast(Expression expression, Type type) {
         return makeCast(expression, type, false);
     }
 
-    public Expression makeConditional(Expression condition, Expression ifTrue, Expression ifFalse) {
-        return new IfExpr(condition, ifTrue, ifFalse);
+    public Expression makeITE(Expression condition, Expression ifTrue, Expression ifFalse) {
```
hernanponcedeleon, line 118: Can we explicitly make the condition a `BoolExp`?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/CmpOp.java
```diff
@@ -0,0 +1,70 @@
...
+            case NEQ -> a.compareTo(b) != 0;
+            case LT, ULT -> a.compareTo(b) < 0;
+            case LTE, ULTE -> a.compareTo(b) <= 0;
+            case GT, UGT -> a.compareTo(b) > 0;
+            case GTE, UGTE -> a.compareTo(b) >= 0;
+        };
+    }
+
+    // Due to constant propagation, and the lack of a proper type system
+    // we can end up with comparisons like "False == 1", thus the following two methods
```
hernanponcedeleon, line 54: Is the following still true? If not (as I expect), can we get rid of these two methods?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -206,14 +206,14 @@ private void processResults(Local e) {
...
-        if (!(exp instanceof IExprBin iBin)) {
+        if (!(exp instanceof IntBinaryExpr iBin)) {
             return;
         }
         Expression base = iBin.getLHS();
         if (base instanceof MemoryObject mem) {
             Expression rhs = iBin.getRHS();
-            //FIXME Address extends IConst
-            if (rhs instanceof IConst ic) {
+            //FIXME Address extends IntLiteral
```
hernanponcedeleon, line 215: I don't get what these FIXMEs mean. Anything we can do about them?

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -92,29 +92,13 @@ public void setInitialValue(int offset, Expression value) {
         initialValues.put(offset, value);
     }
 
-    /**
-     * Updates the initial value at a certain field of this array.
-     *
-     * @param offset Non-negative number of fields before the target.
-     * @param value  New value to be read at the start of each execution.
-     */
-    public void appendInitialValue(int offset, Expression value) {
```
hernanponcedeleon, line 101: Was this never used?

hernanponcedeleon (commented):
I agree we should have a `Constant`  interface. Are you planning to implemented that as part of this PR?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
ThomasHaas: No. The problem is that once aggregates are involved, you cannot really have static Java typing anymore:
```
expr = extract(struct(bool, int), 0)
```
Should be a valid expression for the condition, `expr.getType` will be boolean but the static java type will not be `BoolExpr`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/CmpOp.java
ThomasHaas: Honestly, I'm not entirely sure. I also suspect that this is not true anymore. I can check if this code is ever called.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
ThomasHaas: We can remove it. This was a very old FIXME where we still had the `Address` class that extended `IConst`.
IIRC, this let to problems because `Addr(5) + 8` would get simplified to `5 + 8 = 13`, i.e., it would treat the address as a regular int, do constant folding and thereby killing the address operand.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
ThomasHaas: No, that's why I could remove it :P. It was used when we still had the Boogie parser but became obsolete with the LLVM rework.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: `getValue` and `getAddress` returned the same value.

Btw. you should try to mark both the old and the new code (and enough surrounding context), otherwise I need to look at the file to see what has changed.

ThomasHaas:
> I agree we should have a `Constant` interface. Are you planning to implemented that as part of this PR?

I can add it, but there are questions about its usefulness.

(1) If `Constant` designated all values that are constant throughout each execution but could be different per execution like non-det values, then the `Constant` interface would not have a single method (i.e., it would be an empty interface).
It would just be an indicator/tagging interface. But then one could also do `expr.getRegs().isEmpty()` rather than `expr instanceof Constant` to check constantness. This would also identify expressions like `memObj + 8` as constant despite being a composite expression.

(2) Maybe an interface like `Literal` which only includes `IntLiteral`, `BoolLiteral`, and possibly `MemoryObject` would be more helpful because we could provide a `Object getValue()` method. There is also the question if `MemoryObject` should even have such a method, because technically our `MemoryObjects` do not have values until after processing where they get addresses assigned.

(3) Do we really need a `reduce` method? We could also reuse the `ExpressionSimplifier` which is more powerful in reducing the expressions. The method is definitely more convenient to use and I have not checked how much work it is to replace its usages with a simplifier call.

In summary I think:
- A `Constant` interface for everything is not helpful. A `Expression.isConstant (~ expr.getRegs.isEmpty())` method could replace it and is more general.
- A `Literal` interface might have usages but `MemoryObject` should not implement it. We would have `IntLiteral`, `BoolLiteral`, and later on also `FloatLiteral` and possibly `StringLiteral` implementing it.
- The `reduce` method itself is a little questionable.

hernanponcedeleon:
If I am not wrong, `reduce()` is only used in the `ExpressionSimplifier` (I think both cases can be replace to another call to `visit`), `LoopBound, MemoryAllocation` (both cases are checking some properties of the reduced value, so the same can be done if we use `visit` instead) and `SparseConditionalConstantPropagation` (the method calling `reduce` returns `Expression` so it can be changed to `visit`).

I am all in favor of getting rid of `reduce`.

ThomasHaas:
Do you still want a `Literal` interface or do we skip this for now? I assume it has no purpose if there is no `reduce` method.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/op/CmpOp.java
```diff
@@ -50,8 +50,6 @@ public boolean combine(BigInteger a, BigInteger b){
         };
     }
 
-    // Due to constant propagation, and the lack of a proper type system
-    // we can end up with comparisons like "False == 1", thus the following two methods
     public boolean combine(boolean a, BigInteger b){
         return switch (this) {
             case EQ -> ((b.compareTo(BigInteger.ONE) == 0) == a);
```
hernanponcedeleon, line 55: We should either fully remove both methods (if not use) or keep the comment.

hernanponcedeleon: Rathen than fully removing the visitor, I would add a `Precondition.verify()` to check that the types match (just to be sure removing this does not backfire us if we start using the visitor again before rewritting it), remove the calls to `combine()` and then get rid of those methods.

hernanponcedeleon (approved):

PR #607, opened 2024-01-19
Factories

xeren:
Attempts to instantiate the helper class `EventFactory` and the singleton classes `ExpressionFactory` and `TypeFactory`.  This design choice came from `EventFactory` needing to create expressions, `ExpressionFactory` needing to fetch types, and `TypeFactory` potentially needing to become configurable to different platforms, including type metadata, type layouts, e.g. 32-bit pointers, and in the future, dependent types.

Changes to API:
- Almost all methods of `EventFactory` are no longer static.
- Nested classes become decorators: `EventFactory.withAtomic()`, `.withLlvm()`, `.withX86()`, etc.
- <strike>`static ExpressionFactory.getInstance()`</strike>
- <strike>`static TypeFactory.getInstance()`</strike>
- `Program.getEventFactory()`
- `ProgramBuilder.getEventFactory()`
- `static EventFactory.newInstance(ExpressionFactory)`
- `EventFactory.getExpressionFactory()`
- `static ExpressionFactory.newInstance(TypeFactory)`
- `ExpressionFactory.getTypeFactory()`
- `static TypeFactory.newInstance()`

Controversial API changes:
- `static Intrinsics.fromConfig(Configuration)` returns `Intrinsics.Factory`, which adapts the former interface of `Intrinsics`.
- `MemoryObject.getInitialValue(int)` returns `Optional<Expression>`
- `ExprSimplifier(ExpressionFactory)` instead of `()`
- `ExpressionTransformer(ExpressionFactory)` instead of `()`
- `VisitorBase(T extends EventFactory)` instead of `()`
- `Memory.allocate(IntegerType,int,boolean)` instead of `(int, boolean)`
- `Memory.allocateVirtual(IntegerType,int,boolean,boolean,VirtualMemoryObject)` instead of `(int,boolean,boolean,VirtualMemoryObject)`
- `Location(String,IntegerType,MemoryObject,int)` instead of `(String,MemoryObject,int)`
- `GenericMemoryEvent(Expression,Type,String)` instead of `(Expression,String)`
- `Init(MemoryObject,int,Expression,ExpressionFactory)` instead of `(MemoryObject,int,Expression)`
- `LKMMLock(Expression,Type)` instead of `(Expression)`
- `LKMMUnlock(Expression,Expression)` instead of `(Expression)`


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -310,105 +337,122 @@ public static StoreExclusive newExclusiveStore(Register register, Expression add
...
     // =============================================================================================
 
-    public static class Pthread {
-        private Pthread() {
-        }
+    public Pthread withPthread() {
+        return new Pthread(this);
+    }
+
+    public static final class Pthread extends EventFactory {
```
ThomasHaas, line 344: These decorators feel kinda ugly.
What about the classes like `Pthread` not inheriting from `EventFactory` and renaming `withPthread` to `forPthread`.
Then the user will always be maintaining a pure `EventFactory` instance and use `eventFactory.forPthread().newPthreadEvent`.
At least this stays more true to our current usage pattern.
It also disables strange usages like `eventFactory.withLLVM().withPthread().withX86()` which has no reasonable meaning.

Alternatively, we could forgo the whole partioning/grouping of events and have them all in a single `EventFactory`.
This might result in more verbose naming though.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/TypeFactory.java
```diff
@@ -11,8 +11,6 @@
 
 public final class TypeFactory {
```
ThomasHaas, line 12: With `TypeFactory` not being (treated as) a static class anymore, I think it might be a good idea to have each `Type` know its `TypeFactory`. It at least enables some type checks be made whenever an expression is given. For example in `Alloc`'s constructor you needed to get rid of the type precondition due to the lack of factory access.
Alternatively, it might just be better to put those preconditions that we cannot check inside the constructor anymore into the `EventFactory`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
```diff
@@ -59,7 +59,8 @@ public Function(String name, FunctionType type, List<String> parameterNames, int
 
     @Override
     public Type getType() {
-        return TypeFactory.getInstance().getArchType();
+        final TypeFactory types = program.getEventFactory().getExpressionFactory().getTypeFactory();
+        return types.getArchType();
```
ThomasHaas, line 63: Usages like this suggest that `Program` should probably provide direct getters for the different factories.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/Memory.java
```diff
@@ -16,16 +17,16 @@ public class Memory {
      * @return
      * Points to the created location.
      */
-    public MemoryObject allocate(int size, boolean isStatic) {
+    public MemoryObject allocate(IntegerType addressType, int size, boolean isStatic) {
         Preconditions.checkArgument(size > 0, "Illegal malloc. Size must be positive");
-        MemoryObject address = new MemoryObject(nextIndex++, size, isStatic);
+        MemoryObject address = new MemoryObject(nextIndex++, addressType, size, isStatic);
```
ThomasHaas, line 22: Is this necessary? Since a memory belongs to a single program, we might as well have `Memory` contain a reference to `Program` from which it can obtain the necessary types.

ThomasHaas (commented):
We should carefully think about how to handle all these cases where we relied on globally accessible factories before.
I think this PR shows that making factories non-global comes with a lot of ugly problems.

ThomasHaas:
Btw. if we really want to get this through, I believe its better to make smaller steps.
For example, we can keep the global singleton instances but already add additional getters to, e.g., `Program` (which for now would just return the global instance).
This way, you could first try to reduce the usages of the global/static getters without directly getting rid of all of them.
Then we can discuss the remaining parts where getting rid of global getters is hard and requires (possibly ugly?) API changes.

Also, it is probably easiest to go with one factory at a time: first the type factory, then the expression factory, and lastly the event factory.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
xeren: Although there currently are no overrides, I would propose this inheritance.  From the perspective of a user of `EventFactory`, I would rather like to have just one or two places where it really matters, if I am creating target-specific events, or core events.  E.g. there is this code in `PowerVisitor.java`:
```
  Load load = eventFactory.newRMWLoadExclusive(oldValueRegister, address);
  Store store = eventFactory.newRMWStoreConditional(address, e.getStoreValue(), true);
```
With distinct classes, this code would currently look like
```
  Load load = eventFactory.newRMWLoadExclusive(oldValueRegister, address);
  Store store = eventFactory.forPower().newRMWStoreConditional(address, e.getStoreValue(), true);
```

ThomasHaas: I understand that with distinct classes it looks different, in particular, more explicit. I'm just advocating for the more explicit version. At the end of the day, this is just my opinion. Functionalitywise, I don't see much difference in both approaches.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/TypeFactory.java
xeren: I have added the check to `EventFactory`.  We can add a `Type.getFactory()` later, should we ever need it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -34,17 +34,12 @@ public class MemoryObject extends IntExpr {
...
+        super(addressType);
         this.index = index;
         this.size = size;
         this.isStatic = isStaticallyAllocated;
         this.isThreadLocal = false;
-
-        if (isStaticallyAllocated) {
-            // Static allocations are default-initialized
-            initialValues.put(0, ExpressionFactory.getInstance().makeZero(TypeFactory.getInstance().getArchType()));
-        }
```
xeren, line 47: I was surprised that this caused no errors in the tests.  It should influence the result of `getStaticallyInitializedFields()` and thus the set of `Init` events of non-litmus programs.  Can this safely be removed?

ThomasHaas: What problem do you expect? Static allocations are default initialized to zero in `C`, so why would this affect non-litmus programs? I guess the llvm parser already makes sure to do correct initialization so this code is likely redundant (i.e., nothing will change if you remove it).

hernanponcedeleon:
As discussed.

PR #608, opened 2024-01-26
Expression rework

ThomasHaas:
This PR contains a big rework (not only refactor) of the expression package. It is inspired by the design developed for the LLVM rework which was never implemented.

**Core changes:**
- Added `CastExpression` and `IntSizeCast`. `IntUnaryExpr` no longer represents casts.
- Added `IntegerHelper` class which contains functions to mimic fixed-size BV operations on `BigInteger`, including extension, truncation, and representation normalization (e.g., `-1` == `1` for `bv1`). Most supported operations are representation-agnostic.
- Added `Expression.getOperands` to make traversal of expressions easier.
- Major refactoring (most file changes are due to renaming/moving of classes)
- Added preliminary support for floats (`FloatType` and various float expressions). Floats can be represented in the IR, but we cannot do IEEE-compatible computations nor encode them right now.
- Added `BoolBinaryOp.IFF` to represent equalities on booleans. 

**TODOS:**

- [x] Implement `ExprSimplifier` and make use of it.
- [x] Remove `Expression.reduce` (once the above is done)
- [x] Make more use of `IntegerHelper`: `SCCP` is likely wrong in some edge cases right now (EDIT: `SCCP` uses `ExprSimplifier` now)
- [x] Possibly add floating point types (might be done in a separate PR)


dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryExpr.java
```diff
@@ -0,0 +1,61 @@
...
+import com.dat3m.dartagnan.expression.type.IntegerType;
+
+import java.math.BigInteger;
+
+import static com.google.common.base.Verify.verify;
+
+public class IntUnaryExpr extends UnaryExpressionBase<IntegerType, IntUnaryOp> {
+
+    public IntUnaryExpr(IntUnaryOp operator, Expression operand, IntegerType t) {
+        super(t, operator, operand);
```
hernanponcedeleon, line 15: Shouldn't we check the operand is an integer?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryExpr.java
```diff
@@ -0,0 +1,61 @@
...
+                return new IntLiteral(targetType, value.negate());
+            }
+            case CTLZ -> {
+                if (innerType.isMathematical()) {
+                    throw new UnsupportedOperationException(
+                            String.format("Counting leading zeroes in mathematical integer %s.", inner));
+                }
+                if (value.signum() == -1) {
+                    return new IntLiteral(targetType, BigInteger.ZERO);
+                }
```
hernanponcedeleon, line 39: I don't get this check

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryOp.java
```diff
@@ -0,0 +1,20 @@
...
+
+    @Override
+    public String toString() {
+        return getSymbol();
+    }
+
+    @Override
+    public String getSymbol() {
+        return switch (this) {
+            case CTLZ -> "ctlz ";
```
hernanponcedeleon, line 16: do we really want the extra space here?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/ExpressionHelper.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class ExpressionHelper {
+
+    private ExpressionHelper() {}
+
+    public static void checkSameType(Expression x, Expression y) {
+        Preconditions.checkArgument(x.getType().equals(y.getType()),
+                "The types of %s and %s do not match.", x, y);
+    }
+}
```
hernanponcedeleon, line 14: I think there is code touched by this PR that can use this method and still doesn't

hernanponcedeleon (commented):
I still need to take a closer look at the helper class implementing all the bit manipulation with BigIntegers

dartagnan/src/main/java/com/dat3m/dartagnan/expression/base/NaryExpressionBase.java
```diff
@@ -0,0 +1,34 @@
...
+
+import com.dat3m.dartagnan.expression.Expression;
+import com.dat3m.dartagnan.expression.ExpressionKind;
+import com.dat3m.dartagnan.expression.Type;
+import com.dat3m.dartagnan.program.event.common.NoInterface;
+
+import java.util.List;
+
+@NoInterface
+public abstract class NaryExpressionBase<TType extends Type, TKind extends ExpressionKind> extends ExpressionBase<TType> {
```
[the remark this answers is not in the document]
ThomasHaas: It is used in `GEPExpr` and `ConstructExpr`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/booleans/NonDetBool.java
```diff
@@ -0,0 +1,27 @@
+package com.dat3m.dartagnan.expression.booleans;
+
+import com.dat3m.dartagnan.expression.ExpressionKind;
+import com.dat3m.dartagnan.expression.ExpressionVisitor;
+import com.dat3m.dartagnan.expression.base.LeafExpressionBase;
+import com.dat3m.dartagnan.expression.type.BooleanType;
+
+//TODO instances of this class should be managed by the program
```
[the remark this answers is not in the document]
ThomasHaas: This is a comment that has been there for a long time.
It means that non-deterministic values are "program constants" and should therefore be constructed via something like `program.newConstant()` or `program.newNonDetValue()`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntCmpExpr.java
```diff
@@ -0,0 +1,21 @@
...
+import com.dat3m.dartagnan.expression.base.BinaryExpressionBase;
+import com.dat3m.dartagnan.expression.misc.CmpOp;
+import com.dat3m.dartagnan.expression.type.BooleanType;
+import com.dat3m.dartagnan.expression.utils.ExpressionHelper;
+
+public class IntCmpExpr extends BinaryExpressionBase<BooleanType, CmpOp> {
+
+    public IntCmpExpr(BooleanType type, CmpOp kind, Expression left, Expression right) {
+        super(type, kind, left, right);
+        ExpressionHelper.checkSameType(left, right);
```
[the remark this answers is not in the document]
ThomasHaas: Yeah, will do.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryExpr.java
ThomasHaas: Yes, the signature is also wrong right now (the last parameter is obsolete now).

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryExpr.java
ThomasHaas: Negative numbers have their highest bit set, so there are no leading zeroes.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryOp.java
ThomasHaas: yes, because this is printed as a unary operator `ctlz x`. Without the space it will print `ctlzx`.
We might want to come up with better printing though.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/ExpressionHelper.java
ThomasHaas: Probably. I will do a pass over the whole code to cleanup more things.

ThomasHaas:
> I still need to take a closer look at the helper class implementing all the bit manipulation with BigIntegers

There are certainly some edge cases when computing with non-normalized values that might be wrong.
I wrote a testing class that randomly generates inputs and checks that all operations, extensions, and truncations work correctly (cause bit-fiddling is always error-prone). 
Also, I wasn't sure what normalization to use, so I wrote all sorts of variants, but also made all operations agnostic to the representation so that is shouldn't matter what we do.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntCmpExpr.java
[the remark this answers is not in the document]
ThomasHaas: Meh, we also use this class to make equalities between boolean values.
Not sure what the best solution to this problem is:
- One could generalize the expression class since `==` and `!=` make sense for all types. But then we need to add checks that make sure the operator makes sense for the type (i.e. `==` on structs is fine, but `<` is not).
- Alternatively, one could separately add `==` to booleans (or rather `<=>` as `BoolBinaryOp`). But then we would not be able to make equalities between arbitrary types.
- One could make separate expressions for equalities and the other relational operators.
- Or something else...

ThomasHaas:
I have added `FloatType` and some kinds of float expression, and extended `TypeFactory` and `ExpressionFactory` to allow construction of floating-point types/expressions.
Since we cannot really do IEEE compatible computations on the Java side, `ExprSimplifier` will not touch any float operations (i.e., no constant folding on floats).
There is currently no support for encoding floats, we can just represent them in the IR.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntCmpExpr.java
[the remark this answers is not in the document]
ThomasHaas: I added `BoolBinaryOp.IFF` and now (dis)equalities between booleans are encoded with that operator.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
```diff
@@ -1,178 +1,258 @@
...
+                    || op == LSHIFT || op == RSHIFT || op == ARSHIFT)) {
+                return left;
+            }
+
+            if (lit.isOne() && (op == MUL || op == DIV || op == UDIV)) {
+                return left;
+            }
+
+            if (lit.isZero() && op == MUL && (aggressive || left.getRegs().isEmpty())) {
+                return expressions.makeValue(0, expr.getType());
```
hernanponcedeleon, line 210: don't we have a general makeZero method?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/IntegerHelper.java
```diff
@@ -0,0 +1,225 @@
+package com.dat3m.dartagnan.expression.utils;
+
+import com.google.common.base.Preconditions;
+
+import java.math.BigInteger;
+import java.util.function.BiFunction;
+
+public class IntegerHelper {
+
+    public static final boolean SIGNED_NORMALIZATION = true;
```
hernanponcedeleon, line 10: Shouldn't this be moved to `GlobalSetting` to keep all such configurable options centralized?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/SparseConditionalConstantPropagation.java
```diff
@@ -108,19 +106,26 @@ public void run(Function func) {
...
                 }
                 reachableEvents.add(cur);
 
+                // Invalidate outdated propagation data
+                if (cur instanceof RegWriter rw) {
+                    propagationMap.remove(rw.getResultRegister());
+                    if (propagateCopyAssignments) {
+                        propagationMap.values().removeIf(expr -> expr.getRegs().contains(rw.getResultRegister()));
+                    }
+                }
```
hernanponcedeleon, line 122: Is this related to the bug you mentioned in SCCP?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/IntegerHelper.java
ThomasHaas: Possibly. But honestly, I'm not too sure if we even need this option at all. I kept most code to work over non-normalized values (i.e., it is correct either way).
It is hard to tell what is best to do.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/floats/FloatLiteral.java
```diff
@@ -0,0 +1,59 @@
...
+                && getType().equals(val.getType())
+                && value.equals(val.value)
+                && isNAN == val.isNAN
+                && isInf == val.isInf);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("%s(%s)", getType(), value);
+    }
```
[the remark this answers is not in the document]
ThomasHaas: You should think of this class a prototype for now. It cannot even represent all possible values (e.g., `-0.0`, -INF, SNAN/QNAN (not sure if the distinction is needed) are missing).
I would rather update this class (and all float-related classes/methods) based on demand (i.e., when we encounter something we need to support). The reason I added `NAN` and `INF` was mostly to suggest what one could do.
In fact, those two values cannot even be constructed via the factory right now.
That being said, I can update `toString` to make this prototype a little more complete.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionVisitor.java
```diff
@@ -0,0 +1,70 @@
...
+    default TRet visitBoolBinaryExpression(BoolBinaryExpr expr) { return visitBinaryExpression(expr); }
+    default TRet visitBoolUnaryExpression(BoolUnaryExpr expr) { return visitUnaryExpression(expr); }
+    default TRet visitBoolLiteral(BoolLiteral lit) { return visitLeafExpression(lit); }
+    default TRet visitNonDetBoolExpression(NonDetBool expr) { return visitLeafExpression(expr); }
+
+    // =================================== Floats ===================================
+    default TRet visitFloatBinaryExpression(FloatBinaryExpr expr) { return visitBinaryExpression(expr); }
+    default TRet visitFloatCmpExpression(FloatCmpExpr expr) { return visitBinaryExpression(expr); }
+    default TRet visitFloatUnaryExpression(FloatUnaryExpr expr) { return visitUnaryExpression(expr); }
+    default TRet visitFloatLiteral(FloatLiteral lit) { return visitLeafExpression(lit); }
```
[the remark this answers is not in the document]
ThomasHaas: If we are reworking the nondet values as you suggest in another comment, I can add nondet floats as well.
I think @xeren did some work on changing nondets on a different branch: https://github.com/hernanponcedeleon/Dat3M/commits/nondet/
Maybe that can be integrated here somehow

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -253,6 +254,107 @@ public Expression makeGetElementPointer(Type indexingType, Expression base, List
...
+    }
+
+    public Expression makeCast(Expression expression, Type type, boolean signed) {
+        if (type instanceof BooleanType) {
+            return makeBooleanCast(expression);
+        } else if (type instanceof IntegerType integerType) {
+            return makeIntegerCast(expression, integerType, signed);
+        }
+        throw new UnsupportedOperationException(String.format("Cast %s into %s unsupported.", expression, type));
+    }
```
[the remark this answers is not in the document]
ThomasHaas: Eventually yes.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -253,6 +254,107 @@ public Expression makeGetElementPointer(Type indexingType, Expression base, List
...
+            for (Type fieldType : structType.getDirectFields()) {
+                zeroes.add(makeGeneralZero(fieldType));
+            }
+            return makeConstruct(zeroes);
+        } else if (type instanceof IntegerType intType) {
+            return makeZero(intType);
+        } else if (type instanceof BooleanType) {
+            return makeFalse();
+        } else if (type instanceof FloatType floatType) {
+            return makeValue(BigDecimal.ZERO, floatType);
```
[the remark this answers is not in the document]
ThomasHaas: Will add

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -78,23 +79,14 @@ public BooleanFormula encodeFullProgram() {
...
-                                "Program constant bounds of type %s.",
-                                context.getFormulaManager().getFormulaType(formula)));
+        for (NonDetValue value : context.getTask().getProgram().getConstants()) {
+            Formula formula = context.encodeFinalExpression(value);
+            if (formula instanceof IntegerFormula intFormula && value.getType() instanceof IntegerType intType && !intType.isMathematical()) {
+                IntegerFormulaManager integerFormulaManager = context.getFormulaManager().getIntegerFormulaManager();
+                IntegerFormula min = integerFormulaManager.makeNumber(intType.getMinimumValue(true));
+                IntegerFormula max = integerFormulaManager.makeNumber(intType.getMaximumValue(false));
+                enc.add(integerFormulaManager.greaterOrEquals(intFormula, min));
+                enc.add(integerFormulaManager.lessOrEquals(intFormula, max));
```
hernanponcedeleon, line 89: I think we normally use `imgr` for this (the long version is too verbose). There is a similar change about the Boolean manager in this PR.

ThomasHaas: I agree that the full name is unnecessarily verbose.

I took these changes from @xeren's branch. He removed signedness from nondet values due to unifying them into a single class. For BVs this doesn't matter, but for mathematical integers it does. I guess the intention here was to provide conservative bounds for the integer that are true for both signed and unsigned integers.
That being said, I don't think it reflects the language level semantics correctly.

I don't know how to solve this issue properly though.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/IntegerHelper.java
hernanponcedeleon: If there is no clear win with keeping this, I would simply get rid of it.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
hernanponcedeleon: Can we simply add a `signed` field to the class (even if it is useless for bools)?

ThomasHaas: I think I found a fitting solution. I think it makes sense to add a notion of metadata to these program constants and then we could store signedness there.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/IntegerHelper.java
ThomasHaas: I removed it.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -215,7 +216,7 @@ public Formula lastValue(MemoryObject base, int offset) {
             return formulaManager.getIntegerFormulaManager().makeVariable(name);
         }
         //TODO match this with the actual type.
```
hernanponcedeleon, line 218: I guess we can now get rid of the comment

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -110,106 +109,92 @@ public Formula visitIntLiteral(IntLiteral intLiteral) {
     public Formula visitIntBinaryExpression(IntBinaryExpr iBin) {
         Formula lhs = encode(iBin.getLeft());
         Formula rhs = encode(iBin.getRight());
+        final int bitWidth = iBin.getType().isMathematical() ? 64 : iBin.getType().getBitWidth();
```
hernanponcedeleon, line 112: Why the hardcoded constant? Don't we have a method somewhere to get the arch type width?

ThomasHaas: The constant is somewhat arbitrary. The former code used `32`, but theoretically any constant is wrong I believe.
Since we never generate mathematical integers in any parser anymore (we used to with Boogie), the code is actually never executed.
We might even get rid of the mathematical integer type altogether...

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
ThomasHaas: I made a mistake and reverted my change. This comment is about encoding the content of a memory address (whose type we do not know) rather than encoding the address itself (whose type we know).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -215,8 +216,11 @@ public Formula lastValue(MemoryObject base, int offset) {
         if (useIntegers) {
             return formulaManager.getIntegerFormulaManager().makeVariable(name);
         }
-        //TODO match this with the actual type.
-        return formulaManager.getBitvectorFormulaManager().makeVariable(base.getType().getBitWidth(), name);
+        //TODO match this with the actual type stored at the memory address
+        // (we do not know and guess the arch type right now)
+        TypeFactory types = TypeFactory.getInstance();
+        int archSize = types.getMemorySizeInBytes(types.getArchType());
+        return formulaManager.getBitvectorFormulaManager().makeVariable(archSize, name);
```
hernanponcedeleon: This is only used for litmus tests. I this guessing a problem at all? Will it be once we have support for mixed size concurrency?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -219,7 +219,7 @@ public Formula lastValue(MemoryObject base, int offset) {
         //TODO match this with the actual type stored at the memory address
         // (we do not know and guess the arch type right now)
         TypeFactory types = TypeFactory.getInstance();
-        int archSize = types.getMemorySizeInBytes(types.getArchType());
+        int archSize = types.getMemorySizeInBytes(types.getArchType()) * 8;
```
hernanponcedeleon, line 222: is this pattern used somewhere else? Would it make sense to have a method `getMemorySize()` in the type factory which encapsulates the multiplication??

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
hernanponcedeleon: Are you finally getting rid of this?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
ThomasHaas: If litmus tests declare variables of different sizes, it could be a problem. Also, it is unrelated to mixed-size concurrency. It's just that memory is currently untyped and as long as we support C-like mallocs memory will likely continue to be untyped. For example, if the memory content happened to be of float type, this code would fail.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
ThomasHaas: No, it is never used. But again, this is more or less a randomly chosen value. Just using the constant `64` might be even better than relying on `archType`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: Yes. I just didn't have time yesterday

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
hernanponcedeleon: Litmus tests in the presence of mixed size concurrency will definitely define variables of different types, so this is definitely something we need to support once me move there

ThomasHaas: Mixed-size has not much to do with this. These are the values at a specific memory address.
Since litmus code has only static memory allocations, we can know the exact type in memory.
A mixed-size access will not change the type of memory content, it will just overwrite the memory value partially.

ThomasHaas:
The CI randomly crashed. 
Is there a way to just restart it or do I need to do a fake commit (I always forget)?

hernanponcedeleon:
If you go to `Actions`, there is a `Re-run all jobs` bottom

Issue #610, opened 2024-01-31
Dockerfile failed to build maven


DerGrumpf:
The Goal `org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile` compile failed with error `invalid target release: 17`.

Is there any fix? Trying to use the master branch in the `git clone` command doesn`t change anything.

ThomasHaas:
Hi DerGrumpf,
it seems that the docker image loads java version 11 (openjdk11) but we have updated Dartagnan to require version 17+.
@hernanponcedeleon Do you know how to update the java version in the docker?
I think adding 
```
apt install openjdk-17-jdk-headless
```
OR
```
apt install openjdk-17-jdk openjdk-17-jre
```
might already do the trick?

DerGrumpf:
Haha Yes thx. I update the File for you :)

DerGrumpf:
nvm i didn't have the rights to open a new branch ...

Just add `apt-get install -y openjdk-17-jdk && \ apt-get install -y openjdk-17-jre` after the install RUN command.

PR #611, opened 2024-02-01
Add CI job to build the docker image

hernanponcedeleon:


ThomasHaas:
Can you add the fix from #610 ?

hernanponcedeleon:
> Can you add the fix from #610 ?

Yes ... I just want to check first if the CI would detect the current problem. I will also fix the Dockerfile as part of this PR

PR #612, opened 2024-02-06
Simplifier fix

xeren:
Fixes two simplifications that are happening during Constant Propagation:
- Enable `!true` -> `false` and `!false` -> `true`.
- Disable `x && y` -> `true` and `x || y` -> `true` if `x` equals to `y` while in `aggressive` mode.

The rest should be refactoring.


ThomasHaas (approved):
LGTM

PR #613, opened 2024-02-08
Improved Cutting

ThomasHaas:
This PR cleans up the code related to (static) cutting of relations/constraints from a memory model for Refinement (and also clarifies what "cutting" even means).
The changes are mostly in preparation for online CAAT but they also allow to fix a long-standing issue of treating dependency relations as static (I have not yet added this fix as I need to check its performance impact).

The PR also includes various other changes:
- Refactor/renaming of axioms and definitions
- Changes to `Wmm` to get rid of implicit construction of base relations. This results in changed parsing behaviour for odd cases where a cat-file overwrites a base relation (i.e., (re)defines it before its first usage). For example, if a cat-file defines `ext = ~int`, and then refers to `rfe` for the first time, `rfe` will get defined in terms of the new definition of `ext`. I can fix this behaviour in the parser, if we want to keep the old one (which does not allow one to fully overwrite base relatons).


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -260,71 +209,115 @@ private void simplifyAssociatives(Class<? extends Definition> cls, BiFunction<Re
...
+
+    private Relation makePredefinedRelation(String name) {
+        /*
+            WARNING: The code has possibly unexpected behaviour:
+             - If a relation like `ext` is already defined (different from the standard definition),
+               constructing e.g. `rfe` will refer to this new definition.
+             - When constructing derived relations like fr, the intermediately defined relations like rf^-1
+               will get constructed even if they already exist in the model.
+             TODO: Clarify what the intended behaviour should be.
+         */
```
hernanponcedeleon, line 254: Is it because we internally hardcode the expected definition of things like `rfe`?
I think we should not do this and make use of the explicit definition in the cat file

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -260,71 +209,115 @@ private void simplifyAssociatives(Class<? extends Definition> cls, BiFunction<Re
...
             case IDD -> new DirectDataDependency(r);
             case ADDRDIRECT -> new DirectAddressDependency(r);
             case CTRLDIRECT -> new DirectControlDependency(r);
             case EMPTY -> new Empty(r);
             case RFINV -> {
                 //FIXME: We don't need a name for "rf^-1", this is a normal relation!
                 // This causes models that explicitly mention "rf^-1" to have two versions of the same relation!
-                yield new Inverse(r, getRelation(RF));
+                yield new Inverse(r, getOrCreatePredefinedRelation(RF));
             }
```
hernanponcedeleon, line 275: Based on the comment ... do we even need to support this relation here?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -260,71 +209,115 @@ private void simplifyAssociatives(Class<? extends Definition> cls, BiFunction<Re
...
-            case FR -> composition(r, getRelation(RFINV), getRelation(CO));
-            case MM -> new CartesianProduct(r, Filter.byTag(Tag.MEMORY), Filter.byTag(Tag.MEMORY));
-            case MV -> new CartesianProduct(r, Filter.byTag(Tag.MEMORY), Filter.byTag(Tag.VISIBLE));
-            case IDDTRANS -> new TransitiveClosure(r, getRelation(IDD));
-            case DATA -> intersection(r, getRelation(IDDTRANS), getRelation(MM));
+            case FR -> composition(r, getOrCreatePredefinedRelation(RFINV), getOrCreatePredefinedRelation(CO));
```
hernanponcedeleon, line 276: What is the difference between `getOrCreatePredefinedRelation` and `getOrCreateRelation`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -260,71 +209,115 @@ private void simplifyAssociatives(Class<? extends Definition> cls, BiFunction<Re
...
-            case COI -> intersection(r, getRelation(CO), getRelation(INT));
-            case FRE -> intersection(r, getRelation(FR), getRelation(EXT));
-            case FRI -> intersection(r, getRelation(FR), getRelation(INT));
+            case POLOC -> intersection(r, getOrCreatePredefinedRelation(PO), getOrCreatePredefinedRelation(LOC));
+            case RFE ->  intersection(r, getOrCreatePredefinedRelation(RF), getOrCreatePredefinedRelation(EXT));
+            case RFI -> intersection(r, getOrCreatePredefinedRelation(RF), getOrCreatePredefinedRelation(INT));
+            case COE -> intersection(r, getOrCreatePredefinedRelation(CO), getOrCreatePredefinedRelation(EXT));
+            case COI -> intersection(r, getOrCreatePredefinedRelation(CO), getOrCreatePredefinedRelation(INT));
+            case FRE -> intersection(r, getOrCreatePredefinedRelation(FR), getOrCreatePredefinedRelation(EXT));
+            case FRI -> intersection(r, getOrCreatePredefinedRelation(FR), getOrCreatePredefinedRelation(INT));
```
hernanponcedeleon, line 298: I start to believe that it might be better to get rid of all these. I would be better to have them defined directly in cat (maybe via a proper `include X.cat`)

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -456,7 +459,8 @@ public Knowledge visitProduct(Relation rel, Filter domain, Filter range) {
         }
 
         @Override
-        public Knowledge visitIdentity(Relation rel, Filter set) {
+        public Knowledge visitSetIdentity(SetIdentity id) {
+            Filter set = id.getFilter();
```
hernanponcedeleon, line 463: `final`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -552,18 +557,19 @@ public Knowledge visitControl(Relation rel) {
...
-        public Knowledge visitInternalDataDependency(Relation rel) {
+        public Knowledge visitInternalDataDependency(DirectDataDependency idd) {
             // FIXME: Our "internal data dependency" relation is quite odd an contains all but address dependencies.
             return computeInternalDependencies(EnumSet.of(DATA, CTRL, OTHER));
         }
 
         @Override
-        public Knowledge visitFences(Relation rel, Filter fence) {
+        public Knowledge visitFences(Fences fenceDef) {
+            Filter fence = fenceDef.getFilter();
```
hernanponcedeleon, line 572: `final`?

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
```diff
@@ -68,13 +67,17 @@ public class ExecutionGraph {
...
+        this.ra = ra;
         relationGraphMap = HashBiMap.create();
         filterSetMap = HashBiMap.create();
         constraintMap = HashBiMap.create();
-        this.cutRelations = cutRelations;
+        // TODO: Clean this up
+        cutRelations = refinementModel.computeBoundaryRelations().stream()
+                .filter(r -> r.getName().map(n -> !Wmm.ANARCHIC_CORE_RELATIONS.contains(n)).orElse(true))
+                .map(refinementModel::translateToOriginal)
+                .collect(Collectors.toSet());
```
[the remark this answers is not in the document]
ThomasHaas: I removed the comment already (it was worse before). I just didn't push yet

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/RefinementModel.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+            c.getConstrainedRelations()
+                    .stream().map(orig2BaseRelations::get)
+                    .filter(Objects::nonNull)
+                    .forEach(boundary::add);
+        }
+
+        return boundary;
+    }
+
```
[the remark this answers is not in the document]
ThomasHaas: Yes, it is related to the `Cut` boundary. From the refinement perspective, the boundary relations are those that get eagerly encoded (cut) but are also needed for the lazy part (cause it does compute derived relations from them).
But to make sure that they actually get encoded, we need some axiom on top of them (for otherwise the active sets will be empty), so we place an artificial `ForceEncodeAxiom`.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/RefinementModel.java
```diff
@@ -0,0 +1,126 @@
...
+            final String name = rel.getName().orElse(null);
+            if (name != null && Wmm.ANARCHIC_CORE_RELATIONS.contains(name)) {
+                orig2BaseRelations.put(rel, base.getRelation(name));
+                orig2BaseConstraints.put(rel.getDefinition(), base.getRelation(name).getDefinition());
+            } else {
+                final Relation copy = base.newRelation();
+                rel.getNames().forEach(n -> base.addAlias(n, copy));
+                orig2BaseRelations.put(rel, copy);
+            }
+        }
```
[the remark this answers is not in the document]
ThomasHaas: Yes, in a first step we just copy over all relations (including their names) but place no constraints yet (except for the few core relations that are always there and need special treatment).
In a second step, we copy over all constraints over those relations.

The resulting memory model represents the "lower" part as described by the cut. This lower part is what gets eagerly encoded, the upper part is what the lazy refinement will handle (roughly speaking).

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/Refiner.java
```diff
@@ -26,21 +25,19 @@ public Refiner() { }
     public BooleanFormula refine(DNF<CoreLiteral> coreReasons, EncodingContext context) {
         final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
         List<BooleanFormula> refinement = new ArrayList<>();
-        HashSet<BooleanFormula> addedFormulas = new HashSet<>(); // To avoid adding duplicates
-        // For each symmetry permutation, we will create refinement clauses
```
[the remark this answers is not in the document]
ThomasHaas: This was outdated. I moved the permutation reasoning into the `CoreReasoner` a while ago, so the `Refiner` actually had nothing to do with it anymore. However, the code was still written as if there are permutations involved.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -409,113 +406,60 @@ private RefinementIteration doRefinementIteration(ProverEnvironment prover, WMMS
...
+        final Set<Constraint> constraintsToCut = new HashSet<>();
+        for (Constraint c : original.getConstraints()) {
+            if (c instanceof Axiom ax && ax.isNegated()) {
+                constraintsToCut.add(ax);
+            } else if (c instanceof Difference diff) {
+                final Relation sub = diff.getSubtrahend();
+                final Definition subDef = sub.getDefinition();
+                if (!sub.getDependencies().isEmpty() || subDef instanceof SetIdentity || subDef instanceof CartesianProduct) {
+                    constraintsToCut.add(subDef);
                 }
```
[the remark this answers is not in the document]
ThomasHaas: The dependency check is not for the diff but the RHS of the diff.

I guess the word "semi-derived" makes sense. The two checks are there because CAAT treats them as derived whereas they are base in our `Wmm` (since we have no unary predicates but CAAT has).
It certainly would make sense to add fences as well, although CAAT currently treats them as a base relation but one over which it cannot compute negations.
I would still prefer a rewrite of the form `fencerel(X) -> po;[X];po` and avoid all special reasoning, but that is orthogonal.

Btw. this code would also cut `data/addr/ctrl` if we decide to do so.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -409,113 +406,60 @@ private RefinementIteration doRefinementIteration(ProverEnvironment prover, WMMS
...
-        }
-    }
-
-    private Wmm createDefaultWmm() {
-        Wmm baseline = new Wmm();
-        Relation rf = baseline.getRelation(RF);
-        if (baselines.contains(Baseline.UNIPROC)) {
+    private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
+        // FIXME: This can (in theory) add redundant intermediate relations and/or constraints that
+        //  already exist in the model.
```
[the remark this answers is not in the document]
ThomasHaas: It should just add redundancy that can affect performance. This issue is not specific to biases but is more general with how we handle predefined relations/axioms.
IIRC, this issue already appears in `RISCV`.
The issue is in code like this:
```
let r1 = fr | ... // Implicitily defines a "rf^-1" relation
let r2 = rf^-1 // Explicitly defines another "rf^-1"
```
For biases, it is no practical issue because the baseline model has almost no relations to begin with (so there is hardly a possiblity to create redundancy), unless we cut a lot relations from the original model.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
```diff
@@ -70,7 +75,4 @@ public String toString() {
         return names + " := " + getDefinition().getTerm();
     }
```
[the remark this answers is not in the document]
ThomasHaas: I guess you meant to refer to `isInternal`? This code might have become obsolete now. IIRC, it's only purpose was to let `Refinement` know which relations it should ignore, because (as you said) those involve non-visible events.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -39,155 +38,110 @@ public static class Config {
 
     private static final Logger logger = LogManager.getLogger(Wmm.class);
 
-    public final static ImmutableSet<String> BASE_RELATIONS = ImmutableSet.of(CO, RF, RMW);
+    // These relations are part of every memory model (even the "empty" one) because they are
+    // necessary to specify the anarchic program semantics.
+    public final static ImmutableSet<String> ANARCHIC_CORE_RELATIONS = ImmutableSet.of(CO, RF, RMW);
```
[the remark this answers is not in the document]
ThomasHaas: No, `LOC` is not used to define any anarchic program semantics (by that I mean data/control-flow). We internally use `addr(e1)==addr(e2)` literals and encoding `LOC` just means that we generate `loc(e1, e2) <=> addr(e1)==addr(e2)` formulas.

The other one's are required as follows:
- `RF` is used to define data-flow between memory events.
- `CO` is used to define final values for litmus tests and liveness checks.
- `RMW` is used to define the semantics of conditional stores (which fail if they have no matching rmw load).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -260,71 +209,115 @@ private void simplifyAssociatives(Class<? extends Definition> cls, BiFunction<Re
         }
     }
 
-    private Definition basicDefinition(String name) {
-        Relation r = newRelation(name);
-        return switch (name) {
+    public void removeUnconstrainedRelations() {
+        // TODO: Currently removes all axiom-irrelevant relations and thus may remove
+        //  other relations part of other constraints.
```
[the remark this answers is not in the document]
ThomasHaas: It currently is not a problem because we do not have any other constraints, at least at the point where we use this method.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: Yes, it's because of hardcoded definitions.
I fully agree that we should define all derived relations in the cat-file, however, that means we should have proper support for includes.

Also, the issue would still exist if you programatically tried to generate a `Wmm` rather than using the parser, but this is a rather rare occurence (though the current code for creating biases does it).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: The reason we have this case is because of the definition of `FR`:
```
 case FR -> composition(r, getOrCreatePredefinedRelation(RFINV), getOrCreatePredefinedRelation(CO));
```
Now, I can change this definition to fetch `RF` and manually construct `RF^-1` from it.
This is fine as long as there is no other definition that wants to use `RF^-1` (because this would result in duplicated constructions).
I can do the same for `MM` and `MV` but not for, e.g., `IDDTRANS` because it is used in multiple places.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: The latter doesn't exist. But if it did exist, I would interpret it as "get relation with name X and if it does not exist, create a relation with name X". It would NOT automatically associate any definition to that relation and so would NOT implicitly create any other relations.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: Definitely, but that requires adding support for includes :).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
ThomasHaas: Same as above

ThomasHaas:
There is currently a problem with generating reasons for unnamed relations that have been cut.
I will need to fix this before merging

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/RefinementModel.java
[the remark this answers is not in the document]
hernanponcedeleon: Agree ... I have seen quite a few cases where RA on the whole model takes quite a lot of time, so better to avoid it when possible

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
hernanponcedeleon: Mmmm from the top of my head, I was sure I saw that method while I was reviewing this PR, but you are right.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
hernanponcedeleon: Not really ... we could directly define this directly in each cat file. The question is if ww want to do that in favor of simplifying the code, or we postpone it until we have support for includes.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
```diff
@@ -228,7 +227,10 @@ private RelationGraph createGraphFromRelation(Relation rel) {
 
         // ===== Filter special relations ======
         String name = rel.getNameOrTerm();
-        if (SPECIAL_RELS.contains(name)) {
+        if (cutRelations.contains(rel)) {
+            graph = new DynamicDefaultWMMGraph(name);
+        } else if (SPECIAL_RELS.contains(name)) {
+            // TODO: This case is unnecessary if we decide to cut the special relations upfront
```
hernanponcedeleon, line 233: We are gonna need this for online solving anyway right? Also, IIRC the current solution is "theoretical unsound". I would say lets go for it

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -498,7 +517,7 @@ private static CharSequence generateSummary(RefinementTrace trace, long boundChe
         StringBuilder message = new StringBuilder().append("Summary").append("\n")
                 .append(" ======== Summary ========").append("\n")
                 .append("Number of iterations: ").append(trace.iterations.size()).append("\n")
-                .append("Total native solving time(ms): ").append(totalNativeSolvingTime + boundCheckTime).append("\n")
+                .append("Total native solving time(ms): ").append(totalNativeSolvingTime).append("\n")
                 .append("   -- Bound check time(ms): ").append(boundCheckTime).append("\n")
```
hernanponcedeleon, line 521: The indentation in the bound check time suggests "out of the total native, this was bound". But this is not the case anymore. So I would either keep the `boundCheckTime` for the total time, or change the indentation

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
ThomasHaas: I think you were always in favour of being able to drop in cat-files from herd7, but without includes this would not work anymore. If you are willing to rewrite all cat-models, go for it :P.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: No, it is still part of the total time. By adding it explicitly we accidentally counted it twice before!
This is actually a bug fix rather than a change.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
hernanponcedeleon: Fair point. For compatibility with herd, let's keep the code as it is until we support includes

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -456,8 +471,12 @@ private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
     }
 
     private static void removeFlaggedAxiomsAndReduce(Wmm memoryModel) {
+        // We remove flagged axioms.
+        // NOTE: Theoretically, we could cut them but in practice this causes the whole model to get eagerly encoded,
+        // resulting in the worst combination: eagerly encoded model relations + lazy axiom checks.
+        // The performance on, e.g., LKMM is horrendous!!!!
         List.copyOf(memoryModel.getAxioms()).stream()
-                .filter(Axiom::isNegated)
+                .filter(Axiom::isFlagged)
```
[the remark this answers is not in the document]
ThomasHaas: CAAT cuts negated relations and removes flagged relations, so it won't handle either of them.
I don't know if our eager encoding handles negated axioms correctly, but that is now orthogonal to CAAT.

PR #614, opened 2024-02-10
Add memcpy_s to intrinsics

hernanponcedeleon:


ThomasHaas:
That cannot work, `memcpy_s` works differently and has a different signature.

hernanponcedeleon:
Ohhh I checked that the semantics "are the same" (modulo a few things that I think we can abstract away for the purpose of verification), but I indeed not check the signature

PR #615, opened 2024-02-15
Generate SVCOMP witness if PROGRAM_SPEC is the only property

hernanponcedeleon:
Witnesses only make sense for svcomp properties, all of which are covered by `PROGRAM_SPEC`.
If dartagnan is called e.g. with `--property=liveness,program_spec`, the solver could find a liveness violation and this would result in `getLtlPropertyFromSummary` throwing an `UnsupportedOperationException`.

We should only try to generate the witness if `PROGRAM_SPEC` is the only property (which is the case for svcomp).


ThomasHaas (approved):
LGTM

PR #616, opened 2024-02-15
New Pointer Analysis

xeren:
This PR started as a rework for `FieldSensitiveAlias`.  It now features a third implementation `InclusionBasedPointerAnalysis`.
- It manages pointer modifiers of the type (`int` static offset‘,`List<Integer>` dynamic offsets).  This should make its precision robust against optimizations of the following form.  (`FieldSensitiveAndersen` uses (`int`,`int`) and thus looses precision there)
```
      w = x + y
      address = w + z
      ↓
      address = x + y + z
```
- It stores inclusions, loads, stores and pointer sets in a compact format, such that its memory usage scales better with large arrays.
- Instead of a node for each register or allocated byte, this version manages a node for each complex expression, register writer and phi node.  This means this version not only field sensitive, but also flow sensitive (with the exception of assumptions).  It now requires a prior `Dependency` Analysis.
- It implements a unification rule:  If a variable has only one include edge, it gets merged into that other variable.  This enhances the `mustAlias` queries.
- It implements an acceleration rule:  If a variable directly or indirectly includes itself with a static offset of `c != 0`, then `c` is promoted to a dynamic offset.  Such cycles propagate at most once in total, instead of at most once per allocated byte.  This should also allow the algorithm to terminate even if the program has arrays of unknown/nondeterministic size (if such a feature is ever added in the future).


dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntLiteral.java
```diff
@@ -28,7 +28,7 @@ public BigInteger getValue() {
     }
 
     public int getValueAsInt() {
-        return value.intValueExact();
+        return value.intValue();
     }
```
ThomasHaas, line 32: Why this change? We want to throw an exception here cause that means we should not have used the method in the first place. With the change, truncation errors are swallowed silently. Did you encounter an issue with this?

hernanponcedeleon:
Does the first bullet point mean the new analysis is always more precise than the old ones?
Do we have any concrete example where we get noticeable better performance?

ThomasHaas:
@xeren showed me improved results on EBR with substantial speed up.

Btw. I think this PR does not set the new analysis as default, so the CI still runs with the old one.
I think you should put the new analysis as default to catch any problems.

hernanponcedeleon:
@ThomasHaas would it make sense to add EBR to our CI? IIRC it is supposed to be buggy wrt IMM but correct under hardware models

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,816 @@
...
+        // This cannot be used when a negative alignment occurs, because the analysis would not terminate.
+        if (leftAlignment.size() == 1) {
+            final int alignment = Math.abs(leftAlignment.get(0));
+            for (final Integer a : rightAlignment) {
+                if (a % alignment != 0) {
+                    return false;
+                }
+            }
+            return offset % alignment == 0;
+        }
```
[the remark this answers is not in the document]
ThomasHaas: The issue talks about dynamic indices being non-negative (static indices are fine and occur from `container_of` constructs),
and for those we cannot check if they are non-negative (we don't know!).

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,816 @@
...
+        for (final MemoryObject object : program.getMemory().getObjects()) {
+            totalVariables++;
+            constantMap.put(object, new Variable(object, object.toString()));
+        }
+        for (final RegWriter writer : program.getThreadEvents(RegWriter.class)) {
+            processWriter(writer);
+        }
+        for (final MemoryCoreEvent store : program.getThreadEvents(MemoryCoreEvent.class)) {
+            processMemoryEvent(store);
+        }
```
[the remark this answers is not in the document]
ThomasHaas: The naming here is definitely odd. If this processes only stores, it should look for stores only. If not, the variable should have a different name.

ThomasHaas:
> @ThomasHaas would it make sense to add EBR to our CI? IIRC it is supposed to be buggy wrt IMM but correct under hardware models

I don't know how much time it takes with Yices2. With Z3, running EBR under multiple models is expensive.

ThomasHaas:
I have just tested this analysis on RCU tree and with B=1 it terminates in 2 seconds and with B=2 in 20 seconds. The old analysis took multiple hours IIRC.
However, it computes quite a few empty pointer sets. I also cannot tell how precise it is cause the fast termination may be due to imprecision caused by RCU having the whole memory stored in arrays.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,816 @@
...
+        int result = Math.abs(alignment.get(0));
+        for (final Integer a : alignment.subList(1, alignment.size())) {
+            result = gcd(result, Math.abs(a));
+        }
+        return result;
+    }
+
+    private static int gcd(int left, int right) {
+        // Euclidean algorithm.  I cannot believe this is not part of the standard library.
+        assert left >= 0 && right >= 0;
```
ThomasHaas, line 490: You can use `IntMath.gcd` from google's guava library.

hernanponcedeleon:
It seems that when I rebase the branch, I overwrote the default alias type. It should be fixed by now.

BTW, as Thomas said, the new alias analysis shows to be better than the old one in the benchmarks added in #621, here are the statistics
```
[20.02.2024] 11:56:47 [INFO] AliasAnalysis.fromConfig - Selected alias analysis: FULL
        #Relations: 104
        #Axioms: 4
        #may-edges removed (extended): 86907
        #must-edges added (extended): 156
        total #must|may|exclusive edges: 1689106|2239829|0
        #must|may rf edges: 4|18859
        #must|may co edges: 474|486

[20.02.2024] 11:58:00 [INFO] AliasAnalysis.fromConfig - Selected alias analysis: FIELD_SENSITIVE
        #Relations: 104
        #Axioms: 4
        #may-edges removed (extended): 94306
        #must-edges added (extended): 156
        total #must|may|exclusive edges: 1686959|2287321|0
        #must|may rf edges: 4|21522
        #must|may co edges: 474|486
```

ThomasHaas:
> It seems that when I rebase the branch, I overwrote the default alias type. It should be fixed by now.

I think the default you think about was not changed by @xeren. We have two defaults and the one you changed was for the UI only I think.

hernanponcedeleon:
It seems you are right (even if I'm pretty sure I saw `field_sensitive` in the log when I called dartagnan from the console with no alias options).

We should be able to use `Alias.getDefault()` similarly to what we do with method, target, ... right?

ThomasHaas:
> We should be able to use `Alias.getDefault()` similarly to what we do with method, target, ... right?

I think so. Just change the initializer in `AliasAnalysis.Config` to call `Alias.getDefault()`.

hernanponcedeleon:
I am testing this branch in some new benchmarks and I get a bunch of warnings like this
```
[04.03.2024] 00:14:56 [WARN] InclusionBasedPointerAnalysis.postProcess - empty pointer set for bv64 117:r24 = load(*bv64 117:r23)
```
How should I interpret this @xeren ? It sounds like this load cannot read from anywhere

ThomasHaas:
Yes, this means the are no possible targets to read from. It could be a bug in the analysis but also a correct result.
For example, this warning also appeared in `tree.c` but it turned out that the code does indeed access a NULL pointer.

hernanponcedeleon:
> Yes, this means the are no possible targets to read from. It could be a bug in the analysis but also a correct result. For example, this warning also appeared in `tree.c` but it turned out that the code does indeed access a NULL pointer.

But then this is a bug in the code being analyzed ...

ThomasHaas:
> > Yes, this means the are no possible targets to read from. It could be a bug in the analysis but also a correct result. For example, this warning also appeared in `tree.c` but it turned out that the code does indeed access a NULL pointer.
> 
> But then this is a bug in the code being analyzed ...

Yes, it might be if that memory access is reachable. @xeren Can you add source location information into the warning so we can more easily check what is wrong?

hernanponcedeleon:
@xeren any progress a out this PR? I would like to get this merged soon

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -155,11 +158,13 @@ private void run(Program program) {
...
             final List<Offset<Variable>> edges = queue.remove(variable);
+            logger.trace("dequeue {}", variable);
             algorithm(variable, edges);
         }
-        //printGraph();
+        printGraph();
+        final SyntacticContextAnalysis synContext = SyntacticContextAnalysis.newInstance(program);
         for (final Map.Entry<MemoryCoreEvent, Offset<Variable>> entry : eventAddressSpaceMap.entrySet()) {
-            postProcess(entry);
+            postProcess(entry, synContext);
```
ThomasHaas, line 167: Since you use the syntactic context only inside `postProcess`, I think it would be better to pass the `program` instance and inside `postProcess ` do this:
```
final SyntacticContextAnalysis synContext = logger.isWarnEnabled ? SyntacticContextAnalysis.newInstance(program) : null;
```
IIRC, `SyntacticContextAnalysis` computes quite a lot of data (a whole context stack per event) so it might be better to avoid it if unused. Realistically speaking, the analysis' time/memory consumption won't be noticable though.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -179,7 +184,7 @@ private void processWriter(RegWriter event) {
         } else if (event instanceof Load load) {
             final Offset<Variable> address = getResultVariable(load.getAddress(), load);
             if (address == null) {
-                logger.warn("null pointer address for {}", event);
+                logger.warn("null pointer address for {}: {}", event.getMetadata(SourceLocation.class), event);
```
ThomasHaas, line 187: You can use the static method `SyntacticContextAnalysis.getSourceLocationString(Event)` to get a better location message and a proper one even if there is no source location metadata attached.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/SyntacticContextAnalysis.java
```diff
@@ -100,9 +99,6 @@ public String toString() {
         }
     }
 
-    // We use this enum to track loop nesting
-    private enum LoopMarkerTypes {START, INC, END}
-
```
hernanponcedeleon, line 105: Where are the changes in this class coming from? Are they really related to this PR or is it just a "complementary fix"? Maybe you noticed this now that you print the alias graph.

I remember seeing some problems related to this when generating some FAIL report, but I never got the time to investigate it further.

ThomasHaas: Yes, these problems occured when using `SyntacticContextAnalysis` to debug problems.
The problem was that if you had an empty iteration body (say due to propagating the whole loop body via CP),
then the iteration entry label was both beginning and end of an iteration.
The analysis didn't understand that an event could be both and so it ended up treating it as "end", which caused an exception because there was no matching "begin".

I have not checked if the changes by René are a proper fix (though I discussed this issue with him). Either way, the changes are complementary and unrelated to the alias analysis itself.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
xeren: Usually, `logger.isWarnEnabled()` would be enabled either way.  I attempt to use `Suppliers.memoize(Supplier)`.  That way, at least one error message has to be issued for invoking `SyntacticContextAnalysis`.  And it would be done at most once.

ThomasHaas: I like that solution

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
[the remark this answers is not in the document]
xeren: The method `processMemoryEvent` also processes `GenericMemoryEvent`, i.e. `synchronize_srcu`.  Currently, `Load` is the only special case, since it is both a writer and a memory event.  Hence, there is a comment in `processMemoryEvent`.
To have `Load`'s be processed before `Store`'s allowed me to omit the communication checks for `Load`.  Also, to have `Load`'s be in the same processing order as other writers avoids `replace(Variable,Offset<Variable>)`.
I could name it `processWritersExceptLoads`, if that helps reading the code.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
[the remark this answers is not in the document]
xeren: The guards before this line test if there is a negative value.  So this line is only reachable if all values are non-negative.  I had to add negative values to pass some tests.  The FIXME is there because I would like to have an algorithm that also works for negative values.
I am trying to fix this issue without loosing the precision we are used to by now.  Including the existing analyses, `IBPA` assumes non-negative dynamic indexes / offsets if reading it from a register, regardless of its original C type, which would oftentimes be signed.
There is an interesting theory behind it, but I don't think I will get this implemented soon.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/SyntacticContextAnalysis.java
xeren: My changes assume that iterations of the same loop do not overlap and that nested loops do not touch the boundaries of their outer loop.  This should hold according to `LoopAnalysis`.  However, events that were placed in between iterations are no longer recognized as being inside the loop.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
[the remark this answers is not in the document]
ThomasHaas: You can name it `processNonLoads/processStores` or similar but the name of the loop variable should match.
Currently, there is this odd discrepancy between the loop variable referring to stores specifically but the method seemingly accepting arbitrary memory events.
You could also make this loop filter out the loads rather than hiding it in the called method (the method could then assert that it excepts non-load events).

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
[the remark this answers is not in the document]
ThomasHaas: I think as long as the known edge-cases are documented somewhere (e.g., in the form of FIXMEs) I think the current behavior is fine.
There are papers out their that literally say "sound and precise alias analysis of C code is impossible" :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/SyntacticContextAnalysis.java
ThomasHaas: > However, events that were placed in between iterations are no longer recognized as being inside the loop

Doesn't this mean that the analysis is even more broken now? Or do I misunderstand something?

hernanponcedeleon:
I implemented the flattening pass (*) that I mentioned a while ago on top of this branch (see `alias-flat`) and something seems fishy to me.

The alias analysis produces more precise results when flattening is used (and thus it needs to handle more complex expressions; I would have expected that it would loose precision in some cases) than using the "default" IR
```
java -jar dartagnan/target/dartagnan.jar cat/aarch64.cat --target=arm8 dartagnan/src/test/resources/libvsync/bounded_mpmc_check_empty-opt.ll --bound=1 --program.processing.flattening=true
        #Relations: 16
        #Axioms: 7
        #may-edges removed (extended): 0
        total #must|may|exclusive edges: 81839|82970|0
        #must|may rf edges: 4|308
        #must|may co edges: 40|178
```
```
java -jar dartagnan/target/dartagnan.jar cat/aarch64.cat --target=arm8 dartagnan/src/test/resources/libvsync/bounded_mpmc_check_empty-opt.ll --bound=1 --program.processing.flattening=false
        #Relations: 16
        #Axioms: 7
        #may-edges removed (extended): 0
        total #must|may|exclusive edges: 93322|94770|0
        #must|may rf edges: 4|318
        #must|may co edges: 40|196
```
(*) the pass replaces registers in reg-readers by the expression that register carries withing the block (the pass could be more precise by preserving info between blocks, but this is irrelevant ATM).

ThomasHaas:
> The alias analysis produces more precise results when flattening is used (and thus it needs to handle more complex expressions; I would have expected that it would loose precision in some cases) than using the "default" IR

Your flattening pass is wrong though. Consider
```
r1 = r2;
r2 = ...;
store(&x, r1 + r2);
// ---- After flattening ----
r1 = r2;
r2 = ...;
store(&x, r2 + r2); // Wrong
```
You cannot propagate expressions over register assignments that affect any of the registers contained in the expression.

I have not checked if this case appears in your benchmark example, but for now I would assume your pass is the culprit.
Btw. you also should reset the propagation map for every function (or just make it a local variable), cause theoretically it tries to propagate expressions across different functions (should never happen due to disjoint registers, but still).

hernanponcedeleon:
Can you push your fix so @xeren can take a look?

xeren:
Brief: This analysis is not robust against general flattening.  However, it should be robust against flattening of addition.

I looked into your example.  It features flattening of an expression `r4 = r0 * 2` into `r1 = r4 * 8`.  This analysis is not capable of representing that the variable for `r4` only receives multiples of 2.  Thus, without flattening, any event accessing address `r1` will be assumed to access multiples of 8.  When flattening is enabled, the analysis can deduce from `r1 = (r0 * 2) * 8`, that its result will always be a multiple of 16.  Hence, flattening can still improve the alias results.

hernanponcedeleon:
@xeren I still want to have some documentation explaining how the algorithm works so I can understand it and properly review it

ThomasHaas:
Based on your initial description and my understanding of the algo, let me suggest another (hopefully better) description:
```
The analysis constructs a directed inclusion graph over expressions of the program.
The nodes abstractly represent the values an expression (at a program location) could take over all executions.
There is a node (also called variable in the code) for each memory object (addr), two for each Local (result and rhs), Load (result and addr), and Store (addr and value).
These nodes are not just per expression but also per program location to achieve an SSA-like form (*).

Edges between two nodes A and B describe one of three relationships:
	(1) An edge "A -f_includes-> B" means that "f(A) \subseteq B" holds (where A/B are understood as sets of values). 
	    Rather than arbitrary functions "f", we use multi-linear expressions "f(A) = A + k1*x1 + k2*x2 + ... kn*xn + o" where ki and o are constants.
	    For simplicity, we write "k*x + o" where k=(k1,...,kn) and x=(x1,...,xn) are understood as vectors, and drop the _includes suffix (i.e. write A -f-> B).
	    The (sub)expression "k*x + o" describes the set of values {k*x + o | x \in Z^n} that can be obtained by varying x freely (**).
	(2) An edge "A -stores-> B" means that there exists a store that stores at address-expression A the value-expression B.
	(3) An edge "A -loads-> R" means that there exists a load that loads from address-expression A and puts its result into R (R is the node of a register).

Initially, the graph has the following edges
	- For every "store(A, V)" event, it has a "A -stores-> V" edge
	- For every "R = load(A)" event, it has a "A -loads-> R" edge.
	- For every assignment "R := B + k*x + o", we have a "B -(k*x+o)-> R" edge
	  For too complex assignments that do not match the structure, say "R := h(B1, B2, .., Bn)" where h is complex, we introduce edges "Bi -1x+0-> R".

On this initial graph, two rules are applied until saturation
(1) Transitivity: A -f_includes-> B AND B -g_includes-> C IMPLIES A -fg_includes-> C (notice that we need to compose the labels as well).
(2) Communication: if A -loads-> R and B -stores-> V and A and B can overlap (i.e. the corresponding Load/Store events can alias), then V -0x+0-> R (all values in V are included in R)

It can happen that these rules generate an edge A -f-> B although another edge A -g-> B already exists. 
In this case, the edges get merged by a join operation.
To guarantee termination, the saturation of cyclic inclusion relationships get accelerated:
E.g. if A includes B (B -0x+0-> A) and B includes A+2 (A -0x+2-> B)
then A includes A+2x, B includes A+2x+2 and both include B+2x.

(*) We also generate special phi nodes that can improve precision on code that is not in SSA form.
(**) In the implementation we make an unsound assumption that x is non-negative. 
     This disallows negative (dynamic) indexing into arrays/pointers but gives additional precision in other cases.
     Note that "o" can be negative which is sufficient to support "containerof" functionality.
```

@xeren Is the text I wrote correct? 
@hernanponcedeleon Is this text understandable to you?

hernanponcedeleon:
The text is great @ThomasHaas.

I still need to check that the implementation follows this, but the idea of the algorithm is certainly clear to me now.

xeren:
Your description looks fine to me.  Assignments can also have multi-linear expressions.

The underlying algorithm has much in common with Sui & Lei's: *Fast and Precise Handling of Positive Weight Cycles for Field-sensitive Pointer Analysis* (SAS'19, [DOI](https://doi.org/10.1007/978-3-030-32304-2_3)).  Some exceptions:
- There are no nodes for memory cells.  Instead, if a store may alias with a load, their values are *directly* connected with an include-edge.
- The pointer sets are embedded inside the graph.  There are nodes for base addresses and the graph is transitively-closed.  To get the pointer set of a variable, look at its in-set variables representing base addresses.
- `Offset` uses $\langle o,0,\lbrace 1\rbrace\rangle$ instead of $\langle o,0,\emptyset\rangle$ to denote an entire object $o$, and $\langle o,i,\emptyset\rangle$ instead of $\langle o,i,\lbrace 0\rbrace\rangle$ for a single field $o.f_i$.  Also, we allow negative offsets, since we sometimes get such programs (`container_of`).

ThomasHaas:
> Your description looks fine to me. Assignments can also have multi-linear expressions.

Yes, that's why I have the part about `k*x` being understood as a vector.

> The underlying algorithm has much in common with Sui & Lei's: _Fast and Precise Handling of Positive Weight Cycles for Field-sensitive Pointer Analysis_ (SAS'19, [DOI](https://doi.org/10.1007/978-3-030-32304-2_3)).

I looked for the Datalog-based analysis I told you about yesterday. Its name was not `ccanalyzer` but `cclyzer` from [Structure-Sensitive Points-To Analysis for C and C++](https://yanniss.github.io/cclyzer-sas16.pdf).
There is a newer version for C++ on github ([link](https://github.com/GaloisInc/cclyzerpp)). They also have an [online documentation](https://galoisinc.github.io/cclyzerpp/).
Interestingly, it this computes the call-graph on the fly.

hernanponcedeleon:
> Your description looks fine to me. Assignments can also have multi-linear expressions.

Can you then use this description for the code documentation?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+    private int succeededAddInto = 0;
+
+    // If enabled, the algorithm describes its internal graph to be written into a file.
+    private Graphviz graphviz;
+
+    // When a variable gains an includes-edge, it is added to this queue for later processing.
+    private final LinkedHashMap<Variable, List<Offset<Variable>>> queue = new LinkedHashMap<>();
+
+    // Maps memory events to variables representing their pointer set.
+    private final Map<MemoryCoreEvent, Offset<Variable>> addressVariables = new HashMap<>();
```
ThomasHaas, line 93: The fact that the algorithm uses `Offset<Variable>` rather than `Variable` in many places is confusing.
Is this just a kind of optimization to avoid intermediate variables or something?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+
+    // Maps memory objects to variables representing their base address.
+    // These Variables should always have empty includes-sets.
+    private final Map<MemoryObject, Variable> objectVariables = new HashMap<>();
+
+    // Maps dependent sets of a register to variables representing their phi node's value set.
+    // Singletons are mapped to their output set.
+    // Offsets are for Local output sets; non-singletons and singletons of loads should have zero offset.
+    private final Map<List<RegWriter>, Offset<Variable>> registerVariables = new HashMap<>();
```
ThomasHaas, line 102: Why not use sets as keys? Do you sort the lists or could it happen that you generate the same phi node twice just because the list is ordered differently?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+    private final Dependency dependency;
+
+    // For providing helpful error messages, this analysis prints call-stack and loop information for events.
+    private final Supplier<SyntacticContextAnalysis> synContext;
+
+    private static final class Variable {
+        // Any value contained in these Variables is also contained (+ offset).
+        private final List<Offset<Variable>> includes = new ArrayList<>();
+        // There is some load event using this (+ offset) as pointer set and the other variable as output set.
+        private final List<Offset<Variable>> loads = new ArrayList<>();
```
ThomasHaas, line 114: In the `includes`, a `Offset<Variable>` describes an incoming edge but in `loads` it is an outgoing edge.
This is confusing. Together with the fact that `Offset<Variable>` is sometimes also treated like a variable, I think this very hard to understand.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+        // There is some store event using this (+ offset1) as pointer set and the other variable (+ offset2) as value set.
+        private final List<Offset<Offset<Variable>>> stores = new ArrayList<>();
```
ThomasHaas, line 116: This is effectively a pair of `Offset<Variable>` (which is conceptually a pair of `Variable` if there had been dedicated variables), no? The representation via nested offsets is hard to understand.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+        // All variables that have a direct (includes/loads/stores) link to this.
+        private final HashSet<Variable> seeAlso = new HashSet<>();
```
ThomasHaas, line 118: Use `Set<Variable>`

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+                if (overlaps(load.offset - store.offset, store.alignment, load.alignment)) {
+                    addEdge(store.base.base, load.base, store.base.offset, store.base.alignment);
+                }
+            }
+        }
+    }
+
+    private record Offset<Base>(Base base, int offset, List<Integer> alignment) {
+        @Override public String toString() { return String.format("%s+%d+%sx", base, offset, alignment); }
+    }
```
ThomasHaas, line 472: I would move this declaration directly before or after `Variable` because those two classes are very tightly related.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,897 @@
...
+
+    private Variable newVariable(String name) {
+        totalVariables++;
+        return new Variable(null, name);
+    }
+
+    private static final class Collector implements ExpressionVisitor<Result> {
+
+        final HashSet<MemoryObject> address = new HashSet<>();
+        final HashSet<Register> register = new HashSet<>();
```
ThomasHaas, line 730: Use `Set<...>`

ThomasHaas:
After looking into the code I found the different use-cases of `Offset<...>` very confusing, especially why `Offset<Variable>` is used instead of `Variable` (i.e., the offset itself acts as a variable).
I believe this is done due to the following observation: if `A -f_includes-> B` is the only incoming includes-edge into `B`, then you not only know `f(A) \subseteq B` but even `f(A) = B` because you consider least solutions to the constraint graph.
In those cases you can avoid the variable `B` and use `f(A)` instead (i.e. use `Offset<Variable>` instead of `Variable`).
I guess this also improves precision, because if you kept `B` separately and computed the transitive closure of the graph, you would likely end up with many incoming include-edges and thus forget that `f(A)=B` holds (because you forget that all those edges are derived from a common "bottleneck" through A).
At the same time, `Offset<Variable>` is also used to represent edge labelings + either source or target of that edge (and `Offset<Offset<...>>` was even used to represent a tuple!)This overload of usages makes the algorithm hard to follow. 

Is there a way to make the different use-cases more apparent? Maybe by using different classes, or using interfaces or something?
Maybe one could have `Offset` (non-generic) just describe the edge labeling and `Variable` have two fields `(Variable, Offset)` where you have either have `(this, zero)` for "pure variables" or `(otherVar, offset)` for"derived variables". Alternatively, you could use a separate class `DerivedVariable` which is a tuple `(Variable, Offset)`. 
If you want to keep `Offset<T>` generic, you could also try to make the class abstract (rather than a record) and have `DerivedVariable extends Offset<Variable>` so that derived variables keep the same functionality but look differently (mimicking a typedef).

ThomasHaas:
I feel like your newest change has just shifted the variable vs. edge problem, no?
Previously, you used edges that sometimes represented variables. Now you use variables that sometimes represent edges.

xeren:
I fixed a bug in the analysis, where the communication rule was not correctly implemented.

The rule states that from $(X+o_1\xleftarrow{\text{stores}}W+o_2)$, $(W\supseteq A+o_3)$, $(A+o_4\subseteq R)$, $(R+o_5\xrightarrow{\text{loads}}Y)$ and $((o_2+o_3)\cap(o_4+o_5)\ne\emptyset)$, you can derive $X+o_1\subseteq Y$.

Either $o_2$ or $o_5$ was skipped in the intersection test (depending on where the recently-learned inclusion edge is), resulting in both false positives and false negatives.

> I feel like your newest change has just shifted the variable vs. edge problem, no?
> Previously, you used edges that sometimes represented variables. Now you use variables that sometimes represent edges.

Maybe you were referring to inclusion edges being typed as `DerivedVariable`.  I have now separated both use cases into separate classes.

ThomasHaas:
> I fixed a bug in the analysis, where the communication rule was not correctly implemented.
> 
> The rule states that from (X+o1←storesW+o2), (W⊇A+o3), (A+o4⊆R), (R+o5→loadsY) and ((o2+o3)∩(o4+o5)≠∅), you can derive X+o1⊆Y.
> 
> Either o2 or o5 was skipped in the intersection test (depending on where the recently-learned inclusion edge is), resulting in both false positives and false negatives.

Do you know of an instance of wrong behaviour? For example, does this affect qspinlock or ck_epoch or RCU?

> > I feel like your newest change has just shifted the variable vs. edge problem, no?
> > Previously, you used edges that sometimes represented variables. Now you use variables that sometimes represent edges.
> 
> Maybe you were referring to inclusion edges being typed as `DerivedVariable`. I have now separated both use cases into separate classes.

Yes, that is what I was referring to.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,967 @@
...
+    private final Map<MemoryCoreEvent, DerivedVariable> addressVariables = new HashMap<>();
+
+    // Maps memory objects to variables representing their base address.
+    // These Variables should always have empty includes-sets.
+    private final Map<MemoryObject, Variable> objectVariables = new HashMap<>();
+
+    // Maps dependent sets of a register to variables representing their phi node's value set.
+    // Singletons are mapped to their output set.
+    // Offsets are for Local output sets; non-singletons and singletons of loads should have zero offset.
+    private final Map<List<RegWriter>, DerivedVariable> registerVariables = new HashMap<>();
```
ThomasHaas, line 102: Everytime I read "dependent sets of a register" I have no idea what it means.
Why don't you say something like this this (in particular the first sentence):
```
Maps a set of same-register writers to a variable representing their combined value (output?) sets (~phi node).
For singleton Locals we use derived variables; non-singleton sets and loads are assigned plain variables.
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,967 @@
...
+
+    // When a variable gains an includes-edge, it is added to this queue for later processing.
+    private final LinkedHashMap<Variable, List<IncludeEdge>> queue = new LinkedHashMap<>();
+
+    // Maps memory events to variables representing their pointer set.
+    private final Map<MemoryCoreEvent, DerivedVariable> addressVariables = new HashMap<>();
+
+    // Maps memory objects to variables representing their base address.
+    // These Variables should always have empty includes-sets.
+    private final Map<MemoryObject, Variable> objectVariables = new HashMap<>();
```
ThomasHaas, line 97: I'm wondering if you could also use a `DerivedVariable` here (with trivial modifiers) just for the sake of consistency (unless it is algorithmically important that `MemoryObjects` get plain variables)

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,967 @@
...
+        // Any value contained in these Variables is also contained (+ offset).
+        private final List<IncludeEdge> includes = new ArrayList<>();
+        // There is some load event using this (+ offset) as pointer set and the other variable as output set.
+        private final List<LoadEdge> loads = new ArrayList<>();
+        // There is some store event using this (+ offset1) as pointer set and the other variable (+ offset2) as value set.
+        private final List<StoreEdge> stores = new ArrayList<>();
+        // All variables that have a direct (includes/loads/stores) link to this.
+        private final Set<Variable> seeAlso = new HashSet<>();
+        // If nonnull, this variable represents that object's base address.
+        private final MemoryObject object;
```
ThomasHaas, line 379: Most comments here are outdated. I also think that you can put most of the descriptions on the corresponding record classes (i.e., the explanation of the meaning of the edges).
I would just make sure that these comments (or even the variable names) make clear in which direction the edges go: ingoing or outgoing.

ThomasHaas (commented):
Overall, I like this newer version a lot more. I will do a proper review the next week.

ThomasHaas:
I think you are conflating parallel joining and sequential joining (composition) in your algorithm. These are two different things with different results and it is just accidental that your algo remains correct because you always join parallel edges where one of them is composed.

For example, parallel joining `2x + 2` and `4x + 2` should result in `2x + 2`, i.e., the gcd is used for the dynamic offset and the offsets are not added.
Sequential joining on the other hand would give `2x + 4y + 4` (or `2x + 4` if you combine multiple offsets) because the modifiers just get added. In the case where you would know that both dynamic offsets are the same variable, you would even get `6x + 4`

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,977 @@
...
+        if (left == TOP || right == TOP) {
+            return TOP;
+        }
+        // assert left and right each consist of pairwise indivisible positives
+        final List<Integer> result = new ArrayList<>();
+        for (final Integer i : left) {
+            if (!right.contains(i) && hasNoDivisorsInList(i, right)) {
+                result.add(i);
+            }
+        }
```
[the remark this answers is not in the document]
xeren: This is a bug.  It was supposed to be a guard to avoid adding common values twice.  `hasNoDivisorsInList` was supposed to return `true` if the list contained the value.

ThomasHaas:
Did you intentionally revert your recent changes? If so, you also reverted the changes to the handling of unary expressions, so they are still wrong.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -0,0 +1,977 @@
...
+        for (final MemoryObject object : collector.address) {
+            if (result == null || object != result.address) {
+                addInclude(variable, new IncludeEdge(objectVariables.get(object), RELAXED_MODIFIER));
+            }
+        }
+        for (final Register register : collector.register) {
+            if (result == null || register != result.register) {
+                final DerivedVariable registerVariable = getPhiNodeVariable(register, reader);
+                addInclude(variable, new IncludeEdge(registerVariable.base, RELAXED_MODIFIER));
+            }
```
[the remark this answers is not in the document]
xeren: Such a non-one factor breaks the GEP-like syntax.  The analysis would need to check / assume `f=1` on multiple occasions.

```
r = 4 * *(&y)
do
  *(&z + 3 * r) = 1
  r += 2
while (...)
```
In this example, the analysis is fed a program with a loop.  If the instructions are iterated in order, then the store could have an address of `1*z + 12*x + 0`, but should eventually arrive at `1*z + 6*x + 0`.
You may also consider this example without `* *(&y)`, where .
What I have in mind is to do this in combination with a mechanism for refining index ranges:
- Assume the empty set for `r` initially and create the variable `res(&z + 3 * r)` without inclusion edges.
- Whenever `r` includes `null + a*x + o`, add an inclusion `&z + 3*a*x + o` $\supseteq$ `res(&z + 3 * r)`.

This way, the analysis gains precision for all variables containing proper multiples, like in the example.  It should also get robustness against inlining of multiplications.

That to say, I did not consider robustness this important to the PR.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -842,9 +842,15 @@ yield new Result(null, null, BigInteger.ZERO,
...
+                return null;
+            }
             final Result result = x.getOperand().accept(this);
-            return result == null ? null : x.getKind() != IntUnaryOp.MINUS ? result :
-                    new Result(null, null, result.offset.negate(), result.alignment.isEmpty() ? TOP : result.alignment);
+            final var alignment = new ArrayList<Integer>();
+            for (final Integer a : result.alignment) {
+                alignment.add(-Math.abs(a));
+            }
+            return new Result(null, null, result.offset.negate(), alignment);
```
ThomasHaas, line 853: This is not correct for `-r`, no? It seems that you would return the constant `-0=0` rather than `0 + -1*x`.
I think whenever you have some base `b`, then this base should be considered 1-aligned when doing negation.

hernanponcedeleon:
I tested this on the libvsync benchmarks and all results look good. Unless there are further comments, I will merge.

ThomasHaas (approved):
LGTM

PR #618, opened 2024-02-16
Add libvsync tests

hernanponcedeleon:


ThomasHaas (approved):
LGTM.
Btw. the llvm code has some inline assembly. Not sure what it exactly does (compiler barrier with extra flags?)

hernanponcedeleon:
Yes, just compiler barriers (i.e. nops) required to "match" vmm with c11 builtins

Issue #623, opened 2024-02-23
Getting rid of (unnecessary) init events


ThomasHaas:
**Issue:** 
Init events are frequently a pain in the ass:
(1) Large allocations create a lot of init events which make executions large even if only a small subset of those addresses are used
(2) Allocations of unknown size (e.g. non-deterministic size) cannot be modelled
(3) Arbitrary (out-of-bounds) memory accesses cannot be modelled

A naive solution to those issues is to introduce the idea of an abstract init event that stands for all possible init events.
However, this causes many unexpected problems with the standard CAT models: `loc;loc` can now connect events of different locations by going over the abstract init event (which would be `loc`-related to all memory accesses).
As a consequence, definitions like `fr = rf^-1;co` also become wrong and need to get changed to `fr = rf^-1;co & loc` to recover the intended semantics.
I want to propose an alternative solution that avoids unnecessary init events, similar to how Alloy-based tools handle the issue.

**Idea:**
- Read events may have no rf-edge associated to them, designating a read from uninitialized memory. 
- We dynamically tag such events as `UR` (for `uninit read`).
- We redefine `fr := rf^-1;co | [UR];loc;[W]`, saying that reads from uninitialized memory are `fr`-before all stores to the same location without relying on an init event.

**What values do uninit reads obtain?:**
We could make the value of uninit reads configurable to be either `0` or non-deterministic.
In the latter case, we should probably make sure that different uninit reads from the same location return the same value. This could be achieved by using an uninterpreted function `InitValueOf(addr)`.

**How to handle statically initialized memory?**
We could actually keep init events for truly initialized memory. 
By adding the axiom `empty([UR];loc;[IW])` we would prohibit uninitalized reads from locations that have an init event.
Interestingly, we could also easily detect reads from uninitialized memory (which is UB) by having `flag ~empty([UR])`.
Of course, these init events would still cause problems for issues (1) and (2), however, only when statically initializing (dynamic allocations like `malloc` would not create any init events).

----------------

**Implementation issue: We don't support dynamic tagging!** 
Since we do not support dynamic tagging right now, we can opt to define the relation `[UR]` rather than the set `UR`.
This would be sufficient for all of the above use-cases.

**Issue: We still need to rewrite the CAT-models:**
It is true that we need to rewrite the CAT-models.By changing our predefined notion of `fr`, we already fix most CAT models.
However, relations like `rf^-1;rf` (as used in RISCV) would also require changes. 
[...]

Issue #136, opened 2021-10-08
Allowing dynamically sized mallocs


ThomasHaas:
Right now, Dartagnan overapproximates the set of all possible addresses but to do so, it uses two assumptions:
- There is a fixed and known number of mallocs
- Each malloc has fixed and known size

Why do we do this? Well, if we have a fixed set of addresses, we can compute for each memory event the set of addresses it could potentially affect. We use this information to obtain alias information. However, the encoding of memory accesses doesn't care at all. We can use arbitrary address expressions for each memory event that evaluate to any arbitrary integer without regarding the address space at all. In other words, we are not tied to a fixed memory space in the encoding but rather in the current alias analysis.

Let's assume that we drop the second assumption, that is, we still have a fixed set of mallocs but each of them is dynamically sized. For each malloc k, we can express the malloced memory region as an interval `[base_k, base_k + size_k]`. Now let us create symbolic addresses for each of the bases and let `size_k` be given by any integer expression
We can now encode `0 < base_1, base_1 + size_1 < base_2, base_2 + size_2 < base_3 ...`
This will guarantee that no matter to what value any of the size expressions evaluate, we have non-overlapping memory.
In the absence of UB, ordering the bases in this manner is also sound.

What shall the alias analysis do now? Well, we need to properly track bases. If we have two operations that access different bases, then they can never alias (assuming no undefined behaviour). If they access the same base, we need to compare the offsets to determine whether they can alias or not. If we cannot find any set of possible bases for some memory access, we can still assume the whole address space. If we could establish that all pointers in the program point to some base (i.e. to some malloced object), we could even improve our alias analysis (if we load a ptr and don't know its value, a consequent load from that ptr-addr has to be a `base` so we can still say that accesses with different offsets cannot alias, even if we do not know their bases). This is actually quite frequently the case, unless we have pointers to inside a struct/array.

So if I'm not mistaken, then the only reason we do not have dynamically sized memory is because of our alias analysis. And I believe our alias analyses could be adapted to argue about `base` + `offset` (I think it does it already to some extent).

ThomasHaas:
@hernanponcedeleon and me talked about possible solutions before. The easiest we've come up with works as follows.
- We create init events per read event so that each read is guaranteed to have an associated init. Furthermore, as programs are bounded, this will automatically bound the number of init events we need.
- If we have must-alias information, we can merge multiple init events to a single one.
[...]

Issue #371, opened 2023-01-23
Support for proper unary predicates


ThomasHaas:
This issue is just a reminder that we should eventually support dynamic tagging (= arbitrary unary predicates) and not just static tagging (= our current filters).

hernanponcedeleon:
Once we have that, [here](https://github.com/hernanponcedeleon/Dat3M/blob/95e3c11598d275e31c2a4b973bdab056bff48d9b/dartagnan/src/test/resources/dartagnan-skip.csv#L8) we have a test case

hernanponcedeleon:
Related to #254

ThomasHaas:
Added in #908

PR #624, opened 2024-02-23
Reads without init events

ThomasHaas:
This PR attempts to address #623 .

There is some cleanup to do. 
There is also a problem with removing init events from litmus tests. I need to debug this (I suppose it is related to co-last memory values?)

EDIT: It also partially addresses #136


ThomasHaas:
I need some feedback on how far we want to go in this branch.
(1) If I make sure that we generate all init events as before (even for dynamic allocations), everything works(*) and this PR is pretty much done.
(2) If we don't create init events for dynamic allocations, at least C11 needs some rewriting. The current implementation already changes the definition of `fr` which, together with SC-per-loc, seems to be sufficient for most models to handle to reads from uninitialized memory (I cannot give a 100% guarantee on correctness though!). 
However, `C11` defines it's own `fr` (which we can remove) and does not have SC-per-loc and thus requires the rewriting of an axiom.


(*) As long as the programs are really memory-safe, there should be no change to our current semantics. For programs that can read memory out-of-bounds, we previously could give a spurious PASS as the read had no possible rf-edges and we always required the existence of one. This has changed in this PR: the out-of-bounds reads are possible.

hernanponcedeleon:
A few questions / comments that come to my mind
- the current implementation would return PASS ("be correct" might be an over-claim since the program has UB) for this variant of the `uninitRead.c`, right?
```
int main()
{
    array[SIZE] = 42;
    assert (array[SIZE] == 42);
    return 0;
}
```
At some point I thought handling out-of-memory would not work out-of-box due to our alias analysis relying on defined behavior 
- we need to be sure we properly handle all variants of alloc (malloc, calloc, ...). AFAIR they vary in how (if) the initialize memory which used to be irrelevant for us when we zeroed out all memory
- I don't think the last question is irrelevant now: if one set the new options to false, we return unsound results now, so this still needs to be discussed / fixed
- which is the axiom giving problems in c11? The chain of co and rf in Coh?
- we should add a TODO to the part where you programmatically set the new fr. This should be moved to some common cat once we manage imports

ThomasHaas:
> A few questions / comments that come to my mind
> 
>     * the current implementation would return PASS ("be correct" might be an over-claim since the program has UB) for this variant of the `uninitRead.c`, right?
> 
> 
> ```
> int main()
> {
>     array[SIZE] = 42;
>     assert (array[SIZE] == 42);
>     return 0;
> }
> ```

I think the alias analysis will prohibit those two from aliasing in both the new and the old version, so you will get unsound results in general. In this particular case I don't know what result you will get (possibly PASS for both, but for the wrong reasons).
Anyway, this is simply due to the alias analysis prohibiting this and not due to our anarchic semantics.
For example, if you would load the index from memory, say `int index = SIZE; .... array[index] = 42; assert (...)` the alias analysis will think they can alias and then we still get correct results.
Also note that everything works fine if you access uninitialized memory that has been allocated (dynamic allocs) even if that is UB: the alias analysis handles those correctly!
In that sense, this PR is a conservative extension of our current code: we soundly handle some more types of UB memory accesses than before but we still do not handle all of them. Also, by adding `flag ~empty (UR)` you can now check if the program has UB.

> At some point I thought handling out-of-memory would not work out-of-box due to our alias analysis relying on defined behavior
> 
>     * we need to be sure we properly handle all variants of alloc (malloc, calloc, ...). AFAIR they vary in how (if) the initialize memory which used to be irrelevant for us when we zeroed out all memory

Yes, `calloc` zeroes out the memory and thus would create init events as opposed to `malloc`.

> 
>     * I don't think the last question is irrelevant now: if one set the new options to false, we return unsound results now, so this still needs to be discussed / fixed

Of course, but now we have the possibility to defer the discussion/the fixing. A proper fix would likely need a lot of practical testing.
And there are a lot of things one needs to think about:
(1) For every memory model you need to check if the new `fr` is sufficient or if other changes are necessary. I think in particular C11 and the GPU models need rework. But you could also argue for now that those models are just not compatible with uninit reads and so combining them is a user error rather than a tool error.
(2) What about final memory values if there is no store to that location? Our current encoding requires at least one store to exist and, e.g., litmus tests would fail if you removed init events.


>     * which is the axiom giving problems in c11? The chain of co and rf in Coh?
It's the `Coh` axiom. You need to change it like this I believe:
```
irreflexive ((rf^-1)? ; mo ; rf? ; hb) as Coh // old version
irreflexive ((fr | mo) ; rf? ; hb) as Coh // new version
```
> 
>     * we should add a TODO to the part where you programmatically set the new fr. This should be moved to some common cat once we manage imports

I don't think a TODO is required specifically for `fr` because all relations defined in `getOrCreatePredefinedRelation` need to be revisited once you add imports. In particular, even if you kept the current `fr`, you would still need to remove its definition from `getOrCreatePredefinedRelation` when adding imports.

hernanponcedeleon:
> > * we need to be sure we properly handle all variants of alloc (malloc, calloc, ...). AFAIR they vary in how (if) the initialize memory which used to be irrelevant for us when we zeroed out all memory
> > ```
> 
> Yes, `calloc` zeroes out the memory and thus would create init events as opposed to `malloc`.
> 

Both intrinsics seem to be handled in a similar way. The only diff I see is that `calloc` takes an  extra argument and the size is computed differently, but I don't yet see how we would enforce initialization in the `calloc` case, but not on `malloc`.

> > ```
> > * I don't think the last question is irrelevant now: if one set the new options to false, we return unsound results now, so this still needs to be discussed / fixed
> > ```
> 
> Of course, but now we have the possibility to defer the discussion/the fixing. A proper fix would likely need a lot of practical testing. And there are a lot of things one needs to think about: (1) For every memory model you need to check if the new `fr` is sufficient or if other changes are necessary. I think in particular C11 and the GPU models need rework. But you could also argue for now that those models are just not compatible with uninit reads and so combining them is a user error rather than a tool error. (2) What about final memory values if there is no store to that location? Our current encoding requires at least one store to exist and, e.g., litmus tests would fail if you removed init events.
> 

If the option is not compatible with a given model, we should at least check this and throw an exception if the combination is used. 

From the testing perspective, the minimum would be to fix the cases we already known are wrong. I just run all unit tests and I saw only two problems: (1) related to C11 which might be fixed changing the model as you explained, (2) a timeout on RISCV, maybe related to the `rf^-1 ; rf` (or some similar pattern).

> > ```
> > * we should add a TODO to the part where you programmatically set the new fr. This should be moved to some common cat once we manage imports
> > ```
> 
> I don't think a TODO is required specifically for `fr` because all relations defined in `getOrCreatePredefinedRelation` need to be revisited once you add imports. In particular, even if you kept the current `fr`, you would still need to remove its definition from `getOrCreatePredefinedRelation` when adding imports.

I was under the impression that this was the only remaining non-base relation that reamined there, but you are right we still have a few, e.g. `{rf, co} \cap external`

ThomasHaas:
> Both intrinsics seem to be handled in a similar way. The only diff I see is that `calloc` takes an extra argument and the size is computed differently, but I don't yet see how we would enforce initialization in the `calloc` case, but not on `malloc`.

That should be straightforward by just adding an extra field to `Alloc` saying whether the memory gets zeroed out or not.

> If the option is not compatible with a given model, we should at least check this and throw an exception if the combination is used.

I don't think this is so straightforward because how would you know that the provided model is incompatible? We could 
legitimately have two C11 models, one being the original (incompatible one) and one being the new one.
We don't even throw exceptions if the CAT model and the target architecture are mismatching and that one is a bigger source of errors.
Basically, throwing exceptions based on the CAT model is not feasible because the only way we can detect what memory model is used is by checking its name/filepath, but I don't think this is a smart idea.

> From the testing perspective, the minimum would be to fix the cases we already known are wrong. I just run all unit tests and I saw only two problems: (1) related to C11 which might be fixed changing the model as you explained, (2) a timeout on RISCV, maybe related to the `rf^-1 ; rf` (or some similar pattern).

The question is if the change to the model is a real fix or if it just hides one problem but creates different one's?
All the GPU models are also wrong but you don't see it in the unit tests because litmus tests don't have dynamic allocation.
I can "fix" C11 and leave the original axiom commented out in the file (similar to how we commented out all the lock stuff we don't support) though.
Btw. what RISC-V timed out? That's surprising because the `rf^-1;rf` pattern is only used in the original RISC-V but not in my rewritten version. `rf^-1;rf` would also need to get updated to be compatible with missing init events, though I cannot imagine a case where something breaks.

ThomasHaas:
I changed C11 and reworked `calloc`. Since the unit tests should work now with the new settings, should I enable them by default?

cat/c11.cat
```diff
@@ -20,7 +20,8 @@ let cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
 let crel = REL | (SC & (W | F)) | ACQ_REL
 let ccon = R & CON
 
-let fr = rf^-1 ; mo
+// To be compatible with reads from uninit memory, we use the default definition of fr
+//let fr = rf^-1 ; mo
```
hernanponcedeleon, line 24: Rather than editing this file, I think it is better to have two cat files, one called `_orig` similar to what we did for RISCV.
Since we are going to rely on the default definition of fr what is based on co, we might want to avoid to renaming of `sb=po` and `mo=co`.

Also, it might be a good idea to create two new test classes that extend the C11 ones (locks and lfds) by simply overriding the method selecting the cat file. This was we have all tests run with both models and guaranteeing the same expected result.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
```diff
@@ -41,14 +68,22 @@ public void run(Program program) {
     }
 
     private void processAllocations(Program program) {
+        final ExpressionFactory expressions = ExpressionFactory.getInstance();
+        // FIXME: We should probably initialize depending on the allocation type of the alloc
```
hernanponcedeleon, line 72: I don't get this comment. Isn't the check at line 82 checking this?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
```diff
@@ -13,6 +13,7 @@ public class RelationNameRepository {
     public static final String EXT = "ext";
     public static final String CO = "co";
     public static final String RF = "rf";
+    public static final String UR = "UR";
```
hernanponcedeleon, line 16: isn't `UR` a set and `[UR]` the relation?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/AddInitReadHandling.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.dat3m.dartagnan.program.event.Tag;
+import com.dat3m.dartagnan.wmm.Relation;
+import com.dat3m.dartagnan.wmm.RelationNameRepository;
+import com.dat3m.dartagnan.wmm.Wmm;
+import com.dat3m.dartagnan.wmm.axiom.Emptiness;
+import com.dat3m.dartagnan.wmm.definition.Composition;
+import com.dat3m.dartagnan.wmm.definition.SetIdentity;
+
+// Add empty(UR;loc;[IW]) to enforce that initial writes are read if available.
```
hernanponcedeleon, line 11: `UR` -> `[UR]`?

hernanponcedeleon:
> EDIT: It also partially addresses #136

Out of curiosity (no need to get this done in this PR unless it is trivial): would non-det allocs work out of the box if we simple remove the exception when the size if not constant or what else needs to be implemented to handle those?

hernanponcedeleon:
> I changed C11 and reworked `calloc`. Since the unit tests should work now with the new settings, should I enable them by default? (EDIT: I accidentally enabled the new setting already)

I think it is good to get this option enabled so it is always tested (and I'm trying to come up with an easy way to extend all our test by enabling some options so we get most options tested, maybe not in every PR, but at least in some nightly/weekly job) and we can eventually find problems

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
ThomasHaas: We cannot define dynamic sets (see #371 ) right now, so I had to define `UR` to be the set identity (binary relation) on uninitialized reads. That's why the derived relations use `UR` rather than `[UR]`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/AddInitReadHandling.java
ThomasHaas: See above comment

ThomasHaas:
> > EDIT: It also partially addresses #136
> 
> Out of curiosity (no need to get this done in this PR unless it is trivial): would non-det allocs work out of the box if we simple remove the exception when the size if not constant or what else needs to be implemented to handle those?

There are two more things:
(1) The alias analysis needs to support this. IIRC, the current one won't but the new one in #616 can handle it (maybe with small adaptions)
(2) The encoding of the memory layout cannot be completely static anymore. In particular, `MemoryObject.size` and `MemoryObject.address` have to become proper `Expressions`. Actually, `MemoryObject.address` should go away completely because the `MemoryObject` itself already represents its address. Instead, you would introduce a `Map<MemoryObject, Formula> obj2Addr` in `EncodingContext` that captures the address formulas.
Then `ProgramEncoder.encodeMemory` would need to generate constraints to make sure all objects are non-overlapping and properly aligned. 
A little trick here is to reorder the memory so that all objects with known size come first because then you could still generate constant addresses for them.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
ThomasHaas: By `allocation type` I mean the actual type of object that gets allocated. For mallocs, we allocate `bytes` (cause we don't know better) but for example `alloca` has a proper type. In that case, we could generate init events only at the correct offsets, e.g. every 4 bytes if allocating an `int32` array.

cat/c11.cat
ThomasHaas: I can create a separate cat file. 
> Since we are going to rely on the default definition of fr what is based on co, we might want to avoid to renaming of sb=po and mo=co.

I can also put an explicit definition of `fr` into the cat file that uses `mo` instead of `co`.

> Also, it might be a good idea to create two new test classes that extend the C11 ones (locks and lfds) by simply overriding the method selecting the cat file

And changing the options I guess? Cause the original C11 one won't work without init events.
There are technically three cases, all of which should produce the same results
- Old model with all init events
- New model with all init events
- New model without all init events

ThomasHaas: I added the extra c11 cat file, created two new tests (for locks and lfds) and enabled CAAT on all C11 tests (they work now due to improvements in cutting).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
hernanponcedeleon: I know we cannot yet handle unary predicates. However, isn't the situation and the problem exactly the same that we have with e.g., `[range(r)]`? If so, we should use `[UR]` even if we are "hardcoding the relation" rather than creating it from a filter.

cat/c11.cat
```diff
@@ -20,7 +25,7 @@ let cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
 let crel = REL | (SC & (W | F)) | ACQ_REL
 let ccon = R & CON
 
-// To be compatible with reads from uninit memory, we use the default definition of fr
+// To be compatible with reads from uninit memory, we use our default definition of fr
 //let fr = rf^-1 ; mo
```
hernanponcedeleon: I think we can get rid of this altogether now that we have a copy of the original cat model. Also, I would use `po` and `co` ... in the original one I was using `sb` and `mo` to keep it as faithful to the one I took from herd's repo as possible

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLFDSTest.java
```diff
@@ -0,0 +1,90 @@
...
+        });
+    }
+
+    @Test
+    public void testAssume() throws Exception {
+        AssumeSolver s = AssumeSolver.run(contextProvider.get(), proverProvider.get(), taskProvider.get());
+        assertEquals(expected, s.getResult());
+    }
+
+    // CAAT might not yet work for C11
```
hernanponcedeleon: Does CAAT only work on the new model or can we also enable these tests?

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLFDSTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @Parameterized.Parameters(name = "{index}: {0}, target={1}")
+    public static Iterable<Object[]> data() throws IOException {
+        // Commented ones take too long ATM
+        return Arrays.asList(new Object[][]{
+                // {"dglm", C11, UNKNOWN},
+                {"dglm-CAS-relaxed", C11, FAIL},
+                // {"ms", C11, UNKNOWN},
+                {"ms-CAS-relaxed", C11, FAIL},
+                // {"treiber", C11, UNKNOWN},
+                {"treiber-CAS-relaxed", C11, FAIL},
```
hernanponcedeleon, line 69: Have you tried enabling these test? I think back in the day we commented them out because it took too long, but performance has improve since then

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLocksTest.java
```diff
@@ -0,0 +1,101 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+import static com.dat3m.dartagnan.configuration.Arch.C11;
+import static com.dat3m.dartagnan.utils.ResourceHelper.getTestResourcePath;
+import static com.dat3m.dartagnan.utils.Result.*;
+import static org.junit.Assert.assertEquals;
+
+@RunWith(Parameterized.class)
+public class C11OrigLocksTest extends AbstractCTest {
```
hernanponcedeleon, line 26: Why not t o extend `CLockTest` instead and just override the needed methods? I explicitly wanted to avoid having the test completely decoupled to
1) be sure we have the same expected results for both
2) whenever we add a test, we only need to added to one class

cat/c11.cat
hernanponcedeleon: Do you remember if this was the same reason why we do not have caat enabled for litmus test (at least for those checking safety)

ThomasHaas: Well, we have never enabled CAAT for litmus tests I believe (independent of the memory model). Of course, for C11 we couldn't enable it due to the same problems with cutting. Btw. CAAT is slower on the C11 tests than the assume solver!

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
ThomasHaas: No, it's not the same situation. In the case of `[range(r)]` the trick was to define the combined operator `[range( . )]` directly which also means that the parser has a special rule for that.
To do the same with `[UR]` you would need to change the grammar (or at least the parser) to have a special case just for that one relation. I think that is not a good idea because it is super ambiguous.

Btw. I could also avoid a new base relation and instead define `UR = [R] \ [range(rf)]`. I guess there is not much difference in encoding and verification performance.

cat/c11.cat
ThomasHaas: Do you really want to change everything? We can do it, but it makes it harder to compare both models.
That's also why I kept the original code in comments to more easily see where something was changed.

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLFDSTest.java
ThomasHaas: It should work for both versions, I just missed to remove the comment.

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLFDSTest.java
ThomasHaas: I have not tried it. I think enabling CAAT for C11 was already a change that I should not have made in this PR. These changes are orthogonal to the PR and would better be done separately.

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLocksTest.java
ThomasHaas: Right, that should be possible. Will do

cat/c11.cat
hernanponcedeleon: I kind of disliked to have all that commented code, but I agree that by keeping them we can simply use diff and easy check the changes between those models, so let's do it that way

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLFDSTest.java
hernanponcedeleon: I agree. I just noticed this (did not even remembered it) while reviewing the diff in the code

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLocksTest.java
ThomasHaas: Done

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
ThomasHaas: I replaced `UR` by `[R] \ [range(rf)]` to simplify the code even more.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -203,8 +203,17 @@ private Relation makePredefinedRelation(String name) {
             case CTRLDIRECT -> new DirectControlDependency(r);
             case EMPTY -> new Empty(r);
             case FR ->  {
-                Relation rfinv = addDefinition(new Inverse(newRelation(), getOrCreatePredefinedRelation(RF)));
-                yield composition(r, rfinv, getOrCreatePredefinedRelation(CO));
+                final Relation reads = addDefinition(new SetIdentity(newRelation(), getFilter(Tag.READ)));
+                final Relation rfRange = addDefinition(new RangeIdentity(newRelation(), getOrCreatePredefinedRelation(RF)));
+                final Relation ur = addDefinition(new Difference(newRelation(), reads, rfRange));
+                final Relation rfinv = addDefinition(new Inverse(newRelation(), getOrCreatePredefinedRelation(RF)));
+                //final Relation ur = getOrCreatePredefinedRelation(UR);
```
hernanponcedeleon, line 210: This can go.
Also, it would be useful to add a documentation comment with the new definition of `fr` in plain cat (reading this programmatically is not straight forward)

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11OrigLocksTest.java
```diff
@@ -0,0 +1,33 @@
...
+        super(name, target, expected);
+    }
+
+    @Override
+    protected Configuration getConfiguration() throws InvalidConfigurationException {
+        return Configuration.builder()
+                .copyFrom(super.getConfiguration())
+                .setOption(OptionNames.INIT_DYNAMIC_ALLOCATIONS, "true")
+                .build();
+    }
```
hernanponcedeleon, line 27: You said there are 3 configurations that makes sense. We are missing one where `OptionNames.INIT_DYNAMIC_ALLOCATIONS` is set to false (same for the LFDS test)

ThomasHaas: Well, I said there are (at least) 3 configurations but I didn't intend to have all of them. Technically, I could come up with even more combinations for all tests under all memory models :P. I think at some point you have to draw the line.
Anyway, I can add this one extra case for C11.

hernanponcedeleon: I thought we were missing the variant that was actually enabling the new code. Now I understand we have "new code + new model" and "old code + old model" which are the two most important cases, so feel free to ignore this review comment

PR #626, opened 2024-02-25
Wmm preprocessing

ThomasHaas:
This PR reworks Wmm preprocessing to work similar to how program preprocessing works.
It further adds the following passes
- `RemoveDeadRelations` which removes relations that do not contribute to axioms. This was already executed in the case of Refinement but now it runs also for the eager encoding although it shouldn't ever remove anything (though we could remove flagged axioms if we do not check for CAT properties which would create a lot of dead relations)
- `MergeEquivalentRelations` which attempts to find relations with equivalent relations, replacing them by a single representative. AFAIK, currently it only removes a single `rf^-1` from `riscv-orig.cat` because it matches with another `rf^-1` that was implicitly created by parsing `fr`.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
@@ -13,7 +13,6 @@ import org.sosy_lab.common.configuration.Option;
 import org.sosy_lab.common.configuration.Options;
 
 import java.util.*;
-import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -143,10 +142,10 @@ public class Wmm {
     }
 
     public void removeDefinition(Relation definedRelation) {
-        checkArgument(!(definedRelation.definition instanceof Definition.Undefined),
-                "Already undefined relation %s.", definedRelation);
-        logger.trace("Remove definition {}", definedRelation.definition);
-        definedRelation.definition = new Definition.Undefined(definedRelation);
+        if (!(definedRelation.getDefinition() instanceof Definition.Undefined)) {
+            logger.trace("Remove definition {}", definedRelation.definition);
+            definedRelation.definition = new Definition.Undefined(definedRelation);
+        }
     }
 
     public void addFilter(Filter filter) {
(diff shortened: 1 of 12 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
```diff
@@ -59,8 +59,9 @@ public BooleanFormula getSMTVariable(EncodingContext ctx) {
 
     public BooleanFormula getSMTVariable(Axiom ax, EncodingContext ctx) {
         Preconditions.checkState(this == CAT_SPEC);
-        return ctx.getBooleanFormulaManager()
-                .makeVariable("Flag " + Optional.ofNullable(ax.getName()).orElse(ax.getRelation().getNameOrTerm()));
+        final String varName = ctx.getFormulaManager().escape("Flag " +
+                Optional.ofNullable(ax.getName()).orElse(ax.getRelation().getNameOrTerm()));
+        return ctx.getBooleanFormulaManager().makeVariable(varName);
```
hernanponcedeleon, line 64: Is there any of our examples where the escape is really necessary?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -143,10 +142,10 @@ public Relation addDefinition(Definition definition) {
...
 
     public void removeDefinition(Relation definedRelation) {
-        checkArgument(!(definedRelation.definition instanceof Definition.Undefined),
-                "Already undefined relation %s.", definedRelation);
-        logger.trace("Remove definition {}", definedRelation.definition);
-        definedRelation.definition = new Definition.Undefined(definedRelation);
+        if (!(definedRelation.getDefinition() instanceof Definition.Undefined)) {
+            logger.trace("Remove definition {}", definedRelation.definition);
+            definedRelation.definition = new Definition.Undefined(definedRelation);
+        }
```
hernanponcedeleon, line 148: For consistency, if you use `definedRelation.getDefinition()` in the first case, you should not use `definedRelation.definition` in the second case

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/FlattenAssociatives.java
```diff
@@ -0,0 +1,59 @@
...
+        flattenAssociatives(wmm, Intersection.class, Intersection::new);
+    }
+
+    private void flattenAssociatives(Wmm wmm,
+                                     Class<? extends Definition> cls,
+                                     BiFunction<Relation, Relation[], Definition> constructor
+    ) {
+        final List<Relation> relations = List.copyOf(wmm.getRelations());
+        final List<Constraint> constraints = wmm.getConstraints().stream()
+                .filter(c -> !(c instanceof Definition)).toList();
```
hernanponcedeleon, line 36: Wouldn't make more sense to use `c instanceof Constraint` for the filtering? If at some point `wmm.getConstraints()` could return yet a 3rd class, the current implementation would not have the intended behavior.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/MergeEquivalentRelations.java
```diff
@@ -0,0 +1,139 @@
...
+            return true;
+        }
+        final Definition def1 = r1.getDefinition();
+        final Definition def2 = r2.getDefinition();
+        if (def1.getClass() != def2.getClass()) {
+            return false;
+        }
+
+        if (r1.isRecursive() || r2.isRecursive()) {
+            // For now, we return false for different recursive relations.
```
hernanponcedeleon, line 74: If you have some idea of how this could be changed in the future, please add a TODO ... having the keywork might help when someone has some free time and looks for low-hanging fruits to improve the code

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/RemoveDeadRelations.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+    public static RemoveDeadRelations newInstance() {
+        return new RemoveDeadRelations();
+    }
+
+    @Override
+    public void run(Wmm memoryModel) {
+        final List<Constraint> constraints = List.copyOf(memoryModel.getConstraints());
+        final DependencyCollector collector = new DependencyCollector();
+        constraints.stream().filter(c -> !(c instanceof Definition)).forEach(c -> c.accept(collector));
```
hernanponcedeleon, line 28: Same comment as above about using instanceof Constraint

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
ThomasHaas: Yes and no. There are axioms in Spirv like `flag ~empty (locord \ loc) as AssertLocordSameLoc` which only worked so far because they are named. If you wrote `flag ~empty (locord \ loc)` instead, you would run into an exception.
I noticed this just because some memory model processing failed to preserve axiom names.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/FlattenAssociatives.java
ThomasHaas: `c instanceof Constraint` is true for everything returned by `w..getConstraints`, so no, it would not make more sense :P. The intention here is to get all constraints that are no definitions. 
That being said, I have not taken a proper look into this pass; this is preexisting code that was just moved from `Wmm` to an isolated pass.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/RemoveDeadRelations.java
ThomasHaas: Same answer as above :).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/MergeEquivalentRelations.java
hernanponcedeleon: Then I would remove the "for now" (and "TODO") all together

Issue #532, opened 2023-10-11
Alternatives to partial coherence in the anarchic semantics for PTX


ThomasHaas:
This issue is a summary of what we discussed in #528 

**Context:** PTX requires a partial notion of coherence rather than a total one. We adapted our anarchic semantics to allow coherence to be partial to accomodate this use-case. Our partial encoding simply drops the totality constraint, meaning that coherence is just any acyclic relation over stores to the same address. This causes two issues: (1) coherence need not be transitive and (2) it doesn't need to relate anything (it can just be empty).
Currently, we fix these short-comings by redefining `co = co+` (fixing point 1) and adding an axiom to force `co` to contain certain minimal pairs (e.g., atomic stores) in the PTX model.

**Problem:** The definition of coherence is now awkwardly split between the anarchic semantics and the cat model. Also, we need to switch to the partial coherence notion based on what architecture we are handling (partial for PTX, total for the rest).

**Solutions:** There are two solutions to that problem that allow us to keep the total coherence and avoid any case distinction based on the target architecture (i.e., the CAT model itself can define its coherence to be partial!)

Solution 1: We just use the total coherence order and then define in PTX a partial coherence order from the total order by restricting to `morally-strong` edges and adding causally-related stores:
```let co = ((co & morally-strong) | (([W]; cause; [W]) & loc) | (([IW];loc;[W]) \ id))+```
Here we still rely on the anarchic semantics of `co` and just derive a new relation that fits the needs of the PTX model. 

Solution 2: A more general approach is to define a partial `co` directly in CAT instead of relying on the anarchic semantics at all.
This requires a new primitive in the CAT language, namely "free relations". Suppose `free` was a relation that could take any value, then we can define
```
let co = ([W];free;[W] & loc)+ // co is ANY transitive relation that relates same-address writes
acyclic (co) // co is acyclic
empty (([W];morally-strong;[W] & loc) \ (co | co^-1)) // co is total on morally-strong writes
empty (([W];cause;[W] & loc) \ co) // co relates causally-ordered writes (even if they are not morally strong)
// Alternative to the last axiom by just extending the initial definition:
// let co = ([W];free;[W] & loc | ([W];cause;[W] & loc))+
// NOTE: In both versions the treatment of init writes needs to be added
```
This is a full specification without relying on any anarchic semantics. In a similar way, the user could define any new custom relation.

hernanponcedeleon:
[...]

PR #627, opened 2024-02-26
Free relations in CAT

ThomasHaas:
This PR adds the ability to define free relations in CAT via `let r = free()` (using function call syntax).

For example, with this it is possible to define `co` entirely in CAT:
```
let co = (free() & W*W & loc) // co is a relation on same-address writes
let co = co | [IW];loc;[W] \ id // co orders init writes before all other writes
let co = co+ // co is transitive
empty ((W*W & loc) \ (co | co^-1 | id)) // Co orders every same-address pair of writes ...
acyclic (co) // ... and it is acyclic, meaning it is a total order (per address)
```
Such an encoding would actually be more interesting for partial order `co` as defined in GPU models(*).
This way, we could likely get rid of special case treatment for GPU architectures as suggested in #532 .

More generally, this feature can be used to experiment more easily with the CAT model. For example, we could easily define a total order on (abstract) lock events and encode that `hb` should adhere to that order.

(*) Defining `co` in CAT comes with another issue (as discussed in #532) though: we use the standard notion of `co` to encode liveness properties and final memory values.
One would likely need to add further primitives like `finalVal(co)` and `liveness(co)` to tell the encoding which relation is responsible for defining final values.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
@@ -443,6 +443,21 @@ public class RelationAnalysis {
         }
 
         @Override
+        public Knowledge visitFree(Free def) {
+            final List<Event> visibleEvents = program.getThreadEventsWithAllTags(VISIBLE);
+            EventGraph must = new EventGraph();
+            EventGraph may = new EventGraph();
+
+            for (Event e1 : visibleEvents) {
+                for (Event e2 : visibleEvents) {
+                    may.add(e1, e2);
+                }
+            }
+
+            return new Knowledge(may, enableMustSets ? must : EventGraph.empty());
+        }
+
+        @Override
         public Knowledge visitProduct(CartesianProduct prod) {
             final Filter domain = prod.getFirstFilter();
             final Filter range = prod.getSecondFilter();
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
@@ -227,6 +227,9 @@ public class RelationAnalysis {
         Map<Relation, List<Delta>> qGlobal = new HashMap<>();
         for (Relation r : memoryModel.getRelations()) {
             Knowledge k = r.getDefinition().accept(initializer);
+            if (!enableMustSets) {
+                k = new Knowledge(k.getMaySet(), EventGraph.empty());
+            }
             knowledgeMap.put(r, k);
             if (!k.may.isEmpty() || !k.must.isEmpty()) {
                 qGlobal.computeIfAbsent(r, x -> new ArrayList<>(1))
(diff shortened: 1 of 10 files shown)


ThomasHaas:
What syntax is better to define a new and free relation: `free()` or `new()` (or other alternatives)?

hernanponcedeleon:
I think with `new()` is more clear to understand what this is about

ThomasHaas:
> I think with `new()` is more clear to understand what this is about

Ok, then I will change the syntax to `new()`, but I will keep the internal class name as `Free` since this better describes the semantics.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -442,6 +442,21 @@ public Knowledge visitDefinition(Definition def) {
...
+            EventGraph must = new EventGraph();
+            EventGraph may = new EventGraph();
+
+            for (Event e1 : visibleEvents) {
+                for (Event e2 : visibleEvents) {
+                    may.add(e1, e2);
+                }
+            }
+
+            return new Knowledge(may, enableMustSets ? must : EventGraph.empty());
```
hernanponcedeleon, line 457: Must is always empty do I don't see the point of having the conditional

ThomasHaas: I put it to have it uniform, i.e., if you decide to put a non-empty must-set later on, the option will still work. As long as we have this option, I would keep it as is. If we decide to remove the no must-set option, I would change it.

hernanponcedeleon: I did not see any easy way to specify the must set of a cat-introduced relation. Thus, realistically speaking, I don't think we will ever have a non-empty must set here

ThomasHaas: Realistically speaking we won't have non-empty must-sets. But I would like to avoid giving this one relation a different treatment that is not any more performant. I think it would just break a pattern for no benefit (with some practically unrealistic pitfalls).

hernanponcedeleon: There are a few other relations that do not use the enable must set option. Even more, `visitSyncFence` explicitly uses the empty graph

ThomasHaas: I think this is because those were made by @tonghaining and he was not aware of the existence of that option.
I think the option is implemented badly: the baseline code should simply always compute may/must-sets and the propagator engine should simply discard the must-set if the option is enabled.
Ideally, this would move the handling of that option into a single place into the code and developers don't need to think about the existence of that option when implementing their new relations/definitions.

hernanponcedeleon: I have the feeling that having this fine grained option is useless. Having a way to disable the whole RA might be useful, but AFAIR Implementing this was not as trivial as it sounds. @nagacek mentioned recently that debugging certain things gets harder with so many optimizations.

ThomasHaas: I can see if I can move that option into a central place and if not, I will just remove it.

ThomasHaas: I moved the handling into a central place. Incidentally, I found bugs in RA when computing must-sets for Range/Domain relations.

PR #630, opened 2024-03-03
Fix time printing

ThomasHaas:


dartagnan/src/main/java/com/dat3m/dartagnan/utils/Utils.java
```diff
@@ -18,9 +18,9 @@ public static String toTimeString(long milliseconds) {
         if (hours == 0 && minutes == 0) {
             return String.format("%d.%d secs", seconds, milliseconds);
         } else if (hours == 0) {
-            return String.format("%d:%d mins", minutes, seconds);
+            return String.format("%d:%02d mins", minutes, seconds);
```
hernanponcedeleon, line 21: Wouldn't it be better to use `"%d mins %02d secs", minutes, seconds`?
I personally find that format easier to read

ThomasHaas: I like the other format better, but I can change it. In that case, you would also prefer `%d hours, %02d mins, %02d secs`? Seems kinda verbose to me

hernanponcedeleon: I played around with a few more examples and I'm fine with the current format

PR #631, opened 2024-03-03
Run opt also on llvm files

hernanponcedeleon:


ThomasHaas:
Are you sure you want to always run opt?

hernanponcedeleon:
Why not? vsyncer calls dartagnan using the llvm file directly and thus we loose all optimizations. With the changes in this PR I went from OOM in some SMR benchmarks to handle the full verification in very short time

ThomasHaas:
Because always running `opt` has the big disadvantage that you cannot faithfully share benchmarks anymore, since they will be affected by the optflags and the clang version of the user.

Why don't you just change `vsyncer` to run `opt`? From what I can see, this seems to be a trivial change in [checker_dartagnan.go](https://github.com/open-s4c/vsyncer/blob/main/checker/checker_dartagnan.go), no?

hernanponcedeleon:
What do you exactly mean by "share"? When we share benchmarks between developers, we normally share them in llvm format, which means that they already depend on clang. Moreover, I normally share the `-opt.ll` version which also depends on the opt flags.

If what you mean is that we cannot run the analyzes on the exact same IR, which is the case now when we use llvm files because there are no more optimizations involved (besides the ones implemented in dartagnan), this is mostly done for debugging purposes (*) and can be easily achieved by setting temporary OPT FLAGS to be empty.

(*) I also considered the option of running opt directly on vsyncer. The reason I chose to do this on dartagnan is because we are already doing it for C files and I want the whole analysis of C or llvm code to be as close as possible (it might not be the same becase we have no control on what flags or optimizations were initially used to generate the llvm file). The disadvantages you claim this PR adds are already there when the input file is C.

ThomasHaas:
> What do you exactly mean by "share"? When we share benchmarks between developers, we normally share them in llvm format, which means that they already depend on clang. Moreover, I normally share the `-opt.ll` version which also depends on the opt flags.

You running clang/llvm and the Dartagnan user running clang/llvm is very different. What is important is the replicability: I don't need to care about what settings you had when generating the code, because Dartagnan will behave identical on my side.

> If what you mean is that we cannot run the analyzes on the exact same IR, which is the case now when we use llvm files because there are no more optimizations involved (besides the ones implemented in dartagnan), this is mostly done for debugging purposes (*) and can be easily achieved by setting temporary OPTFLAGS to be empty.

Yes, I'm mostly concerned with debugging. It is possible to clear OPTFLAGS, but that is annoying on various levels:
- When running from the IDE, changing environment variables back-and-forth is annoying.
- Unit tests that run Dartagnan as a whole need to ignore OPTFLAGS or they will be affected by the user's setup.
- How many times has one of us forgotten to set the correct OPTFLAGS?

Also, LLVM opt is not idempotent. If the user puts in a `.c` to generate the `.ll` code and then later uses that `.ll` directly as input, the result may be different. So if you sent me your `.ll` code generated by standard settings and I run Dartagnan on that also using standard settings, I can still get different code.
Also, have you checked what happens if the user uses the `output/...-opt.ll` as input? It might cause a crash due to reading and writing to the same file? Or do you end up with another version `output/...-opt-opt.ll`?
The fact that each Dartagnan invocation will always generate temporary files doesn't seem too nice as well.

> (*) I also considered the option of running opt directly on vsyncer. The reason I chose to do this on dartagnan is because we are already doing it for C files and I want the whole analysis of C or llvm code to be as close as possible (it might not be the same becase we have no control on what flags or optimizations were initially used to generate the llvm file). The disadvantages you claim this PR adds are already there when the input file is C.

C code needs to be treated relative to the user, simply because it is affected by the user's CPU architecture, his standard library and ultimately his compiler version. There is no way around it. I don't buy the argument that because Dartagnan cannot guarantee reproducibility on `.c` files that it is ok to also not provide reproducability on `.ll` files.

In summary:
- Dartagnan should have some input format that it can consume as is (by default). This guarantees deterministic reproducibility on different machines.
- Dartagnan CAN have additional options to preprocess the program such as running `opt` but those should be optional. One could argue that `OPTFLAGS=empty` should be considered the default and so Dartagnan will reproduce results correctly. If we go with that argument though, you should make sure to not invoke `opt` and not produce additional files if `OPTFLAGS` is empty. Alternatively, there could be a commandline option `--run-opt` that makes Dartagnan run `opt` before verification.
- Unit tests should not be affected by `OPTFLAGS` settings (or any other environment settings), so the unit tests need to get adapted (unless we use the alternative option `--run-opt`).

hernanponcedeleon:
Ok, I buy your argument of trying to keep the sources of non-determinism to a minimum.

PR #633, opened 2024-03-09
Fix svcomp iterative approach

hernanponcedeleon:


ThomasHaas:
Is this PR meant to only change the output string? I don't see how this is related to SVCOMP

hernanponcedeleon:
The code iterates over the console lines to find the verification output (one line before the changed code in this PR) and then it uses this to decide if the iterative approach should continue by incrementing the bound or not.

ThomasHaas:
Ok I see. Checking against details of the output like the `(ms)` part sounds not so clever to me :P

ThomasHaas (approved):
LGTM

Issue #523, opened 2023-10-02
[BUG] Incorrect RA computation for setminus relation


natgavrilenko:
In a setminus relation, the superset should have a dependency to the complement to ensure correct computation of (regular) relation analysis sets.

Scenario:
- May and must sets computed for the superset (complement hasn't been processed yet).
- May and must sets of the superset are copied to the may and must sets of the parent setminus relation.
- May and must sets computes for the complement.
- May and must sets of the setminus relation (and its parents) are not reduced according to the information from the complement (propagation always increments sets).

Impact:
- Not updating may set in a setminus relation can lead to unnecessary computation.
- Not updating must set in a setminus relation can lead to an incorrect verification result.

Possible solutions:
- Add a dependency a superset to the complement to ensure that the complement is computed first.
- Propagate decremental updates to a setminus and its parent relations.

ThomasHaas:
The "decremental updates" solution is not going to work: decremental updates are a lot harder to do in general and AFAIK impossible in recursive relations.
A temporary solution with a (fake) dependency seems to be the best for now.

A more sophisticated solution is by not "propagating" at all and instead computing the may/must set of a relation only once all its dependencies are computed already (except in the case of recursion, but then there cannot be set differences inside the stratum). This was how it was originally done and it should also be more efficient anyway.

natgavrilenko:
Yes, I agree, precomputing will be the best solution.

ThomasHaas:
We have a temporary workaround code implemented since #524 . We should still keep this issue open.

hernanponcedeleon:
We this issue be closed now?

ThomasHaas:
I guess we can close it.

PR #635, opened 2024-03-12
Fix RA fix

ThomasHaas:
This PR fixes a bug in the fix from #523.
The fix is still ugly but now it should be correct at least :).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
@@ -216,16 +216,16 @@ public class RelationAnalysis {
 
     private void run() {
         logger.trace("Start");
-        Wmm memoryModel = task.getMemoryModel();
-        Map<Relation, List<Definition>> dependents = new HashMap<>();
+        final Wmm memoryModel = task.getMemoryModel();
+        final Map<Relation, List<Definition>> dependents = new HashMap<>();
         for (Relation r : memoryModel.getRelations()) {
             for (Relation d : r.getDependencies()) {
                 dependents.computeIfAbsent(d, k -> new ArrayList<>()).add(r.getDefinition());
             }
         }
         // ------------------------------------------------
-        Initializer initializer = new Initializer();
-        Map<Relation, List<Delta>> qGlobal = new HashMap<>();
+        final Initializer initializer = new Initializer();
+        final Map<Relation, List<Delta>> qGlobal = new HashMap<>();
         for (Relation r : memoryModel.getRelations()) {
             Knowledge k = r.getDefinition().accept(initializer);
             if (!enableMustSets) {
@@ -238,7 +238,7 @@ public class RelationAnalysis {
             }
         }
         // ------------------------------------------------
-        Propagator propagator = new Propagator();
+        final Propagator propagator = new Propagator();
         for (Set<DependencyGraph<Relation>.Node> scc : DependencyGraph.from(memoryModel.getRelations()).getSCCs()) {
             logger.trace("Regular analysis for component {}", scc);
             Set<Relation> stratum = scc.stream().map(DependencyGraph.Node::getContent).collect(toSet());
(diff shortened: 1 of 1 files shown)


ThomasHaas:
> The removal in the `toAdd` delta seems to explicitly handle the two bullet points of the "Impact" section in #523.
> 
> Is this what that issue refers as "propagate decremental updates"? If so, didn't you say that solution would not work for recursive relations?

No, decremental updates are if you propagate a deletion, i.e., you try to reduce already computed results.
What this "fix" does is to reduce the incremental update before applying it so that we never generate too large results in the first place.

The problematic situation is with a relation `let c = a \ b`.
What happens right now is that when `a` gets incremented by `delta(a)`, the propagator will sent the update `delta(c) = delta(a) \ b`. This is fine if `b` was already fully computed, but that may not necessarily be the case, e.g., `b` might still be empty. If now `b` gets incremented by `delta(b)`, we would need to propagate a decremental update to `c` to delete some elements, but that does not work in general (so we don't do it).
However, the updates are processed along a stratification meaning that when we merge `delta(c)` into `c` (i.e., perform the update), we actually know that `b` is fully computed (cause `b` must be in a smaller stratum since we don't allow differences in recursion).
So we can fix the proper update to `delta(c) <- delta(c) \ b` before applying it.
In short: an update to `a` may propagate a too large update to `c` but we can later on correct this update when we actually merge it into `c`.

hernanponcedeleon:
"We don't allow difference in recursion" you mean "The right hand side of a difference cannot be recursive"?

Does this means that this solution should be fully sound as long as we don't allow those?

What about the comment saying that the push solution does not properly respect the stratification? Does it refers to this or something different?

ThomasHaas:
> "We don't allow difference in recursion" you mean "The right hand side of a difference cannot be recursive"?

Yes.

> Does this means that this solution should be fully sound as long as we don't allow those?

I believe so.

> What about the comment saying that the push solution does not properly respect the stratification? Does it refers to this or something different?

Our current "push-based" approach to propagation does not respect the stratification properly in the sense that it can push updates from a lower stratum to a higher stratum before all "intermediate" strata were processed (i.e., any stratum that is inbetween the lower and the higher one). This does not matter if there are no differences/negations involved because then any chaotic iteration strategy will give you the correct result.
A "pull-based" approach would not have this problem with differences because an update for the higher stratum gets computed by pulling data from all lower strata, which are all fully computed by that point in time.

The fix essentially makes sure that at least for a difference, we locally also "pull" changes. What I mean by this is that when `c` (in `let c = a \ b`) gets processed, we take all the updates that were pushed onto `c` (from `a`) and reduce them by pulling data from `b`.
An alternative fix would be to make sure that `b` is always before `a` in the stratification (I haven't tested this approach).

So the comment is meant to say 
```
Our whole propagation strategy is push-based and thus fails to respect stratification properly.
This does not matter for positive relations (stratification is irrelevant) but can cause wrong results on negative relations (i.e., differences), hence we introduce this "ugly" fix.
A pull-based propagation approach would not have needed this fix in the first place.
```

The more I think about this though, the less I'm sure about the "pull-based" approach being better in general.
It clearly wins for non-recursive relations, but it feels more complicated to do on recursive one's.
In fact, when I implemented propagation in the CAAT solver, I ended up having a hybrid approach that kinda does both. It actually handles differences pretty much the same way this fix does.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
```diff
@@ -409,6 +422,19 @@ public static final class Delta {
             may = maySet;
             must = mustSet;
         }
+
+        private static Delta combine(List<Delta> deltas) {
+            if (deltas.size() == 1) {
+                return deltas.get(0);
+            }
+            EventGraph mayDelta = new EventGraph();
+            EventGraph mustDelta = new EventGraph();
```
natgavrilenko, line 431: It will always create a new object, even if must sets are disabled and always point to the signleton. Probably, will not have any visible impact in practice.

natgavrilenko (approved):

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationAnalysis.java
ThomasHaas: Yes, to streamline the handling of the `enableMustSets` option, I changed the code to just always compute the must sets and then just filter them out in the core algorithm before propagating.
This way the option is handled at a central place using only a few lines of code.
I believe the performance impact is marginal.

PR #638, opened 2024-03-16
Allow unary relation operators (^-1, *, +, etc.) in recursive definitions

ThomasHaas:
This PR removes a restriction we imposed on recursive definitions that is not necessary anymore.


hernanponcedeleon:
Where was this restriction initially coming from?

ThomasHaas:
It was an arbitrary restriction in the original implementation of `Wmm` and its relation analysis. IIRC, there was a dedicated `getMaySetRecursive` methods and it was never implemented for the unary operators.

PR #639, opened 2024-03-16
Improved memory handling

ThomasHaas:
This PR improves handling all around memory, partially addressing the needs of #136:
- `MemoryObject` now has an`allocationSite` (an `Alloc` event) which is non-null for dynamic allocations.
- `MemoryObject.address` is gone (the `MemoryObject` itself already represents an address).
- `Alloc` is now a proper core event (also moved to the `core` package) and does not get replaced by `Local` anymore (the alias analysis has been adapted to support this).
- `ProgramEncoder.encodeMemory` now assigns actual addresses to memory objects. For now, it still generates a static memory layout (i.e., fixed addresses) but now it is also possible to encode complex memory layouts (in particular dynamic one's).
- `MemoryObject` does not automatically add initial values for static allocations. This is responsibility of the parser frontend: `ProgramBuilder` will now explicitly add init values for litmus tests.
- `MemoryAllocation` does not distinguish between litmus and non-litmus code anymore. It always only generates init events for explicitly initialized events. As above, responsibility of generating initial values is in the parser rather than the allocation pass.
- There is some minor cleanup
-- `MemoryObject.cVar` is now just `MemoryObject.name` to be more general.
-- `MemoryObject.isAtomic` is gone since it was unused.
-- `MemoryObject.index` is renamed to `MemoryObject.id` (it was never an index to anything)

As far as I see, the only step that is missing to support dynamically-sized allocations is extending `MemoryObject.size` to be a proper `Expression` (as discussed in #136). 
The biggest obstacle to this change is the current alias analysis. However, I think the new one in #616 might have an easier time supporting dynamic sizes.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Alloc.java
```diff
@@ -103,4 +121,16 @@ protected String defaultString() {
...
+    @Override
+    public <T> T accept(EventVisitor<T> visitor) {
+        return visitor.visitAlloc(this);
+    }
+
+    @Override
+    public BooleanFormula encodeExec(EncodingContext ctx) {
+        return ctx.getBooleanFormulaManager().and(
+                super.encodeExec(ctx),
+                ctx.equal(ctx.result(this), ctx.encodeExpressionAt(allocatedObject, this)));
```
hernanponcedeleon, line 134: Shouldn't you use `getAllocatedObject()` instead of `allocatedObject` to be sure this is not null?

ThomasHaas: I can do, but I guess both versions result in an exception either way.

PR #644, opened 2024-03-30
builtin workgroupsize & nest index

tonghaining:


dartagnan/src/test/resources/spirv/workgroupsize-unset.spv.dis
```diff
@@ -0,0 +1,29 @@
+; @ Output: forall (%size[0]==2 and %size[1]==3 and %size[2]==4)
+; SPIR-V
```
hernanponcedeleon, line 2: Isn't this missing the configuration? or is the intended behavior that the config is programmatically set by line 21 in the spirv code?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
```diff
@@ -381,7 +396,8 @@ private Thread createThreadFromFunction(Function function, int x, int y, int z)
         Memory memory = program.getMemory();
 
         // Create thread-local variables
-        BuiltIn builtIn = (BuiltIn) getDecoration(DecorationType.BUILT_IN);
+        BuiltIn builtIn = (BuiltIn) getDecoration(BUILT_IN);
+        builtIn.setDimension(threadGrid.get(0), threadGrid.get(1), 1);
```
hernanponcedeleon, line 400: Why is the last parameter always 1?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
```diff
@@ -591,5 +607,7 @@ public void setThreadGrid(List<Integer> threadGrid) {
             throw new ParsingException("Thread grid dimensions must be less than 128");
         }
         this.threadGrid = threadGrid;
+        BuiltIn builtIn = (BuiltIn) this.helperDecorations.getDecoration(BUILT_IN);
+        builtIn.setDimension(threadGrid.get(0), threadGrid.get(1), 1);
```
hernanponcedeleon, line 611: Why is the last parameter always 1?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
```diff
@@ -44,21 +59,43 @@ public void addDecoration(String id, String... params) {
...
             switch (decoration) {
                 case "LocalInvocationId", "GlobalInvocationId" -> {
-                    int size = TypeFactory.getInstance().getMemorySizeInBytes(type) / 3;
-                    // TODO: Use array element type instead of arch type
-                    memObj.setInitialValue(0, FACTORY.makeValue(x, TYPE_ARCH));
-                    memObj.setInitialValue(size, FACTORY.makeValue(y, TYPE_ARCH));
-                    memObj.setInitialValue(size * 2, FACTORY.makeValue(z, TYPE_ARCH));
+                    if (x < 0 || y < 0 || z < 0) {
+                        throw new ParsingException("Illegal BuiltIn hierarchy ('%d', '%d', '%d')", x, y, z);
+                    }
```
hernanponcedeleon, line 68: isn't zero also a wrong value for hierarchy?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
```diff
@@ -44,21 +59,43 @@ public void addDecoration(String id, String... params) {
...
+                    MemoryObject memObj = (MemoryObject) expr;
+                    IntegerType elementType = getArrayElementType(type, 3);
```
hernanponcedeleon, line 70: where does the 3 magic constant come from? is it because configs have 3 entries? if so, please document

dartagnan/src/test/resources/spirv/workgroupsize-unset.spv.dis
tonghaining: Based on the [document](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/WorkgroupSize.html) I think `WorkgroupSize` decoration shall only change the value of the `%gl_WorkGroupSize` (initialized with either `OpSpecConstantComposite` or `OpConstantComposite`) to the value of `WorkgroupSize` instead of setting value to `WorkgroupSize`. This test is to show when config is not given, the value of `%gl_WorkGroupSize` will not be updated.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
tonghaining: I understand our config as:

- `threadGrid.get(0)`: thread number of a subworkgroup
- `threadGrid.get(1)`: subworkgroup number of a workgroup
- `threadGrid.get(2)`: workgroup number of a shader

I think this is also consistent with we using them [here](https://github.com/hernanponcedeleon/Dat3M/blob/spirv-visitor/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java#L378). I used 1 here for the missing dimension of WorkgroupSize.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
tonghaining: Same as above.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
tonghaining: Yeah. I'll update it

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
tonghaining: No it was because the dimension of WorkgroupSize is always 3.

dartagnan/src/test/resources/spirv/workgroupsize-unset.spv.dis
hernanponcedeleon: I still don't get it. How does the builtin relates to the config? I would expect that we need to sync between the values this variable gets and the config. For example, the `MP.spv.dis` one has `@Config: 2, 1, 1` and 
```
          %5 = OpSpecConstant %uint 1
          %6 = OpSpecConstant %uint 1
          %7 = OpSpecConstant %uint 1
%gl_WorkGroupSize = OpSpecConstantComposite %v3uint %5 %6 %7
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorSpirvOutput.java
```diff
@@ -99,38 +102,36 @@ private Expression acceptAssertionValue(SpirvParser.AssertionValueContext ctx) {
...
-                result.addAll(unrollType(arrayType.getElementType()));
-            }
-            return result;
-        } else if (type instanceof AggregateType aggregateType) {
-            List<Type> result = new ArrayList<>();
-            for (Type fieldType : aggregateType.getDirectFields()) {
-                result.addAll(unrollType(fieldType));
+    private int getOffset(Type type, List<Integer> indexes) {
+        int offset = 0;
+        for (int index : indexes) {
```
natgavrilenko, line 120: This looks good for a valid input, but I think we should also validate the length and the value of indexes. 

Here is an example test for illegal indexes in a vector (the only correct one would be "%var[0][0]"). I would also add a test for a structure and for using indexes with for a scalar type.

@Test
    public void testIllegalVectorIndex() {
        // TODO: The first case does throw an exception, 
        //  but it would be nice to make it more informative, 
        //  maybe something like "Illegal assertion for variable '%myvar', index out of bounds"
        doTestIllegalVectorIndex("%var[0][0][0]==0",
                "Unexpected index for type bv64");
        doTestIllegalVectorIndex("%var[0]==0", "Some informative error");
        doTestIllegalVectorIndex("%var[1][0]==0", "Some informative error");
        doTestIllegalVectorIndex("%var[0][1]==0", "Some informative error");
    }

    private void doTestIllegalVectorIndex(String ast, String error) {
        String input = """
            ; @Input: %var = {{0}}
            ; @Output: forall (<<ast>>)
                           OpCapability Shader
                    %ext = OpExtInstImport "GLSL.std.450"
                           OpMemoryModel Logical GLSL450
                           OpEntryPoint GLCompute %main "main"
                           OpSource GLSL 450
                   %void = OpTypeVoid
                   %func = OpTypeFunction %void
                 %uint64 = OpTypeInt 64 0
                %v1_type = OpTypeVector %uint64 1
                %v2_type = OpTypeVector %v1_type 1
               %ptr_type = OpTypePointer Uniform %v2_type
                    %var = OpVariable %ptr_type Uniform
                   %main = OpFunction %void None %func
                  %label = OpLabel
                           OpReturn
                           OpFunctionEnd
                 """.replace("<<ast>>", ast);

        try {
            // when
            localParse(input);
            fail("Should throw exception");
        } catch (ParsingException e) {
            // then
            assertEquals(error, e.getMessage());
        }
    }

natgavrilenko (commented):
The output part looks good to me, except that I would add validation for illegal number of indexes and illegal index values.
For buildins, I will review and test on Monday.

dartagnan/src/test/resources/spirv/workgroupsize-unset.spv.dis
tonghaining: So in this example, the value of `%gl_WorkGroupSize` is [1,1,1], and the `threadGrid` should be [1,1,1] or [2,1,1]?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorSpirvOutput.java
tonghaining: Added, thanks!

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
natgavrilenko: I see that you set BuildIn dimensions here so that we can distinguish the situation when config is not set explicitly and we use default (1,1,1). 

I would suggest to keep it simple and use whatever config we have after parsing headers, either default (1,1,1) or whatever is explicitly set by a user. Otherwise, in case of missing annotations, we can get weird situations where thread groupId doesn't agree with workgroup size.

And if dimensions are constant, then we can set them as final attributes in the BuiltIn constructor.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
```diff
@@ -294,10 +296,23 @@ public Expression addExpression(String name, Expression value) {
         if (types.containsKey(name) || expressions.containsKey(name)) {
             throw new ParsingException("Duplicated definition '%s'", name);
         }
+        if (preThreadDecoration.containsKey(name)) {
```
natgavrilenko, line 299: I would probably create another expression transformer for global variables (or alter the existing one) to keep all decorations at the same place, but your solution should work fine too. Let keep it this way for now, we can refactor it later if we have time.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
```diff
@@ -15,13 +17,15 @@
...
 
     private static final ExpressionFactory FACTORY = ExpressionFactory.getInstance();
-    private static final IntegerType TYPE_ARCH = TypeFactory.getInstance().getArchType();
 
     private final Map<String, List<String>> mapping = new HashMap<>();
 
     private int x = -1;
     private int y = -1;
     private int z = -1;
+    private int wgDimensionX = -1;
```
natgavrilenko, line 26: Maybe rename it to something more generic, e.g. dimensionX? The BuiltIn class can also use these values later for other decorations.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/ProgramBuilderSpv.java
tonghaining: Okay

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
tonghaining: Renamed

PR #648, opened 2024-04-14
OpBitwise

tonghaining:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBits.java
```diff
@@ -42,50 +43,96 @@ public Event visitOpShiftRightArithmetic(SpirvParser.OpShiftRightArithmeticConte
...
+            return createEventsForArrayType(id, typeCtx.getText(), op1, op2, op);
         }
         throw new ParsingException("Illegal result type for '%s'", id);
     }
 
-    private Expression getOperandInteger(String id, String opId) {
-        Expression op = builder.getExpression(opId);
-        if (op.getType() instanceof IntegerType) {
-            return op;
+    private Event createEventsForArrayType(String id, String type, Expression op1, Expression op2, IntBinaryOp op) {
```
natgavrilenko, line 95: I don't think that we can blindly have an array type in a register, at least not without further changes. Tried a small test below, and we get an UnsupportedOperationException in encoding, which makes sense. 

We will definitely need arrays later, but I suggest to split this commit and postpone the array part until after the deadline.

; @Config: 1, 1, 1
; SPIR-V
; Version: 1.0
; Schema: 0
               OpCapability Shader
        %ext = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpSource GLSL 450
       %void = OpTypeVoid
       %func = OpTypeFunction %void
       %uint = OpTypeInt 32 0
     %v3uint = OpTypeVector %uint 3
    %v24uint = OpTypeVector %uint 24
   %ptr_uint = OpTypePointer Input %uint
 %ptr_v3uint = OpTypePointer Input %v3uint
%ptr_v24uint = OpTypePointer Output %v24uint
         %c0 = OpConstant %uint 0
         %c1 = OpConstant %uint 1
         %c2 = OpConstant %uint 2
        %arr = OpConstantComposite %v3uint %c0 %c1 %c2
        %out = OpVariable %ptr_v24uint Output
       %main = OpFunction %void None %func
      %label = OpLabel
  %ptr_out_x = OpAccessChain %ptr_uint %out %c0
       %test = OpBitwiseAnd %v3uint %arr %arr
               OpStore %ptr_out_x %test
               OpReturn
               OpFunctionEnd

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBitsTest.java
```diff
@@ -0,0 +1,204 @@
...
+import org.junit.Test;
+
+import java.util.List;
+
+import static com.dat3m.dartagnan.expression.integers.IntBinaryOp.*;
+import static com.dat3m.dartagnan.expression.integers.IntUnaryOp.MINUS;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class VisitorOpsBitsTest {
```
natgavrilenko, line 21: Nice tests!

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBits.java
ThomasHaas: Yes, we have no encoding for arrays (nor structs). IIRC, there is a pass that can eliminate struct-typed registers, but I'm not sure if it can also handle array types.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBits.java
```diff
@@ -42,50 +43,96 @@ public Event visitOpShiftRightArithmetic(SpirvParser.OpShiftRightArithmeticConte
...
+        checkTypeAlignment(id, idType, op1);
+        checkTypeAlignment(id, idType, op2);
+        Register register = builder.addRegister(id, typeId);
+        Local event = EventFactory.newLocal(register, EXPR_FACTORY.makeBinary(op1, op, op2));
+        return builder.addEvent(event);
+    }
+
+    private void checkTypeAlignment(String id, Type idType, Expression op) {
+        Type opType = op.getType();
+        if (idType != opType) {
```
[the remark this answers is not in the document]
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBits.java
tonghaining: I just find we have 3 tests in benchmark will need array supporting like in [here](https://github.com/hernanponcedeleon/Dat3M/blob/spirv-visitor/dartagnan/src/test/resources/parsers/program/spirv/global_barrier.spv.dis#L306). Where is the struct-typed registers handled? Maybe we can try that?

natgavrilenko: All these tests will also need OpCompositeExtract, which we don't have yet either. It shouldn't be a very big deal to add support them all, but I expect that testing all corner cases can take a few days.

I would finish the "easy" part with bitwise operation for scalars + atomic min/max + OpSelect. But of course this is up to you.

ThomasHaas: > I just find we have 3 tests in benchmark will need array supporting like in [here](https://github.com/hernanponcedeleon/Dat3M/blob/spirv-visitor/dartagnan/src/test/resources/parsers/program/spirv/global_barrier.spv.dis#L306). Where is the struct-typed registers handled? Maybe we can try that?

The pass is called `RegisterDecomposition` and it is a pretty trivial one: it splits locals of the form `{ T1, T2 } r = (a, b)` into `T1 r1 = a; T2 r2 = b`, and replaces every "extract" performed on the struct into direct register accesses.
It works only as long as the struct is fully local (i.e., it is never stored to memory).
I believe it should also work on array-typed registers out-of-the-box.
We only required it for a very special case because LLVM already removes almost all aggregates for us, so it is likely faulty in a more general setting.

In case you need to load/store arrays from/into memory, I think there is no easy solution.
If you have litmus-style assertions over the arrays, I think you also need to update the assertions.
That being said, I agree with @natgavrilenko that you should probably focus on the things that are already sufficiently supported in the backend and finish the "easy" parts first.

tonghaining: Okay I'll only keep the scalar part then, thanks!

PR #650, opened 2024-04-16
Make parseMetadataAttachment more robust in the absence of some metadata

hernanponcedeleon:
The `parseMetadataAttachment` was throwing an exception if some metadata was missing. This was causing problems with vsyncer where some llvm pass was removing (or not generating, not too sure) some metadata. Anyway, our code should not rely on the assumption that this info is available.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
@@ -320,9 +320,13 @@ public class VisitorLlvm extends LLVMIRBaseVisitor<Expression> {
                     scope = (SpecialMdTupleNode) metadataSymbolTable.get(scope.<MdReference>getField("scope").orElseThrow().mdName());
                 }
                 assert scope.nodeType() == SpecialMdTupleNode.Type.DIFile;
+                // https://llvm.org/docs/LangRef.html#difile suggests that "file" and "directory" 
+                // are mandatory fields and thus the ElseThrow
                 final String filename = scope.<MdGenericValue<String>>getField("filename").orElseThrow().value();
                 final String directory = scope.<MdGenericValue<String>>getField("directory").orElseThrow().value();
-                final int lineNumber = diLocationNode.<MdGenericValue<BigInteger>>getField("line").orElseThrow().value().intValue();
+                // Field "line" is optional. When missing we assume value 0
+                final int lineNumber = diLocationNode.<MdGenericValue<BigInteger>>getField("line")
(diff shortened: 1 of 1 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -309,20 +309,28 @@ private List<Metadata> parseMetadataAttachment(List<MetadataAttachmentContext> m
...
-                final int lineNumber = diLocationNode.<MdGenericValue<BigInteger>>getField("line").orElseThrow().value().intValue();
+                final String filename = scope.getField("filename").isPresent()
+                        ? scope.<MdGenericValue<String>>getField("filename").get().value()
+                        : "";
+                final String directory = scope.getField("directory").isPresent()
+                        ? scope.<MdGenericValue<String>>getField("directory").get().value()
+                        : "";
+                final int lineNumber = scope.getField("line").isPresent()
+                        ? diLocationNode.<MdGenericValue<BigInteger>>getField("line").get().value().intValue()
+                        : 0; // Assuming line 0 when value not present
```
ThomasHaas, line 333: You can just use `map.(n -> n.value).orElse(defaultValue)` (*) in place of the original `orElseThrow()`.
What precisely is happening though? Do we get just some of the fields rather than all?
I thought those are special metadata nodes where all fields are required, otherwise the IR is ill-formed (maybe I'm misremembering though).

(*) written from memory, so might need some adaption :P):

db7: Line is optional, see LLVM parser: https://github.com/llvm/llvm-project/blob/main/llvm/lib/AsmParser/LLParser.cpp#L5049

```c++
/// parseDILocationFields:
///   ::= !DILocation(line: 43, column: 8, scope: !5, inlinedAt: !6,
///   isImplicitCode: true)
bool LLParser::parseDILocation(MDNode *&Result, bool IsDistinct) {
#define VISIT_MD_FIELDS(OPTIONAL, REQUIRED)                                    \
  OPTIONAL(line, LineField, );                                                 \
  OPTIONAL(column, ColumnField, );                                             \
  REQUIRED(scope, MDField, (/* AllowNull */ false));                           \
  OPTIONAL(inlinedAt, MDField, );                                              \
  OPTIONAL(isImplicitCode, MDBoolField, (false));
  PARSE_MD_FIELDS();
#undef VISIT_MD_FIELDS
```

ThomasHaas (approved):
I probably would have used `-1` as default value for missing line numbers, but `0` should also be fine.
Anyways, LGTM.

Issue #647, opened 2024-04-13
Unexpected liveness violation


db7:
The attached file 
[test.ll.txt](https://github.com/hernanponcedeleon/Dat3M/files/14965701/test.ll.txt)
gives unexpected results with dat3m.

> Note I had to rename the file as `test.ll.txt` to make github accept it, you might want to remove the `.txt` suffix.

I'm using this command to run the file:
```
java -jar /usr/share/dat3m/dartagnan/target/dartagnan.jar \
  test.ll \
  --property=program_spec,liveness \
  --modeling.threadCreateAlwaysSucceeds=true \
  --encoding.wmm.idl2sat=true \
  --solver=yices2 \
  --target=imm \
  /usr/share/dat3m/cat/imm.cat \
  --bound=5
```

This is the [commit](https://github.com/hernanponcedeleon/Dat3M/commit/b57432899) I'm using.

The output of the command above tells me:

```
[13.04.2024] 07:55:22 [INFO] RefinementSolver.runInternal - Specification violation found.
[13.04.2024] 07:55:22 [INFO] RefinementSolver.runInternal - Summary
 ======== Summary ========
Number of iterations: 3
Total native solving time: 1.632 secs
   -- Bound check time: 0.0 secs
Total CAAT solving time: 0.169 secs
   -- Model extraction time: 0.29 secs
   -- Population time: 0.106 secs
   -- Consistency check time: 0.1 secs
   -- Reason computation time: 0.25 secs
   -- Refining time: 0.25 secs
   -- #Computed core reasons: 28
   -- #Computed core reduced reasons: 26
   -- Min model size (#events): 44
   -- Average model size (#events): 48
   -- Max model size (#events): 51

[13.04.2024] 07:55:22 [INFO] RefinementSolver.runInternal - Verification finished with result FAIL
============ Liveness violation found ============
	E923:	@check-spsc.c#40
=================================================
FAIL
Total verification time: 11.627 secs
```

Now, also interesting is that if I use bound == 6, the output of dat3m changes, giving me a program spec violation.

```
[13.04.2024] 08:00:30 [INFO] RefinementSolver.runInternal - Specification violation found.
[13.04.2024] 08:00:30 [INFO] RefinementSolver.runInternal - Summary
 ======== Summary ========
Number of iterations: 17
Total native solving time: 4.577 secs
   -- Bound check time: 0.0 secs
Total CAAT solving time: 0.370 secs
   -- Model extraction time: 0.108 secs
[...]

PR #654, opened 2024-04-18
Improved reporting and detection of CAAT non-termination

ThomasHaas:
This PR adds a check to prevent CAAT from not terminating.
It further tries to figure out the source of non-termination and report it to the user.
Also, the `Dependency` analysis now reports accesses to uninitialized registers with source information (this is related to #647)

I might also add source location reporting for reachable bounds, so the user knows which loops are not unrolled.
Are there any other places that would benefit from better reporting? If so, I would add them.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
@@ -410,13 +410,10 @@ public class RefinementSolver extends ModelChecker {
     }
 
     private boolean checkProgress(List<RefinementIteration> trace) {
-        if (trace.size() < 2) {
+        if (trace.size() < 2 || trace.getLast().isConclusive()) {
             return true;
         }
         final RefinementIteration last = trace.get(trace.size() - 1);
-        if (last.isConclusive()) {
-            return true;
-        }
         final RefinementIteration prev = trace.get(trace.size() - 2);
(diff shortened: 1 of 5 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -301,6 +306,54 @@ private void runInternal(SolverContext ctx, ProverEnvironment prover, Verificati
...
+
+        final DNF<CoreLiteral> lastReasons = propertyTrace.getFinalIteration().inconsistencyReasons;
+        for (Conjunction<CoreLiteral> reason : lastReasons.getCubes()) {
+            for (CoreLiteral literal : reason.getLiterals()) {
+                if (literal instanceof RelLiteral relLit && doesViolateAliasingRules(relLit, alias)) {
+                    final Event e1 = relLit.getSource();
+                    final Event e2 = relLit.getTarget();
+
+                    final StringBuilder builder = new StringBuilder();
+                    builder.append("Found aliasing problem between:\n");
```
hernanponcedeleon, line 327: Maybe you can specify that the solver found an execution where both alias but the alias analysis says this is impossible.

ThomasHaas: I can update the error to `Found unexpected aliasing between: `, which I think is perfectly understandable to us.

hernanponcedeleon (approved):

PR #656, opened 2024-04-19
Better symmetry reasoning in CAAT

ThomasHaas:
This PR improves symmetry reasoning in CAAT for benchmarks with large symmetries.

EDIT: To streamline some code and avoid special corner cases, I also added a `NormalizeFenceRelations` pass that rewrites `fencerel(F)` to `po;[F];po`. CAAT no longer supports the standard `fencerel`.

PR #660, opened 2024-04-23
CAAT dataraces/cat_spec

ThomasHaas:
This PR adds support for checking `cat_spec` with CAAT, in particular, for checking data races.
Since data races are negatively formulated conditions, cutting is necessary.
Typically, this cutting approach would cut off so much of the memory model that most of it needs to get encoded eagerly again.
The trick this PR adds is a memory model rewriting step that makes sure that cutting affects only the parts of the model that are "truly negative", i.e., those that appear under an odd number of negations.
For example, in `~empty(dr),  dr = conf \ hb`, the `hb` is actually "positive" and the rewriting makes sure it does not get cut.
More generally, the rewriting is irrespective of `cat_spec` and data races, but applies to all memory models that have nested negations. It's just that data races are a canonical example of nested negations where the complex relation (`hb`) is positive.

Aside from this, the PR improves CAAT core reasoning and fixes a non-termination issue (for highly symmetric benchmarks) recently introduced in #656 .


dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -537,14 +543,107 @@ private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
...
+            final Relation r = diff.getDefinedRelation();
+            final Relation a = diff.getMinuend();
+            final Relation b = diff.getSubtrahend();
+
+            // (1) Create b' (if not already existing)
+            final Relation bPrime;
+            if (replacements.containsKey(b)) {
+                bPrime = replacements.get(b);
+            } else {
+                final String bPrimeName = b.getName().map(n -> n + "#POS").orElse("__POS" + counter++);
```
hernanponcedeleon, line 636: Why the different `#POS` and `__POS`?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -180,10 +180,15 @@ private void runInternal(SolverContext ctx, ProverEnvironment prover, Verificati
 
         // ------------------------ Preprocessing / Analysis ------------------------
 
-        removeFlaggedAxioms(memoryModel);
+        // TODO: This is a reasonable transformation for all solvers, however,
+        //  our current processing pipelines (WmmProcessor/ProgramProcessor) are unaware of the property
+        //  so we cannot perform property-aware transformation in those pipelines right now.
+        removeFlaggedAxiomsIfNotNeeded(task);
```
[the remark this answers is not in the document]
ThomasHaas: It means that this transformation of the memory model (deleting all flagged axioms) could be used for all solvers, including the eager one. So a reasonable choice would be to have a `WmmProcessor` that does this transformation generally, however, this `WmmProcessor` would need to be aware of the specification we check for (if `cat_spec` don't touch the memory model, otherwise delete flagged axioms).
And this "specification-awareness" is something that we do not have right now.
It would be easy if the `WmmProcessor` took as input not the `Wmm` but the whole `VerificationTask`.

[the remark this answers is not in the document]
hernanponcedeleon: Ok, then please use "both our methods, i.e. eager and lazy" ... Solver is a term quote overloaded in our code base

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -537,14 +543,107 @@ private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
...
+        if (!task.getProperty().contains(Property.CAT_SPEC)) {
+            List.copyOf(task.getMemoryModel().getAxioms()).stream()
+                    .filter(Axiom::isFlagged)
+                    .forEach(task.getMemoryModel()::removeConstraint);
+        }
+    }
+
+    /*
+    The constraints/relations of the Wmm can be categorised into positive and negative,
+    depending on whether the number of negations applied to the constraint/relation is even (=positive)
```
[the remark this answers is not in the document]
ThomasHaas: Things like `[~X]` are technically also negations but our current `Wmm` treats `~X` as base relation anyway so in a sense it is already "cut" off correctly.
The other two cases are as you said: negated axioms and relation differences and that is why I only consider them in the code right now.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -537,14 +543,107 @@ private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
...
+        wmm.getAxioms().forEach(ax -> ax.accept(collector));
+        return new PolaritySeparator(positives, negatives);
+    }
+
+    /*
+        This pass rewrites the memory model so that no positive constraints appear on the rhs-side of a difference.
+        Suppose "r = a \ b" where r and a are negative and b is positive.
+        For example, this is the case if there is a constraint "~empty(r)"
+        We rewrite the model to
+            let b`= free()
```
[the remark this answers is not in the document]
ThomasHaas: They relate in the sense that conceptually `b' == b`. It's just that we cannot enforce this equality eagerly because that would mean we need to encode `b` which is what we want to avoid. Instead, we learn the connection between those two variables lazily in CAAT (we actually do NOT learn the precise equality though).

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -537,14 +543,107 @@ private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
...
+                return (polarity ? positives : negatives).add(c);
+            }
+        };
+
+        wmm.getAxioms().forEach(ax -> ax.accept(collector));
+        return new PolaritySeparator(positives, negatives);
+    }
+
+    /*
+        This pass rewrites the memory model so that no positive constraints appear on the rhs-side of a difference.
```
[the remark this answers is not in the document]
ThomasHaas: This is always possible. I can write the short proof of why this works in the comments.
The core idea is similar to how one can encode any fixed point instead of the largest one for recursive equations because the best choice the solver can make is chose the least fixed point. In a similar vein, we allow the solver to choose `b' != b`, but this choice is suboptimal and he would be better off choosing `b' = b`.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -537,14 +543,107 @@ private static void addBiases(Wmm memoryModel, EnumSet<Baseline> biases) {
...
+        final PolaritySeparator separator = computePolaritySeparator(wmm);
+        final Set<Difference> negDiff = separator.negatives().stream()
+                .filter(Difference.class::isInstance).map(Difference.class::cast)
+                .filter(diff -> !separator.negatives().contains(diff.getSubtrahend().getDefinition()))
+                .collect(Collectors.toSet());
+
+        final Map<Relation, Relation> replacements = new HashMap<>();
+        int counter = 0;
+        for (Difference diff : negDiff) {
+            // r = a \ b where r and a are negative and b is positive.
```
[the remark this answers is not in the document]
ThomasHaas: If `r` is also positive, then it follows that `b` must also be negative and then the transformation is not applied.
The definition of `negDiff` find differences where `b` is non-negative and so `r/a` must be non-positive.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: Oh, a mistake.

ThomasHaas: Actually, this was intended. I like to put the prefix `__` for all internally generated stuff and use `#` for annotations/decorations of existing stuff.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -605,13 +608,54 @@ private boolean process(Constraint c) {
...
             let r = a \ b' // b' gets cut, but it is effectively a base relation so this causes no large encoding.
-            empty (r & b) // b is now in a positive position and thus not need to get cut
+            empty (r & b) // b is now in a positive position and thus not need to get cut.
+
+        We claim that this model is equisatisfiable to the original memory model.
+        To understand this intuitively, it is helpful so think of the fresh b' as the original b but with "some slack".
+        We will argue that the best choice for b' is precisely b.
+        Any larger choice for b' will make r smaller, but since r is non-positive this reduces the chances
+        of satisfying negated axioms (e.g., if "~empty(r)" is true for some small r, then it is also true for all larger r).
+        On the other hand, if b' is chosen too small, then r gets so large that it violates "empty(r & b)".
```
hernanponcedeleon, line 624: I had completely missed the fact that you also transformed the axiom. Now it makes much more sense

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -605,13 +608,54 @@ private boolean process(Constraint c) {
...
+        We prove this formally:
+
+        "NEW => ORIGINAL":
+        Suppose M' is a consistent execution of the new memory model.
+        Then from "let r = a \ b'" it follows that M'(r) \subseteq M'(a).
```
hernanponcedeleon, line 629: Is `M'(r)` the restriction of `r` to the pairs of executed events in `M'`?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -605,13 +608,54 @@ private boolean process(Constraint c) {
...
+        Furthermore, from "empty (r & b)" it follows that M'(r) \subseteq ~M'(b).
+        Combining both, we get M'(r) \subseteq M'(a) & ~M'(b) = M'(a) \ M'(b)
```
hernanponcedeleon, line 631: I don't follow how you derive `~M'(b) = M'(a) \ M'(b)`

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -605,13 +608,54 @@ private boolean process(Constraint c) {
...
+        Now consider an execution M of the original memory model that matches with M' on the base relations
+        (modulo b' which does not exist in the original model).
+        Then M(a) = M'(a) and M(b) = M'(b), so that M'(r) \subseteq M'(a) \ M'(b) = M(a) \ M(b) = M(r).
```
hernanponcedeleon, line 634: "Then M(a) = M'(a) and M(b) = M'(b)" -> why? I don't think you are assuming `a` and `b` to be base

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -605,13 +608,54 @@ private boolean process(Constraint c) {
...
+        of satisfying negated axioms (e.g., if "~empty(r)" is true for some small r, then it is also true for all larger r).
+        On the other hand, if b' is chosen too small, then r gets so large that it violates "empty(r & b)".
+        We prove this formally:
+
+        "NEW => ORIGINAL":
+        Suppose M' is a consistent execution of the new memory model.
+        Then from "let r = a \ b'" it follows that M'(r) \subseteq M'(a).
+        Furthermore, from "empty (r & b)" it follows that M'(r) \subseteq ~M'(b).
+        Combining both, we get M'(r) \subseteq M'(a) & ~M'(b) = M'(a) \ M'(b)
+        Now consider an execution M of the original memory model that matches with M' on the base relations
```
hernanponcedeleon, line 632: Can you call it "anarchic execution" or "candidate execution"? It took me a while to realize you were not meaning "consistent"

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -605,13 +608,54 @@ private boolean process(Constraint c) {
...
+        i.e., the difference between "~empty(r)" and "flag ~empty(r)" is irrelevant for the argumentation.
+
+        NOTE 2: This rewriting is always possible.
+
+        NOTE 3: There is a minor caveat in the new memory model.
+        If an execution under the original memory model violates, e.g., "flag ~empty(r)" with multiple edges {e1, ..., ek},
+        then for the same execution the new memory model could report only a subset of those violations.
+        The reason is that the new memory model has a some slack in the choice of b':
+        it will choose b' such "flag ~empty(r)" holds (i.e., we have a violation), but it will not guarantee to
+        make the choice in such a way that it maximizes the number of violations.
```
hernanponcedeleon, line 656: The SMT solver does not gives us this guarantee either, right?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: The original axioms are still there. I only add a new axiom.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: No, its the interpretation of `r` in the given execution `M'`. `r` can take many values depending on the concrete execution (i.e., depending on the values taken by the base relations like `rf/co`). Without an execution `M'`, `r` is just a syntactic element / a variable.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: You should read it as `M'(a) & ~M'(b) = M'(a) \ M'(b)`. The `&` there is to be understood as an intersection.
Maybe I will put some parenthesis to make this clear.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: This is because if you know all base relations, then you also know all derived relations.
In particular, if `M` and `M'` agree on the base relations they must also agree on derived relations.
The only exception is for derived relations that depend on `b'` because this one does not appear in `M` and so they "disagree" there. However, neither `a` nor `b` depend on `b'` and so both executions agree on those.
Put differently, both executions must agree up to the point where I do the rewriting and only from there on there can be a difference.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
ThomasHaas: No, the text is meant to be seen from the SMT solver perspective. The SMT solver wants to find a satisfying assigment and so it will make sure that `~empty(r)` holds, but it will not care "how non-empty" `r` is.
That being said, I think the default search strategy of solvers prefer `false` over `true`, meaning the solver will automatically prefer to choose a small `b'` and thus a large `r`, maximizing the number of violations.

Btw., for a similar reason, if you have two flagged axioms, the solver might report the violation of only one of them even if both are violated in the original memory model.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
hernanponcedeleon: Is this true: "the flagged axiom is still there, it is just that caat ignores it"

ThomasHaas: Since the flagged axiom is negative, it also gets cut by the cutting algorithm. So CAAT doesn't know that it even exists.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
hernanponcedeleon: > Btw., for a similar reason, if you have two flagged axioms, the solver might report the violation of only one of them even if both are violated in the original memory model.

This is what I meant by "The SMT solver does not gives us this guarantee either"

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
hernanponcedeleon: Is it the case that if the relation was named, you add `#POS` and if the relation was unnamed you add `__POS`?

ThomasHaas: Yes. In the latter case I also add a number to avoid duplicated names.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -622,23 +623,32 @@ private boolean process(Constraint c) {
...
-        Combining both, we get M'(r) \subseteq M'(a) & ~M'(b) = M'(a) \ M'(b)
-        Now consider an execution M of the original memory model that matches with M' on the base relations
+        Suppose M' is a consistent execution under the new memory model.
+        Then from "let r = a \ b'" it follows that
+            M'(r) \subseteq M'(a).
+        Furthermore, from "empty (r & b)" it follows that
+            M'(r) \subseteq ~M'(b).
+        Combining both, we get
+            M'(r) \subseteq (M'(a) & ~M'(b)) = (M'(a) \ M'(b)).
+        Now consider an execution M under the original memory model that matches with M' on the base relations
```
hernanponcedeleon, line 639: "execution" -> "candidate execution"

ThomasHaas: Why do you prefer the term candidate execution? I think we never use this anywhere. I added an introductory sentence explaining what an execution of a memory model is in-line with our usage of that term in the two OOPSLA papers.

hernanponcedeleon: Most people would parse "execution" as "consistent executions", thus is why I normally prefer to use "candidate execution" or "behavior" so people don't understand it wrong.

This is just a pedantic comment from my side. As you said, the comment explains how executions should be interpret.

PR #662, opened 2024-04-25
Implement option to get dartagnan version

hernanponcedeleon:
- Implement option to get dartagnan version
- Update version in pom files (missing from last release)
- Rename assume method to eager in svcomp script

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
@@ -27,12 +27,12 @@ public class GitInfo {
     }
 
     public static void logGitInfo() {
-        logger.info("Git branch: " + properties.getProperty("git.branch"));
-        logger.info("Git commit ID: " + properties.getProperty("git.commit.id"));
+        logger.info("Git branch: " + properties.getProperty("git.branch", "unknown"));
+        logger.info("Git commit ID: " + properties.getProperty("git.commit.id", "unknown"));
     }
 
     public static String getGitId() {
-        return properties.getProperty("git.commit.id");
+        return properties.getProperty("git.commit.id", "unknown");
     }
 
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
@@ -11,31 +11,28 @@ import com.dat3m.dartagnan.Dartagnan;
 
 public class GitInfo {
 
-    private static final Logger logger = LogManager.getLogger(Dartagnan.class);
+    private static final Logger logger = LogManager.getLogger(GitInfo.class);
 
     static Properties properties = new Properties();
 
-    public static void CreateGitInfo() throws IOException {
+    public static void initGitInfo() throws IOException {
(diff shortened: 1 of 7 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
```diff
@@ -11,19 +11,31 @@
...
 public class GitInfo {
 
-	private static final Logger logger = LogManager.getLogger(Dartagnan.class);  
+    private static final Logger logger = LogManager.getLogger(Dartagnan.class);
 
-	static Properties properties = new Properties();
-	
-	public static void CreateGitInfo() throws IOException {
-       try (InputStream is = Dartagnan.class.getClassLoader()
+    static Properties properties = new Properties();
```
ThomasHaas, line 16: Why do we have this static field if it gets effectively reassigned/overwritten in all operations?
Can't we load this once and then have the `CreateGitInfo/getGidId` (btw. the first method's name should be lower case) read out the data?

Alternatively, you can remove this field and provide a `private static Properties loadGitProperties()` method that loads the properties object (possibly `null` on failure?). Then we can later decide if this helper function should load the properties only once and cache the result.

hernanponcedeleon: > Why do we have this static field if it gets effectively reassigned/overwritten in all operations? 

The method is static otherwise the compiler complains about `Cannot make a static reference to the non-static field properties`. Also, I don't think loading information (which is what `properties.load` does) counts as reassigning/overwritting

ThomasHaas: No, I mean why don't you just have a local `Properties properties = new Properties()` in each function to avoid this "global" variable?

hernanponcedeleon: If we move the variable to each method, we need to repeat the loading each time. Now the init() method takes care of the loading and the other methods just query the properties they need.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
```diff
@@ -11,31 +11,28 @@
...
-            properties.load(is);
-            return properties.getProperty("git.commit.id");
-        }
+    public static void logGitInfo() {
+        logger.info("Git branch: " + properties.getProperty("git.branch"));
+        logger.info("Git commit ID: " + properties.getProperty("git.commit.id"));
+    }
+
+    public static String getGitId() {
+        return properties.getProperty("git.commit.id");
```
ThomasHaas, line 35: I think you should do a `getProperty(..., defaultValue)` (maybe "unknown" as default?) because the loading of properties may fail in which the normal `getProperty` would return NULL.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/GitInfo.java
ThomasHaas: Yeah, that's fine. Either do it once globally or each time locally; those were my two suggestions :).

ThomasHaas (approved):
LGTM

PR #663, opened 2024-04-29
[DRAFT] Add transformation pass to make xchg side effect free when the stored value does not change

hernanponcedeleon:
This is a prototype implementation based on an idea from @JonasOberhauser. However, we are still not convinced (thus the draft status) if the pass is sound (probably not for arbitrary WMM). 

TODO
- [ ] Add the new litmus tests to the CI
- [ ] Add taslock


ThomasHaas:
I think this transformation only makes sense if we also extend liveness handling to special-handle the dummy store.
If it is just about detecting side-effect-free iterations, a more canonical approach would be to change the instrumentation produced by `DynamicSpinLoopDetection` to have
```
r = load(x);
store(x, v); // Must be RMW with the load
sideEffect = sideEffect | (r != v);
```
This would avoid the dummy altogether. A benefit (or a disadvantage) would be that we can detect this pattern without actually relying on special `Xchg` events: we could do this even for manually placed LL/SC implementations of Xchg, or Xchg's coming from C-litmus or SPIR-V (maybe we don't want to though?).

ThomasHaas:
Btw. I think the dummy store must be wrong in a case like this:
```
T1:
x = 1;
xchg(&x, 1); // Only produces dummy
store(&y, 1, REL);

T2:
if (load(&y, ACQ) == 1) {
    assert (x == 0); // Should fail, but it won't I think
}
```
The problem is that T2 must observe the dummy store because it is the most recent one. Yet, it cannot observe it, so we do not get any executions at all! I think the dummy store must also go to a dummy address to avoid it participating in `co` and `fr` and thus hiding other stores.

hernanponcedeleon:
At least for IMM I get FAIL with both the pass enabled/disabled for the program below (not sure which mo you intended for xchg so I just used SC)
```
#include <stdlib.h>
#include <pthread.h>
#include <assert.h>
#include <stdatomic.h>

atomic_int x, y;

void *thread_1(void *unused)
{
    x = 1;
    atomic_exchange(&x, 1); // Only produces dummy
    atomic_store_explicit(&y, 1, memory_order_release);
    return NULL;
}

void *thread_2(void *unused)
{
    if (atomic_load_explicit(&y, memory_order_acquire) == 1) {
        assert (x == 0); // Should fail, but it won't I think
    }
    return NULL;
}

int main()
{
    pthread_t t1, t2;

    pthread_join(t1, NULL);
    pthread_join(t2, NULL);

    return 0;
}

```

ThomasHaas:
The example works because your change to `RelationAnalysis` was not as you intended: you disable rf-edges whose read event has tag `NO_READ` instead of whose write event has that tag. Because of that, the dummy store was still readable.
If you make the dummy unreadable, the example becomes PASS, which is wrong.

hernanponcedeleon:
> The example works because your change to `RelationAnalysis` was not as you intended: you disable rf-edges whose read event has tag `NO_READ` instead of whose write event has that tag. Because of that, the dummy store was still readable. If you make the dummy unreadable, the example becomes PASS, which is wrong.

I just fix the `RelationAnalysis` and I still get the same result with/without the pass

ThomasHaas:
Did you rerun `mvn clean install`? Cause I can reproduce my claim. If you get fail, just print the witness graph and see that it reads from the dummy (cause it cannot read from anywhere else).

JonasOberhauser:
> Btw. I think the dummy store must be wrong in a case like this:
> 
> ```
> T1:
> x = 1;
> xchg(&x, 1); // Only produces dummy
> store(&y, 1, REL);
> 
> T2:
> if (load(&y, ACQ) == 1) {
>     assert (x == 0); // Should fail, but it won't I think
> }
> ```
> 
> The problem is that T2 must observe the dummy store because it is the most recent one. Yet, it cannot observe it, so we do not get any executions at all! I think the dummy store must also go to a dummy address to avoid it participating in `co` and `fr` and thus hiding other stores.

Yes, I came to the same conclusion (that this kind of executions would give wrong results). 

I think it's best to keep the store and allow rf from it, just treat it differently for side effect analysis. The same is true for matching lock/unlock pairs

JonasOberhauser:
> I think this transformation only makes sense if we also extend liveness handling to special-handle the dummy store. If it is just about detecting side-effect-free iterations, a more canonical approach would be to change the instrumentation produced by `DynamicSpinLoopDetection` to have
> 
> ```
> r = load(x);
> store(x, v); // Must be RMW with the load
> sideEffect = sideEffect | (r != v);
> ```
> 
> This would avoid the dummy altogether. A benefit (or a disadvantage) would be that we can detect this pattern without actually relying on special `Xchg` events: we could do this even for manually placed LL/SC implementations of Xchg, or Xchg's coming from C-litmus or SPIR-V (maybe we don't want to though?).

Yes, that would be my suggestion too

JonasOberhauser:
> I think this transformation only makes sense if we also extend liveness handling to special-handle the dummy store.

I would say that probably you need to adjust the liveness check anyways. Reading from a store such that all co successors have the same value.

ThomasHaas:
> > I think this transformation only makes sense if we also extend liveness handling to special-handle the dummy store.
> 
> I would say that probably you need to adjust the liveness check anyways. Reading from a store such that all co successors have the same value.

If the dummy store was to a dummy address, then it would not affect the co-maximality of the true stores and thus not prevent liveness violation detection.
That being said, extending liveness detection is probably better (and more principled).

JonasOberhauser:
> > > I think this transformation only makes sense if we also extend liveness handling to special-handle the dummy store.
> > 
> > 
> > I would say that probably you need to adjust the liveness check anyways. Reading from a store such that all co successors have the same value.
> 
> If the dummy store was to a dummy address, then it would not affect the co-maximality of the true stores and thus not prevent liveness violation detection. That being said, extending liveness detection is probably better (and more principled).

The issue is that it can potentially give you false positives, e.g., coi ordering is not preserved

hernanponcedeleon:
We will need some other approach for this, thus closing for now

PR #665, opened 2024-04-30
Propagate correct memory order tags when compiling to C11

hernanponcedeleon:


ThomasHaas (approved):
These are some of the changes from #663, right? I was about to write that we should do those in a separate PR, but you were faster.
LGTM.

hernanponcedeleon:
> These are some of the changes from #663, right? 

Yes, I just copied the changes from that branch.

ThomasHaas:
For some reason `linuxrwlock` fails the assertion under `vmm` but runs fine on `imm`. This is not supposed to be the case, right?

hernanponcedeleon:
I am actually taking a look to why this is the case. The violation has this pattern where `a:T1 -rlx-> T2 -acq-> T3 -rel-> b:T2` and it seems there is some missing propagation from `a:T1` to `b:T2` (I have seen this pattern in LKMM before one of Jonas fixes).

hernanponcedeleon:
VMM does not necessarily preserve some dependencies. For example, `ppo` has this `[Marked];(dep | coi | fri);[W & Marked]`, but it does not preserve orderings when the target is a load.

I used vsyncer-optimize by initially setting all memory accesses as SC and it returned the maximally relaxed version added in the latest commit.

ThomasHaas:
Oh, that's interesting. Have you asked why even data/address dependencies between loads are not preserved? It reminds me of DecAlpha's memory model.

hernanponcedeleon:
> Oh, that's interesting. Have you asked why even data/address dependencies between loads are not preserved? It reminds me of DecAlpha's memory model.

Maybe @JonasOberhauser can explain in more detail, but AFAIR, his view is that developers should not rely on this kind of (fragile?) dependencies for orderings and use instead barriers.

ThomasHaas (approved):
LGTM

JonasOberhauser:
> I understand the argument, but that implies that dependencies to writes are less fragile (definitely true for ctrl-deps)? Anyways, it doesn't need to bother us, so we can merge.


I'm adding those mostly to ensure no OOTA with a local argument (based on reasoning about ppo), instead of having to add no-OOTA as an ad-hoc axiom.

JonasOberhauser:
@hernanponcedeleon  have you confirmed it is because of the lack of R-dep>R ordering? Maybe add such ordering rules in a test run to see if it resolves the issue, would surprise me a little TBH

ThomasHaas:
I think the issue is that a relaxed RMW breaks a REL/ACQ sequence.
The counter example has this pattern:
- The RWThread releases the write lock.
- The RThread observes this release with a relaxed FADD (thus not synchronizing).
- The RThread then observes its own relaxed FADD with an acquire FSUB, and then enters the CS, seemingly without synchronization.

So in short, I think the culprit pattern is this:
```
T1:
W(&x, 1, REL);

T2:
RMW(&x, someOp, RLX); // Reads and overwrites the release write with a relaxed one
R(&x, ACQ); // Observes the relaxed write and thus fails to synchronize
```
If the second acquire load was an acquire fence, it would likely work again (due to `[R];po;[Acq & F];po` in `ppo`).

ThomasHaas:
Here is the test:

```C
atomic_int x, y;

void *thread_1(void *unused)
{
    atomic_store_explicit(&x, 1, memory_order_relaxed);
    atomic_store_explicit(&y, 1, memory_order_release);
    return NULL;
}

void *thread_2(void *unused)
{
    atomic_fetch_add_explicit(&y, 0, memory_order_relaxed);
    if (atomic_load_explicit(&y, memory_order_acquire) == 1) {
        int r1 = atomic_load_explicit(&x, memory_order_relaxed);
        assert (r1 == 1);
    }
}

int main()
{
	pthread_t t1, t2;

	pthread_join(t1, NULL);
	pthread_join(t2, NULL);

	return 0;
}
```
This test fails under `vmm` but succeeds under `imm`.

JonasOberhauser:
> I think the issue is that a relaxed RMW breaks a REL/ACQ sequence. The counter example has this pattern:
> 
> * The RWThread releases the write lock.
> * The RThread observes this release with a relaxed FADD (thus not synchronizing).
> * The RThread then observes its own relaxed FADD with an acquire FSUB, and then enters the CS, seemingly without synchronization.
> 
> So in short, I think the culprit pattern is this:
> 
> ```
> T1:
> W(&x, 1, REL);
> 
> T2:
> RMW(&x, someOp, RLX); // Reads and overwrites the release write with a relaxed one
> R(&x, ACQ); // Observes the relaxed write and thus fails to synchronize
> ```
> 
> If the second acquire load was an acquire fence, it would likely work again (due to `[R];po;[Acq & F];po` in `ppo`).

Thanks. The issue isn't relaxed per se. `RMW(&x, someOp, REL);` should have the same issue.
I suppose the issue here is that generally we can speculate through rfe;ppo;rfi . It is a combination of two things that provides order in these specific cases:
- the RMW's stored value depends on the read value, therefore compilers usually don't speculate 
- hardware usually doesn't speculate through RMWs (even though we have found a bug in BOOM doing exactly that).


However, the inc can be implemented as a CAS loop where the value to be stored is known before doing the read, at most 

```
    while (t = x, ! x.cas(t,t+1));
```

and perhaps some future hardware could then observe the execution
x==0
  read x 0 // speculatively
  write x 1 
read x 1 (through forwarding with acquire) & commit
read y 0  & commit
commit read x 0, write x1

I think the compiler won't do this speculation because of the VMM implied compiler barriers. Current hardware specs supported by VMM don't allow this speculation either as far as I'm aware. So in principle it might be sound in VMM to add this kind of ordering for the time being.

Either way, it seems a bit iffy.

PR #666, opened 2024-04-30
Log a warning instead of throwing an exception when opt is not availa…

hernanponcedeleon:
…ble in the system


ThomasHaas (commented):
Does this even work because you still return the path to the opt file (which was not generated)? Shouldn't you return the original file instead?
Also, I would make it more clear that we continue without optimizations. Something like this:
```
Failed to run opt (llvm optimizations). Continuing without optimizations.
```

hernanponcedeleon:
> Does this even work because you still return the path to the opt file (which was not generated)? Shouldn't you return the original file instead? 

Good catch. I tested it, but in my case it was working because I have the `-opt.ll` file from a previous run.

ThomasHaas (approved):
LGTM

PR #667, opened 2024-04-30
Add AssignmentInlining pass

ThomasHaas:
This PR adds a pass that inlines assignments:
```
r = someExpr
RegReader(r) // uses r
//------- becomes --------
r = someExpr // Now dead
RegReader(someExpr) // uses "someExpr" directly
```
This inlining is only performed if the register `r` is used only once so as to avoid duplicating `someExpr`.
The PR also adds more rewriting to `ExprSimplifier` to simplify ITE expressions obtained after inlining.


dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
```diff
@@ -23,8 +25,39 @@ public ExprSimplifier(boolean aggressive) {
...
+    // For now, it only rewrites "ITE(cond, x, y) op z" to "ITE(cond, x op z, y op z)".
+    // Returns NULL, if no rewriting is applicable.
+    private Expression tryGeneralRewrite(BinaryExpression expression) {
+        final Expression left = expression.getLeft().accept(this);
+        final Expression right = expression.getRight().accept(this);
+        if (!(left instanceof ITEExpr) && !(right instanceof ITEExpr)
+                // We don't rewrite "ITE op ITE" for now
+                || (left instanceof ITEExpr && right instanceof ITEExpr)) {
+            return null;
+        }
```
hernanponcedeleon, line 38: isn't the check simple if you check 
```
if one of them is ITE, 
  then apply the transformation
  else return null
```
?

You might even (I am not completely sure) use `if left instanceof ITEExpr ite` and avoid the explicit cast

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -35,7 +35,12 @@ public class ProcessingManager implements ProgramProcessor {
     @Option(name = DEAD_ASSIGNMENT_ELIMINATION,
             description = "Performs dead code elimination.",
             secure = true)
-    private boolean dce = true;
+    private boolean performDce = true;
+
+    @Option(name = ASSIGNMENT_INLINING,
+            description = "Performs inlining of assignments to avoid intermediary variables.",
```
hernanponcedeleon, line 41: "of assignments" -> "of assignments which are used only once"

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -106,15 +113,26 @@ private ProcessingManager(Configuration config) throws InvalidConfigurationExcep
                         FunctionProcessor.chain(
                                 ResolveLLVMObjectSizeCalls.fromConfig(config),
                                 sccp,
-                                dce ? DeadAssignmentElimination.fromConfig(config) : null,
-                                RemoveDeadCondJumps.fromConfig(config)
+                                dce,
+                                removeDeadJumps
                         ), Target.FUNCTIONS, true
                 ),
+                DebugPrint.withHeader("After flattening", Printer.Mode.ALL),
```
hernanponcedeleon, line 120: I guess this was for debugging. Please remove

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
ThomasHaas: I apply this transformation only if exactly one of the expressions is an ITE but not both.
So I have to capture the cases `ITE op x` and `x op ITE` and both are different (because op my not be commutative)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private AssignmentInlining() { }
+
+    public static AssignmentInlining newInstance() {
+        return new AssignmentInlining();
+    }
+
+    @Override
+    public void run(Function function) {
+        if (!function.hasBody() || function.getProgram().getFormat() == Program.SourceLanguage.LITMUS ) {
```
hernanponcedeleon, line 39: This looks like a hack.

What is the problem? I assume it has something to do with final assertions. If that if the case, we should make the pass work with those (we will eventually need this for spirv) instead of skipping it

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/SparseConditionalConstantPropagation.java
```diff
@@ -65,6 +66,9 @@ public static SparseConditionalConstantPropagation fromConfig(Configuration conf
 
     @Override
     public void run(Function func) {
+        if (!func.hasBody() || func.getProgram().getFormat() == Program.SourceLanguage.LITMUS) {
```
hernanponcedeleon, line 69: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
```diff
@@ -232,6 +275,26 @@ public Expression visitITEExpression(ITEExpr expr) {
...
         }
 
+        if (trueCase instanceof BoolLiteral tLit && falseCase instanceof BoolLiteral fLit) {
+            if (tLit.getValue() == fLit.getValue()) {
+                if (aggressive || cond.getRegs().isEmpty()) {
+                    return tLit;
+                } else if (tLit.getValue()) {
+                    return expressions.makeBoolBinary(cond, BoolBinaryOp.OR, expressions.makeTrue());
+                } else {
+                    return expressions.makeBoolBinary(cond, BoolBinaryOp.AND, expressions.makeFalse());
```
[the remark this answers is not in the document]
ThomasHaas: Yes, but I won't put an extra comment there because the whole class preserves dependencies (if `aggressive=false`) and so I would have to put that comment into every single optimization if we want to be consistent.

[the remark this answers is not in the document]
hernanponcedeleon: I was worried someone would read the `\/ true` or `/\ false` and get completely lost (as I did when I just saw the diff), but the class is documented at the top about this, so this should be enough

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
ThomasHaas: Both this pass and SCCP are sound for litmus tests (and litmus-style assertions). However, we deliberately do avoid optimizations on litmus code and that's why we added the NOOPT tag.
It is currently easier to just skip over litmus than to check for the NOOPT tag all the time.
In fact, it might be even better to have a separate processing pipeline for litmus code that does not even have these passes in the first place.

If we decide to optimize litmus code then we should first get rid of the NOOPT tag or put a more clear semantics to it (i.e., not blindly place it over the whole code). Updating the passes is then just a matter of removing the litmus check and possibly adding special-case handling of NOOPT tags if we still need them for some reason.

hernanponcedeleon: > However, we deliberately do avoid optimizations on litmus code and that's why we added the NOOPT tag. 

It is not true the NOOPT tag is just about litmus tests. We also need this for e.g. fake dependencies and spin-termination events.

> It is currently easier to just skip over litmus than to check for the NOOPT tag all the time. In fact, it might be even better to have a separate processing pipeline for litmus code that does not even have these passes in the first place.

But then instead of choosing to disable just SCCP and AssignmentInlining, we could check the format when calling `ProcessingManager.run`. If we do so, we can remove the part that tags every litmus event as NOOPT and check if every other occurrence of NOOPT is indeed necessary.

ThomasHaas: > > However, we deliberately do avoid optimizations on litmus code and that's why we added the NOOPT tag.
> 
> It is not true the NOOPT tag is just about litmus tests. We also need this for e.g. fake dependencies and spin-termination events.

Ok, it is true that there are other use-cases. While I think most if not all of them can be removed (e.g., fake dependencies are preserved anyways by the simplifier), if we want to support NOOPT for arbitrary events, I have to add support for this. I think most passes ignore NOOPT currently, so they all need fixing.

hernanponcedeleon (approved):

PR #670, opened 2024-05-08
Witness

hernanponcedeleon:
This PR allows do distinguish between generating the witness violation in the `*.dot` or `*.png` format. The later requires a dependency on `graphviz` that we don't want to have in `vsyncer`.

I also tried to unify all witness related options (including svcomp) as a single option, however I did not find a way to avoid the extra `WITNESS_ORIGINAL_PROGRAM_PATH` options because we need to know the path to the original c file.

I also restructure the folders related to witness code and did some formatting (thus the large diff).


dartagnan/src/main/java/com/dat3m/dartagnan/witness/WitnessType.java
```diff
@@ -0,0 +1,20 @@
...
+        return NONE;
+    }
+
+    public static boolean generateGraphviz(WitnessType type) {
+        return type.equals(DOT) || type.equals(PNG);
+    }
+
+    public static boolean convertToPng(WitnessType type) {
+        return type.equals(PNG);
+    }
```
ThomasHaas, line 18: You could make these methods non-static.

ThomasHaas: You also need to change the call from `convertToPng(witnessType)` to `witnessType.convertToPng()` (the method does not take a parameter anymore).

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -225,10 +225,7 @@ private static void generateWitnessIfAble(VerificationTask task, ProverEnvironme
             try {
                 WitnessBuilder w = WitnessBuilder.of(modelChecker.getEncodingContext(), prover,
                         modelChecker.getResult(), summary);
-                if (w.canBeBuilt()) {
-                    // We can only write witnesses if the path to the original C file was given.
-                    w.build().write();
-                }
+                w.build().write();
```
ThomasHaas, line 228: How come the `w.canBeBuilt()` check became obsolete? Was it never really relevant in the first place?

hernanponcedeleon: It was "irrelevant" in the sense that we could in principle create a witness without having the `programfile` attribute. However, in such a case we will not get the points in SVCOMP because either the tool checking the witness format will complain, or some of the validators will complain (dartagnan would).

I decided to remove this because anyway we never call dartagnan directly for SVCOMP, bur rather programmatically and [the code guarantees we pass the other option](https://github.com/hernanponcedeleon/Dat3M/blob/12df7a05e57301754625d17ef08d503964543655/svcomp/src/main/java/com/dat3m/svcomp/SVCOMPRunner.java#L135). Currently, dartagnan throws an exception if called directly and the option if missing. Thus, this needs to be fixed anyway and the easiest is probably to put the `w.canBeBuilt()` back.

ThomasHaas (approved):
LGTM

PR #671, opened 2024-05-09
Add atomicity axiom to sc.cat

hernanponcedeleon:
We are currently failing this `java -jar dartagnan/target/dartagnan.jar cat/sc.cat benchmarks/locks/ttas.c` due to the missing atomicity axiom in SC.

### DIFF
--- a/cat/sc.cat
+++ b/cat/sc.cat
@@ -8,3 +8,6 @@ let com = (rf | fr | co)
 
 (* Sequential consistency *)
 acyclic po | com as sc
+
+(* Atomicity *)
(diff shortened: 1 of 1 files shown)


ThomasHaas (approved):
I guess this is for user-friendliness so they don't have to use svcomp.cat?
Fine by me.

hernanponcedeleon:
It is not obvious by the name the svcomp.cat is SC+atomicity (not only of instructions but also of blocks). Most of the people will see SC.cat and assume it is stronger than any of the other models we have (which right now it is not since it violates atomicity)

PR #672, opened 2024-05-09
Add 4 litmus tests demonstrating unordered failing cmpxchg. 

hernanponcedeleon:
This PR adds 4 litmus test from [here](https://lkml.org/lkml/2024/5/1/952).

While testing this I noticed we are missing a RMW tag to read in the failing CAS case.
This can be seen in this rule of `linux-kernel.cat` + the comment in the `*-ordered-*` litmus tests
```
([M] ; po? ; [RMW] ; fencerel(After-atomic) ; [M]) |
```


ThomasHaas (approved):
LGTM

PR #675, opened 2024-05-14
Check all three properties by default

hernanponcedeleon:


ThomasHaas (approved):
LGTM

hernanponcedeleon:
> This also shows a problem in regards to litmus code assertions: it is not guaranteed that you can even check all three properties at the same time! Only for C-code you know that all properties are of the same type.

You are right, but we don't have any real user checking litmus tests. I still think we should use all 3 properties as default and maybe just improve the error msg (i.e., tell the user what to do) when mixed type properties are used
```
The set of properties [Program specification, Liveness, CAT specification] are of mixed type (safety and reachability properties). Cannot encode mixed properties into a single SMT-query.
```
->
```
The set of properties [Program specification, Liveness, CAT specification] are of mixed type (safety and reachability properties). Cannot encode mixed properties into a single SMT-query. Please select a different set of properties.
```

ThomasHaas:
Fine with me

PR #678, opened 2024-05-20
Remove support for LISA format

hernanponcedeleon:
The LISA format was hardly used and with [this](https://github.com/paulmckrcu/litmus/commit/4e999bd29f3a43830f554d7bc6560b2673f9584b), I think we can consider it dead.


ThomasHaas:
Do we gain much from deleting it now? IIRC, I used LISA for a demonstration in my FDL presentation because I didn't wanna bother with assembly code, so it was useful to me.

hernanponcedeleon:
If we see future use cases of this, sure, we can keep it. I just don't want to keep code around that is trivially useless.

If the advantage was that you can show case examples without having to rely on hard assembly syntax, I guess the advantage over C is that you can play with different barriers.

What if we have support for barriers as inline assembly? I think that would cover that case. I already have a prototype that could handle e.g., this code
```
void *thread_1(void *unused)
{
    x = 1;
    asm volatile ("mfence" ::: "memory");
    a = y;
    return NULL;
}

void *thread_2(void *unused)
{
    y = 1;
    asm volatile ("mfence" ::: "memory");
    b = x;
    return NULL;
}

```

ThomasHaas:
Apart from the advantage of playing around with tagging, it also avoids the overhead of creating `main` + `pthread_create`.
This in turn reduces the generated witnesses to only contain the relevant code and no thread creation overhead (the witnesses look litmus-like). 
That being said, I currently have no plans to use LISA. I'm just saying that it was useful.

hernanponcedeleon:
I don't believe the overhead of a adding a `main` + `pthread_create` is worse than the annoyance of having to write threads in columns or adding a bunch of instructions (load register, jump, label) to add any interesting control flow.

The witness argument is indeed a good one. We are planning to improve the witness generation soon, so I will keep in mind this to see if we can do something about this.

Anyway, if we ever need this again, we can pick the few needed classes from the history.

ThomasHaas:
Yeah, my argument is not that writing LISA is easier than C (it certainly isn't). It's just that it is easier and more general than all the other litmus languages we support while also having nice witness graphs.

ThomasHaas (approved):
LGTM

PR #679, opened 2024-05-21
Fences as inline assembly

hernanponcedeleon:
This PR basic support for parsing fences given as inline assembly.


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -394,6 +394,8 @@ public Expression visitCallInst(CallInstContext ctx) {
             //FIXME ignore side effects of inline assembly
             if (resultRegister != null) {
                 block.events.add(newLocal(resultRegister, program.newConstant(returnType)));
+            } else {
+                visitChildren(ctx);
```
ThomasHaas, line 398: Just call `ctx.inlineAsm().accept(this)` to traverse down. 

Also, we should probably log a warning when executing the first branch because we interpret the inline assembly as just some unconstrained value.

ThomasHaas (approved):
LGTM

PR #681, opened 2024-05-22
Replace several back jumps with via forward jumping

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public void run(Program program) {
+        program.getFunctions().forEach(f -> backToForwardJumps(f, Event::getGlobalId));
+    }
+
+    private void backToForwardJumps(Function function, ToIntFunction<Event> linId) {
+        for (Label label : function.getEvents(Label.class)) {
+            final List<CondJump> backJumps = label.getJumpSet().stream()
+                    .filter(j -> linId.applyAsInt(j) > linId.applyAsInt(label))
+                    .toList();
```
ThomasHaas, line 55: There is no need for the `linId` thing. Just use `getGlobalId` directly (I know you got this code from the `LoopFormVerificattion` pass).
Also, you could add a `.sorted()` to get the backjumps correctly ordered. Then the last element will be your final backjump.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
```diff
@@ -0,0 +1,78 @@
...
+
+            if (backJumps.size() > 1) {
+                final CondJump last = backJumps.stream().max(comparingInt(linId)).get();
+
+                final Label forwardLabel = EventFactory.newLabel("Forward_to_" + label.getGlobalId());
+                final CondJump gotoHead = EventFactory.newGoto(label);
+
+                final Label continueLabel = EventFactory.newLabel("Continue_after_" + last.getGlobalId());
+                final CondJump gotoContinue = EventFactory.newGoto(continueLabel);
```
ThomasHaas, line 64: Add `__` prefixes to internally generated names. Also, I think you could call the labels: `__continueLoop` (or `__repeatLoop`) and `__breakLoop`, indicating their purposes (in fact, you didn't even have to use explicit gotos to get multiple backjumps: a `continue` in a normal loop is a backjump).
Lastly, using `globalId` in the names is a bad idea because it will immediately cause mismatches.
I typically use a counter and add `#<counter>` as a suffix to give unique names (technically we do not even need unique names, but it makes CTRL+F searches easier in the IR).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
```diff
@@ -0,0 +1,78 @@
...
+
+                last.insertAfter(gotoContinue);
+                gotoContinue.insertAfter(forwardLabel);
+                forwardLabel.insertAfter(gotoHead);
+                gotoHead.insertAfter(continueLabel);
```
ThomasHaas, line 69: I think there is an overload to insert a whole sequence: `last.insertAfter(List.of(...))`;

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
```diff
@@ -0,0 +1,78 @@
...
+    ...
+    if Y goto Forward_to_L
+    goto Continue_after_last
+    Forward_to_L:
+    goto L
+    Continue_after_last
+    More Code
+    ...
+*/
+public class MultipleBackJumpsToForwardJumps implements ProgramProcessor {
```
ThomasHaas, line 40: This could just be a `FunctionProcessor` or if you want to keep it a `ProgramProcessor`, you need to manually run an `IdReassignment` pass (you can call it in `run`) to restore correct global ids.

Also, the pass name is oddly specific (and I think not fully correct). Just call it `NormalizeLoops`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    private void backToForwardJumps(Function function, ToIntFunction<Event> linId) {
+        for (Label label : function.getEvents(Label.class)) {
+            final List<CondJump> backJumps = label.getJumpSet().stream()
+                    .filter(j -> linId.applyAsInt(j) > linId.applyAsInt(label))
+                    .toList();
+
+            if (backJumps.size() > 1) {
+                final CondJump last = backJumps.stream().max(comparingInt(linId)).get();
```
ThomasHaas, line 58: `LoopFormVerification` requires that the unique backjump is also unconditional and this pass may fail to achieve that if `backjumps.size()==1`. The check `backjumps.size() == 1 && backjumps.get(0).isGoto()` would be fine. 
Btw. the only language level construct that can produce conditional backjumps are `do-while` loops.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
```diff
@@ -0,0 +1,78 @@
...
+
+                last.insertAfter(gotoContinue);
+                gotoContinue.insertAfter(forwardLabel);
+                forwardLabel.insertAfter(gotoHead);
+                gotoHead.insertAfter(continueLabel);
+
+                for(CondJump j : backJumps) {
+                    final CondJump forwardJump = EventFactory.newJump(j.getGuard(), forwardLabel);
+                    j.replaceBy(forwardJump);
+                }
```
ThomasHaas, line 74: Don't replace the jumps because it will not retain their metadata. Just change the jump target directly.
For this you need to call `j.updateReferences(Map.of(j.getLabel(), forwardLabel))`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MultipleBackJumpsToForwardJumps.java
hernanponcedeleon: You mean `backJumps.size() > 1 || (backjumps.size() == 1 && backjumps.get(0).isGoto())`, right?

I could also use `backJumps.size() > 0`. This will do the simplification even for the case where we have a single unconditional backjump, but I don't expect this to create any performance issue and it keeps the code simpler.

ThomasHaas: I meant to skip the code if my proposed conditions holds:
```
if (backjumps.size() == 1 && backjumps.get(0).isGoto()) {
   return;
}
```
Technically there is no problem in always doing this transformation but it has the ugly consequence that the pass is not idempotent but I think it should be just for sanity reasons.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -85,6 +85,9 @@ private ProcessingManager(Configuration config) throws InvalidConfigurationExcep
         final FunctionProcessor removeDeadJumps = RemoveDeadCondJumps.fromConfig(config);
         programProcessors.addAll(Arrays.asList(
                 printBeforeProcessing ? DebugPrint.withHeader("Before processing", Printer.Mode.ALL) : null,
+                ProgramProcessor.fromFunctionProcessor(
+                    NormalizeLoops.newInstance(), Target.FUNCTIONS, true
+                ),
```
ThomasHaas, line 90: This will not apply to litmus-style code in case you also intend to run it on such code (e.g., SPIRV).
IIRC, back in the days the distinction between functions and threads was made to just avoid running passes on litmus code altogether. But I'm not sure if this is still necessary.
The cleanest approach would probably be to construct a separate processing pipeline for litmus code anyways.

hernanponcedeleon: I hardly believe anybody will ever write litmus benchmarks that would require this pass to be applied (the same probably applies to other passes we have).

The new code follows the same pattern we use for the whole class: if the pass is a `FunctionProcessor`, then use `Target.FUNCTIONS` until threads are created. After that, we use `Target.THREADS`.

That being said, you are right that eventually we might want to have different processing pipelines for the different formats.

ThomasHaas (approved):
LGTM

PR #683, opened 2024-05-27
Alias live cycle detection stats

xeren:
This PR modifies the debug `addInto` statistics in `InclusionBasedPointerAnalysis`.  Attempts to add temporary elements during cycle detection are separated from those adding edges to the inclusion graph.  This should slightly improve insights into the analysis process.


ThomasHaas (approved):
LGTM

PR #684, opened 2024-05-28
UI update

ThomasHaas:
This PR updates our UI and consists of a few features that I implemented a while ago for live demonstrations:
- The UI allows selection of a property to check for (only a single property)
- Configuration parameters can be passed (e.g. `encoding.locallyConsistent=false`)
- The fonts of the two code panes (program code / wmm code) is scalable by using ctrl and +/-.
- The UI can generate a witness and render it (assuming `dot` can be invoked to generate the `.png`).
- If no program file is loaded, we assume C-code by default now (previously the format was unspecified and we got an error).
- EDIT: I added a zoom option to displayed witnesses using +/- keys. The zoom is not really smooth and the quality gets kinda bad, but it is still better than no zoom at all  I guess :).

I also reformatted all UI files and replaced some switch statements by switch expressions.


hernanponcedeleon:
A few things I noticed while testing
- We get `ERROR: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 0` when running litmus tests from the console
- A witness is generated even when we get UNKNOWN. This is probably something we want to avoid to mimic console behavior.
- It would be good to add the witness windows as a listener to the other buttons such that the current window is closed when we start a new verification run (*)

(*) This is just a suggestion because one could argue we should keep old windows opened and the user could close them when not needed (specially if the verification takes long). However, I think the ui is mostly useful for examples with short verification times and I think ending up with multiple opened windows might confuse the user about what was the last generated witness.

ThomasHaas:
> A few things I noticed while testing
> 
>     * We get `ERROR: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 0` when running litmus tests from the console

I cannot reproduce this. Are you maybe missing my last commit to this branch?

> 
>     * A witness is generated even when we get UNKNOWN. This is probably something we want to avoid to mimic console behavior.

Console also produces a witness for UNKNOWN.

> 
>     * It would be good to add the witness windows as a listener to the other buttons such that the current window is closed when we start a new verification run (*)
> 
> 
> (*) This is just a suggestion because one could argue we should keep old windows opened and the user could close them when not needed (specially if the verification takes long). However, I think the ui is mostly useful for examples with short verification times and I think ending up with multiple opened windows might confuse the user about what was the last generated witness.

If the user is confused by multiple windows, he should just close them. And for all short examples, the user could always close all windows and just rerun the verification to get the desired witness.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -215,6 +212,24 @@ public static void main(String[] args) throws Exception {
...
 
+    public static File generateExecutionGraphFile(VerificationTask task, ProverEnvironment prover, ModelChecker modelChecker,
+                                                  WitnessType witnessType)
+            throws InvalidConfigurationException, SolverException, IOException {
+        Preconditions.checkArgument(modelChecker.hasModel(), "No model to generate.");
+
+        final ExecutionModel m = ExecutionModel.withContext(modelChecker.getEncodingContext());
+        m.initialize(prover.getModel());
+        final SyntacticContextAnalysis synContext = newInstance(task.getProgram());
+        final String name = task.getProgram().getName().substring(0, task.getProgram().getName().lastIndexOf('.'));
```
hernanponcedeleon, line 223: The error I see in the console is due to this line. Somehow, the program has no name

ThomasHaas: Oh. You originally said that you got an error when running *from the console* which is what I couldn't reproduce.
You meant to say that you get an error when running *from the UI* I guess. That one I can reproduce.

hernanponcedeleon: Sorry ... By console I meant the display from the UI where we are supposed to see the result

ThomasHaas: Fixed

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -165,33 +172,17 @@ public static void main(String[] args) throws Exception {
...
-                        case LAZY:
-                            modelChecker = RefinementSolver.run(ctx, prover, task);
-                            break;
-                        default:
-                            throw new InvalidConfigurationException("unsupported method " + o.getMethod());
-                    }
+                    modelChecker = switch (o.getMethod()) {
+                        case EAGER -> AssumeSolver.run(ctx, prover, task);
+                        case LAZY -> RefinementSolver.run(ctx, prover, task);
+                    };
```
hernanponcedeleon, line 178: I know this cover all cases right now, but why not keep the default case just in case we even add a new method and forget to update this?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -215,6 +206,27 @@ public static void main(String[] args) throws Exception {
         }
     }
 
+    public static File generateExecutionGraphFile(VerificationTask task, ProverEnvironment prover, ModelChecker modelChecker,
+                                                  WitnessType witnessType)
+            throws InvalidConfigurationException, SolverException, IOException {
+        Preconditions.checkArgument(modelChecker.hasModel(), "No model to generate.");
```
hernanponcedeleon, line 212: "to generate" -> "to generate execution graph"

ui/src/main/java/com/dat3m/ui/Dat3M.java
```diff
@@ -4,111 +4,161 @@
...
+
+    @Override
+    public void actionPerformed(ActionEvent event) {
+        String command = event.getActionCommand();
+        if (ControlCode.TEST.actionCommand().equals(command)) {
             runTest();
-            if(testResult != null){
+            if (testResult != null) {
                 optionsPane.getConsolePane().setText(testResult.getVerdict());
+                if (optionsPane.getOptions().showWitness() && testResult.hasWitness()) {
```
hernanponcedeleon, line 86: I find annoying that if the "generate witness" button is marked (I would expect user to click this once and keep it marked rather than selecting/unselecting each time they run the verification), we generate a witness for the UNKNOWN result.

Rather than checking if there is a witness, IMO we should check if the result is FAIL.

I don't think the argument "we also generate a witness if you pass the option --witness=png to the console and the result is UNKNOWN" is valid for the simple reason that it creates a file that does not disturb the user. Contrary to this, when running on the CI a new windows will pop up.

ui/src/main/java/com/dat3m/ui/options/OptionsPane.java
```diff
@@ -73,34 +86,38 @@ public OptionsPane(){
...
-        Method method = (Method)methodPane.getSelectedItem();
-        Solvers solver = (Solvers)solverPane.getSelectedItem();
-        return new UiOptions(target, method, bound, solver, timeout, cflags);
+        boolean showViolationGraph = showViolationField.isSelected();
+        String cflags = cflagsField.getText().strip();
+        String config = configField.getText().strip();
+        Arch target = (Arch) targetPane.getSelectedItem();
+        Method method = (Method) methodPane.getSelectedItem();
+        Solvers solver = (Solvers) solverPane.getSelectedItem();
+        EnumSet<Property> properties = EnumSet.of((Property) propertyPane.getSelectedItem());
```
hernanponcedeleon, line 116: Since we just have a few properties (TBH, I would filter out the SVCOMP data races), wouldn't it makes more sense to have this as a button (similar to showViolationGraph) rather than a selector? This way the user can check more than one property at a time.

ui/src/main/java/com/dat3m/ui/options/OptionsPane.java
```diff
@@ -120,23 +137,32 @@ private void mkGrid(){
         cflagsPane.add(new JLabel("CFLAGS: "));
         cflagsPane.add(cflagsField);
 
+        JPanel configPane = new JPanel(new FlowLayout(LEFT));
+        configPane.add(new JLabel("Config:  "));
```
hernanponcedeleon, line 141: I think "Config" is confusing. I would use "Options" which is how we refer to those in the README

ui/src/main/java/com/dat3m/ui/Dat3M.java
ThomasHaas: I can change that. But your original argument was that it was not aligned with the console behaviour although it was.

ui/src/main/java/com/dat3m/ui/options/OptionsPane.java
ThomasHaas: I think there are plenty of things you could do with the UI, especially when considering user-friendliness.
The features I added in this PR are mostly related to doing live demonstrations (like enlarging fonts/zooming), i.e., the user is me.

Radio buttons come with the problem that not a single one may be selected, so you need error reporting as well.
Similarly, multiple properties fail for litmus code in many case so you also need error reporting there (I guess the exception error in the console might be enough).
None of that is particularly hard, but should be done in a different PR IMO.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
hernanponcedeleon: By complain you mean "raise a warning" or it won't even compile?

ui/src/main/java/com/dat3m/ui/Dat3M.java
hernanponcedeleon: My bad based on what I actually wrote. What I meant was "it is different in the sense in the UI something happened that annoys the user whole in the console it does not"

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: It won't compile. Switches always need to be exhaustive.

hernanponcedeleon (approved):

PR #685, opened 2024-05-28
Add support for memcpy_s

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1293,12 +1294,17 @@ private List<Event> inlineNonDet(FunctionCall call) {
     //FIXME: The following support for memcpy, memcmp, and memset is unsound
     // For proper support, we need at least alias information and most likely also proper support for mixed-sized accesses
 
-    // Handles both std.memcpy and llvm.memcpy
-    private List<Event> inlineMemCpy(FunctionCall call) {
+    // Handles std.memcpy, std.memcpy_s and llvm.memcpy
+    private List<Event> memCpy(FunctionCall call, boolean c11) {
```
ThomasHaas, line 1298: How does `c11` play into this? From the documentation, it seems that `c99` does not even have the `_s` version of `memcpy`. Also, why did you get rid of the "inline"-suffix in the method name?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1334,18 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+    }
+
+    private List<Event> inlineMemCpyC99(FunctionCall call) {
+        return memCpy(call, false);
+    }
+
+    private List<Event> inlineMemCpyLLVM(FunctionCall call) {
+        return memCpy(call, false);
+    }
+
```
ThomasHaas, line 1348: You could also use a lambda instead of these 3 version, i.e., rather than referencing the extra method `Intrinsics::inlineMemCpyLLVM` you could directly write `call -> memCpy(call, false)`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1293,12 +1294,17 @@ private List<Event> inlineNonDet(FunctionCall call) {
...
 
-    // Handles both std.memcpy and llvm.memcpy
-    private List<Event> inlineMemCpy(FunctionCall call) {
+    // Handles std.memcpy, std.memcpy_s and llvm.memcpy
+    private List<Event> memCpy(FunctionCall call, boolean c11) {
+        // https://en.cppreference.com/w/c/string/byte/memcpy
+        // In C99 version, signature is
+        // memcpy( void *restrict dest, const void *restrict src, size_t count );
+        // In C11 version, signature is
+        // memcpy_s( void *restrict dest, rsize_t destsz, const void *restrict src, rsize_t count );
```
ThomasHaas, line 1303: These are two different functions and the former still exists in `c11` as far as I see.
You should rather add a TODO/FIXME that says that we treat `memcpy_s` simply as `memcpy` (they have different semantics!).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: Just to make it a bit more clear that this method was not directly called from the same place all other `inlineX` are called.
But I guess I can put back the inline in the name and use lambdas as you suggest in the other comment.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: As far as I can see, the differences only occur in cases where there are UB ... thus, from the point of view of verification (at least the kind of verification we do where we assume the absence of those type of UB we cannot check), I would argue that they have the "same" semantics and it makes sense to use the same model for them

ThomasHaas: I don't think this is true precisely because `memcpy_s` does more checks to avoid UB in a few cases.
So a program might have UB with `memcpy` but does not have UB with `memcpy_s`.
The whole point of the extra parameter is to make checks possible before invoking UB (similar to `memset_chk`).

PR #686, opened 2024-05-28
Add support for memcpy_s

hernanponcedeleon:
Somehow, I ended up closing #685 and I don't find a way to open it again.

This PR should give a proper implementation of `memcpy_s`, including the runtime checks to avoid UB.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
@@ -209,6 +209,7 @@ public class Intrinsics {
         LKMM_FENCE("__LKMM_FENCE", false, false, false, true, Intrinsics::handleLKMMIntrinsic),
         // --------------------------- Misc ---------------------------
         STD_MEMCPY("memcpy", true, true, true, false, Intrinsics::inlineMemCpy),
+        STD_MEMCPYS("memcpy_s", true, true, true, false, Intrinsics::inlineMemCpyS),
         STD_MEMSET(List.of("memset", "__memset_chk"), true, false, true, false, Intrinsics::inlineMemSet),
         STD_MEMCMP("memcmp", false, true, true, false, Intrinsics::inlineMemCmp),
         STD_MALLOC("malloc", false, false, true, true, Intrinsics::inlineMalloc),
@@ -1323,6 +1324,58 @@ public class Intrinsics {
         if (call instanceof ValueFunctionCall valueCall) {
             // std.memcpy returns the destination address, llvm.memcpy has no return value
             replacement.add(EventFactory.newLocal(valueCall.getResultRegister(), dest));
+        }
+
+        return replacement;
+    }
+
+    private List<Event> inlineMemCpyS(FunctionCall call) {
+        final Function caller = call.getFunction();
+        final Expression dest = call.getArguments().get(0);
+        final Expression destsz = call.getArguments().get(1);
+        final Expression src = call.getArguments().get(2);
+        final Expression countExpr = call.getArguments().get(3);
+
+        if (!(countExpr instanceof IntLiteral countValue)) {
+            final String error = "Cannot handle memcpy_s with dynamic count argument: " + call;
+            throw new UnsupportedOperationException(error);
+        }
+        final int count = countValue.getValueAsInt();
+
+        // Runtime checks
+        final Expression nullExpr = expressions.makeZero(types.getArchType());
+        final Expression destIsNull = expressions.makeEQ(dest, nullExpr);
+        final Expression srcIsNull = expressions.makeEQ(src, nullExpr);
+        final Expression countGtDestsz = expressions.makeGT(countExpr, destsz, false);
+        final Expression overlap = expressions.makeOr(
+                expressions.makeGTE(expressions.makeAdd(src, countExpr), dest, false),
+                expressions.makeGTE(expressions.makeAdd(dest, countExpr), src, false));
+
+        // Runtime full condition
+        final Expression c1 = expressions.makeOr(destIsNull, srcIsNull);
+        final Expression c2 = expressions.makeOr(c1, countGtDestsz);
+        final Expression c3 = expressions.makeOr(c2, overlap);
+        final Expression value = expressions.makeITE(c3, expressions.makeZero(types.getArchType()), reg);
+
+        final List<Event> replacement = new ArrayList<>(2 * count + 1);
+        for (int i = 0; i < count; i++) {
+            final Expression offset = expressions.makeValue(i, types.getArchType());
+            final Expression srcAddr = expressions.makeAdd(src, offset);
+            final Expression destAddr = expressions.makeAdd(dest, offset);
+            // FIXME: We have no other choice but to load ptr-sized chunks for now
+            final Register reg = caller.getOrNewRegister("__memcpy_s_" + i, types.getArchType());
+
+            replacement.addAll(List.of(
+                    EventFactory.newLoad(reg, srcAddr),
+                    EventFactory.newStore(destAddr, value)
+            ));
+        }
+        if (call instanceof ValueFunctionCall valueCall) {
+            // Returns zero on success and non-zero value on error
+            Register resultRegister = valueCall.getResultRegister();
+            final Expression ok = expressions.makeZero((IntegerType)resultRegister.getType());
+            final Expression error = expressions.makeOne((IntegerType)resultRegister.getType());
+            replacement.add(EventFactory.newLocal(resultRegister, expressions.makeITE(c3, ok, error)));
         }
 
         return replacement;
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
@@ -1396,7 +1396,7 @@ public class Intrinsics {
 
         // Condition 2: dest != NULL && destsz <= RSIZE_MAX && (src == NULL || count > destsz || overlap(src, dest)) 
         // ----> return error > 0 and zero out [dest, dest+destsz)
-        // The first two are guaranteed by not matching c1
+        // The first two are guaranteed by not matching cond1
         final Expression cond2 = expressions.makeOr(expressions.makeOr(srcIsNull, invalidCount), overlap);
         CondJump skipE2 = EventFactory.newJump(expressions.makeNot(cond2), success);
         CondJump skipRest2 = EventFactory.newGoto(end);
@@ -1408,10 +1408,10 @@ public class Intrinsics {
         for (int i = 0; i < destsz; i++) {
             final Expression offset = expressions.makeValue(i, types.getArchType());
             final Expression destAddr = expressions.makeAdd(dest, offset);
-            final Expression value = expressions.makeZero(types.getArchType());
-            replacement.addAll(List.of(
-                EventFactory.newStore(destAddr, value)
-            ));
+            final Expression zero = expressions.makeZero(types.getArchType());
+            replacement.add(
+                EventFactory.newStore(destAddr, zero)
+            );
         }
         replacement.addAll(List.of(
             retError2,
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
@@ -1375,7 +1375,7 @@ public class Intrinsics {
             Register resultRegister = valueCall.getResultRegister();
             final Expression ok = expressions.makeZero((IntegerType)resultRegister.getType());
             final Expression error = expressions.makeOne((IntegerType)resultRegister.getType());
-            replacement.add(EventFactory.newLocal(resultRegister, expressions.makeITE(c3, ok, error)));
+            replacement.add(EventFactory.newLocal(resultRegister, expressions.makeITE(c3, error, ok)));
         }
 
         return replacement;
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
@@ -1358,17 +1358,13 @@ public class Intrinsics {
         final Expression srcIsNull = expressions.makeEQ(src, nullExpr);
         // We assume RSIZE_MAX = 2^64-1
         final Expression rsize_max = expressions.makeValue(BigInteger.ONE.shiftLeft(64).subtract(BigInteger.ONE), types.getArchType());
-        final Expression invalidDestsz = expressions.makeGT(
-                expressions.makeCast(destszExpr, types.getArchType()), rsize_max, false);
-        final Expression countGtMax = expressions.makeGT(
-                expressions.makeCast(countExpr, types.getArchType()), rsize_max, false);
+        final Expression invalidDestsz = expressions.makeGT(destszExpr, rsize_max, false);
+        final Expression countGtMax = expressions.makeGT(countExpr, rsize_max, false);
         final Expression countGtdestszExpr = expressions.makeGT(countExpr, destszExpr, false);
         final Expression invalidCount = expressions.makeOr(countGtMax, countGtdestszExpr);
-        // src and dest are pointers (int64), count is usually interpreted as int32, thus the cast
-        final Expression countExprCast = expressions.makeCast(countExpr, types.getArchType());
         final Expression overlap = expressions.makeAnd(
-                expressions.makeGTE(expressions.makeAdd(src, countExprCast), dest, false),
-                expressions.makeGTE(expressions.makeAdd(dest, countExprCast), src, false));
+                expressions.makeGTE(expressions.makeAdd(src, countExpr), dest, false),
+                expressions.makeGTE(expressions.makeAdd(dest, countExpr), src, false));
 
         final List<Event> replacement = new ArrayList<>();
         
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
@@ -1337,12 +1337,11 @@ public class Intrinsics {
         final Register resultRegister = ((ValueFunctionCall)call).getResultRegister();
         final Function caller = call.getFunction();
         final Expression dest = call.getArguments().get(0);
-        // This parameter has type rsize_t/size_t which we model as types.getArchType(), thus the cast
-        final Expression destszExpr = expressions.makeCast(call.getArguments().get(1), types.getArchType());
+        final Expression destszExpr = call.getArguments().get(1);
         final Expression src = call.getArguments().get(2);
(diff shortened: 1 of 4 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1329,86 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+            // FIXME: We have no other choice but to load ptr-sized chunks for now
+            final Register reg = caller.getOrNewRegister("__memcpy_s_" + i, types.getArchType());
+            // We should zero out the entire destination range [dest, dest+destsz)
+            // Here we handle [dest, dest+count), the remaining are handled by the loop below
+            final Expression value = expressions.makeITE(c4, expressions.makeZero(types.getArchType()), reg);
+
+            replacement.addAll(List.of(
+                    EventFactory.newLoad(reg, srcAddr),
+                    EventFactory.newStore(destAddr, value)
+            ));
```
ThomasHaas, line 1384: There are three cases for `memcpy_s`:
- The inputs are valid and you do a standard memcpy (+ generate a success error code).
- The inputs are invalid and in particular `dest == NULL`, then the method only produces an error code without any memory accesses.
- The inputs are invalid but at least `dest != NULL` (*), then the range `[dest, dest + destsz]` gets zeroed which only requires memory stores (no loads).

I don't think your implementation is correct. You have to use proper branching to make the case distinctions rather than trying to use ITE.

Btw. how is this function even used in practice? Does the compiler magically transform `memcpy` to `memcpy_s` or does the user specficially call the latter? If so, to what extent can the user recover from an error or does he just terminate the program but with an error code that is better than segfault?
Basically, I'm wondering if failure of `memcpy_s` is a bug in pretty much all cases. If so, we could also generate an `assert (boundCheck) ### memcpy_s error` and don't even bother too much about the failure details (we could also provide an option to toggle this behaviour).

(*) Technically, `destsz` could also be invalid but only if it is larger than `RSIZE_MAX` which is impossible in the implementation anyway.

ThomasHaas:
I think we should have a unit test related to `memcpy_s`. This one seems to be not so straightforward.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1329,86 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+        final Expression srcIsNull = expressions.makeEQ(src, nullExpr);
+        // We assume RSIZE_MAX = 2^64-1
+        final Expression rsize_max = expressions.makeValue(BigInteger.ONE.shiftLeft(64).subtract(BigInteger.ONE), types.getArchType());
+        final Expression gtMax = expressions.makeOr(
+                expressions.makeGT(countExpr, rsize_max, false),
+                expressions.makeGT(destszExpr, rsize_max, false));
+        final Expression countGtdestszExpr = expressions.makeGT(countExpr, destszExpr, false);
+        final Expression overlap = expressions.makeOr(
+                expressions.makeGTE(expressions.makeAdd(src, countExpr), dest, false),
+                expressions.makeGTE(expressions.makeAdd(dest, countExpr), src, false));
```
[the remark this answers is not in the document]
hernanponcedeleon: Good catch. I also think it would be correct if we use `makeAnd`. I will go with this.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: Ok, I missed the "(if both dest and destsz are valid)". After re-reading, this is my understanding
- `dest == NULL` or `invalid destsz` ----> return error > 0
- `dest != NULL && valid destsz && (src == NULL || count > destsz || overlap(src, dest))` ----> return error > 0 and zero out `[dest, dest+destsz)`

> I don't think your implementation is correct. You have to use proper branching to make the case distinctions rather than trying to use ITE.

Why do you think we need branches? I don't see this.

> Btw. how is this function even used in practice? Does the compiler magically transform `memcpy` to `memcpy_s` or does the user specficially call the latter? 

The user explicitly uses the later. This is used in libvsync and the whole reason of this PR. It is then the responsibility of the developer to check if the returned value != 0.

> If so, to what extent can the user recover from an error or does he just terminate the program but with an error code that is better than segfault? Basically, I'm wondering if failure of `memcpy_s` is a bug in pretty much all cases. If so, we could also generate an `assert (boundCheck) ### memcpy_s error` and don't even bother too much about the failure details (we could also provide an option to toggle this behaviour).

This is up to the user. If we want to be precise about its semantics, I don't think we can make this.
If we don't want to be precise, I would simply treat `memcpy_s` as `memcpy` and get minimal changes to the current code.

> 
> (*) Technically, `destsz` could also be invalid but only if it is larger than `RSIZE_MAX` which is impossible in the implementation anyway.

ThomasHaas: > Ok, I missed the "(if both dest and destsz are valid)". After re-reading, this is my understanding
> 
>     * `dest == NULL` or `invalid destsz` ----> return error > 0
> 
>     * `dest != NULL && valid destsz && (src == NULL || count > destsz || overlap(src, dest))` ----> return error > 0 and zero out `[dest, dest+destsz)`

Yes, that seems right. Importantly, in neither of those two cases you load from `src`!

> 
> 
> > I don't think your implementation is correct. You have to use proper branching to make the case distinctions rather than trying to use ITE.
> 
> Why do you think we need branches? I don't see this.

You just listed the different cases. One case performs loads and stores, another only stores, and the third one has no memory accesses. There is no way to get this behaviour without three different control-flow branches.
Your solution tried to merge cases (1) and (2) by always performing the loads and just ignoring their results to mimic case (2). These redundant loads are already questionable and you should avoid them, but you definitely cannot have redundant stores so case (3) must be different than the other two.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,108 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+        final Expression srcIsNull = expressions.makeEQ(src, nullExpr);
+        // We assume RSIZE_MAX = 2^64-1
+        final Expression rsize_max = expressions.makeValue(BigInteger.ONE.shiftLeft(64).subtract(BigInteger.ONE), types.getArchType());
+        final Expression invalidDestsz = expressions.makeGT(
+                expressions.makeCast(destszExpr, types.getArchType()), rsize_max, false);
+        final Expression countGtMax = expressions.makeGT(
+                expressions.makeCast(countExpr, types.getArchType()), rsize_max, false);
+        final Expression countGtdestszExpr = expressions.makeGT(countExpr, destszExpr, false);
+        final Expression invalidCount = expressions.makeOr(countGtMax, countGtdestszExpr);
+        // src and dest are pointers (int64), count is usually interpreted as int32, thus the cast
```
hernanponcedeleon, line 1367: This and the other casts in the pass really bother me. Without them, I was getting different BVs sizes, but I think the compiler and the type system should take care of giving me the correct types. So there is something that still seems fishy here. I suspect my system might not be properly recognizing `memcpy_s`

ThomasHaas: You can see this problem in your `.ll` file:
```
declare dso_local i32 @memcpy_s(...) #3 // varargs declaration because the function is undeclared in the C-code
//... The call is thus a varags call with ugly function type casts:
  %73 = call i32 (i32*, i64, i32*, i64, ...) bitcast (i32 (...)* @memcpy_s to i32 (i32*, i64, i32*, i64, ...)*)(i32* %71, i64 12, i32* %72, i64 12), !dbg !86
```
This function is declared in `string.h` according to cppreference and you are not including that one.

hernanponcedeleon: Even if I add `string.h`, function `memcpy_s` ends up with an empty body in the llvm file. There is this

> As with all bounds-checked functions, memcpy_s only guaranteed to be available if __STDC_LIB_EXT1__ is defined by the implementation and if the user defines __STDC_WANT_LIB_EXT1__ to the integer constant 1 before including [<string.h>](https://en.cppreference.com/w/c/string/byte).

and by trying the code from [here](https://stackoverflow.com/questions/47867130/stdc-lib-ext1-availability-in-gcc-and-clang), I get `#error "__STDC_LIB_EXT1__ not defined"`, so we might need a solution that works even if the implementation of `memcpy_s` is not available.

However, if I remove the casts from the code, I am now not able to reproduce the parsing problems I was getting before. What I don't understand is that even if the llvm code show different types in the call arguements
```
  %8 = call i32 (i8*, i64, i32*, i64, ...) bitcast (i32 (...)* @memcpy_s to i32 (i8*, i64, i32*, i64, ...)*)(i8* null, i64 12, i32* %7, i64 12), !dbg !27
  %15 = call i32 (i32*, i64, i8*, i64, ...) bitcast (i32 (...)* @memcpy_s to i32 (i32*, i64, i8*, i64, ...)*)(i32* %14, i64 12, i8* null, i64 12), !dbg !34
  %44 = call i32 (i32*, i64, i32*, i64, ...) bitcast (i32 (...)* @memcpy_s to i32 (i32*, i64, i32*, i64, ...)*)(i32* %42, i64 8, i32* %43, i64 16), !dbg !60
  %73 = call i32 (i32*, i64, i32*, i64, ...) bitcast (i32 (...)* @memcpy_s to i32 (i32*, i64, i32*, i64, ...)*)(i32* %71, i64 12, i32* %72, i64 12), !dbg !86
  %102 = call i32 (i32*, i64, i32*, i64, ...) bitcast (i32 (...)* @memcpy_s to i32 (i32*, i64, i32*, i64, ...)*)(i32* %100, i64 12, i32* %101, i64 12), !dbg !112
```
we parse all of them as bv64
```
11:      bv32 r8 <- call memcpy_s(bv64(0), bv64(12), bv64 r7, bv64(12))
30:      bv32 r15 <- call memcpy_s(bv64 r14, bv64(12), bv64(0), bv64(12))
107:     bv32 r44 <- call memcpy_s(bv64 r42, bv64(8), bv64 r43, bv64(16))
184:     bv32 r73 <- call memcpy_s(bv64 r71, bv64(12), bv64 r72, bv64(12))
261:     bv32 r102 <- call memcpy_s(bv64 r100, bv64(12), bv64 r101, bv64(12))
```

hernanponcedeleon: > The count/size parameters are more dangerous. Since your code uses `sizeof(...)` everywhere, you will get a consistent type `size_t` which equals 64 bit on most systems. If you were to put constants that are interpreted as 32 bit ints, then you would see the problem again. 

The type of those is `rsize_t` which AFAIK is defined as `typedef size_t rsize_t;`. Thus, the compiled should take care of the casting even if constant are used. However, for 32 bits archs, we probably get into problems.

> Always upcasting those parameters to 64 bit is probably fine though.

Since we overall assume 64 bits arch, I guess we should use the casts. I will put those two back.

> 
> EDIT: Btw. even in your newest .ll file, the function is not properly recognized.

If you mean that we have `declare dso_local i32 @memcpy_s(...) #3` with no body, I am aware of this. This is what I meant above with `"__STDC_LIB_EXT1__ not defined"` and apparently is not mandatory to have them implemented

> These functions are defined in ISO 9899:2011 Annex K and they are optional to implement. Before using them you must check if __STDC_LIB_EXT1__ is defined.

ThomasHaas: I'm not talking specifically about the body, but the whole signature. Since you had a commit to include `string.h` and to regenerate the llvm file, I thought you found a way to get this function recognized.
It might make sense you just declare the function in the benchmark to get the right signature (not strictly necessary though).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,106 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+        // This parameter has type rsize_t/size_t which we model as types.getArchType(), thus the cast
+        final Expression destszExpr = expressions.makeCast(call.getArguments().get(1), types.getArchType());
+        final Expression src = call.getArguments().get(2);
+        // This parameter has type rsize_t/size_t which we model as types.getArchType(), thus the cast
+        final Expression countExpr = expressions.makeCast(call.getArguments().get(3), types.getArchType());
+
+        if (!(countExpr instanceof IntLiteral countValue)) {
+            final String error = "Cannot handle memcpy_s with dynamic count argument: " + call;
+            throw new UnsupportedOperationException(error);
+        }
```
ThomasHaas, line 1349: This is not going to work. You generate the cast expression and then check for it to be a literal, which should always fail. The reason it works right now is that the cast is recognized as NOOP and optimized away immediately.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,106 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+        Label check1 = EventFactory.newLabel("__memcpy_s_check_1");
+        Label check2 = EventFactory.newLabel("__memcpy_s_check_2");
+        Label success = EventFactory.newLabel("__memcpy_s_success");
+        Label end = EventFactory.newLabel("__memcpy_s_end");
+
+        // dest == NULL or destsz > RSIZE_MAX ----> return error > 0
+        final Expression c1 = expressions.makeOr(destIsNull, invalidDestsz);
+        CondJump skipE1 = EventFactory.newJump(expressions.makeNot(c1), check2);
+        CondJump skipRest1 = EventFactory.newGoto(end);
+        Local retError1 = EventFactory.newLocal(resultRegister, expressions.makeOne((IntegerType)resultRegister.getType()));
```
ThomasHaas, line 1382: I think you should refactor this and avoid the magic constants:
```
Expression errorCodeFail = expressions.makeOne((IntegerType)resultRegister.getType());
Expression errorCodeSuccess = expressions.makeZero(...);
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,106 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+            check1,
+            skipE1,
+            retError1,
+            skipRest1));
+
+        // dest != NULL && destsz <= RSIZE_MAX && (src == NULL || count > destsz || overlap(src, dest)) 
+        // ----> return error > 0 and zero out [dest, dest+destsz)
+        // The first two are guaranteed by not matching c1
+        final Expression c2 = expressions.makeOr(srcIsNull, invalidCount);
+        final Expression c3 = expressions.makeOr(c2, overlap);
```
ThomasHaas, line 1393: I would combine these into one line. The intermediary `c2` is confusing. I intuitively thought that it designates a case like `c1` and `c3` but it was just an intermediary step to construct `c3`.
Also, you might want to rename those, cause `cx` could mean `condition x` or `case x` (if one thinks there are multiple cases to check for).
Maybe just have `c1 = condCase1` and `c3 = caseCond2 (getting rid of the intermediate c2)` and then just add a prefix to the comments: `Case 1: dest == null || ...`, `Case 2: dest != null && ...`, `Case 3: All inputs are valid`
I think that is easier to follow with one condition expression per case, and the fact that there are three cases to consider.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,106 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+            final Register reg = caller.getOrNewRegister("__memcpy_" + i, types.getArchType());
+
+            replacement.addAll(List.of(
+                    EventFactory.newLoad(reg, srcAddr),
+                    EventFactory.newStore(destAddr, reg)
+            ));
+        }
+        replacement.addAll(List.of(
+            retSuccess,
+            end));
```
ThomasHaas, line 1429: Maybe it is just me, but I like to put the last brackets on a separate line:
```Java
       replacement.addAll(List.of(
            retSuccess,
            end
       ));
```
It's more symmetrical and matches better with how the standard curly brackets are placed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,116 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+            check2,
+            skipE2
+        ));
+        for (int i = 0; i < destsz; i++) {
+            final Expression offset = expressions.makeValue(i, types.getArchType());
+            final Expression destAddr = expressions.makeAdd(dest, offset);
+            final Expression value = expressions.makeZero(types.getArchType());
+            replacement.addAll(List.of(
+                EventFactory.newStore(destAddr, value)
+            ));
```
ThomasHaas, line 1414: Just call `replacement.add`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,116 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+        CondJump skipRest2 = EventFactory.newGoto(end);
+        Local retError2 = EventFactory.newLocal(resultRegister, errorCodeFail);
+        replacement.addAll(List.of(
+            check2,
+            skipE2
+        ));
+        for (int i = 0; i < destsz; i++) {
+            final Expression offset = expressions.makeValue(i, types.getArchType());
+            final Expression destAddr = expressions.makeAdd(dest, offset);
+            final Expression value = expressions.makeZero(types.getArchType());
```
ThomasHaas, line 1411: No need to be so indirect and call it `value`, just call it `zero`. It's not a big deal but I think everytime I saw the generated store I intuitively looked for `value` to see what gets stored and everytime it was just zero :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1328,6 +1331,116 @@ private List<Event> inlineMemCpy(FunctionCall call) {
...
+        replacement.addAll(List.of(
+            check1,
+            skipE1,
+            retError1,
+            skipRest1
+        ));
+
+        // Condition 2: dest != NULL && destsz <= RSIZE_MAX && (src == NULL || count > destsz || overlap(src, dest)) 
+        // ----> return error > 0 and zero out [dest, dest+destsz)
+        // The first two are guaranteed by not matching c1
```
ThomasHaas, line 1399: "c1" -> "cond1"

ThomasHaas (approved):
LGTM

PR #689, opened 2024-06-07
Add support for cttz (count trailing zeros) intrinsic

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -240,18 +240,33 @@ public Formula visitIntUnaryExpression(IntUnaryExpr iUn) {
...
+                    // enc = extract(bv, 0, 0) == 1 ? 0 : (extract(bv, 1, 1) == 1 ? 1 : extract ... extract(bv, 63, 63) == 1? 63 : 64)
+                    int bvLength = bvmgr.getLength(bv);
+                    BitvectorFormula bv1 = bvmgr.makeBitvector(1, 1);
+                    BitvectorFormula enc = bvmgr.makeBitvector(bvLength, bvLength);
+                    for(int i = bvLength - 1; i >= 0; i--) {
+                        BitvectorFormula bvi = bvmgr.makeBitvector(bvLength, i);
+                        BitvectorFormula bvbit = bvmgr.extract(bv, i, i);
+                        enc = booleanFormulaManager.ifThenElse(bvmgr.equal(bvbit, bv1), bvi, enc);
+                    }
+                    return enc;
```
ThomasHaas, line 267: You need to count from `i = 0`, no?

hernanponcedeleon: You mean the loop should start from `i=0` and increment? I don't think so (assuming the comment at line 258 is correct).

The difference between comments 258 and 243 is the sequence used from the extract method and the implementation of both intrinsics differ only on that.

If the implementation is indeed wrong, we should definitely add an assertion to `cttz.c` to catch this.

ThomasHaas: Ok, I see that line `BitvectorFormula bvbit = bvmgr.extract(bv, i, i);` makes this correct. On first sight, it looked identical to the iteration of `ctlz` which didn't make sense to me.

ThomasHaas (approved):
LGTM

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
xeren: It looks correct.  The expression is built in reversed test order.  After the last iteration, `enc` first checks the least-significant bit.

PR #690, opened 2024-06-07
Assertion expressions

xeren:
This PR replaces `AbstractAssert` with `Expression`.  It primarily removes duplicated code.  It also changes the handling of user assertions: `PropertyEncoder` now completely controls their encoding (as opposed to `ModelChecker` having to inject them into the spec field first).


dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
```diff
@@ -19,9 +18,12 @@ public class Program {
 
     public enum SourceLanguage { LITMUS, LLVM, SPV }
 
+    public enum SpecificationType { EXISTS, FORALL, NOT_EXISTS, ASSERT }
+
     private String name;
-    private AbstractAssert spec;
-    private AbstractAssert filterSpec; // Acts like "assume" statements, filtering out executions
+    private SpecificationType specificationType = SpecificationType.ASSERT;
```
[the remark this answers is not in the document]
ThomasHaas: It seems so. I kinda dislike the distinction between `forall` and `assert`, since the latter is conceptually an instance of the former. But I guess it still is an improvement of what we had before.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -397,4 +397,18 @@ private static void printWarningIfThreadStartFailed(Program p, EncodingContext e
             }
         }
     }
+
+    private static void printSpecification(StringBuilder sb, Program program) {
+        sb.append("Condition ").append(program.getSpecificationType()).append(" ");
```
ThomasHaas, line 402: print as lower case to match the previous behaviour and the litmus input format.

ThomasHaas (commented):
The output format of the condition has regressed:
- The output does not attach Tid prefixes, resulting in specs that look like `(X == 4) && (X == 5)` but actually mean `(0:X == 4) && (1:X == 5)`.
- A sequence of conditions `A && B && C` is now printed with parenthesis for all subexpressions: `((A && B) && C)` which causes a lot visual overhead for many tests.

Comparison:
```
// Previously:
Condition exists (ok==bv64(1) && x==bv64(1) && y==bv64(1) && z==bv64(2) && 0:bv64 X2==bv64(0) && 0:bv64 X6==bv64(0) && 1:bv64 X1==bv64(1) && 1:bv64 X5==bv64(0) && 2:bv64 X2==bv64(1) && 2:bv64 X5==bv64(0))

// Now (notice the double occurence of X5)
Condition EXISTS ((((((((((ok == bv64(1)) && (x == bv64(1))) && (y == bv64(1))) && (z == bv64(2))) && (bv64 X2 == bv64(0))) && (bv64 X6 == bv64(0))) && (bv64 X1 == bv64(1))) && (bv64 X5 == bv64(0))) && (bv64 X2 == bv64(1))) && (bv64 X5 == bv64(0)))
```
I think we need a smart printer for expressions that avoids parenthesis for nested associative operations and prints function ids in front of registers (configurable).

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionPrinter.java
```diff
@@ -0,0 +1,98 @@
...
+import com.dat3m.dartagnan.expression.misc.ExtractExpr;
+import com.dat3m.dartagnan.expression.misc.GEPExpr;
+import com.dat3m.dartagnan.expression.misc.ITEExpr;
+import com.dat3m.dartagnan.program.Register;
+
+import java.util.stream.Collectors;
+
+public final class ExpressionPrinter implements ExpressionVisitor<String> {
+
+    private final boolean qualified;
```
ThomasHaas, line 17: This should get a more descriptive name: "qualified" does not in the slightest tell you what is going on here.
Just call it `printRegsWithFuncIds` or similar.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionPrinter.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    public String visit(Expression expr) {
+        final ExpressionKind kind = this.kind;
+        this.kind = expr.getKind();
+        final String inner = expr.accept(this);
+        final boolean direct = kind == null || kind == this.kind || expr.getOperands().isEmpty();
+        final String result = direct ? inner : "(" + inner + ")";
+        this.kind = kind;
+        return result;
```
ThomasHaas, line 32: Can one of the variables (probably the local one) be renamed to something like `parentKind`?
I found this very hard to read because the two kinds that get compared have the same names.

Btw. for all non-associative operations, this code creates ambiguous outputs (e.g. for `-` and `\`).
It might not be too bad, since we rarely have those nested directly, but it deserves at least a `TOOD/FIXME` comment.

ThomasHaas (approved):
LGTM

Issue #586, opened 2023-12-02
Issue with GEPToAddition and alignment


ThomasHaas:
Our `GEPToAddition` pass produces wrong offsets that do not respect alignment.
The underlying issue is that `size(struct {a, b}) != size(a) + size(b)` due to alignment requirements.

This results in misinterpretation of [this](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/goblint-regression/06-symbeq_15-list_entry_nr.c?ref_type=heads) benchmark.

ThomasHaas:
Fixed by #691

PR #691, opened 2024-06-10
Fix size and offset computations of types

ThomasHaas:
This PR improves computations of types sizes, alignments, and offsets.  
Previously, we had wrong computations in many cases, however, it only rarely lead to issues in benchmarks that made use of `offset_of` compiler primitives or similar ways to compute pointer offsets without GEP.
As a result, issue #586 should also be resolved.

### DIFF
--- a/benchmarks/c/miscellaneous/offsetof.c
+++ b/benchmarks/c/miscellaneous/offsetof.c
@@ -1,0 +1,28 @@
+#include <stddef.h>
+#include <dat3m.h>
+
+#define container_of(ptr, type, member) ({                      \
+        const typeof( ((type *)0)->member ) *__mptr = (ptr);    \
+        (type *)((char *)__mptr - offsetof(type,member));})
+
+typedef struct {
+    int x; // 4 byte padding
+    struct {
+      long a;
+      char b; // 7 byte padding
+    };
+    int z; // 4 byte padding
+} myStruct_t; // Total size: 32 bytes, because all 4 members are 8-byte aligned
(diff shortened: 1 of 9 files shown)


hernanponcedeleon (commented):
I could not spot anything fundamentally wrong, but there are way too many computations going on. 

Can you add some basic tests with a structure where the compiler adds some padding. Maybe we can even have a second structure will the same members, but add `__attribute__((packed));` to avoid the padding.

ThomasHaas:
> I could not spot anything fundamentally wrong, but there are way too many computations going on.
> 
> Can you add some basic tests with a structure where the compiler adds some padding. 

All LLVM IR structs are padded by default but you won't see the padding in the IR expliclity. So in fact, many non-trivial benchmarks have padded structs and our internal type size computation was almost always wrong (it was always missing trailing padding). The difficult part is making those tests fail, because wrong size/offset computations simply do not matter unless you have `offsetof` somewhere. I will see what I can do.

>Maybe we can even have a second structure will the same members, but add `__attribute__((packed));` to avoid the padding.

IIRC, adding `__attribute__((packed))` will result in IR code that we cannot handle, because the LLVM IR declaration of the struct uses extra brackets `<...>` ([LangRef](https://llvm.org/docs/LangRef.html#t-struct)). We might parse the code, but we certainly ignore the packedness!

ThomasHaas:
I added a test that breaks on development (if run without opt passes) but runs correctly on this PR.

PR #692, opened 2024-06-10
Fix addresses in witnesses

ThomasHaas:
This PR fixes the printing of addresses in witnesses.
Furthermore, the witness will now print every address that belongs to an object as `obj + offset`.
Addresses that do not belong to an object get an "[OOB]" tag for "out-of-bounds".

EDIT: I also added a fix for Refinements inconclusiveness analysis to give proper feedback.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
@@ -47,10 +47,10 @@ import org.sosy_lab.common.configuration.Option;
 import org.sosy_lab.common.configuration.Options;
 import org.sosy_lab.java_smt.api.*;
 
+import java.math.BigInteger;
 import java.text.DecimalFormat;
 import java.util.*;
 import java.util.function.BiPredicate;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static com.dat3m.dartagnan.GlobalSettings.REFINEMENT_GENERATE_GRAPHVIZ_DEBUG_FILES;
@@ -241,7 +241,7 @@ public class RefinementSolver extends ModelChecker {
         if (smtStatus == SMTStatus.UNKNOWN) {
             // Refinement got no result (should not be able to happen), so we cannot proceed further.
             logger.warn("Refinement procedure was inconclusive. Trying to find reason of inconclusiveness.");
-            analyzeInconclusiveness(task, propertyTrace, analysisContext);
+            analyzeInconclusiveness(task, analysisContext, solver.getExecution());
             throw new RuntimeException("Terminated verification due to inconclusiveness (bug?).");
         }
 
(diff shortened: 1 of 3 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ExecutionGraphVisualizer.java
```diff
@@ -188,18 +186,36 @@ private ExecutionGraphVisualizer addThreadPo(Thread thread, ExecutionModel model
...
+            obj = entry.getKey();
+            objAddress = nextObjAddr;
+        }
+
+        if (obj == null) {
+            return address + " [OOB]";
+        } else if (address.equals(objAddress)) {
+            return obj.toString();
+        } else {
+            final boolean isOOB = address.compareTo(objAddress.add(BigInteger.valueOf(obj.size()))) >= 0;
```
hernanponcedeleon, line 206: I don't get this check .... seems to detect some "non-trivial" OOB

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -322,44 +322,38 @@ private void analyzeInconclusiveness(VerificationTask task, RefinementTrace prop
...
+        final Map<BigInteger, Set<EventData>> addr2Events = new HashMap<>();
+        model.getAddressReadsMap().forEach((addr, reads) -> addr2Events.computeIfAbsent(addr, key -> new HashSet<>()).addAll(reads));
+        model.getAddressWritesMap().forEach((addr, writes) -> addr2Events.computeIfAbsent(addr, key -> new HashSet<>()).addAll(writes));
+        model.getAddressInitMap().forEach((addr, init) -> addr2Events.computeIfAbsent(addr, key -> new HashSet<>()).add(init));
+
+        for (Set<EventData> sameLocEvents : addr2Events.values()) {
+            final List<EventData> events = sameLocEvents.stream().sorted().toList();
+
+            for (int i = 0; i < events.size() - 1; i++) {
+                for (int j = 1; j < events.size(); j++) {
```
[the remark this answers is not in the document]
ThomasHaas: Yeah

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ExecutionGraphVisualizer.java
ThomasHaas: This check captures accesses that seem to belong to object `A` but actually go beyond `A`.
There are two cases where this happens:
(1) You have two adjacent memory objects `A` and `B` that have a gap inbetween them (say, due to alignment requirements), and the accessed address is precisely in this gap. The code will first think that the access goes to `A` but this additonal check figures out that the access really goes into the "gap".
(2) The access goes far beyond the last memory object `Z`, i.e., `Z + sizeof(Z) <= addr`. Again, the previous code will think it is an access to `Z` and the additional check sees that it actually goes beyond `Z` and thus is OOB.

PR #693, opened 2024-06-12
Easier handling of nondet choices

ThomasHaas:
This PR adds a new `NonDetChoice` event that essentially plays the role of `__VERIFIER_nondet_XXX` calls.
The difference is that `NonDetChoice` works with arbitrary types and does not rely on intrinsic functions to be declared, which makes it easier to use programatically (e.g., linux CAS compilation can emit this event rather than relying on calling `__VERIFIER_nondet_bool` which must get declared globally beforehand).
`__VERIFIER_nondet_XXX` calls can now be inlined early by translating them all to `NonDetChoice` events.
There is a dedicated pass that finally resolves those choice events by generating program constants.

The major reason why I made this PR is because currently it just sucks to programatically generate non det choices.
While we currently have only a single use-case, namely Linux CAS compilation, I also require this to generate code instrumentation for k-boundedness checks of loops.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/svcomp/NonDetChoice.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/svcomp/NonDetChoice.java
@@ -1,0 +1,46 @@
+package com.dat3m.dartagnan.program.event.lang.svcomp;
+
+import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.event.AbstractEvent;
+import com.dat3m.dartagnan.program.event.Event;
+import com.dat3m.dartagnan.program.event.RegWriter;
+import com.google.common.base.Preconditions;
+
+public class NonDetChoice extends AbstractEvent implements RegWriter {
+
+    protected Register register;
+    protected boolean isSigned;
+
+    public NonDetChoice(Register register, boolean isSigned) {
+        this.register = Preconditions.checkNotNull(register);
+        this.isSigned = isSigned;
+    }
+
+    protected NonDetChoice(NonDetChoice other) {
+        super(other);
+        this.register = other.register;
+        this.isSigned = other.isSigned;
+    }
+
+    public boolean isSigned() { return isSigned; }
+
+    @Override
+    protected String defaultString() {
+        return String.format("%s <- *", register);
+    }
+
+    @Override
+    public Event getCopy() {
+        return new NonDetChoice(this);
+    }
+
+    @Override
+    public Register getResultRegister() {
+        return register;
+    }
+
+    @Override
+    public void setResultRegister(Register reg) {
+        this.register = reg;
+    }
+}
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
@@ -6,7 +6,6 @@ import com.dat3m.dartagnan.expression.ExpressionFactory;
 import com.dat3m.dartagnan.expression.Type;
 import com.dat3m.dartagnan.expression.integers.IntBinaryOp;
 import com.dat3m.dartagnan.expression.integers.IntLiteral;
-import com.dat3m.dartagnan.expression.type.FunctionType;
 import com.dat3m.dartagnan.expression.type.IntegerType;
 import com.dat3m.dartagnan.expression.type.TypeFactory;
 import com.dat3m.dartagnan.program.Function;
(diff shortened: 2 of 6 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/svcomp/NonDetChoice.java
```diff
@@ -0,0 +1,46 @@
...
+        super(other);
+        this.register = other.register;
+        this.isSigned = other.isSigned;
+    }
+
+    public boolean isSigned() { return isSigned; }
+
+    @Override
+    protected String defaultString() {
+        return String.format("%s <- *", register);
```
hernanponcedeleon, line 29: Why not using a more self explaining string, like `non-det()`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1278,38 +1265,41 @@ private void inlineLate(Function function) {
...
+            nonDetType = types.getIntegerType(bits);
+        }
 
-        final NonDetValue value = (NonDetValue) call.getFunction().getProgram().newConstant(types.getIntegerType(bits));
-        value.setIsSigned(signed);
-        return List.of(EventFactory.newLocal(register, expressions.makeCast(value, register.getType(), signed)));
+        final Register nonDetReg = call.getFunction().getOrNewRegister("__r_nondet_" + suffix, nonDetType);
+        return List.of(
+                EventFactory.Svcomp.newSignedNonDetChoice(nonDetReg, signed),
+                EventFactory.newLocal(result, expressions.makeCast(nonDetReg, result.getType(), signed))
```
hernanponcedeleon, line 1301: Do we really need this cast?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveNonDetChoices.java
```diff
@@ -0,0 +1,46 @@
...
+import com.dat3m.dartagnan.expression.Type;
+import com.dat3m.dartagnan.expression.processing.ExpressionInspector;
+import com.dat3m.dartagnan.program.Function;
+import com.dat3m.dartagnan.program.Program;
+import com.dat3m.dartagnan.program.event.EventFactory;
+import com.dat3m.dartagnan.program.event.core.Local;
+import com.dat3m.dartagnan.program.event.lang.svcomp.NonDetChoice;
+import com.dat3m.dartagnan.program.misc.NonDetValue;
+
+public class ResolveNonDetChoices implements ProgramProcessor {
```
hernanponcedeleon, line 13: Why is this a program processor if we just propagate the call to each function?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorLKMM.java
```diff
@@ -69,8 +65,8 @@ public List<Event> visitLKMMCmpXchg(LKMMCmpXchg e) {
         loadFail.addTags(Tag.RMW);
         Load loadSuccess;
         return eventSequence(
-                newValueFunctionCall(havocRegister,nondetBoolFunction,List.of()),
-                newJump(expressions.makeBooleanCast(havocRegister), success),
+                EventFactory.Svcomp.newNonDetChoice(havocRegister),
```
hernanponcedeleon, line 68: Probably it does not make much sense to keep having this as an svcomp event

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/svcomp/NonDetChoice.java
ThomasHaas: I didn't wanna use anything with parenthesis `()` cause it looks like a function call (and this is not call). Without the parenthesis, `r <- non_det` looks the same as a standard register assignment, so it was not an option either.
So `*` ended up being the most clear choice while also being standard notation.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: Definitely for nondeterministic booleans, because the register will be `bv8` instead of `bv1/bool`. Apart from that case, I don't think it is needed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveNonDetChoices.java
ThomasHaas: Because on hand it only runs on threads (it is almost always wrong to run it on functions) and one the other hand it affects the program by generating new program constants.
In particular, the pass could never run on an isolated function that does not belong to some program.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorLKMM.java
ThomasHaas: It doesn't make sense, but that is the current state for all our special events, including spinloops and loop bounds, so I kept it at that. In fact, some might argue that it does more closely belong to SVCOMP than the other two events, because it actually finds usage there.

hernanponcedeleon: Yeah, all those events initially came from `VERIFIER_X` annotations and only later we end up using them for other purposes or having them automatically added. We might move them in the future, but this is no blocker for this PR.

PR #695, opened 2024-06-13
Model locks as spinloops to avoid wrong liveness bugs

hernanponcedeleon:
Our current mocking of locks is unsound for liveness.

This PRs fixes this by modeling them as spinlocks.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorBase.java
```diff
@@ -57,33 +48,35 @@ public List<Event> visitInitLock(InitLock e) {
...
         Expression address = e.getAddress();
         String mo = e.getMo();
 
         Load rmwLoad = newRMWLoadWithMo(dummy, address, mo);
+        Label spinLoopHead = newLabel("__spinloop_head");
         return eventSequence(
+                spinLoopHead,
                 rmwLoad,
-                newTerminator(expressions.makeNEQ(dummy, one)),
+                newJump(expressions.makeNot(dummy), spinLoopHead),
```
ThomasHaas, line 79: Why does unlock even spin? Unlocking an already unlocked mutex is UB so it would make more sense to put an assertion to check that. Alternatively, if we do not want to check for UB, the load operation can go completely

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorLKMM.java
```diff
@@ -206,15 +207,18 @@ public List<Event> visitLKMMXchg(LKMMXchg e) {
 
     @Override
     public List<Event> visitLKMMLock(LKMMLock e) {
-        Register dummy = e.getFunction().newRegister(e.getAccessType());
-        Expression nonzeroDummy = expressions.makeBooleanCast(dummy);
+        Type type = types.getBooleanType();
+        Register dummy = e.getFunction().newRegister(type);
```
ThomasHaas, line 211: Same as for the other locking primitives: you load a bool but store integers which is dangerous.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
The failure is not really related to the witness encoding. We even return PASS when verifying `lazy-01.c` when the program is really buggy. I still don't understand why.

ThomasHaas:
> EDIT: shouldn't we revert the order of the jump and the store to actually model an exchange inside a loop?

You don't want to do an Xchg but rather a CAS. And I think it is totally fine to move the store part to outside the loop.

> EDIT2: taking a look to other compilation visitors ... can we simply force the lock-acquire to succeed with an assume? I don't think this would interfere with liveness.

No, because that would make all executions that try to acquire an already acquired lock trivially safe. Also, you would not be able to detect deadlocks involving locks.
Assumes are only fine for litmus code.

hernanponcedeleon:
> You don't want to do an Xchg but rather a CAS. And I think it is totally fine to move the store part to outside the loop.

Why? Isn't this also a valid implementation?
```
while(xchg(x,1) != 0){} // lock
store(x, 0) // unlock
```

> No, because that would make all executions that try to acquire an already acquired lock trivially safe. Also, you would not be able to detect deadlocks involving locks. Assumes are only fine for litmus code.

Then we have a problem with other compilations, like [arm8](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java#L59).

ThomasHaas:
> > You don't want to do an Xchg but rather a CAS. And I think it is totally fine to move the store part to outside the loop.
> 
> Why? Isn't this also a valid implementation?
> 
> ```
> while(xchg(x,1) != 0){} // lock
> store(x, 0) // unlock
> ```

It is, but do you really want to use a loop implementation that we cannot detect as spinloop? Just use the easiest one, i.e., CAS loops. If you even put the store outside of the loop, we have a static spinloop and don't unroll the loop at all.

> > No, because that would make all executions that try to acquire an already acquired lock trivially safe. Also, you would not be able to detect deadlocks involving locks. Assumes are only fine for litmus code.
> 
> Then we have a problem with other compilations, like [arm8](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java#L59).

I'm not surprised.

hernanponcedeleon:
The compilation of locks can introduce loops now. IIRC, some of the inline assembly from the linux kernel also uses loops and we used to manually unroll them. At some point I would like to change this.

Because of the above, would it make sense to move the `LoopFormVerification` pass after compilation?

ThomasHaas:
So you went with boolean registers again? This is dangerous because it relies on implicit conversions (the initial value in memory is an integer!), especially if the lock variables are statically initialized. In this case it might not even be true that value "unlocked == 0". Indeed, this is not the case for statically initialized rwlocks and this is relevant for SVCOMP!
Not sure if `pthread_mutex_t` can ever have a non-null initial value though.

 Also, many pthread functions are implemented directly during inlining (`Intrinsics.java`) and some of them, in particular, `pthread_mutex_trylock` still assume an integer at the location. Implicit conversion will do its work, but it is still fishy.
 That being said, it should be easier to just implement lock/unlock/init also in `Intrinsics` with LLVM atomics and then let the default compilation scheme translate it for hardware models (a disadvantage is that you would need to put a proper CAS inside the loop, making it a dynamic spinloop, but you could generate a loop bound annotation to force only a single unrolling).
Then the only use of dedicated lock events is for litmus tests and you can (probably) use boolean values and assumes.

ThomasHaas:
> Because of the above, would it make sense to move the `LoopFormVerification` pass after compilation?

Yes, the pass should come after all loops in the program have been generated because it also annotates the loop headers so that they can still be detected after unrolling. IIRC, this is currently only used to generate context information.

hernanponcedeleon:
> So you went with boolean registers again? This is dangerous because it relies on implicit conversions (the initial value in memory is an integer!), especially if the lock variables are statically initialized. In this case it might not even be true that "unlocked == 0". Indeed, this is not the case for statically initialized rwlocks and this is relevant for SVCOMP! Not sure if `pthread_mutex_t` can ever have a non-null initial value though.
> 

Aghhh I had in my mind your comment "you use boolean for acquire, but int for release" and I forgot about this part of the comment. I will probably go will all ints, but keeping lock events (see below).

> Also, many pthread functions are implemented directly during inlining (`Intrinsics.java`) and some of them, in particular, `pthread_mutex_trylock` still assume an integer at the location. Implicit conversion will do its work, but it is still fishy. That being said, it should be easier to just implement lock/unlock/init also in `Intrinsics` with LLVM atomics and then let the default compilation scheme translate it for hardware models (a disadvantage is that you would need to put a proper CAS inside the loop, making it a dynamic spinloop, but you could generate a loop bound annotation to force only a single unrolling). Then the only use of dedicated lock events is for litmus tests and you can (probably) use boolean values and assumes.
> 

I tried to do this (the code is in `lock-mocking-v2`), but I think this is not trivial. The mutex type is actually a union type (which even IIRC is architecture dependent). 

Because of this (I suspect), I get liveness violations due to mixed-size accesses for correct programs like this
```
#include <pthread.h>

pthread_mutex_t mutex;

void *thread(void *unused)
{
    pthread_mutex_lock(&mutex);
    pthread_mutex_unlock(&mutex);
    return NULL;
}

int main()
{
    pthread_t t1, t2;

    pthread_create(&t1, NULL, thread, NULL);
    pthread_create(&t2, NULL, thread, NULL);

    pthread_join(t1, NULL);
    pthread_join(t2, NULL);

    return 0;
}

```

ThomasHaas:
I don't see how lock events are related to what you say below. Either way, you should end up with almost the same code after processing.

    Also, many pthread functions are implemented directly during inlining (Intrinsics.java) and some of them, in particular, pthread_mutex_trylock still assume an integer at the location. Implicit conversion will do its work, but it is still fishy. That being said, it should be easier to just implement lock/unlock/init also in Intrinsics with LLVM atomics and then let the default compilation scheme translate it for hardware models (a disadvantage is that you would need to put a proper CAS inside the loop, making it a dynamic spinloop, but you could generate a loop bound annotation to force only a single unrolling). Then the only use of dedicated lock events is for litmus tests and you can (probably) use boolean values and assumes.

    I tried to do this (the code is in lock-mocking-v2), but I think this is not trivial. The mutex type is actually a union type (which even IIRC is architecture dependent).

The mutex type is a union, yes, but this only affects the initial value of the mutex location and should only make problems if that value is non-zero (e.g., due to a static initializer).
As long as all accesses are over pthread_mutex_XYZ calls (and the initial value is zero), we should be fine if we handle them consistently, even with boolean types (one needs to double-check the encoding generated for reading bools as ints though).
So your example should work correctly, if you don't mess up the implementation.

hernanponcedeleon:
I put back the int types and I get the same problem as in the other branch. The mutex is globally defined, so it is initialized to zero. However the read from the acquire gets value 9223372036854775808. If I initialized the mutex with `pthread_mutex_init` in `main`, all is good. I could not yet find out if accessing the lock without explicitly calling `pthread_mutex_init` is UB.

ThomasHaas:
> I put back the int types and I get the same problem as in the other branch. The mutex is globally defined, so it is initialized to zero. However the read from the acquire gets value 9223372036854775808. If I initialized the mutex with `pthread_mutex_init` in `main`, all is good. I could not yet find out if accessing the lock without explicitly calling `pthread_mutex_init` is UB.

It is UB. You need to use at least the `PTHREAD_MUTEX_INITIALIZER` macro for static initialization (if you don't call `pthread_mutex_init)`. But that is interesting, can you show me the .ll code?

hernanponcedeleon:
```
ponce@A2204107822HW:~/git/Dat3M$ cat benchmarks/c/miscellaneous/mutex.c 
#include <pthread.h>

pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;

void *thread(void *unused)
{
    pthread_mutex_lock(&mutex);
    pthread_mutex_unlock(&mutex);
    return NULL;
}

int main()
{
    pthread_t t1, t2;

    pthread_create(&t1, NULL, thread, NULL);
    pthread_create(&t2, NULL, thread, NULL);

    pthread_join(t1, NULL);
    pthread_join(t2, NULL);

    return 0;
}
ponce@A2204107822HW:~/git/Dat3M$ java -jar dartagnan/target/dartagnan.jar cat/svcomp.cat benchmarks/c/miscellaneous/mutex.c --witness=png
[14.06.2024] 16:33:06 [INFO] GitInfo.logGitInfo - Git branch: lock-mocking-v2
[14.06.2024] 16:33:06 [INFO] GitInfo.logGitInfo - Git commit ID: 6a86ab798f51a494ffd5f286fac7a0ea72ff0a01
...

[14.06.2024] 16:33:07 [INFO] RefinementSolver.runInternal - Verification finished with result FAIL
============ Liveness violation found ============
        E28:    @mutex.c#7
        E54:    @mutex.c#7
FAIL
Total verification time: 1.129 secs
```
[mutex-opt.zip](https://github.com/user-attachments/files/15839487/mutex-opt.zip)

ThomasHaas:
I see, on your machine the first entry is `i32` whereas mine is `i64`. This causes the `i64` read to only read the lower half as 0 and fill up the rest. 

```
    public BooleanFormula equal(Formula left, Formula right) {
        //FIXME: We should avoid the automatic conversion, or standardize what conversion we expect.
        // For example, when encoding rf(w, r), we always want to convert the store value type to the read value type
        // for otherwise, we might get an under-constrained value for val(r) (e.g., its upper bits might be unconstrained,
        // if we truncate it to smaller size of the store value).
        if (left instanceof IntegerFormula l) {
```

ThomasHaas (approved):
LGTM

PR #697, opened 2024-06-13
Better mem2reg

ThomasHaas:
This PR (1) adds a new `RemoveDeadNullChecks` pass which is a very simple pass that eliminates null checks and (2) improves `Mem2Reg` to also work on complex allocations (arrays and aggregates).
The `RemoveDeadNullChecks` pass is necessary to enable `Mem2Reg` in some circumstances: `Mem2Reg` only works if the allocated memory is solely used as an address in Loads/Stores, but NULL checks are non-address usages.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
@@ -8,6 +8,7 @@ import com.dat3m.dartagnan.expression.misc.ITEExpr;
 import com.dat3m.dartagnan.expression.processing.ExprTransformer;
 import com.dat3m.dartagnan.program.Function;
 import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.analysis.LoopAnalysis;
 import com.dat3m.dartagnan.program.event.Event;
 import com.dat3m.dartagnan.program.event.EventVisitor;
 import com.dat3m.dartagnan.program.event.RegReader;
@@ -15,9 +16,12 @@ import com.dat3m.dartagnan.program.event.RegWriter;
 import com.dat3m.dartagnan.program.event.core.Alloc;
 import com.dat3m.dartagnan.program.event.core.Local;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.math.BigInteger;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import static com.dat3m.dartagnan.expression.integers.IntBinaryOp.*;
@@ -33,9 +37,12 @@ import static com.dat3m.dartagnan.expression.integers.IntCmpOp.*;
       - Some operations (like addition) preserve positiveness.
 
     TODO: The pass is very naive: it globally assigns a signedness to registers rather than per program point.
+    TODO 2: This pass only runs correctly on unrolled code and will skip functions with loops.
  */
 public class RemoveDeadNullChecks implements FunctionProcessor {
 
+    private final static Logger logger = LogManager.getLogger(RemoveDeadNullChecks.class)
+;
     private RemoveDeadNullChecks() { }
 
     public static RemoveDeadNullChecks newInstance() { return new RemoveDeadNullChecks(); }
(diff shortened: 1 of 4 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
```diff
@@ -0,0 +1,167 @@
...
+
+import static com.dat3m.dartagnan.expression.integers.IntBinaryOp.*;
+import static com.dat3m.dartagnan.expression.integers.IntCmpOp.*;
+
+/*
+    This pass tries to remove unnecessary NULL checks in a very simple manner:
+    it tries to figure out the signedness of expressions and if they are always positive, then they cannot be NULL.
+
+    The signedness is determined as follows:
+      - Registers written to by Allocs as well as MemoryObject expressions are always positive (allocations always succeed!)
```
hernanponcedeleon, line 31: Should it be "we assume allocation to always succeed"

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private enum Sign {
+        UNKNOWN,
+        NON_NEG,
+        POS;
+
+        static Sign meet(Sign a, Sign b) {
+            return a.compareTo(b) >= 0 ? b : a;
+        }
+    }
```
hernanponcedeleon, line 51: This means the order of the enum is important, right? If so, we should document this

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
```diff
@@ -0,0 +1,167 @@
...
+        public Expression visitIntCmpExpression(IntCmpExpr cmp) {
+            if (cmp.getRight() instanceof IntLiteral lit && lit.isZero() && cmp.getLeft().accept(signChecker) == Sign.POS) {
+                // Simplify "expr cop 0" if <expr> is known to be positive.
+                final IntCmpOp op = cmp.getKind();
+                final boolean valueOnNonNull = op == NEQ || op == GT || op == GTE || op == UGT || op == UGTE;
+                return ExpressionFactory.getInstance().makeValue(valueOnNonNull);
+            }
+            return cmp;
+        }
+    }
```
hernanponcedeleon, line 84: Would it make sense to have a dual pattern for computing expression that are definitely negative?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
```diff
@@ -0,0 +1,167 @@
...
+        public Sign visitIntBinaryExpression(IntBinaryExpr expr) {
+            final Sign leftSign = expr.getLeft().accept(this);
+            final Sign rightSign = expr.getRight().accept(this);
+            if (leftSign == Sign.UNKNOWN || rightSign == Sign.UNKNOWN) {
+                return Sign.UNKNOWN;
+            }
+
+            // --- Both subexpressions are (at least) non-negative ---
+            final IntBinaryOp op = expr.getKind();
+            if (leftSign == Sign.POS && op == ADD) {
```
hernanponcedeleon, line 151: Should this be `(leftSign == Sign.POS || rightSign == Sign.POS) && op == ADD`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
ThomasHaas: No, allocation events always succeed. This is not directly related to `malloc` which could fail (though we assume it does not).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
ThomasHaas: Possible but probably useless.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadNullChecks.java
ThomasHaas: I guess that would be slightly more accurate

PR #698, opened 2024-06-14
Fixed copying of event tags

ThomasHaas:
This PR fixes a problem we had for a very long: when copying events, we did not copy the set of tags but instead shared the same set instance. This could lead to very subtle and hard-to-debug bugs, so we now make proper copies instead.

Memory-wise, the copies can be wasteful if they all end up storing the exact same set of tags (which is common), so copy-on-write sets or persistent sets might offer a more memory-efficient solution.
I don't use them yet, but I added a `TagSet` class to abstract away from the underlying implementation and use sorted lists for now (they are more efficient than hash sets).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/AbstractEvent.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/AbstractEvent.java
@@ -16,7 +16,7 @@ import java.util.*;
 public abstract class AbstractEvent implements Event {
 
     private final MetadataMap metadataMap = new MetadataMap();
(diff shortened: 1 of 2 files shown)


xeren:
`java.util.concurrent.CopyOnWriteArraySet` might fit into this role.

ThomasHaas:
Yes, it's a possibility, because our sets are quite small. For larger sets I would try JavaSMT's persistent map.
Either way, it is currently no bottleneck AFAIK and easy to exchange later (if needed).

hernanponcedeleon (approved):

PR #699, opened 2024-07-01
Avoid exception in SyntacticContextAnalysis

hernanponcedeleon:
The current analysis throws an exception if function calls or loops do not satisfy certain well-formedness properties.

Since the analysis is not used for verification, but rather to give useful feedback to the user, whenever those conditions are not satisfied, we should give up the analysis instead of not giving the verification answer.


ThomasHaas (approved):
LGTM.

PR #702, opened 2024-07-17
Refactor relation analysis interface

DIvanov503:
Prepare the RA for having multiple implementations. Main changes:

- Make `RelationAnalysis` an interface, put its implementation into `NativeRelationAnalysis`
- Collect RA pieces from other components into `NativeRelationAnalysis`
- - Arrange knowledge closure computation from `Constraint's as visitors
- - Move a piece from the encoder in a method
- Minor refactors

Points I'm not so sure about:

- Naming of the new visitors and the method
- The use of duplicated options


dartagnan/src/main/java/com/dat3m/dartagnan/configuration/RelationMethod.java
```diff
@@ -0,0 +1,19 @@
+package com.dat3m.dartagnan.configuration;
+
+import java.util.Arrays;
+
+public enum RelationMethod implements OptionInterface {
+	NATIVE;
```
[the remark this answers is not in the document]
ThomasHaas: I thought the Datalog implementation might only do the bottom-up RA, no? If so, you either have to be able to disable XRA or do XRA externally on top of the Datalog RA.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -0,0 +1,1669 @@
...
+
+    @Option(name = ENABLE_RELATION_ANALYSIS,
+            description = "Derived relations of the memory model ",
+            secure = true)
+    private boolean enable = true;
+
+    @Option(name = ENABLE_MUST_SETS,
+            description = "Tracks relationships of the memory model, which exist in all execution that execute the two participating events.",
+            secure = true)
+    private boolean enableMustSets = true;
```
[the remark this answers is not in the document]
ThomasHaas: I don't think we need any fine-grained control. I think we should have at most(!)
- NONE/DISABLED (which we probably never use)
- STANDARD/NATIVE (what we do right now)
- DATALOG (alternative implementation)

And then possibly an extra option to enable/disable XRA (enabled by default).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/EmptyRelationAnalysis.java
```diff
@@ -0,0 +1,82 @@
...
+import com.dat3m.dartagnan.wmm.utils.EventGraph;
+import org.sosy_lab.common.configuration.Configuration;
+import org.sosy_lab.common.configuration.InvalidConfigurationException;
+
+import java.util.*;
+
+import static com.dat3m.dartagnan.program.event.Tag.*;
+import static java.util.stream.Collectors.toSet;
+
+public class EmptyRelationAnalysis extends NativeRelationAnalysis {
```
hernanponcedeleon, line 20: The "empty" in the name tell me nothing. We need a more meaningful name. What about "coarse"?

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/RelationMethod.java
```diff
@@ -3,15 +3,15 @@
 import java.util.Arrays;
 
 public enum RelationMethod implements OptionInterface {
```
hernanponcedeleon, line 5: "RelationAnalysisMethod" is probably a better name

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -648,7 +648,8 @@ public Void visitSyncBarrier(SyncBar syncBar) {
 
         @Override
         public Void visitSyncFence(SyncFence syncFenceDef) {
-            final Relation syncFence = syncFenceDef.getDefinedRelation();;
+            final Relation syncFence = syncFenceDef.getDefinedRelation();
+            ;
```
hernanponcedeleon, line 652: This `,` is useless, isn't it?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -649,9 +649,8 @@ public Void visitSyncBarrier(SyncBar syncBar) {
         @Override
         public Void visitSyncFence(SyncFence syncFenceDef) {
             final Relation syncFence = syncFenceDef.getDefinedRelation();
-            ;
             final boolean idl = !context.useSATEncoding;
-            final String relName = syncFence.getName().get(); // syncFence is base, it always has a name
+            final String relName = syncFence.getName().orElseThrow(); // syncFence is base, it always has a name
```
hernanponcedeleon, line 653: Any special reason for the change? the comment explains that `get()` is always guaranteed to succeed.

DIvanov503: The reason is to suppress the warning by the IDE and marginally improve readability. `orElseThrow()` is equivalent to `get()` but makes the intent more explicit

DIvanov503: See [this](https://bugs.openjdk.org/browse/JDK-8140281)

PR #706, opened 2024-07-29
Fix an incorrect term pattern in SyncWith

DIvanov503:


hernanponcedeleon:
Thanks for the fix @DIvanov503

PR #708, opened 2024-08-07
Datalog backed relation analysis

DIvanov503:


hernanponcedeleon:
FYI @DIvanov503 https://souffle-lang.github.io/install

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -82,7 +82,7 @@ private Dartagnan(Configuration config) throws InvalidConfigurationException {
 
     public static void main(String[] args) throws Exception {
 
-        initGitInfo();
+//        initGitInfo();
```
DIvanov503, line 85:
```suggestion
        initGitInfo();
```

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
```diff
@@ -39,6 +39,8 @@ public static void configure(Configuration config) throws InvalidConfigurationEx
 
     private static final boolean USE_TEST_PATH = isJUnitTest();
 
+    public static final boolean CHECK_DATALOG = true; // Enables comparing basic RA in datalog against native
```
DIvanov503, line 42:
```suggestion
    public static final boolean CHECK_DATALOG = false; // Enables comparing basic RA in datalog against native
```

hernanponcedeleon:
@DIvanov503 I killed the pipeline after 14hs. There seems to be a problem with `linux-kernel.cat`. I still don't get why the timeout from `AbstractLitmusTest` is not triggering.

I tried a few examples manually and they all finished relatively fast. I tried specially the ones having large number of events or using RCU features which I thought could make the analysis harder. I guess we need a script to run examples one by one in the console to find out which test(s) are not finishing.

ThomasHaas:
I wouldn't try to fix the native RA in this PR (especially if the source of the problem is unclear). If you have a proper fix, put it into a separate PR.

DIvanov503:
I agree but I

- want to know that my fix works before merging
- don't want to write tests for my fix

Hence I want the pipeline to pass using the datalog RA for checking if both implementations compute the same results. Once it's done and I'm convinced about the fix, it can go to a different PR.

hernanponcedeleon:
As discussed with @DIvanov503, this was supersed by #743, this cloosing.

PR #713, opened 2024-08-13
Fix Printer OutOfBoundsException

natgavrilenko:


dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
```diff
@@ -113,7 +113,9 @@ private void appendEvent(Event event){
             if(!(event instanceof Label)) {
                 result.append("   ");
             }
-            result.append(padding, idSb.length(), padding.length());
+            if (idSb.length() < padding.length()) {
```
hernanponcedeleon, line 116: What happens if this condition is false? Do we just miss the padding (i.e., the output good look weird)?

Do you have some example where the printer is already failing? No need to add a test, I just want to see how it looks visually with the fixed code.

natgavrilenko: Example: any program with 6-digit event id. If this condition is false, it will just miss the initial padding. 

In my opinion, the program still looks reasonably fine:
278070:260:l86.l88:
278071:   === Calling hlist_del ===
278072:   === Calling __hlist_del ===
278073:   bv64 262:r7 = load(*bv64 260:r124)

versus the "small" event numbers:
[0] thread bv32 main(bv32 r0, bv64 r1)
0:       ThreadStart
1:       bv64 r7 <- stackalloc({ bv64, bv64 }, bv64(1))

PR #716, opened 2024-08-15
Implement Observer Threads

AndrejFinki:


ThomasHaas:
Is there a need to open PRs? I think you can just merge into your branch directly without creating PRs.

PR #721, opened 2024-08-28
Smtlib2

hernanponcedeleon:
This PR creates a wrapper around the `ProverEnvironment` which allows to dump the encoding to an smtlib2 file.
Attached are the encodings of ticketlock both with eager and lazy methods, and this is the output of z3
```
> z3 output/ticketlock-eager.smt2 
unsat
unsat
> z3 output/ticketlock-lazy.smt2 
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
sat
unsat
unsat
```
[output.zip](https://github.com/user-attachments/files/16785768/output.zip)

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
@@ -11,7 +11,10 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+
 import org.sosy_lab.java_smt.api.*;
 import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
 
@@ -21,12 +24,14 @@ public class ProverWithTracker implements AutoCloseable {
 
     private final FormulaManager fmgr;
     private final ProverEnvironment prover;
-    private final String fileName; 
+    private final String fileName;
+    private final Set<String> declarations;
 
     public ProverWithTracker(SolverContext ctx, String fileName, ProverOptions... options) {
         this.fmgr = ctx.getFormulaManager();
         this.prover = ctx.newProverEnvironment(options);
         this.fileName = fileName;
+        this.declarations = new HashSet<>();
         init();
     }
 
@@ -48,37 +53,6 @@ public class ProverWithTracker implements AutoCloseable {
     public void close() throws Exception {
         removeDuplicatedDeclarations(fileName);
         prover.close();
-    }
-
-    // Each call to fmgr.dumpFormula(f) defines all variables used in f.
-    // We might end up with several definitions of the declarations.
-    // This might not be the most performant implementation, but duming the 
-    // smtlib2 file is intended only for debugging or trying to understand 
-    // how to optimize the encoding, thus performance is not an issue. 
-    private void removeDuplicatedDeclarations(String fileName) {
-        try {
-
-            BufferedReader reader = new BufferedReader(new FileReader(fileName));
-            List<String> newLines = new ArrayList<>();
-            String line;
-            while ((line = reader.readLine()) != null) {
-                // We only skip repeated declarations
-                if (!line.contains("declare-fun") || !newLines.contains(line)) {
-                    newLines.add(line);
-                }
-            }
-            reader.close();
-
-            BufferedWriter writer = new BufferedWriter(new FileWriter(fileName));
-            for (String newLine : newLines) {
-                writer.write(newLine);
-                writer.newLine();
-            }
-            writer.close();
-
-        } catch (IOException e) {
-            e.printStackTrace();
-        }
     }
 
     public void addConstraint(BooleanFormula f) throws InterruptedException {
@@ -135,10 +109,21 @@ public class ProverWithTracker implements AutoCloseable {
         try {
             writer = new FileWriter(file, true);
             PrintWriter printer = new PrintWriter(writer);
-            printer.append(content);
+            printer.append(removeDuplicatedDeclarations(content));
             printer.close();
         } catch (IOException e) {
             e.printStackTrace();
         }
+    }
+
+    private String removeDuplicatedDeclarations(String content) {
+        String output = "";
+        for(String line : content.split("\n")) {
+            if(line.contains("declare-fun") && !declarations.add(line)) {
+                continue;
+            }
+            output += line + "\n";
+        }
+        return output;
     }
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
@@ -38,35 +38,39 @@ public class ProverWithTracker implements AutoCloseable {
     }
     
     private void init() {
-        StringBuilder description = new StringBuilder();
-        LocalDate currentDate = LocalDate.now();
-        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("YYYY-MM-DD");
-        description.append("Generated on: " + currentDate.format(formatter) + "\n");
-        description.append("Generator: Dartagnan\n");
-        description.append("Application: Bounded model checking for weak memory models\n");
-        description.append("Publications: \n" +
-            "- Hernán Ponce de León, Florian Furbach, Keijo Heljanko, " +
-            "Roland Meyer: Dartagnan: Bounded Model Checking for Weak Memory Models " +
-            "(Competition Contribution). TACAS (2) 2020: 378-382\n" +
-            "- Thomas Haas, Roland Meyer, Hernán Ponce de León: " +
-            "CAAT: consistency as a theory. Proc. ACM Program. Lang. 6(OOPSLA2): 114-144 (2022)"
-        );
-        try {
-            Files.deleteIfExists(Paths.get(fileName));
-            write("(set-info :smt-lib-version 2.6)\n");
-            write("(set-logic ALL)\n");
-            write("(set-info :category \"industrial\")\n");
-            write("(set-info :source |\n" + description + "\n|)\n");
-            write("(set-info :license \"https://creativecommons.org/licenses/by/4.0/\")\n");
-        } catch (IOException e) {
-            e.printStackTrace();
+        if(dump()) {
+            StringBuilder description = new StringBuilder();
+            LocalDate currentDate = LocalDate.now();
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("YYYY-MM-DD");
+            description.append("Generated on: " + currentDate.format(formatter) + "\n");
+            description.append("Generator: Dartagnan\n");
+            description.append("Application: Bounded model checking for weak memory models\n");
+            description.append("Publications: \n" +
+                "- Hernán Ponce de León, Florian Furbach, Keijo Heljanko, " +
+                "Roland Meyer: Dartagnan: Bounded Model Checking for Weak Memory Models " +
+                "(Competition Contribution). TACAS (2) 2020: 378-382\n" +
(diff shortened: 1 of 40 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -0,0 +1,144 @@
...
+        removeDuplicatedDeclarations(fileName);
+        prover.close();
+    }
+
+    // Each call to fmgr.dumpFormula(f) defines all variables used in f.
+    // We might end up with several definitions of the declarations.
+    // This might not be the most performant implementation, but duming the 
+    // smtlib2 file is intended only for debugging or trying to understand 
+    // how to optimize the encoding, thus performance is not an issue. 
+    private void removeDuplicatedDeclarations(String fileName) {
```
ThomasHaas, line 58: Is the file only valid if it contains no duplicates or is this just to reduce the file?

hernanponcedeleon: At least both z3 and cvc5 (the solvers I tried) complain.

ThomasHaas: I see. But then I believe you should follow a different approach that only generates the file at the very end (there is limited use for partial/invalid intermediate files).
Also, the way you remove duplicates is error-prone: it will fail if symbols are popped and then redeclared (unlikely to happen in the way we use the solver though).
I think you could just track the sequence of prover calls (push, pop, check, assert(f)) in a list and in the end generate the SMT-lib from that.

hernanponcedeleon: What is the advantage of writing the file only at the end? Keeping track of everything results is a more "complex" implementation (my first version of the code was doing this) and I don't see the point of this without a clear advantage.

> there is limited use for partial/invalid intermediate files

I disagree about this. The whole reason I implemented this is because I have been waiting for the end of the same verification run for the last two weeks. There are caat iterations that take over 12 hs and I could simply take the intermediate file and check if some other solver is faster (i.e., _is the encoding really hard or just this particular tool struggles about this?_).

ThomasHaas: If you can use the intermediate file it means your smtlib is not invalid, no? That's why I was asking. Or do you run a script to make it valid before testing another solver?

hernanponcedeleon: Ok, you are right that the scenario I described will not work out of the box with the current implementation. 

If I want to have this, I could either call `removeDuplicatedDeclarations` at the end of `write` (decreases performance) or keep track of all declared variables (increases memory foot print). Since this functionality is mostly for debugging, I don't mind the performance hit and I would be inclined to this solution.

ThomasHaas: I don' think tracking variables is really expensive. You can traverse formulas, so technically you only need a single formula (a single pointer) that contains all variables and then just traverse it. Anyway, a correct implementation does not get around tracking the stack of push/pop operations: you need to do it at least within the duplication deletion phase.
Well, or you just hope that the interaction we produce with the solver never produces a history on which your algorithm breaks (which might be a fair assumption).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -135,10 +109,21 @@ private void write(String content) {
         try {
             writer = new FileWriter(file, true);
             PrintWriter printer = new PrintWriter(writer);
-            printer.append(content);
+            printer.append(removeDuplicatedDeclarations(content));
```
ThomasHaas, line 112: This cannot work, can it? A single formula won't contain duplicates, it will contain duplicates wrt. the existing file content. I I think you need to traverse the whole file every single time you append something.

hernanponcedeleon: Notice that I redefined the `removeDuplicatedDeclarations`. Now the class keep tracks of declarations, and each single formula will check if the variables that it is trying to define were already defined by some previous one, and if so, it skips them.

Attached are the smtlib2 files generated from the last commit. They seem correct to me (at least the overall structure makes sense).

[output.zip](https://github.com/user-attachments/files/16796300/output.zip)

ThomasHaas: Oh sry, I didn't see this. Then it is fine (though it still has the same theoretical issue I mentioned already).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private String removeDuplicatedDeclarations(String content) {
+        String output = "";
+        for(String line : content.split("\n")) {
+            if(line.contains("declare-fun") && !declarations.add(line)) {
+                continue;
+            }
+            output += line + "\n";
+        }
+        return output;
```
ThomasHaas, line 122: Use `StringBuilder`. You could probably even return the string builder and append it into the file without constructing a string at all.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
ThomasHaas: No. If you have
```
push
declare x
pop
declare x
assert (P(x))
```
then both declaration are valid/necessary, because the former gets popped before the redeclaration. Your procedure would avoid printing the second declaration, generating an invalid smtlib file.
I don't think this happens in our current usage of the solver, so you get away with "ignoring the stack structure".

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
```diff
@@ -55,21 +55,27 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
         SymmetryEncoder symmetryEncoder = SymmetryEncoder.withContext(context);
 
         logger.info("Starting encoding using " + ctx.getVersion());
+        prover.write("; Program encoding");
         prover.addConstraint(programEncoder.encodeFullProgram());
+        prover.write("; Memory model encoding");
```
ThomasHaas, line 60: Seeing that you only generate comments, a `prover.writeComment` method might be nicer to use (which automatically add the `; `-prefix). Also, I think writing anything but comments defeats the purpose of the class. But that is up to you.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/DataRaceSolver.java
```diff
@@ -59,16 +59,20 @@ private void run() throws InterruptedException, SolverException, InvalidConfigur
 		SymmetryEncoder symmetryEncoder = SymmetryEncoder.withContext(context);
 
         logger.info("Starting encoding using " + ctx.getVersion());
+        prover.write("; Program encoding");
 		prover.addConstraint(programEncoder.encodeFullProgram());
+        prover.write("; Memory model encoding");
```
ThomasHaas, line 64: Since you are touching this file already, can you reformat it?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -109,7 +109,7 @@ public static void performStaticWmmAnalyses(VerificationTask task, Context analy
         analysisContext.register(RelationAnalysis.class, RelationAnalysis.fromConfig(task, analysisContext, config));
     }
 
-    protected void saveFlaggedPairsOutput(Wmm wmm, WmmEncoder encoder, ProverEnvironment prover, EncodingContext ctx, Program program) throws SolverException {
+    protected void saveFlaggedPairsOutput(Wmm wmm, WmmEncoder encoder, ProverWithTracker prover, EncodingContext ctx, Program program) throws SolverException {
```
ThomasHaas, line 112: This is one of the cases where the use-site could keep working on `ProverEnvironment` if it was implemented.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -0,0 +1,160 @@
...
+        prover.pop();
+    }
+
+    public void write(String content) {
+        if(dump()) {
+            File file = new File(fileName);
+            FileWriter writer;
+            try {
+                writer = new FileWriter(file, true);
+                PrintWriter printer = new PrintWriter(writer);
```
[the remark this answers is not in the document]
hernanponcedeleon: I don't get what you mean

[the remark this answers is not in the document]
ThomasHaas:
```
try (FileWriter writer = new FileWriter(...);
       PrintWriter = new PrintWriter(...)) {
       // ...
}
```
You need to double-check the syntax for declaring two resources (not sure if semicolon is correct here).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -0,0 +1,185 @@
...
+        long start = System.currentTimeMillis();
+        boolean result = prover.isUnsatWithAssumptions(fs);
+        long end = System.currentTimeMillis();
+        if(dump()) {
+            write("(push)\n");
+            for(BooleanFormula f : fs) {
+                write(fmgr.dumpFormula(f).toString());
+            }
+            write("(set-info :status " + (result ? "unsat" : "sat") + ")\n");
+            write("(check-sat)\n");
```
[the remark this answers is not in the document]
hernanponcedeleon: The `status` this is not needed for debugging, but I am planning to eventually submit some of the encodings to SMT-COMP and there it is mandatory (including their ordering).

Because I need to put the status before the `check-sat`, I can only put up to the `push` before starting solver in dat3m side. If we end up with an scenario where really the smt solver does not finish, we just need to manually add a final `check-sat` and try other solver. However this scenario will not be that often. One of my main goals is to have files where dat3m eventually finished and being able to identify which iterations were slow and hopefully get some insight from smt solver developers.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -0,0 +1,185 @@
...
+
+    public boolean isUnsat() throws SolverException, InterruptedException {
+        long start = System.currentTimeMillis();
+        boolean result = prover.isUnsat();
+        long end = System.currentTimeMillis();
+        if(dump()) {
+            write("(set-info :status " + (result ? "unsat" : "sat") + ")\n");
+            write("(check-sat)\n");
+            write("; Original solving time: " + (end - start) + " ms");
+        }
```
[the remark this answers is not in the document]
hernanponcedeleon: I need the status, so I cannot move this before the actual solving (see answer above)

hernanponcedeleon:
I have no clue why the diff from the browser shows changes related to PRs recently merged to `development`. I double checked that my rebase did not miss some of the changes, e.g., the `Location` class removed in #725 is not in this branch https://github.com/hernanponcedeleon/Dat3M/tree/smtlib2/dartagnan/src/main/java/com/dat3m/dartagnan/program/memory

I don't get those locally. Attached if the diff I get locally wrt `development`.



[smtlib.patch](https://github.com/user-attachments/files/16898622/smtlib.patch)

ThomasHaas (approved):
LGTM

Issue #719, opened 2024-08-21
Implement spinloop detection for ARM exclusive pairs


natgavrilenko:
Consider the attached 'spinloop.litmus' test. The loop in thread P0 has no side effects, because this loop terminates whenever STLXR is executed. The value of W0 is 0 if the store was successful and 1 if the store failed. However, the test terminates with UNKNOWN result for any unrolling bound. The manually unrolled test 'spinloop-unrolled.litmus' works as expected.

The current implementation of spinloop detection fails to detect this spinloop. The reason is that dartagnan internally represent STLXR as a pair of instructions in the same branch, while spinloop detection essentially analyses the branches withing a loop. 

Note: I would avoid changing the internal representation of ARM exclusive pair to contain branches. Representing such event pairs using branches would generate artificial control dependencies, which are used in the ARM consistency model. I would rather implement a special handling of RmwStoreExclusive events in the spinloop detection method.

[spinloop.txt](https://github.com/user-attachments/files/16695825/spinloop.txt)
[spinloop-unrolled.txt](https://github.com/user-attachments/files/16695826/spinloop-unrolled.txt)

ThomasHaas:
Your code actually shows a problem in the parser for AARCH64. The registers W0 is parsed as X0, making the loop truly side-effectful as the address of the exclusive load (X0) gets (wrongly) updated by the status of the exclusive store (W0).
That being said, we would still fail to detect the side-effectfree iteration where the store fails (but fixing this is easy, though insufficient due to the parser bug).

hernanponcedeleon:
Register W0 is the lowest 32 bits of register X0 (which is 64 bits). The fact that the exclusive store partially updates the address of the load looks to me like a problem in the test per se.

I think @db7 generated this program from some real assembly. Diogo can you confirm if the assembly really uses W0/X0 for both the addresses and the execution status of the exclusive store, or this was a problem when translating from assembly to pasm with kittens?

ThomasHaas:
Oh, I'm not familiar with the register names in assembly code, but that certainly sounds wrong.

db7:
@hernanponcedeleon I can check that.

ThomasHaas:
Yes, but W0 is the lower part of X0, so the address gets updated if W0 is changed. This is surely wrong.

hernanponcedeleon:
```
 LC00:
 LDAXR   X3, [X0]
 STLXR   W0, X2, [X0]
 CBZ     W0, LC01
[...]

PR #722, opened 2024-08-30
Improved liveness detection for store exclusives

ThomasHaas:
This PR addresses issue #719 while also refactoring some code related to that.


ThomasHaas:
Does anybody know why commits from over 3 months ago are shown in this PR?

hernanponcedeleon:
Probably you created you branch before I forced pushed to development a few days ago (I had the same issue when I opened the smtlib PR, which was based on a local copy of depelopment that was in turns based rebased on top of master that had one extra commit, thus the forced pushed)

db7:
It would be nice to have this branch *or* `arm-asm` branch merged, because rebasing one on another always creates conflicts.

PR #724, opened 2024-09-02
Add support to CAAT for SyncBar and SyncFence relations

ThomasHaas:
CAAT still does not work with virtual memory objects, so it will produce wrong results on the PTX7.5 litmus tests.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
@@ -473,7 +473,7 @@ public class RefinementSolver extends ModelChecker {
         // TODO: We should probably automatically cut all "unknown relation",
         //  i.e., use a white-list of known relations instead of a blacklist of unknown one's.
         return def instanceof LinuxCriticalSections // LKMM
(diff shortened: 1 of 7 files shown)


hernanponcedeleon:
You should at least enable caat in the spirv tests, e.g. https://github.com/hernanponcedeleon/Dat3M/blob/9a23ab07c017edd2d9b698c279913ee214b4df5a/dartagnan/src/test/java/com/dat3m/dartagnan/spirv/benchmarks/SpirvRacesTest.java#L123

ThomasHaas:
Some of those tests also use virtual memory objects, so I cannot enable all of them.

hernanponcedeleon:
Is the comment "refinement still fails to understand virtual memory" outdated? If not, in which benchmarks is this causing problems?

ThomasHaas:
The comment is still valid and it will produce wrong results on PTX7.5 litmus tests. I haven't checked the details, but I believe the `ExecutionModel` just does not capture virtual accesses correctly and this causes Refinement to produce wrong results.

ThomasHaas:
I should add: I think virtual addressing was added in a very ad hoc manner and has many flaws. If we really want to support it in the long run, we should properly think it through before trying to fit in the ad hoc solution into all the solver backends.

PR #725, opened 2024-09-02
Refactor/Rename Location to align with its actual semantics

ThomasHaas:
During the discussions in #703 I realized why the `Location` class always felt off: its name suggested it represents an address, but it never did; it always represented the final value in memory.
This PR fixes this long-standing mistake.


hernanponcedeleon:
> it always represented the final value in memory. 

"Final" or "latest"?

ThomasHaas:
> > it always represented the final value in memory.
> 
> "Final" or "latest"?

Is there a difference in interpretation? I think "final", "last", and "latest" are all the same, no?

hernanponcedeleon:
"Final" suggests there are no further changes, "latest" does not. 

The fact that part of the code evaluates the value at a given event, sounds to me like you mean the later.

ThomasHaas:
No, the code does throw an error when trying to evaluate at a specific event. I have not changed the meaning at all: the expression represents the co-maximal value of the execution, i.e., the final value in memory. This is how it always was for litmus code.

PR #726, opened 2024-09-04
Backwards reaching definitions

xeren:
This PR decouples an interface for `Dependency` and `UseDefAnalysis`, and adds an alternative implementation for it, `BackwardsReachingDefinitionsAnalysis` (BRDA).  It takes into account, that readers on uninitialized registers are unlikely, while final writers are frequent, although rarely useful.  This means, that the propagated data is smaller in BRDA, than in the existing forward-directed implementations.
The more noteworthy feature is, that I wrote it to support loops.  This means that it, as well as the dependent alias analysis could potentially be done before unrolling.


ThomasHaas:
`UseDefAnalysis` also supported loops. But I guess your new analysis might be more efficient?

xeren:
Yes, it should have the capabilities of both implementations and be more efficient.

ThomasHaas:
Also, why do you use `BranchEquivalence` rather than `ExecutionAnalysis`? I think the latter is more correct because it also considers that instructions may fail to execute.

ThomasHaas:
@hernanponcedeleon Do you remember on which benchmark `UseDefAnalysis` ran out of memory? I think you had some huge benchmarks where this was a problem. If so, it would be interesting to test this new analysis for efficiency.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,388 @@
...
+            analysis.analyzeMust(branch);
+        }
+        return analysis;
+    }
+
+    public static BackwardsReachingDefinitionsAnalysis injectForProgram(Program program, Context analysisContext) {
+        final BranchEquivalence branch = analysisContext.get(BranchEquivalence.class);
+        final BackwardsReachingDefinitionsAnalysis analysis = forProgram(program, branch);
+        analysisContext.register(BackwardsReachingDefinitionsAnalysis.class, analysis);
+        analysisContext.register(ReachingDefinitionsAnalysis.class, analysis);
```
ThomasHaas, line 121: Why do you need to register the concrete class here? If not needed, I would prefer to move the registering to calling code so as to not break the pattern we have.

xeren:
> Also, why do you use BranchEquivalence rather than ExecutionAnalysis? I think the latter is more correct because it also considers that instructions may fail to execute.

Because the `cfImpliesExec` condition is only needed in the over approximation (the 'may' relation). `BranchEquivalence` on the other hand, more precisely the `areMutuallyExclusive` relation, is only needed in the under approximation (the 'conditionally-must' relation).  Since the later is optional, BE is not mandatory.

xeren:
By the way, using the analysis has the side effect, that final register encodings are omitted, if the register is not used by the spec.  This could be a small issue fixed.

ThomasHaas:
> > Also, why do you use BranchEquivalence rather than ExecutionAnalysis? I think the latter is more correct because it also considers that instructions may fail to execute.
> 
> Because the `cfImpliesExec` condition is only needed in the over approximation (the 'may' relation). `BranchEquivalence` on the other hand, more precisely the `areMutuallyExclusive` relation, is only needed in the under approximation (the 'conditionally-must' relation). Since the later is optional, BE is not mandatory.

I'm not sure I understand. It seems you could just use `ExecutionAnalysis.areMutuallyExclusive` in your code to both simplify it and be more accurate (in theory). Conceptually, when analysing dataflow you care about executed instructions rather than instructions in the control-flow.
It likely won't make a practical difference in your use-case, because the `ExecutionAnalysis` is based only on control-flow but I don't see why your analysis should be aware of that and "optimize" for it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
xeren: It is indeed not required anywhere and can be removed.

hernanponcedeleon:
> This means, that the propagated data is smaller in BRDA, than in the existing forward-directed implementations.

Which part of our pipelines are (positively) affected by this? E.g., does the alias analysis potentially becomes more precise? Do we get smaller mayset in idd?

ThomasHaas:
I think this statement is not about precision but just about memory consumption. The forward-directed implementation kept information about (last) register writes even if those registers were never going to get used again.

xeren:
> Which part of our pipelines are (positively) affected by this? E.g., does the alias analysis potentially becomes more precise? Do we get smaller mayset in idd?

Compared to `UseDefAnalysis` and `Dependency`, the results of BRDA for any `RegReader` should be exactly the same.  So, any dependent analysis should be relatively unaffected, up until the encoder.  `idd` stays the same.  With *propagated data*, I just meant the internal updates of BRDA and `Dependency`.  At best, there could be side effects on BRDA's querying performance, based on the data layout I chose, i.e. the time it needs to fetch a list of writers.

Since `UseDefAnalysis` and `Dependency` are unused now, I probably should add the following options, before merging.
- `program.processing.loopBounds.useDefAnalysis` = {`FORWARD`, `BACKWARD`}
- `program.analysis.reachingDefinitions` = {`FORWARD`, `BACKWARD`}

ThomasHaas:
If the analysis work fine, I don't think we need to keep around the unused options.

hernanponcedeleon:
Keeping two implementations of the analysis might be good: if we encounter any example where one is too slow, we can try another (similar to what happened with the alias analysis and `wsq.c`). 

However, keeping options for each single pass making use of the analysis seems too fine-grained without much clear benefit. I would rather **always** use forward or **always** us backwards.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+        final var analysis = new BackwardsReachingDefinitionsAnalysis();
+        final Set<Register> finalRegisters = new HashSet<>();
+        analysis.initialize(function, finalRegisters);
+        analysis.run(function, finalRegisters);
+        analysis.postProcess();
+        return analysis;
+    }
+
+    /**
+     * Analyzes an entire set of threads.
```
hernanponcedeleon, line 99: "an entire set of threads." -> "the entire program (after thread creation)" ... you are not passing any threads as parameter

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+     * <p>
+     * Optionally queries {@link ExecutionAnalysis} for pairs of writers appearing together in an execution.
+     * @param program Contains a set of threads to be analyzed.  Additionally-defined functions are ignored.
```
hernanponcedeleon, line 102: It sounds like the analysis is intended to be run only after threads has been created. If so, maybe we should add some check somewhere.

EDIT: Now I am confused. The comment says that functions are ignored, but below you have one case where you iterate over both threads and functions.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+        final ExecutionAnalysis exec = analysisContext.get(ExecutionAnalysis.class);
+        final var analysis = new BackwardsReachingDefinitionsAnalysis();
+        final Set<Register> finalRegisters = finalRegisters(program);
+        for (Function function : program.isUnrolled() ? program.getThreads() :
+                Iterables.concat(program.getThreads(), program.getFunctions())) {
+            analysis.initialize(function, finalRegisters);
+            analysis.run(function, finalRegisters);
+        }
+        analysis.postProcess();
+        if (exec != null && program.isUnrolled()) {
```
hernanponcedeleon, line 116: Can `exec` really be null?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+        @Override
+        public List<RegWriter> getMayWriters() {
+            return info.mayWriters.stream().filter(w -> w.getResultRegister().equals(register)).toList();
+        }
+
+        @Override
+        public List<RegWriter> getMustWriters() {
+            return info.mustWriters.stream().filter(w -> w.getResultRegister().equals(register)).toList();
+        }
+
```
hernanponcedeleon, line 177: What is the difference between may and must?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+        for (Event event : function.getEvents()) {
+            if (event instanceof RegReader reader) {
+                final Set<Register> usedRegisters = new HashSet<>();
+                for (Register.Read read : reader.getRegisterReads()) {
+                    usedRegisters.add(read.register());
+                }
+                readerMap.put(reader, new ReaderInfo(usedRegisters));
+            }
+        }
+        readerMap.put(null, new ReaderInfo(finalRegisters));
```
hernanponcedeleon, line 228: Why not process everything in the same loop iteration? I don't see any dependency from the second loop tot eh first one.

Also, what are these `X.put(null, ...)` representing?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    private void run(Function function, Set<Register> finalRegisters) {
+        //For each register used after this state, all future users.
+        final State currentState = new State();
+        for (Register finalRegister : finalRegisters) {
+            final var info = new StateInfo();
+            info.mayReaders.add(null);
```
hernanponcedeleon, line 236: You are abusing the usage of `null` for some special purpose that is not clear to me at all. Is this really necessary?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+            for (RegReader reader : entry.getValue().mayReaders) {
+                readerMap.get(reader).uninitialized.add(entry.getKey());
+            }
+        }
+    }
+
+    private void postProcess() {
+        // build the reverse association
+        for (Map.Entry<RegWriter, Readers> entry : writerMap.entrySet()) {
+            if (entry.getKey() == null) {
```
hernanponcedeleon, line 263: cant you do `entry.hasKey()`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+                mayWriters.add(entry.getKey());
+            }
+        }
+        final Comparator<RegWriter> byGlobalId = Comparator.comparingInt(RegWriter::getGlobalId);
+        for (Map.Entry<RegReader, ReaderInfo> entry : readerMap.entrySet()) {
+            entry.getValue().mayWriters.sort(byGlobalId);
+        }
+        // set the final flag
+        for (Readers writer : writerMap.values()) {
+            writer.mayBeFinal = writer.readers.remove(null);
```
hernanponcedeleon, line 278: once again passing null ...

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+            return true;
+        }
+        target.mayReaders = target.copyOnWrite ? new HashSet<>(target.mayReaders) : target.mayReaders;
+        target.copyOnWrite = false;
+        target.mayReaders.addAll(source.mayReaders);
+        return true;
+    }
+
+    private void analyzeMust(ExecutionAnalysis exec) {
+        // Require that function is unrolled.
```
hernanponcedeleon, line 363: Can this be programmatically checked? Maybe in the caller?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
ThomasHaas: The analysis can be run during processing where `ExecutionAnalysis` is not available, but it can also be run afterwards. In the former case it should subsume `UseDefAnalysis` and in the latter case `Dependency`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
ThomasHaas:
```
r = 5;
if (*)
   r = 3;
x = r; // r=5 is may but not must. r=3 is must.
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
ThomasHaas: The `null` entry is likely for reads from uninitialized registers.
I agree about the loop: either you merge them or you keep them and iterate over `getEvents(RegWriter.class)` resp. `getEvents(RegReader.class)` to skip the `instanceof` checks.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
ThomasHaas: There is no `hasKey`, is there? But technically, the entry is valid and just represents the special value of uninitialized registers.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
ThomasHaas: I would assume this is free because `ExecutionAnalysis` already assumes this. In fact, at this point in time we are only working over threads.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -0,0 +1,384 @@
...
+public class BackwardsReachingDefinitionsAnalysis implements ReachingDefinitionsAnalysis {
+
+    private final Map<RegReader, ReaderInfo> readerMap = new HashMap<>();
+    private final Map<RegWriter, Readers> writerMap = new HashMap<>();
+
+    @Override
+    public Writers getWriters(RegReader reader) {
+        Preconditions.checkNotNull(reader, "reader is null");
+        final ReaderInfo result = readerMap.get(reader);
+        Preconditions.checkArgument(result != null, "reader %s has not been analyzed.", reader);
```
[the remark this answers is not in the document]
xeren: Valid, if `reader.getFunction()` had been analyzed by `this`.  I considered more likely, that a user would try to query events of another function or program.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
ThomasHaas: Btw. do we get no must information at all without the `ExecutionAnalysis`?
Would a dominator tree analysis suffice to generate must information?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -390,16 +391,18 @@ public Void visitAddressDependency(DirectAddressDependency addrDirect) {
...
 
         private Void visitDirectDependency(Relation r) {
-            Dependency dep = context.getAnalysisContext().get(Dependency.class);
-            EncodingContext.EdgeEncoder edge = context.edge(r);
+            final ReachingDefinitionsAnalysis definitions = context.getAnalysisContext()
+                    .get(ReachingDefinitionsAnalysis.class);
+            final EncodingContext.EdgeEncoder edge = context.edge(r);
             encodeSets.get(r).apply((writer, reader) -> {
-                if (!(writer instanceof RegWriter rw)) {
+                if (!(writer instanceof RegWriter wr) || !(reader instanceof RegReader rr)) {
```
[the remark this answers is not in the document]
xeren: I have the parameter of `ReachingDefinitionsAnalysis.getWriters` be `RegReader` to clarify, that non-RegReaders never participate in this relation.  Also, that got rid of some IDE warnings about 'suspicious calls to [collection methods]' in the implementations.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
hernanponcedeleon: I thought both implementations were exactly the same (I missed the different usage of `mayWriters` and `mustWriters`), thus the question

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
hernanponcedeleon: > The `null` entry is likely for reads from uninitialized registers. 

Whatever this is used for, it should be documented, I don't want someone reading the code to have to guess

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
xeren: Yes, it is.  But to keep safe, I left this comment.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
```diff
@@ -209,14 +213,17 @@ public boolean mayBeFinal() {
...
+    private void initializeWriterMap(Collection<Event> events) {
+        for (Event event : events) {
             if (event instanceof RegWriter writer) {
                 writerMap.put(writer, new Readers());
             }
         }
-        writerMap.put(null, new Readers());
-        for (Event event : function.getEvents()) {
+        writerMap.put(INITIAL_WRITER, new Readers());
+    }
```
ThomasHaas, line 223: I don't like this change. I think the original request was not about saving a call to `function.getEvents`.
Just use `for (RegWriter writer : function.getEvents(RegWriter.class))`. You can even use `function.getEvents(RegWriter.class).foreach(w -> writerMap.put(w, new Readers())` if you prefer.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
xeren: Yep, since there are no dependencies, both versions are valid.  I consider my version better in terms of cache optimization, but not by much.

In `writerMap`, `null` stands for the initial writer(s).  In `readerMap`, it denotes the final reader(s).  Both cases don't have actual `Event` instances.  In this analysis, they behave *as if* they had.  For readability, I added named null constants.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BackwardsReachingDefinitionsAnalysis.java
xeren: It does not exist, even in Java 22.

hernanponcedeleon:
Unless I missed something, not all comments were fixed. From the top of my head:
- we still have fine grained options that I don't like
- some of the preconditions / checks need to be streamlined

ThomasHaas (approved):
LGTM

PR #727, opened 2024-09-05
Add unrolling bound to program spec encoding

natgavrilenko:


dartagnan/src/test/java/com/dat3m/dartagnan/spirv/basic/SpirvAssertionsTest.java
```diff
@@ -78,30 +77,23 @@ public static Iterable<Object[]> data() throws IOException {
...
                 {"loop-struct-cond-suffix.spv.dis", 2, PASS},
-                {"loop-struct-cond-suffix.spv.dis", 3, PASS},
-                {"loop-struct-cond-sequence.spv.dis", 2, FAIL},
+                {"loop-struct-cond-sequence.spv.dis", 2, UNKNOWN},
                 {"loop-struct-cond-sequence.spv.dis", 3, PASS},
-                {"loop-struct-cond-sequence.spv.dis", 4, PASS},
-                {"loop-struct-cond-nested.spv.dis", 2, FAIL},
+                {"loop-struct-cond-nested.spv.dis", 2, UNKNOWN},
                 {"loop-struct-cond-nested.spv.dis", 3, PASS},
-                {"loop-struct-cond-nested.spv.dis", 4, PASS},
```
hernanponcedeleon, line 104: - Why are you removing some tests?

- Was it the case that we were giving wrong results before (due to not really evaluating the condition after threads terminated), thus some previous FAIL are now UNKNOWN?

ThomasHaas: It seems the Natalia removed tests with unnecessarily high bounds, i.e., we don't need to use B=4 if B=3 already gives PASS.

hernanponcedeleon (approved):

PR #728, opened 2024-09-06
Use correct smtlib2 syntax for push/pop

hernanponcedeleon:
The correct smtlib syntax for `push` is `(push <numeral>)` (see section 3.9.9 of [SMTLIBTutorial.pdf](https://smtlib.github.io/jSMTLIB/SMTLIBTutorial.pdf)). Similar to `pop`.

While the solvers I tested before (z3, cvc5) interpret `(push)` as `(push 1)`, yices2 complains about the syntax.


ThomasHaas (approved):
LGTM

PR #729, opened 2024-09-09
Fix InclusionBasedPointerAnalysis.Top

xeren:
Sets `TOP` to its theoretical definition. This should make `IBPA` sound for programs with negative dynamic indexing.


dartagnan/src/test/java/com/dat3m/dartagnan/miscellaneous/AnalysisTest.java
```diff
@@ -112,7 +112,7 @@ public void fieldinsensitive0() throws InvalidConfigurationException {
 
     @Test
     public void full0() throws InvalidConfigurationException {
-        program0(FULL, MAY, MAY, NONE, NONE, NONE, NONE);
+        program0(FULL, NONE, MAY, NONE, NONE, NONE, NONE);
```
ThomasHaas, line 115: This the analysis get more precise here? I thought it would only be less precise with this change.

PR #733, opened 2024-09-11
Add support for Parametric

tonghaining:


hernanponcedeleon:
@tonghaining as a test case for the PR, you can remove the hack we have for linux-kernel.cat and original code from [here](https://github.com/torvalds/linux/blob/8d8d276ba2fb5f9ac4984f5c10ae60858090babc/tools/memory-model/linux-kernel.cat#L88)

ThomasHaas:
I'm wondering if we even need semantic representations of these parametric calls in the `Wmm` class.
I feel like this feature can be implemented purely in the parser by inlining at parsing time.
I might try this later.

ThomasHaas:
I think there is a question about semantics of parametric definitions: do they act like macros or like proper functions?
Consider this case:
```
let a = ...
let f(r) = r | a
let a = ...
let b = f(c) // Does f(c) use the new or the old definition of "a"?
```
If it acts like macros, then the parametric relation is evaluated in the current namespace (it sees updated definitions), if it acts like a function then the evaluation is wrt. the namespace at definition time of the function (it sees only the definitions that existed when the function was defined).

ThomasHaas:
This PR should be subsumed by #736

hernanponcedeleon:
The implementation in #736 is working properly, thus I am closing this PR.

hernanponcedeleon:
The implementation in https://github.com/hernanponcedeleon/Dat3M/pull/736 is working properly, thus I am closing this PR.

PR #736, opened 2024-09-12
Added support for includes and functions in cat

ThomasHaas:
~~This PR is marked as DRAFT simply because there are currently no tests related this feature, and no existing `.cat`-files were updated. Otherwise, the PR is functional.~~

Added functionality:
- `include "filePath"` now has semantics in `.cat` files.
- Semantically, the included file (if it exists) is simply inlined into the including file (as if copy&pasted). If the file does not exist, we give a warning and treat it as NOOP.
- The `filePath` is resolved relative to an optionally provided `--cat.include=<includeDirectory>` which defaults to `DAT3M_HOME/cat`. 
- Includes can be nested but there is no detection of recursive includes: those will likely cause recursion until out of memory.
- Support for functions (see below)

If anyone (@hernan-poncedeleon ? :)) is interested in rewriting existing `.cat` files to make use of this feature, you are welcome to add to this PR. In particular, a simple `basic.cat` with definitions such as `let poloc = po & loc` would be nice to have so that we can remove a lot of hardcoded definitions from the `Wmm` class.

~~**Possible other feature**: I want to take a look into adding parametric definitions as in #733 to this PR but with a less invasive approach (i.e., purely implemented in the parser).~~

EDIT: I implemented functions/function calls in `.cat`. The support is very extensive:
- Multiple parameter are allowed: `let f(x, y) = x | y` (though 0 parameters are currently unsupported)
- Functions also work for unary sets:
```
let f(x, y) = x | y
let hb = f(com, po)
let M = f(R, W) 
```
- Functions capture the context in which they were defined:
```
let z = oldDef
let f(x, y) = x | y | z // refers to the above defined z
let z = newDef
let a = f(b, c) // uses `oldDef` for z
```
- Functions can be redefined just like relations
- Functions can take complex arguments: `let hb = f(po, f(co, f(rf, f(fr))))`
- Functions can call other functions: `let g(x) = f(x, x)`. They cannot call themselves (recursion in functions is not allowed)


hernanponcedeleon:
> If anyone (@hernan-poncedeleon ? :)) is interested in rewriting existing `.cat` files to make use of this feature, you are welcome to add to this PR. In particular, a simple `basic.cat` with definitions such as `let poloc = po & loc` would be nice to have so that we can remove a lot of hardcoded definitions from the `Wmm` class.

Do we want to keep treating `ctrl, data, addr` as "base" relations or should we define them as
```
let data = (idd^+) & (M * M)
let addr = (addrDirect | ((idd^+);addrDirect)) & (M * M)
let ctrl = ((idd^+);ctrlDirect) & (M * V)
```
Put it differently, do we want to get rid of all relations that make use of operators in `Wmm.makePredefinedRelation()`?

ThomasHaas:
Possibly. The awkward part is that some of those base relations like `idd` are magical in that they argue about non-visible events. I'm not sure if we want to expose them.

cat/linux-kernel.cat
```diff
@@ -18,7 +18,7 @@
  * It can be replaced by include "cos.cat" for tests that do not use locks.
  *)
 
-include "lock.cat"
+//include "lock.cat"
```
hernanponcedeleon, line 21: I am fine with modifying / cleaning some of the cat files (I am actually doing while making better use of the includes). The only model I don't want to touch is lkmm: the version in the kernel tree is updated every now and then and when I update our version I want to keep the required changes at minimum (with parametric relations the only remaining change is moving the def of Marked from the bell file to the cat one IIRC).

What is the problem with lock.cat? We should leave the include as it is and we either remove the file from our repo, or we commented out in lock.cat whatever is causing problems (the later might have the advantage of still performing some sanity checks in the form of flagged axioms).

ThomasHaas: > What is the problem with lock.cat? 

```
let RL = try RL with emptyset
let RU = try RU with emptyset
...
let all-possible-rfe-lf =
	let possible-rfe-lf r =
		let pair-to-relation p = p ++ 0
		in map pair-to-relation ((LKW * {r}) & loc & ext)
	in map possible-rfe-lf (LF \ range(rfi-lf))
...
with rfe-lf from cross(all-possible-rfe-lf)
...
let rf = rf | ([IW | UL] ; singlestep(co) ; lk-rmw^-1)
...
show co,rf,fr
```
There is a lot of unsupported stuff. And most definitions are some very roundabout way to add lock events to `rf/co`, something which we would probably do differently.

> We should leave the include as it is and we either remove the file from our repo, or we commented out in lock.cat whatever is causing problems (the later might have the advantage of still performing some sanity checks in the form of flagged axioms).
We can also delete that file. But including the sanity checks seems like a bad idea: you pay the price for all `cat_spec` verifications for something that you should never need to verify.

hernanponcedeleon: Yes, I spent some more time this morning cleaning up the models / Wmm.class and came up to the conclusion that the best is to comment the whole content of `lock.cat` and add inside it an include to `cos.cat` (which is the one where I define all basic stuff).

I am running some tests right now and hopefully will push (make use of parametric and include) soon.

ThomasHaas: Hmm, I wouldn't abuse `cos.cat` for the basic stuff because that file is a standard one in `herd7` with the special purpose of generating coherence (and `cos.cat` is used in more models than just LKMM).
I would rather have a completely own `basic.cat` (or whatever you want to call it) and you could include that. You could then have `cos.cat` just include `basic.cat` so you don't need to change the memory model includes.

hernanponcedeleon: We don't yet have a `cos.cat` file in our repo. Copying the file from herd's repo might cause problems similar to `lock.cat` (i.e. unsupported stuff). Adding an empty one that just includes `basic.cat` seems pointless.

My goal is to keep all models using `cos.cat` untouched. When we run those models with dat3m, it would include our own `cos.cat`. If one would want to use our models with herd, it would have to run something like `herd8 -model cat/X.cat -include /herdrepo/...`, thus picking up the one that herd needs

ThomasHaas: I still think it would be cleaner to have it just include `basic.cat`. For example, in those models that herd7 does not support/have (GPU models), we would also like to include those definitions but including `cos.cat` suggests that this is somehow related to herd7 or that herd7 supports those files somehow.
Apart from that, `cos.cat` can include `basic.cat` and have a comment/explanation of what is going on here (i.e. that this is a mock that is different to herd7's version). Just having such a comment is IMO enough to justify that.

hernanponcedeleon: Done

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
```diff
@@ -19,34 +19,23 @@ public class RelationNameRepository {
...
-    public static final String COE = "coe";
-    public static final String COI = "coi";
-    public static final String FRE = "fre";
-    public static final String FRI = "fri";
     public static final String MFENCE = "mfence";
     public static final String ISH = "ish";
     public static final String ISB = "isb";
     public static final String SYNC = "sync";
     public static final String ISYNC = "isync";
     public static final String LWSYNC = "lwsync";
```
ThomasHaas, line 31: Can we somehow even get rid of these fence definitions? It's unfortunate that the relation has the same name as the event tag, so we probably cannot define `let mfence = po;[mfence];po`.
Well, maybe it works (I haven't tried).

hernanponcedeleon: I considered moving each definition to `X.fences.cat` and get rid of those, but then some of the parsers were complaining. Maybe I can move the string to each particular parser (similar to how `RefinementSolver` defines the strings it needs)

ThomasHaas: I see. The `RefinementSolver` thing needs to be solved differently anyways because the biases will only work if the user has defined all those relations (which might not be the case!). Unless that is intended, but then we need to generate proper error messages if the relations are missing.

hernanponcedeleon: How hard would it be to take a second cat model as input for the bias?

ThomasHaas: Probably not too hard.

hernanponcedeleon: I got rid of all fence definitions in the code.

I leave it to you to fix the `RefinementSolver` :P

ThomasHaas: Can we define `fencerel(r)` rather than having special handling for it?

hernanponcedeleon: https://github.com/hernanponcedeleon/Dat3M/pull/736/commits/3b8c0dfaeaa1f5a57385dafd433ebd17dfe67474 best commit ever

ThomasHaas: Yep, very nice. I thought you would just define `let fencerel(r) = po;[r];po` though. But I don't mind either way.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -112,19 +113,20 @@ public static void main(String[] args) throws Exception {
...
+        logger.info("Program path: {}", fileProgram);
 
         File fileModel = new File(Arrays.stream(args).filter(a -> a.endsWith(".cat")).findFirst()
                 .orElseThrow(() -> new IllegalArgumentException("CAT model not given or format not recognized")));
-        logger.info("CAT file path: " + fileModel);
+        logger.info("CAT file path: {}", fileModel);
 
-        Wmm mcm = new ParserCat().parse(fileModel);
+
+        Wmm mcm = new ParserCat(Path.of(o.getCatIncludePath())).parse(fileModel);
```
hernanponcedeleon, line 123: I don't think this works if you add an include path different than where the main cat file is, e.g., `java -jar dartagnan/target/dartagnan.jar cat/tso.cat --cat.include=/home/ponce/git/Dat3M/ litmus/AARCH64/extra2/IRIW+poaps+LL.litmus`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -11,39 +14,92 @@
...
+            );
+        }
+    }
+
+    VisitorCat(Path includePath) {
+        this.includePath = includePath;
         this.wmm = new Wmm();
     }
 
+
```
hernanponcedeleon, line 67: remove

dartagnan/src/main/java/com/dat3m/dartagnan/utils/options/BaseOptions.java
```diff
@@ -75,4 +75,11 @@ public abstract class BaseOptions {
     private boolean smtlib=false;
 
     public boolean getDumpSmtLib() { return smtlib; }
+
+    @Option(
+            name=CAT_INCLUDE,
+            description = "The directory used to resolve cat include statements. Defaults to Dartagnan's cat directory."
```
hernanponcedeleon, line 81: "Defaults to Dartagnan's cat directory" -> "Defaults to $DAT3M_HOME/cat"

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: Yes, because all includes will be resolved w.r.t. the provided inclusion path so `basic.cat` won't be found.
What solution do you suggest? A include resolution strategy that checks multiple locations?
I think creating a robust resolution strategy is not easy.

hernanponcedeleon: Are you saying that here the problem is that it cannot find `basic.cat` in `--cat.include=/home/ponce/git/Dat3M/`? How is this different from not finding e.g., `filters.cat` (from `tso.cat`) when not passing `--cat.include` and just logging a warning?

I am fine with "all includes to be assumed to be in the same path", what I don't want is the assumption that the model file I pass as `cat/X.cat` should also be in the include path and that the line I commented above looks like.

Am I missing something?

hernanponcedeleon: This cmd does not work
`java -jar dartagnan/target/dartagnan.jar cat/tso.cat --cat.include=/home/ponce/git/Dat3M litmus/AARCH64/extra2/IRIW+poaps+LL.litmus`
this one does
`java -jar dartagnan/target/dartagnan.jar cat/tso.cat --cat.include=/home/ponce/git/Dat3M/cat litmus/AARCH64/extra2/IRIW+poaps+LL.litmus`

> if I provide an invalid path it fails to include
fails means you see a warning or you get an exception?

ThomasHaas: I get an exception due to usage of undefined relations (e.g. `fr`). Which is to be expected because `basic.cat` fails to include.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/RelationNameRepository.java
ThomasHaas: > I leave it to you to fix the `RefinementSolver` :P

Ok, this is not super straightforward due to cutting. I would rather not fix this in this PR and let the biases be only semi-functional for now.

ThomasHaas: Actually, I don't think it will work at all because the relations are queried on the baseline model (which is empty most of the time). So the feature might be non-functional.

hernanponcedeleon: > I would rather not fix this in this PR and let the biases be only semi-functional for now.

Do you at least have an idea of who this should be fixed?

ThomasHaas: I think we need the functionality to intersect/merge two memory models. Then we can generate a Wmm by cutting the original one and then merge into this one the bias model.

hernanponcedeleon: Ok. Is there anything missing for the PR or can I merge once the CI is green?

ThomasHaas: I think nothing is missing.

ThomasHaas: As I said above, the original memory model also needs to get cut in many cases and the cut part needs to be part of the baseline model. So you need to be able to merged two models (the cut model and the parsed baseline model).

hernanponcedeleon: > the original memory model also needs to get cut 

Is this something that is only needed after the changes in this PR because, what needs to be cut, before used to be considered base relations?

ThomasHaas: No, originally we first cut the model and on the resulting model we then programatically added the biases. For the later part we relied on `getPredefinedRelation` for a few cases like `fr` and `po-loc`. I guess the construction of those few relations could be moved into the bias-generating code though.

hernanponcedeleon: I did this and now biases seem to be functional. Unless you have some objections about my last commit, I think we can finally merge

ThomasHaas: I think it's fine. Technically, `fr` gets created even if it is not needed in some cases, but I guess that shouldn't matter much.

PR #739, opened 2024-09-18
Pass the C file to the svcomp runner rather than the optimized llvm

hernanponcedeleon:
In #666 we decided to log a warning instead of throwing an exception if `opt` failed. However, if it failed, then the `*-opt.ll` file (which was still being used by the svcomp wrapper) does not exists.

This PR solves the problem by using directly the C file rather than the llvm code.


ThomasHaas (approved):
LGTM

PR #741, opened 2024-09-20
Some improvements to aarch64 pseudo-assembly

hernanponcedeleon:
This PR adds support for
- return instructions
- nop instructions
- hexa constant
- registers name using underscored names
- special handling for zero registers (similar to RISCV)

It also enables some more passes for litmus code.

TODO
- [x] Fix special treatment of zero registers in `LazyRelationAnalysis` and `NativeRelationAnalysis` #743

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
@@ -7,12 +7,14 @@ import com.dat3m.dartagnan.expression.ExpressionFactory;
 import com.dat3m.dartagnan.expression.Type;
 import com.dat3m.dartagnan.expression.integers.IntLiteral;
 import com.dat3m.dartagnan.expression.type.FunctionType;
+import com.dat3m.dartagnan.expression.type.IntegerType;
 import com.dat3m.dartagnan.expression.type.TypeFactory;
 import com.dat3m.dartagnan.program.*;
 import com.dat3m.dartagnan.program.Thread;
 import com.dat3m.dartagnan.program.Program.SourceLanguage;
 import com.dat3m.dartagnan.program.event.Event;
 import com.dat3m.dartagnan.program.event.EventFactory;
+import com.dat3m.dartagnan.program.event.RegWriter;
 import com.dat3m.dartagnan.program.event.core.Label;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.dat3m.dartagnan.program.event.metadata.OriginalId;
@@ -85,6 +87,27 @@ public class ProgramBuilder {
         for (Function func : Iterables.concat(program.getThreads(), program.getFunctions())) {
             if (func.hasBody()) {
                 func.getEvents().forEach(e -> e.setMetadata(new OriginalId(e.getGlobalId())));
+            }
+        }
+    }
+
+    public static void replaceZeroRegisters(Program program, List<String> zeroRegNames) {
+        for (Function func : Iterables.concat(program.getThreads(), program.getFunctions())) {
+            if (func.hasBody()) {
+                for (String zeroRegName : zeroRegNames) {
+                    Register zr = func.getRegister(zeroRegName);
+                    if (zr != null) {
+                        for (RegWriter rw : func.getEvents(RegWriter.class)) {
+                            if (rw.getResultRegister().equals(zr)) {
+                                Register dummy = rw.getThread().getOrNewRegister("__zeroRegDummy_" + zr.getName(), zr.getType());
+                                rw.setResultRegister(dummy);
+                            }
+                        }
+                        // This comes after the loop to avoid the renaming in the initialization event
+                        Event initToZero = EventFactory.newLocal(zr, expressions.makeGeneralZero(zr.getType()));
+                        func.getEntry().insertAfter(initToZero);
+                    }
+                }
             }
         }
     }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
@@ -17,8 +17,12 @@ import com.dat3m.dartagnan.program.event.arch.StoreExclusive;
 import com.dat3m.dartagnan.program.event.core.Label;
 import com.dat3m.dartagnan.program.event.core.Load;
 
+import java.math.BigInteger;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+
+import static com.dat3m.dartagnan.parsers.program.utils.ProgramBuilder.replaceZeroRegisters;
 
 public class VisitorLitmusAArch64 extends LitmusAArch64BaseVisitor<Object> {
 
@@ -53,9 +57,10 @@ public class VisitorLitmusAArch64 extends LitmusAArch64BaseVisitor<Object> {
         visitVariableDeclaratorList(ctx.variableDeclaratorList());
         visitInstructionList(ctx.program().instructionList());
         VisitorLitmusAssertions.parseAssertions(programBuilder, ctx.assertionList(), ctx.assertionFilter());
-        return programBuilder.build();
+        Program prog = programBuilder.build();
+        replaceZeroRegisters(prog, Arrays.asList("XZR, WZR"));
+        return prog;
     }
-
 
     // ----------------------------------------------------------------------------------------------------------------
     // Variable declarator list, e.g., { 0:EAX=0; 1:EAX=1; x=2; }
@@ -212,6 +217,12 @@ public class VisitorLitmusAArch64 extends LitmusAArch64BaseVisitor<Object> {
     }
 
     @Override
+    public Object visitReturn(LitmusAArch64Parser.ReturnContext ctx) {
+        Label end = programBuilder.getEndOfThreadLabel(mainThread);
+        return programBuilder.addChild(mainThread, EventFactory.newGoto(end));
+    }
+
+    @Override
     public Expression visitExpressionRegister64(LitmusAArch64Parser.ExpressionRegister64Context ctx) {
         Expression expr = programBuilder.getOrNewRegister(mainThread, ctx.register64().id, archType);
         if(ctx.shift() != null){
@@ -255,4 +266,12 @@ public class VisitorLitmusAArch64 extends LitmusAArch64BaseVisitor<Object> {
         programBuilder.addChild(mainThread, EventFactory.newLocal(result, expressions.makeAdd(register, expr)));
         return result;
     }
+
+    @Override
+    public Expression visitImmediate(LitmusAArch64Parser.ImmediateContext ctx) {
+        final int radix = ctx.Hexa() != null ? 16 : 10;
+        BigInteger value = new BigInteger(ctx.constant().getText(), radix);
+        return expressions.makeValue(value, archType);
+    }
+
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
@@ -96,7 +96,7 @@ public class ProcessingManager implements ProgramProcessor {
                                 ComplexBlockSplitting.newInstance(),
                                 BranchReordering.fromConfig(config),
                                 Simplifier.fromConfig(config)
-                        ), Target.FUNCTIONS, true
+                        ), Target.ALL, true
                 ),
                 ProgramProcessor.fromFunctionProcessor(NormalizeLoops.newInstance(), Target.ALL, true),
                 RegisterDecomposition.newInstance(),
--- a/dartagnan/src/main/antlr4/LitmusAArch64.g4
+++ b/dartagnan/src/main/antlr4/LitmusAArch64.g4
@@ -76,6 +76,8 @@ instruction
     |   branchRegister
     |   branchLabel
     |   fence
+    |   return
+    |   nop
     ;
 
 mov locals [String rD, int size]
@@ -129,6 +131,14 @@ branchRegister locals [String rV, int size]
 
 branchLabel
     :   label Colon
+    ;
+
+return
+    :   Ret
+    ;
+
+nop
+    :   Nop
     ;
 
 loadInstruction locals [String mo]
@@ -252,7 +262,7 @@ location
     ;
 
 immediate
-    :   Num constant
+    :   Num Hexa? constant
     ;
 
 label
@@ -263,6 +273,18 @@ assertionValue
     :   location
     |   threadId Colon register64
     |   constant
+    ;
+
+Hexa
+    :   '0x'
+    ;
+
+Ret
+    :   'ret'
+    ;
+
+Nop
+    :   'nop'
     ;
 
 Locations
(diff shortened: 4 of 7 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -96,7 +96,7 @@ private ProcessingManager(Configuration config) throws InvalidConfigurationExcep
                                 ComplexBlockSplitting.newInstance(),
                                 BranchReordering.fromConfig(config),
                                 Simplifier.fromConfig(config)
-                        ), Target.FUNCTIONS, true
+                        ), Target.ALL, true
```
ThomasHaas, line 99: I predict that the linux litmus tests will break :).
IIRC, not a single pass respects the `IfAsJump` construct that is used to track proper ctrl-deps.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
```diff
@@ -53,9 +57,32 @@ public Object visitMain(LitmusAArch64Parser.MainContext ctx) {
...
+    /*
+    The "xzr" register plays a special role in AArhc64:
+      1. Reading accesses always return the value 0.
+      2. Discards data when written.
+     TODO: The below code is a simple fix to guarantee point 1. above.
+      Point 2. might also be resolved: although we do not prevent writing to xzr,
+      the value of xzr is never read after the transformation so its value is effectively 0.
+      However, the exists/forall clauses could still refer to that register and observe a non-zero value.
+     */
+    private void replaceXZRRegister(Program program) {
```
ThomasHaas, line 76: Why not just also do the replacement on the specification?

Is that register only used in local computations or can it also be used in memory operations?
Specifically, if a load is performed into that register, then the load value is obviously discarded, but does the load otherwise still provide consistency guarantees (e.g. ACQ)? Will operations reading from the special register afterwards carry dependencies from the load?
If the latter is true, then replacing the register simply by `0` is unsound as it breaks data deps.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
```diff
@@ -53,9 +57,32 @@ public Object visitMain(LitmusAArch64Parser.MainContext ctx) {
...
+        final ExpressionVisitor<Expression> xzrReplacer = new ExprTransformer() {
+            @Override
+            public Expression visitRegister(Register reg) {
+                return reg.getName().equals("xzr") ? expressions.makeGeneralZero(reg.getType()) : reg;
+            }
+        };
+        program.getThreadEvents(RegReader.class)
+                .forEach(e -> e.transformExpressions(xzrReplacer));
+    }
```
ThomasHaas, line 85: I think RISCV has the same situation (and maybe even Power?), just with a different register name.
Maybe it makes sense to put this functionality into `ProgramBuilder` with an additional `Register` parameter for the replacement. Then we can reuse that function.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
hernanponcedeleon: It seems it didn't

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
hernanponcedeleon: > Why not just also do the replacement on the specification?

Done

> Is that register only used in local computations or can it also be used in memory operations? Specifically, if a load is performed into that register, then the load value is obviously discarded, but does the load otherwise still provide consistency guarantees (e.g. ACQ)? Will operations reading from the special register afterwards carry dependencies from the load? If the latter is true, then replacing the register simply by `0` is unsound as it breaks data deps.

I could not find any proper answer by taking a look to the documentation. The closer I got was
```
Register '31' is one of two registers depending on the instruction context:

For instructions dealing with the stack, it is the stack pointer, named rsp
For all other instructions, it is a "zero" register, which returns 0 when read and discards data when written - named rzr (xzr, wzr)
``` 
but it is not really clear which instructions could actually use the registers, i.e. the documentation of loads was not clear about xzr being allowed or not.

When I get some more time, I will try to come up with some litmus tests and check what herd7 says.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
hernanponcedeleon: Done

ThomasHaas: I don't like your solution. Specifying the 0-register on construction of `ProgramBuilder` does nothing but introduce unnecessary state. Also, as soon as there are multiple such registers to replace (e.g. of different sizes), the solution won't work.
Just make the replacement function public and let it take the register name as input: it's less code, less state, and more general.
The litmus parser should just make sure to call that function right before `.build()`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusAArch64.java
ThomasHaas: I expected not much information to be out there, so that is fine :).
But I find the point about the stack pointer interesting. So in some contexts is actually represents a non-zero value.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
I clarified with Luc about the zero registers (see https://github.com/herd/herdtools7/issues/988). Memory orderings are preserved (thus we need to keep loads targeting zero registers, but use a dummy register), but there are no data/addr/ctrl dependencies. Thus, we need to replace `regWriters` of the form `zr <- e` with `dummy <- e ; zr <- 0`. Also, since zero registers are not necessarily initialized explicitly in litmus code, the pass adds `zr <- 0` before any `RegReader` using `zr`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -78,6 +81,16 @@ public Program build() {
...
+    public void setZeroRegister(String registerName) {
+        for(Thread thread : program.getThreads()) {
+            for(Register register : thread.getRegisters()) {
+                if(register.getName().equals(registerName)) {
+                    register.setAsZeroRegister();
+                }
+            }
+        }
+    }
+
```
ThomasHaas, line 93: Why are we back to such stateful approach?
Why not just provide the 0-register (its name) to the `ReplaceZeroRegisters` pass on construction and just invoke that pass inside the parser?
I think it is fair to say that this particular program transformation is something the parser or the program builder should do. You could even have the program builder take an optional set of passes as input that are run as post-processing (of the parsing) when building the final program. I would keep it simple and just invoke it directly from the parser though.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ReplaceZeroRegisters.java
```diff
@@ -0,0 +1,50 @@
...
+import com.dat3m.dartagnan.program.event.RegWriter;
+import com.dat3m.dartagnan.program.event.core.Local;
+
+/*
+    This pass replaces every RegWritter
+        zr <- e
+    targeting a zero register to
+        dummy <- e; zr <- 0
+    and for every RegReader reading zr, it inserts an event zr <- 0 before it
+*/
```
ThomasHaas, line 18: (1) Typo
(2) Since dependencies are not preserved, just replace all read-usages of `zr` directly by `0`. Alternatively, if you really want to keep the register and let SCCP do the replacement, just set it once to `0` at the beginning of the thread/function. You guarantee that the register is never overwritten, so no need to reset it all the time.

hernanponcedeleon:
Tested by @db7 on tests generated from real assembly. @ThomasHaas unless you have further comments I will merge

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ReplaceZeroRegisters.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public void run(Function function) {
+
+        for (RegWriter rw : function.getEvents(RegWriter.class)) {
+            if (rw.getResultRegister().isZeroRegister()) {
+                Register zr = rw.getResultRegister();
+                Register dummy = rw.getThread().newRegister(zr.getType());
+                rw.setResultRegister(dummy);
+                rw.insertAfter(EventFactory.newLocal(zr,
+                        ExpressionFactory.getInstance().makeZero((IntegerType) zr.getType())));
```
ThomasHaas, line 34: I think the replacement register should have a name to understand its intention, e.g., `__zeroRegDummy_<origRegName>` or `__ignored`.
Also, you can probably reuse a single replacement register (per function) since all usages should be type-consistent.
Of course, this assumes that you change the pass to take as input the replacement register (or rather its name), then it can perform per function a lookup on that register name, construct a same-type dummy and do the replacement.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ReplaceZeroRegisters.java
```diff
@@ -0,0 +1,50 @@
...
+        for (RegReader rr : function.getEvents(RegReader.class)) {
+            if (rr.getRegisterReads().stream().anyMatch(r -> r.register().isZeroRegister())) {
+                for (Register zr : rr.getRegisterReads().stream().filter(r -> r.register().isZeroRegister())
+                        .map(r -> r.register()).toList()) {
+                    Local lzr = EventFactory.newLocal(zr,
+                            ExpressionFactory.getInstance().makeZero((IntegerType) zr.getType()));
+                    rr.getPredecessor().insertAfter(lzr);
+                }
+            }
+        }
```
ThomasHaas, line 47: As mentioned before, either replace directly with `0` or simply construct a single instruction `zr <- 0` at the beginning of the function (after the ThreadStart event though).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
hernanponcedeleon: After the bad experience we had with the boogie parser, I thought we decided against doing any kind of magic in the parser. The parser should parse, the transformation pipeline should transform.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ReplaceZeroRegisters.java
hernanponcedeleon: Again, I don't want passes to do more magic than they should (thus I don't want to replace all read-usages by zero). I know CP will eventually do this, but
1) that is the purpose of CP, thus there is no magic here.
2) ATM we are not even applying CP to litmus tests (we only have zero register for litmus code), thus the local assignment is explicitly visible. I don't think we gain much performance by removing those local events, so I don't see the point.

In principle, having a single assignment at the beginning will not work because nothing prevents us of set two different registers as zero one (this was the whole purpose of your proposal of moving the `replaceZeroRegister` (now `setZeroRegister`) to the program builder). I could have code to get all different zero registers and initialize each of them once, but this will probably result in more complicated code with almost null benefit in terms of performance.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
ThomasHaas: Function inlining etc. are a totally different thing: they are semantic-preserving transformations that we do either out of necessity or optimizations.
Zero registers, however, are a different thing because what the parser produces does simply not represent the correct semantics and we require an extra pass to "make it correct".
If zero registers were a feature of our core language, I wouldn't mind, but I think they are not and I don't see why they should be. I mean, in this PR you can mark a register as "zero" but it will not be enforced at all unless the register was specifically marked so during parsing time.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ReplaceZeroRegisters.java
ThomasHaas: > Again, I don't want passes to do more magic than they should (thus I don't want to replace all read-usages by zero). I know CP will eventually do this, but
> 
>     1. that is the purpose of CP, thus there is no magic here.
> 
>     2. ATM we are not even applying CP to litmus tests (we only have zero register for litmus code), thus the local assignment is explicitly visible. I don't think we gain much performance by removing those local events, so I don't see the point.

I don't understand how it is more magic to just set the registers to zero once and for all. Isn't this the more canonical solution? I don't care about SCCP or any optimization/performance. In this case, whether the register is set again and again does not really matter, but it is a more convoluted solution with a few more disadvantages (requires more code, bloats up printing or even witness generation in the future).

I think it is a lot more magic to have a "isZero" flag that must be set in the parser for some pass to come along and enforce zeroness exactly once. Afterwards, the flag is useless and actually wrong, because we never again enforce that the register holds only zero values: I can create "zr <- 1" events and no one will complain that I write non-zero values to zero registers.
And this is IMO the biggest problem of this approach: you introduce state into the core logic (`Register` class) that far outlives its single use-case and thus opens up the possiblity of inconsistency (cause zero-ness is never enforced again).

> In principle, having a single assignment at the beginning will not work because nothing prevents us of set two different registers as zero one (this was the whole purpose of your proposal of moving the `replaceZeroRegister` (now `setZeroRegister`) to the program builder). I could have code to get all different zero registers and initialize each of them once, but this will probably result in more complicated code with almost null benefit in terms of performance.

I don't understand. What stops you from initializating every zero register to 0? Just pass multiple register names to the pass on creation, or just let the pass run multiple pass. In fact, the reason I say this is because it results in overall less code:
```Java
foreach (String zeroRegName : zeroRegNames) {
   Register zeroReg = func.getRegisterByName(zeroRegName);
   if (zeroReg != null) {
        Event initToZero = EventFactory.newLocal(zeroReg, expressions.getGeneralZero(zeroreg.getType()))
        func.getEntry().insertAfter(initToZero);
   }
}
```
This piece of code subsumes your doubly-nested RegReader loop while being more clear and concise (and no nested loops!). And it allows you to get rid of all the modifications to `Register` and elsewhere, easily saving 10+ lines of code.

hernanponcedeleon:
@ThomasHaas as we discussed, I ended up fully moving the transformation to the parsing rather than trying to add zero registers to the core language to avoid problems with all the other passes.

AFAIK the current behavior should be equivalent to herd7, e.g. this program
```
RISCV xzr
{
0:x6=x; 0:x8=y;
x=1; y=0;
}
 P0          ;
 lw x0,0(x6) ;
 sw x0,0(x8) ;
exists (y=0)
```
generates a witness where the read value is one, but the value is discarded

ThomasHaas:
Sounds good. I will approve once the CI is done.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
I am waiting for @db7 to finish some testing and if there are no problems, I will merge

db7:
> I am waiting for @db7 to finish some testing and if there are no problems, I will merge

The test is still running. The generation of test cases is still quite inefficient.

PR #744, opened 2024-09-28
Add OpenCL Model

tonghaining:


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorOpenCL.java
```diff
@@ -0,0 +1,137 @@
...
+import com.dat3m.dartagnan.program.event.core.RMWStore;
+import com.dat3m.dartagnan.program.event.lang.catomic.*;
+import com.dat3m.dartagnan.program.event.metadata.MemoryOrder;
+
+import java.util.List;
+
+import static com.dat3m.dartagnan.program.event.EventFactory.*;
+import static com.dat3m.dartagnan.program.event.EventFactory.OpenCL.newOpenCLBarrier;
+
+public class VisitorOpenCL extends VisitorBase {
```
hernanponcedeleon, line 24: Do we really need a new visitor? At first glance the only difference with the c11 one is the visitor method for the control barrier and how events are tagged. I feel those could be moved to the c11 one, or at least make this visitor extend the c11 one and thus avoiding having the reimplement some of the event visitors

hernanponcedeleon (commented):
A few comments based on a quick look to the changes. Will take a closer look next week

cat/opencl-herd.cat
```diff
@@ -0,0 +1,153 @@
...
+let sc3 = [F]; po; fr; po; [F]
+let sc4 =          fr; po; [F]
+let sc5 = [F]; po; fr
+
+let scp = (SC * SC) & (ghb | lhb | co | sc1 | sc2 | sc3 | sc4 | sc5)
+
+let sc6 = rf^-1; (ghb | lhb); [W]
+
+let unv = ~0
+let ifexists(s) = unv; [s]; unv
```
ThomasHaas, line 131: IIRC, we have `_` as set of all events. So you can use `let unv = _*_`.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -76,7 +76,7 @@ public class Dartagnan extends BaseOptions {
     private static final Logger logger = LogManager.getLogger(Dartagnan.class);
 
     private static final Set<String> supportedFormats =
-            ImmutableSet.copyOf(Arrays.asList(".litmus", ".c", ".i", ".ll", ".spv.dis"));
+            ImmutableSet.copyOf(Arrays.asList(".litmus", ".c", ".i", ".ll", ".spv.dis", ".cl"));
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
```diff
@@ -24,6 +25,7 @@ public Program parse(File file) throws Exception {
         if (needsClang(file)) {
             file = compileWithClang(file, "");
             file = applyLlvmPasses(file);
+            file = applyDemangling(file);
```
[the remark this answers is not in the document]
tonghaining: Removed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorOpenCL.java
tonghaining: Moved into C11 visitor.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CoreCodeVerification.java
```diff
@@ -38,7 +41,8 @@ public static CoreCodeVerification fromConfig(Configuration config) {
             Assume.class, Assert.class,
             ThreadCreate.class, ThreadArgument.class, ThreadStart.class,
             ControlBarrier.class, // For PTX and Vulkan
-            BeginAtomic.class, EndAtomic.class
+            BeginAtomic.class, EndAtomic.class,
+            C11Init.class, OpenCLInit.class, OpenCLBarrier.class // For OpenCL
```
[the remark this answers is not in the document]
tonghaining: Yeah the Inits seems to be unnecessary after adding the TagSet to MemoryObject. ControlBarrier is enough, I removed all these three classes.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -54,6 +56,12 @@ public class MemoryObject extends LeafExpressionBase<Type> {
     public boolean isThreadLocal() { return this.isThreadLocal; }
     public void setIsThreadLocal(boolean value) { this.isThreadLocal = value;}
 
+    public boolean isAtomic() { return this.isAtomic; }
+    public void setIsAtomic(boolean value) { this.isAtomic = value;}
+
+    public String getMemorySpace() { return this.memorySpace; }
+    public void setMemorySpace(String value) { this.memorySpace = value; }
```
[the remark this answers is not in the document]
tonghaining: TagSet Added :D

Why "memory scope"?

cat/opencl-herd.cat
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -271,13 +272,27 @@ public static String extractLoadMo(String cMo) {
     }
 
     // =============================================================================================
-    // =========================================== PTX =============================================
```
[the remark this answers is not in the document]
tonghaining: Splitted back.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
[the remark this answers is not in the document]
ThomasHaas: I just saw the `memory_scope` parameters in the litmus code and thought those were related to this. In the end, I don't know what is "correct" :P

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorOpenCL.java
```diff
@@ -0,0 +1,137 @@
...
+    public List<Event> visitOpenCLBarrier(OpenCLBarrier e) {
+        Event entryFence = EventFactory.newControlBarrier(e.getName() + "_entry", e.getId());
+        entryFence.addTags(Tag.OpenCL.ENTRY_FENCE, C11.MO_RELEASE);
+        Event exitFence = EventFactory.newControlBarrier(e.getName() + "_exit", e.getId());
+        exitFence.addTags(Tag.OpenCL.EXIT_FENCE, C11.MO_ACQUIRE);
+        return tagList(e, eventSequence(
+                entryFence,
+                newOpenCLBarrier(e.getName(), e.getId()),
+                exitFence
+        ));
```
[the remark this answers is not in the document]
ThomasHaas: Ok, I saw that you removed it already.

cat/opencl-herd.cat
```diff
@@ -0,0 +1,153 @@
...
+(*******************)
+(* Synchronisation *)
+(*******************)
+
+let Acq = (ACQ | SC | ACQ_REL) & (R | F | RMW)
+let Rel = (REL | SC | ACQ_REL) & (W | F | RMW)
+
+(* Release sequence *)
+let rs = (_ * RMW) | wi
+let max_co = co? & rs & ~((co & ~rs) ; co)
```
ThomasHaas, line 46: Just a side-note: I'm not a big fan of the complementation if differences are enough: `a & ~b = a \ b`.
I know that herd's version uses the complementation so we can keep it here, but maybe we should consider adding a Wmm transformation pass to rewrite the model automatically.
Of course, the pass can be added in a different PR. I just wanna mention it here :).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -252,7 +288,14 @@ public Expression visitReAtomicFetchOp(LitmusCParser.ReAtomicFetchOpContext ctx)
...
         Expression value = returnExpressionOrOne(ctx.value);
-        Event event = EventFactory.Atomic.newFetchOp(register, getAddress(ctx.address), value, ctx.op, ctx.c11Mo().mo);
+        Expression address = getAddress(ctx.address);
+        Event event = EventFactory.Atomic.newFetchOp(register, address, value, ctx.op, ctx.c11Mo().mo);
+        addMemoryObjectTags(event, address);
+        if (ctx.openCLScope() != null) {
+            event.addTags(ctx.openCLScope().scope);
+        } else {
+            event.addTags(Tag.OpenCL.DEFAULT_SCOPE);
+        }
```
ThomasHaas, line 298: Does this mean regular C11 litmus code always comes with OpenCL tagging? I don't think this will cause issues, but I'm wondering if the parser should already be aware of the target architecture to prevent unnecessary tagging.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -577,4 +734,17 @@ private Expression assignToReturnRegister(Register register, Expression value) {
...
         return value;
     }
+
+    private void addMemoryObjectTags(Event event, Expression address) {
+        if (address instanceof Register reg) {
+            MemoryObject object = programBuilder.getLocFromReg(reg);
+            if (object != null) {
+                event.addTags(object.getFeatureTags());
+            }
+        } else if (address instanceof BinaryExpressionBase<?,?> binExpr) {
```
ThomasHaas, line 744: Don't check against `BinaryExpressionBase` but rather against the interface `BinaryExpression`. All the `...Base` classes are meant as implementation aids and not as interface (I even gave them a special `NoInterface` attribute to mark this), meaning no code should ever refer to them.
Btw. is there a reason that this is restricted to binary expressions?

tonghaining: Updated to all expression. It was I saw in one tests binary expression is used, there is no reason to restrict it to binary expressions XD

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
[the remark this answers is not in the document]
hernanponcedeleon: It is still there

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
ThomasHaas: See @hernanponcedeleon comment about the OpenCL header to distinguish between OpenCL and pure C11.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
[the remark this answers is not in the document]
tonghaining: Both are removed now

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -44,6 +45,11 @@ public class ProgramBuilder {
     private final Map<Integer, Function> id2FunctionsMap = new HashMap<>();
     private final Map<Integer, Map<String, Label>> fid2LabelsMap = new HashMap<>();
     private final Map<String, MemoryObject> locations = new HashMap<>();
+    private final Map<Register, MemoryObject> reg2LocMap = new HashMap<>();
+    private final Map<Integer, Map<String, IntegerType>> id2RegTypeMap = new HashMap<>();
+    private final Map<Integer, Map<String, Expression>> id2RegConstMap = new HashMap<>();
+    private final Map<Integer, Map<String, String>> id2RegLocPtrMap = new HashMap<>();
+    private final Map<Integer, Map<String, String>> id2RegLocValMap = new HashMap<>();
```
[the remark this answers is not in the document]
tonghaining: Updated

tonghaining:
Sounds good to me.
> * For Litmus having OpenCL in the first line, we add scopes, address spaces, and any other opencl tag which are explicitly stated in the syntax and we default to some specific ones when not explicitly stated. This should make the porting og C litmus so OpenCL easy (just change the first line and add @wg0, dev0 to all tests). Otherwise, we request explicit mention of the needed tags. This might make the porting harder and I am not sure it has any benefit.

One more change we need is to add "global" for all variables for C litmus

cat/c11-partialsc.cat
```diff
@@ -0,0 +1,109 @@
...
+let asw = IW * (M \ IW)
+let sb = po
+let mo = co & (~NAL * ~NAL)
+
+let cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
+let crel = REL | (SC & (W | F)) | ACQ_REL
+let ccon = R & CON
+
+// To be compatible with reads from uninit memory, we use our default definition of fr
+//let fr = rf^-1 ; mo
```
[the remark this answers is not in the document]
hernanponcedeleon: The change below should fix this, right?
```
// To be compatible with reads from uninit memory, we use our default definition of fr
//let fr = rf^-1 ; mo
include "basic.cat" // Remove the previous include!!!
let fr = fr & (_ * ~NAL)
```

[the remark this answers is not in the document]
ThomasHaas: I don't know what tagging is necessary. Anyway, I think you don't have mallocs for now (it's litmus code) so init events are there.

dartagnan/src/main/antlr4/LitmusC.g4
```diff
@@ -239,6 +261,28 @@ c11Mo returns [String mo]
...
+openCLFenceFlags
+    :   openCLFenceFlag Bar openCLFenceFlag
+    |   openCLFenceFlag
+    ;
+
+openCLFenceFlag returns [String flag]
+    :   OpenCLFenceFlagGL    {$flag = OpenCL.GLOBAL_SPACE;}
+    |   OpenCLFenceFlagLC    {$flag = OpenCL.LOCAL_SPACE;}
+    ;
+
```
hernanponcedeleon, line 285: The rules refers to `flag`, but the return value to addresses. Would it make sense to merge this somehow, i.e. use `openCLSpace ` wherever we are using `openCLFenceFlag `?

dartagnan/src/main/antlr4/LitmusC.g4
```diff
@@ -239,6 +261,28 @@ c11Mo returns [String mo]
...
+openCLSpace returns [String space]
+    :   OpenCLGlobalSpace  {$space = OpenCL.GLOBAL_SPACE;}
+    |   OpenCLLocalSpace   {$space = OpenCL.LOCAL_SPACE;}
+    ;
+
+openCLScope returns [String scope]
+    :   OpenCLMemoryScopeWI    {$scope = OpenCL.WORK_ITEM;}
+    |   OpenCLMemoryScopeWG    {$scope = OpenCL.WORK_GROUP;}
+    |   OpenCLMemoryScopeDEV   {$scope = OpenCL.DEVICE;}
+    |   OpenCLMemoryScopeALL   {$scope = OpenCL.ALL;}
```
hernanponcedeleon, line 273: Is all intended for host/device share memory? We are not using this at all at the moment, right? (I am not suggesting getting rid of it if the model uses it)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -274,29 +275,33 @@ public Label getEndOfThreadLabel(int tid) {
...
+            case VULKAN -> ScopeHierarchy.ScopeHierarchyForVulkan(scopeIds[0], scopeIds[1], scopeIds[2]);
+            case OPENCL -> ScopeHierarchy.ScopeHierarchyForOpenCL(scopeIds[0], scopeIds[1]);
+            default -> throw new UnsupportedOperationException("Unsupported architecture: " + arch);
+        };
 
-    public void newScopedThread(Arch arch, String name, int id, int ...scopeIds) {
         if(id2FunctionsMap.containsKey(id)) {
-            throw new MalformedProgramException("Function or thread with id " + id + " already exists.");
+            Thread thread = (Thread) id2FunctionsMap.get(id);
+            thread.setScopeHierarchy(scopeHierarchy);
```
hernanponcedeleon, line 289: Are there cases where we really want to set the scope on an already existing thread?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -346,4 +351,14 @@ public void addSwwPairThreads(int threadId0, int threadId1) {
...
+
+    // ----------------------------------------------------------------------------------------------------------------
+    // OpenCL
+    public void setReg2LocMap(Register reg, MemoryObject loc) {
+        reg2LocMap.put(reg, loc);
+    }
+
+    public MemoryObject getLocFromReg(Register reg) {
+        return reg2LocMap.get(reg);
+    }
```
hernanponcedeleon, line 363: Where is this being used? If it is used for some initialization of the form `r0:&a` can we somehow create some local events at the beginning of the corresponding thread with `r0 <- &a`. I think this is what we are doing with threads having parameters of the form `P0(atomic_t *x, atomic_t *y)` ... they are transformed to `x <- &x` where the left hand side is a register.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -46,6 +47,9 @@ public VisitorLitmusC(){
     public Program visitMain(LitmusCParser.MainContext ctx) {
         //FIXME: We should visit thread declarations before variable declarations
         // because variable declaration refer to threads.
+        if (ctx.LitmusLanguage().getText().equals("OPENCL")) {
+            isOpenCL = true;
+        }
```
hernanponcedeleon, line 52:
```suggestion
            isOpenCL = ctx.LitmusLanguage().getText().equals("OPENCL");
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -165,21 +173,44 @@ public Object visitThread(LitmusCParser.ThreadContext ctx) {
...
+        boolean atomicity = ctx.pointerTypeSpecifier().atomicTypeSpecifier() != null
+                || ctx.pointerTypeSpecifier().basicTypeSpecifier().AtomicInt() != null;
+        if (!atomicity) {
+            object.addFeatureTag(C11.NON_ATOMIC_LOCATION);
+        }
+        if (isOpenCL && ctx.pointerTypeSpecifier().openCLSpace() != null) {
+            object.addFeatureTag(ctx.pointerTypeSpecifier().openCLSpace().space);
+        }
+        programBuilder.setReg2LocMap(register, object);
+        programBuilder.addChild(currentThread, EventFactory.newLocal(register, object));
```
hernanponcedeleon, line 210: This is what I was suggesting above. Why do we even need the reg2Loc map if we are already doing this?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -252,7 +283,10 @@ public Expression visitReAtomicFetchOp(LitmusCParser.ReAtomicFetchOpContext ctx)
     public Expression visitC11AtomicOp(LitmusCParser.C11AtomicOpContext ctx) {
         Register register = getReturnRegister(true);
         Expression value = returnExpressionOrOne(ctx.value);
-        Event event = EventFactory.Atomic.newFetchOp(register, getAddress(ctx.address), value, ctx.op, ctx.c11Mo().mo);
+        Expression address = getAddress(ctx.address);
+        Event event = EventFactory.Atomic.newFetchOp(register, address, value, ctx.op, ctx.c11Mo().mo);
+        addMemoryObjectTags(event, address);
+        addScopeTag(event, ctx.openCLScope());
```
hernanponcedeleon, line 289: Shouldn't this be conditional on the instruction actually having a scope? I think right now this might be `null`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -577,4 +724,27 @@ private Expression assignToReturnRegister(Register register, Expression value) {
...
+            MemoryObject object = programBuilder.getLocFromReg(reg);
+            if (object != null) {
+                event.addTags(object.getFeatureTags());
+            }
+        } else {
+            for (Expression add : address.getOperands()) {
+                addMemoryObjectTags(event, add);
+            }
+        }
+    }
```
hernanponcedeleon, line 739: I don't really understand what this method is about

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopeHierarchy.java
```diff
@@ -32,6 +32,14 @@ public static ScopeHierarchy ScopeHierarchyForPTX(int gpu, int cta) {
         return scopeHierarchy;
     }
 
+    public static ScopeHierarchy ScopeHierarchyForOpenCL(int dev, int wg) {
+        ScopeHierarchy scopeHierarchy = new ScopeHierarchy();
+        scopeHierarchy.scopeIds.put(Tag.OpenCL.ALL, 0);
+        scopeHierarchy.scopeIds.put(Tag.OpenCL.DEVICE, dev);
+        scopeHierarchy.scopeIds.put(Tag.OpenCL.WORK_GROUP, wg);
+        return scopeHierarchy;
+    }
```
hernanponcedeleon, line 41: Somewhere you were defining a hierarchy with two elements (dev and wg) and now there are three. Shouldn't there be a match between that in this?

dartagnan/src/main/java/com/dat3m/dartagnan/program/Thread.java
```diff
@@ -56,6 +56,10 @@ public Set<Thread> getSyncSet() {
         return syncSet.get();
     }
 
+    public void setScopeHierarchy(ScopeHierarchy scopeHierarchy) {
+        this.scopeHierarchy = Optional.of(scopeHierarchy);
+    }
+
```
hernanponcedeleon, line 62: Related to a previous comment: do we really need to explicitly add the hierarchy to a thread that did not have one? This sounds to me like we are doing things wrong: either the hierarchy is set when creating the thread, of we don't need a hierarcy at all.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
```diff
@@ -118,12 +120,24 @@ public List<Event> visitAtomicXchg(AtomicXchg e) {
...
+        Event entryFence = EventFactory.newControlBarrier(e.getName() + "_entry", e.getId());
+        entryFence.addTags(Tag.OpenCL.ENTRY_FENCE, C11.MO_RELEASE);
+        Event exitFence = EventFactory.newControlBarrier(e.getName() + "_exit", e.getId());
+        exitFence.addTags(Tag.OpenCL.EXIT_FENCE, C11.MO_ACQUIRE);
+        return tagList(e, eventSequence(
+                entryFence,
+                exitFence
+        ));
+    }
+
```
hernanponcedeleon, line 140: What is the point of having this split into an entry and exit barrier? Wouldn't we get the same behavior by having a single barrier tagged both as entry/exit and rel/acq?

dartagnan/src/test/java/com/dat3m/dartagnan/comparison/AbstractComparisonTest.java
```diff
@@ -0,0 +1,133 @@
...
+    protected abstract Provider<Arch> getTargetProvider();
+    protected Provider<Wmm> getTargetWmmProvider() {
+        return Providers.createWmmFromArch(getTargetProvider());
+    }
+    protected Provider<EnumSet<Property>> getPropertyProvider() {
+        return Provider.fromSupplier(() -> EnumSet.of(Property.PROGRAM_SPEC));
+    }
+    protected long getTimeout() { return 10000; }
+    // List of tests that are known to show bugs in the compilation scheme and thus the expected result should be FAIL instead of PASS
+    protected List<String> getCompilationBreakers() { return emptyList(); }
```
hernanponcedeleon, line 79: This is probably a copy/paste error from the test that compares a C-like program and its compiled version. This can be probably be removed.

dartagnan/src/test/resources/dartagnan-skip.csv
```diff
@@ -3,6 +3,7 @@
 
 // sb (po) is different for C11 than hardware models. In particular it is not a total order (due to C11 not guaranteeing the evaluation order of function parameters). See https://github.com/herd/herdtools7/issues/394 for a discussion
 litmus/C11/auto/linearisation.litmus
+litmus/OPENCL/portedFromC11//auto/linearisation.litmus
```
hernanponcedeleon, line 6: Remove extra `/`

hernanponcedeleon (commented):
Comment about the code. I still have not check in detail the cat models and the tests.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
ThomasHaas: From what I understand, this code tags events that access specially-marked locations. I.e., if location `x` is marked, then `load(&x)` gets marked as well. The issue is that a load may refer to `x` like this: `r <- &x; load(r)` in which case the accessed location is obscured by the register (and this is I think also the reason for why `getLocFromReg` exists).

Now, this approach is only feasible if there are no unclear aliasing relations, i.e., it is statically known which memory event accesses which location. This might be viable for litmus code but I don't think we can extend this to general OpenCL code.
A workaround might be to compute the tags within the CAT by observing the init event that belongs to the address:
```
let myDynamicTag = [range([IW & myStaticTag]; loc)] 
// myDynamicTag relates events to itself that access an address whose init event is marked myStaticTag
```
This way only only init events need to get static tagging.
In case there are no init events, e.g. due to dynamic allocation, then there is also no tagging. I don't know if OpenCL allows to dynamically allocate tagged memory.

hernanponcedeleon: This seems to be analogous to what we have in real spirv code. There were some tags (related to storage classes IIRC) that were actually related to the memory object being accessed. If this is the case, we should probably handle this in the same way it is currently handled by the spirv code.

ThomasHaas: I took a quick glance at the SPIRV code and it seems that the parser relies on a sort of explicit tagging of pointers used in memory accesses. Put differently, in SPIRV pointers are explicitly tagged rather than implicitly by the memory they point to. This also suggests that you could technically have two same-address pointers but with different storage classes (might be UB though?). In other words, storage classes are not per address but per access (like access types) and so a memory access has the same storage class independent of the actually accessed address.
I think this makes sense, because the compiler generates different assembly based on the storage classes and so it must statically be able to infer them.

I don't know if the tagging @tonghaining implemented here is supposed to represent the same kind of thing.
Basically the question is: is the tagging supposed to model different assembly generated for different code (-> static tagging via tagged pointers) or is it supposed to model the hardware doing different things (for the same instruction) depending on the internal memory layout (-> dynamic tagging based on accessed address)?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
tonghaining: This is for the case e.g. we have declaration: 
`{0:r1 = 0;}`
which will create the thread P0 while initializing r1. It happens before we visiting threads and getting to know their scopeHierarchy.

dartagnan/src/main/antlr4/LitmusC.g4
tonghaining: We are using them now, we have litmus from herd/overhauling testing about memory scope.

dartagnan/src/main/antlr4/LitmusC.g4
tonghaining: Do you mean changing the returning `flag` to `space `to be consistent with `openCLSpace`? Or using `openCLSpace` to replace `openCLFenceFlag`? 
`openCLSpace` is `global`/`local` while `openCLFenceFlag` is `CLK_GLOBAL_MEM_FENCE`/`CLK_LOCAL_MEM_FENCE`, is there a good way to do that?

dartagnan/src/main/antlr4/LitmusC.g4
```diff
@@ -18,22 +18,30 @@ variableDeclaratorList
...
-    |   typeSpecifier? varName (Equals Ast? (Amp? varName | LPar Amp? varName RPar))?                                   # globalDeclaratorLocationLocation
-    |   typeSpecifier? t = threadId Colon n = varName (Equals Ast? (Amp? varName | LPar Amp? varName RPar))?            # globalDeclaratorRegisterLocation
-    |   typeSpecifier? varName LBracket DigitSequence? RBracket (Equals initArray)?                                     # globalDeclaratorArray
+    |   typeSpecifier? t = threadId Colon n = varName (Equals initConstantValue)?                                                           # globalDeclaratorRegister
+    |   typeSpecifier? varName (Equals Ast? (Amp? varName | LPar Amp? varName RPar))?                                                       # globalDeclaratorLocationLocation
+    |   typeSpecifier? t = threadId Colon n = varName (Equals Ast? (Amp? varName | LPar Amp? varName RPar))?                                # globalDeclaratorRegisterLocation
+    |   typeSpecifier? varName LBracket DigitSequence? RBracket (Equals initArray)?                                                         # globalDeclaratorArray
```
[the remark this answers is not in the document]
hernanponcedeleon: Ok, I was checking the diff and not the actual file. The file actually looks better after your changes so keep them.

dartagnan/src/main/antlr4/LitmusC.g4
hernanponcedeleon: > `openCLFenceFlag` is `CLK_GLOBAL_MEM_FENCE`/`CLK_LOCAL_MEM_FENCE`, is there a good way to do that?

Shouldn't then this be 

```
openCLFenceFlag returns [String flag]
    :   OpenCLFenceFlagGL    {$flag = CLK_GLOBAL_MEM_FENCE;}
    |   OpenCLFenceFlagLC    {$flag = CLK_LOCAL_MEM_FENCE;}
    ;
```
???

dartagnan/src/main/antlr4/LitmusC.g4
hernanponcedeleon: What is the difference between `device` and `all`? I would expect the first to allow all threads in the same device to communicate, meaning that only `all` would mean we are talking about sync between threads in different devices, or host/device. I don't think our litmus format allows us to define host code. Can we specify that threads belong to different devices?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
hernanponcedeleon: The situation should be similar for PTX or Vulkan visitor. How are we handling similar situations there? Are threads visited before initialization?

tonghaining: I think we also did it wrong for them. Their declaration will throw a 'thread not exist' exception during register initialization. We did not have a relevant test for this scenario for either PTX or Vulkan visitor.

hernanponcedeleon: If what you say is true, I would expect an exception in code like [this](https://github.com/hernanponcedeleon/Dat3M/blob/2d6ef05b23f6921d8effe9cee84749071a600cf8/litmus/PTX/Nvidia/Atom-SB.litmus#L5)

tonghaining: Ahh I was wrong, in PTX and Vulkan visitor, we visit the thread declarators before visiting the variable declarator.

dartagnan/src/main/antlr4/LitmusC.g4
tonghaining: Why do we return flags like that? I meant the token of `OpenCLFenceFlagGL` is `CLK_GLOBAL_MEM_FENCE`, and `OpenCLFenceFlagLC` is `CLK_LOCAL_MEM_FENCE`.

hernanponcedeleon: Good point. I guess this is an abuse of code coming from "similar" situations like c11 memory orders. However, it makes sense for c11 because there is a difference between the tokens and the tags, e.g. `memory_order_acquire` and `ACQ`.

You should get rid of the flags and use the token (please also fix all other cases where we have the same issue)

dartagnan/src/main/antlr4/LitmusC.g4
tonghaining: Yes now the opencl litmus format only allow device code. We are using the `P0@wg 0, dev 0` to specify which device the thread belong to.

dartagnan/src/test/resources/dartagnan-skip.csv
tonghaining: Removed

dartagnan/src/test/java/com/dat3m/dartagnan/comparison/AbstractComparisonTest.java
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
ThomasHaas: Ok, then both memory and pointers have tags and these are different. Memory accesses inherit the tags of their pointers and so pointers need to carry this extra information like they do for SPIRV.
Interestingly, this tag inheritance cannot always be done statically because of [genericAddressSpace](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/genericAddressSpace.html).

Now, assuming that an object in space X is only accessible with pointers to space X (and otherwise it is UB), there is really not much difference between computing the tagging based on the pointer type vs. based on the actually accessed address (since they will match). The former can be done statically in many cases except for "genericAddressSpace" while the latter is always fully dynamic.
I think the dynamic tagging is easier to implement in CAT as I described above and does not require special pointer types(*). On the other hand, we cannot detect UB based on mismatching tagging between pointer and memory (constructing such a mismatch in real OpenCL code is likely impossible anyways).

(*) We will eventually want to have proper pointer types with address spaces, but this is a more invasive and complex change.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
hernanponcedeleon: This looks much better now

dartagnan/src/main/java/com/dat3m/dartagnan/program/ScopeHierarchy.java
hernanponcedeleon: I just noticed this is analogous to `ScopeHierarchyForVulkan`, thus it is fine

dartagnan/src/main/java/com/dat3m/dartagnan/program/Thread.java
tonghaining: The scopeHierarchy is changed back to final now after reordering the visit of thread declarator and variable declarator

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
tonghaining: In addScopeTag() it checks if the scope context is `null`, if it's `null` and `isOpenCL`, we add the default scope.

dartagnan/src/main/antlr4/LitmusC.g4
tonghaining: I think the process is similar for OpenCL? 
For token `global`(from memory event) and `CLK_GLOBAL_MEM_FENCE`(from fence) we should add `GLOBAL` tag;
for token `local` and `CLK_LOCAL_MEM_FENCE` we should add `LOCAL` tag.

hernanponcedeleon: I don't understand what you mean

tonghaining: I meant for example we got this one fence:
```atomic_work_item_fence(CLK_GLOBAL_MEM_FENCE,memory_order_release,memory_scope_device);```
Then we should add `GLOBAL` tag to it in our visitor, and `CLK_GLOBAL_MEM_FENCE` is not a tag.

Comment (hernanponcedeleon) [reply dartagnan/src/main/antlr4/LitmusC.g4:285]: Now I am confused. Your current visitor seems not to be adding `GLOBAL` (which should claim should be added) but it is actually adding `CLK_GLOBAL_MEM_FENCE` which should not (the situation is even worse, because if the instruction uses `CLK_GLOBAL_MEM_FENCE  | CLK_LOCAL_MEM_FENCE `, it adds both thins as a single string/tag)

Comment (tonghaining) [reply dartagnan/src/main/antlr4/LitmusC.g4:285]: It is using the `OpenCLFenceFlagContext().flag` in the current visitor, which I think should be `GLOBAL`/`LOCAL`. Why do you say it adds both as a single string when fence has both spaces? I think they were added one by one in iteration.

Comment (hernanponcedeleon) [reply dartagnan/src/main/antlr4/LitmusC.g4:285]: > It is using the `OpenCLFenceFlagContext().flag` in the current visitor, which I think should be `GLOBAL`/`LOCAL`. 

Ok, I now understand why both `openCLSpace` and `openCLFenceFlag` make use of `OpenCL.GLOBAL_SPACE/OpenCL.LOCAL_SPACE`. However, for the first one we can simply used the token and get rid of the `{$space = OpenCL.GLOBAL_SPACE;}` part, right?

> Why do you say it adds both as a single string when fence has both spaces? I think they were added one by one in iteration.

I misread the visitor, I thought `openCLFenceFlag()` was being used in `fence.addTags()` instead of `flagCtx.flag`. Sorry about the confusion

File: dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java (hunk 24 above)
Comment (hernanponcedeleon) [reply dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java:140]: Even the natural language spec talks about two fences, so better to keep it like this.

File: cat/c11.cat [hunk 28]
```diff
@@ -15,18 +15,22 @@ C "C++11"
...
 include "c11_los.cat"
 
+// dynamic_tag relates events to itself that access an address whose init event is marked static_tag
+let dynamic_NAL = [range([IW & NAL]; loc)]
+let dynamic_AL = [range([IW & ~NAL]; loc)]
+
 let asw = IW * (M \ IW)
 let sb = po
-let mo = co & (~NAL * ~NAL)
+let mo = dynamic_NAL; co; dynamic_NAL
```
hernanponcedeleon, line 24: shouldn't this be `dynamic_AL`?

I think it would be better to define a parametric definition
```
// dynamic_tag relates events to itself that access an address whose init event is marked X
let dynamic_tag(X) = [range([IW & X]; loc)]
```

cat/c11-partialsc.cat
```diff
@@ -0,0 +1,110 @@
...
+include "basic.cat"
+let fr = fr & (_ * ~NAL)
```
hernanponcedeleon, line 29: Now that I see this again, I think it would be better to replace all these lines by `let fr = rf^-1;mo | ([R] \ [range(rf)]);loc;[W]`

cat/opencl.cat
```diff
@@ -92,8 +98,8 @@ irreflexive rf; (ghb | lhb) as O-Rf
 let vis(hb) = (W * R) & hb & loc & ~((hb & loc); [W]; hb)
 
 (* A non-atomic load can only read from a store that is visible. *)
-empty (rf; [NAL & GLOBAL])\vis(ghb) as O-NaRfG
-empty (rf; [NAL & LOCAL])\vis(lhb) as O-NaRfL
+empty (rf; dynamic_NAL & dynamic_GLOBAL)\vis(ghb) as O-NaRfG
+empty (rf; dynamic_NAL & dynamic_LOCAL)\vis(lhb) as O-NaRfL
```
hernanponcedeleon, line 102: I don't remember what is the precedence of `;` wrt `&`, but better to use `()` here

cat/opencl.cat
```diff
@@ -49,27 +55,27 @@ let rs_prime = wi | (unv; [R & W])
...
 
 (* Release-acquire synchronisation *)
-let ra_sw(r) = ([r & Rel]; Fsb?; [W \ WI]; rs?; [r]; rf; [R \ WI]; sbF?; [Acq & r]) & incl & ~wi
+let ra_sw(r) = (r & [Rel]; Fsb?; [W \ WI]; rs?; r; rf; [R \ WI]; sbF?; [Acq] & r) & incl & ~wi
 
 (* Barrier synchronisation *)
 // in OpenCL a barrier results in two fence operations: entry and exit fences:
 // https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_work_group_functions
-let bar_sw(r) =  (EF * XF) & syncbar & ~wi & swg & (r * r)
+let bar_sw(r) = r & [EF]; syncbar & ~wi & swg; r & [XF]
```
hernanponcedeleon, line 63: I think you need parenthesis

tonghaining: Added

cat/opencl.cat
tonghaining: Added

cat/c11.cat
tonghaining: Updated

cat/c11-partialsc.cat
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
```diff
@@ -87,7 +88,11 @@ private void createInitEvents(Program program) {
                 final String threadName = "Init_" + nextThreadId;
                 final Thread thread = new Thread(threadName, initThreadType, paramNames, nextThreadId,
                         EventFactory.newThreadStart(null));
-                thread.append(EventFactory.newInit(memObj, field));
+                if (program.getArch() == Arch.C11 || program.getArch() == Arch.OPENCL) {
+                    thread.append(EventFactory.newC11Init(memObj, field));
+                } else {
+                    thread.append(EventFactory.newInit(memObj, field));
+                }
```
hernanponcedeleon, line 95: I think I would rather change `newC11Init(x,y)` by `newInit(x,y, boolean)`, implement `newCInit(x,y) = `newInit(x,y,false)` and avoid the if here.

dartagnan/src/test/java/com/dat3m/dartagnan/comparison/C11AndOpenCLTest.java
```diff
@@ -0,0 +1,43 @@
...
+import com.dat3m.dartagnan.utils.rules.Provider;
+import com.dat3m.dartagnan.utils.rules.Providers;
+import com.dat3m.dartagnan.wmm.Wmm;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+
+@RunWith(Parameterized.class)
+public class C11AndOpenCLTest extends AbstractComparisonTest {
```
hernanponcedeleon, line 13: You should also add a test that compares these test for races. You can simply create a new class that extends this one and override the propertyProvider method (assuming `LitmusOpenClRacesTest` does not run on litmus/OPENCL/portedFromC11/)

litmus/OPENCL/herd/global_barrier.litmus
```diff
@@ -0,0 +1,120 @@
...
+  atomic_store_explicit(f0, 1, memory_order_release);
+
+  atomic_work_item_fence(CLK_LOCAL_MEM_FENCE, memory_order_relaxed, memory_scope_work_group);
+
+  atomic_work_item_fence(CLK_LOCAL_MEM_FENCE, memory_order_relaxed, memory_scope_work_group);
+
+  atomic_work_item_fence(CLK_LOCAL_MEM_FENCE, memory_order_relaxed, memory_scope_work_group);
+
+  atomic_work_item_fence(CLK_LOCAL_MEM_FENCE, memory_order_relaxed, memory_scope_work_group);
+
```
hernanponcedeleon, line 25: Even if the original test has the fence repeated several times, this seems useless and we should clean the tests

cat/c11-partialsc.cat
```diff
@@ -0,0 +1,113 @@
...
+let sb = po
+let mo = dynamic_tag(~NAL); co; dynamic_tag(~NAL)
+
+let cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
+let crel = REL | (SC & (W | F)) | ACQ_REL
+let ccon = R & CON
+
+// To be compatible with reads from uninit memory, we use our default definition of fr
+// let fr = rf^-1 ; mo
+include "basic.cat"
```
hernanponcedeleon, line 31: Remove these two

cat/c11-partialsc.cat
```diff
@@ -0,0 +1,113 @@
...
+let vis = ([W] ; hbl ; [R]) & ~(hbl; [W]; hbl)
+
+(* consistent_atomic_rf *)
+irreflexive (rf ; hb) as Rf
+
+(* consistent_non_atomic_rf  *)
+empty ((rf ; dynamic_tag(NAL)) \ vis) as NaRf
+
+(* Consistency of RMWs *)
+empty rmw & (fre; coe) as atomic
```
hernanponcedeleon, line 78: Move this below the comment below

cat/c11.cat
```diff
@@ -1,71 +1,60 @@
...
+let mo = dynamic_tag(~NAL); co; dynamic_tag(~NAL)
 
 let cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
 let crel = REL | (SC & (W | F)) | ACQ_REL
 let ccon = R & CON
 
 // To be compatible with reads from uninit memory, we use our default definition of fr
-//let fr = rf^-1 ; mo
+// let fr = rf^-1 ; mo
+include "basic.cat"
```
hernanponcedeleon, line 31: Remove

cat/c11-partialsc.cat
```diff
@@ -0,0 +1,113 @@
...
+
+(* locks_only_consistent_locks *)
+// irreflexive ([LS] ; lo^-1 ; [LS] ; ~(lo ; [UL] ; lo)) as Lo2
+
+(* data_races *)
+let cnf = ((W * _) | (_ * W)) & loc
+let dr = ext & (cnf \ hb \ (hb^-1) \ (A * A))
+
+(* unsequenced_races *)
+let ur = int & ((W * M) | (M * W)) & loc & ~id & ~(sb+) & ~((sb+)^-1)
```
hernanponcedeleon, line 95: Aren't you missing some `~flag` related to races? Same for c11.cat

cat/opencl.cat
```diff
@@ -0,0 +1,128 @@
...
+// LOCAL: local memory
+// EF: entry fence of barrier
+// XF: exit fence of barrier
+
+// dynamic_tag relates events to itself that access an address whose init event is marked X or Fence tagged with X
+let dynamic_tag(X) = [range([IW & X]; loc)] | [X & F]
+
+let mo = dynamic_tag(~NAL); co; dynamic_tag(~NAL)
+let sb = po
+let rb = (rf^-1; mo) \ id
```
hernanponcedeleon, line 32: You need to add the thing to deal with memory not having initial events and you don't need the \id ... this is for handling RMW as a single event.

In short, replace this by the definition of `fr` in c11.cat

cat/opencl.cat
```diff
@@ -0,0 +1,128 @@
...
+
+(* Inclusive scopes *)
+let incl = (swg & (WG * WG)) | (sdv & (DV * DV)) | (ALL * ALL)
+
+(*******************)
+(* Synchronisation *)
+(*******************)
+
+let Acq = (ACQ | SC | ACQ_REL) & (R | F)
+let Rel = (REL | SC | ACQ_REL) & (W | F)
```
hernanponcedeleon, line 44: A like the names `Acq` and `Rel`. Can you please rename `cacq` and etc in the c11 variants to make them patch this?

cat/opencl.cat
```diff
@@ -0,0 +1,128 @@
...
+(* Release-acquire synchronisation *)
+let ra_sw(r) = ((r & [Rel]); Fsb?; [W \ WI]; rs?; r; rf; [R \ WI]; sbF?; ([Acq] & r)) & incl & ~wi
+
+(* Barrier synchronisation *)
+// in OpenCL a barrier results in two fence operations: entry and exit fences:
+// https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_work_group_functions
+let bar_sw(r) = (r & [EF]); syncbar & ~wi & swg; (r & [XF])
+
+(* Allowed to synchronise on the other region *)
+let scf = (SC * SC) | ((dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)); F*F;(dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)))
```
hernanponcedeleon, line 63: The original definition was `let scf = mo_sc^2 | (G & L & F)^2`, since your definition of dynamic tag includes fences, I believe you can use `(dynamic_tag(GLOBAL) & dynamic_tag(LOCAL) * (dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)`

cat/opencl.cat
```diff
@@ -0,0 +1,128 @@
...
+
+(* Visible side effects *)
+let vis(hb) = (W * R) & hb & loc & ~((hb & loc); [W]; hb)
+
+(* A non-atomic load can only read from a store that is visible. *)
+empty (rf; (dynamic_tag(NAL) & dynamic_tag(GLOBAL))) \ vis(ghb) as O-NaRfG
+empty (rf; (dynamic_tag(NAL) & dynamic_tag(LOCAL))) \ vis(lhb) as O-NaRfL
+
+(* Consistency of RMWs *)
+empty rmw & (fre; coe) as atomic
```
hernanponcedeleon, line 103: Move this below (in all models)

cat/opencl.cat
```diff
@@ -0,0 +1,128 @@
...
+acyclic (SC*SC) & scp & incl as O-Sscoped
+
+(***************)
+(*    Races    *)
+(***************)
+
+let cnf = ((W * W) | (W * R) | (R * W)) & loc
+
+let dr = cnf & ~(ghb | lhb) & ~(ghb | lhb)^-1 & ~wi & ~incl
+flag ~empty dr as data_race
```
hernanponcedeleon, line 123: There is no notion of unsequenced races in opencl?

cat/opencl.cat
```diff
@@ -0,0 +1,128 @@
...
+
+(* Barrier divergence *)
+let bsw = bar_sw(dynamic_tag(GLOBAL)) | bar_sw(dynamic_tag(LOCAL))
+let bd = [EF] & ((~wi & swg); unv) \ (bsw ; unv)
+flag ~empty bd as barrier_divergence
```
hernanponcedeleon, line 128: Do we have examples where we test for this property? If not, I assume at least in the paper there should be some

hernanponcedeleon (commented):
We are almost there ;)

Please answer the review with new commits (i.e. do not rebase) so it is easier to review the next time (otherwise I need to go over a large diff)

hernanponcedeleon:
Btw ... Don't forget to add opencl as a target in the README

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -152,34 +157,57 @@ public Object visitGlobalDeclaratorArray(LitmusCParser.GlobalDeclaratorArrayCont
...
         for(LitmusCParser.ExpressionContext expressionContext : ctx.expression())
             expressionContext.accept(this);
-
-        scope = currentThread = -1;
         return null;
     }
 
     @Override
     public Object visitThreadArguments(LitmusCParser.ThreadArgumentsContext ctx){
         if(ctx != null){
```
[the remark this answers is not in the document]
tonghaining: After we formalize the OpenCL Litmus format this is redundant, removed.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -281,7 +311,19 @@ public Expression visitReAtomicAddUnless(LitmusCParser.ReAtomicAddUnlessContext
     public Expression visitReC11AtomicXchg(LitmusCParser.ReC11AtomicXchgContext ctx) {
         Register register = getReturnRegister(true);
         Expression value = (Expression) ctx.value.accept(this);
-        Event event = EventFactory.Atomic.newExchange(register, getAddress(ctx.address), value, ctx.c11Mo().mo);
+        Expression address = getAddress(ctx.address);
+        Event event = EventFactory.Atomic.newExchange(register, address, value, ctx.c11Mo().mo);
+        programBuilder.addChild(currentThread, event);
+        return register;
+    }
```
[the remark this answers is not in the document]
tonghaining: Added

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -481,6 +580,11 @@ public Object visitNreAssignment(LitmusCParser.NreAssignmentContext ctx){
         Expression value = (Expression)ctx.re().accept(this);
         if(variable instanceof MemoryObject || variable instanceof Register){
             Event event = EventFactory.newStoreWithMo(variable, value, C11.NONATOMIC);
+            if (variable instanceof Register reg) {
+                if (isOpenCL) {
+                    event.addTags(Tag.OpenCL.DEFAULT_WEAK_SCOPE);
+                }
+            }
```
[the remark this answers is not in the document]
tonghaining: Updated, `WI` should be added to all the weak events. I think originally I was trying to use it for registers like `*x`..

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
tonghaining: Will some other Archs also have unique tags for Init? Then this boolean might be not enough?

litmus/OPENCL/herd/global_barrier.litmus
tonghaining: Removed, also for the global_barrier_mo.litmus

cat/c11-partialsc.cat
tonghaining: Removed

cat/c11-partialsc.cat
tonghaining: Moved

cat/c11.cat
tonghaining: removed

cat/c11-partialsc.cat
tonghaining: Added

cat/opencl.cat
tonghaining: Updated

cat/opencl.cat
tonghaining: Renamed

cat/opencl.cat
tonghaining: Updated to `((dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)); unv; (dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)))`

cat/opencl.cat
tonghaining: Moved

cat/opencl.cat
tonghaining: `ur` appears in herd version, overhauling doesn't have one. Shall we add?

cat/opencl.cat
tonghaining: We don't have, the paper omitted the barrier divergence part.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
ThomasHaas: What is this boolean supposed to mean? I would avoid adding any language specific stuff.

cat/c11-orig.cat
```diff
@@ -16,9 +16,9 @@ let asw = IW * (M \ IW)
 let sb = po
 let mo = co
 
-let cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
-let crel = REL | (SC & (W | F)) | ACQ_REL
-let ccon = R & CON
+let Cacq = ACQ | (SC & (R | F)) | ACQ_REL | (F & CON)
+let Crel = REL | (SC & (W | F)) | ACQ_REL
+let Ccon = R & CON
```
hernanponcedeleon, line 21: What I mean was to use Rel, Acq, Con

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
hernanponcedeleon: > What is this boolean supposed to mean? I would avoid adding any language specific stuff.

I don't mean to the Init class, but to the factory. I don't see any difference between having `EventFactory.newInit(x, y, boolean)` and `EventFactory.newInit(x, y)` rather than `EventFactory.newC11Init(x, y)` and `EventFactory.newInit(x, y)` and what I propose simplifies the code (not only in the MemoryAllocation, but also in the factory where we avoid the duplicated code of this methods)

cat/opencl.cat
```diff
@@ -60,7 +60,7 @@ let ra_sw(r) = ((r & [Rel]); Fsb?; [W \ WI]; rs?; r; rf; [R \ WI]; sbF?; ([Acq]
 let bar_sw(r) = (r & [EF]); syncbar & ~wi & swg; (r & [XF])
 
 (* Allowed to synchronise on the other region *)
-let scf = (SC * SC) | ((dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)); F*F;(dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)))
+let scf = (SC * SC) | ((dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)); unv; (dynamic_tag(GLOBAL) & dynamic_tag(LOCAL)))
```
hernanponcedeleon, line 63: you don't need to `unv`, do you?

cat/opencl.cat
hernanponcedeleon: Yes please. Otherwise, once you add a unit test to compare both models for races, this might cause problems

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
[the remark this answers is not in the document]
tonghaining: Hmm, as the pipeline complains, we need the null check for LKMM tests. I have added it back

cat/opencl.cat
tonghaining: added

cat/c11-orig.cat
tonghaining: Okay, I misunderstood..

cat/opencl.cat
tonghaining: I think we need? Otherwise it becomes `([DG] & [DL]);([DG] & [DL])` which is `([DG] & [DL] & [DG] & [DL])`

hernanponcedeleon: [x];unv;[y] should give you all pairs in unv (which is pretty much anything) restricted to pairs starting with x and ending with y. I fail to see how this is different than [x];[y]

hernanponcedeleon: > I think we need? Otherwise it becomes `([DG] & [DL]);([DG] & [DL])` which is `([DG] & [DL] & [DG] & [DL])`

I don't this this is true. The first one allows you to have pairs (x,y) with x!=y, the second ones does not

tonghaining: `unv` is defined as `_ * _`, so I think `[X]; unv; [Y]` should be `X * Y`? And `[X];[Y]` is like `X & Y`

cat/opencl-herd.cat
```diff
@@ -0,0 +1,156 @@
+OpenCL
+(* OpenCL Memory Model *)
+
+(*
+* This model is based on:
+* https://github.com/herd/herdtools7/blob/master/herd-www/cat_includes/opencl_base.cat
+* https://github.com/herd/herdtools7/blob/master/herd-www/cat_includes/opencl.cat
+*)
```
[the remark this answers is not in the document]
tonghaining: Yes, it yields the same [result](https://github.com/hernanponcedeleon/Dat3M/pull/744/commits/175bcfe587c77763ba208972aa28bea3c0c3ffa6#diff-968a9d8daa09e4d799a4b086bd5aaf07d463888ff08c3c976c36703322fc3f13) as we tested on herd

cat/opencl.cat
hernanponcedeleon: > `unv` is defined as `_ * _`, so I think `[X]; unv; [Y]` should be `X * Y`?

This I agree

> And `[X];[Y]` is like `X & Y`

This I don't. The first uses composition so it is clearly a relation. The second one is the intersection of tags, thus, it is a tag

hernanponcedeleon: I see that the change I propose makes `overhauling/example6` fail, but I suspect that the problem is that our AtomiceFence event misses the VISIBLE tag

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemoryAllocation.java
ThomasHaas: I still don't know what the boolean is parameter is supposed to do? Is it about copying the feature tags from the `MemoryObject`? I think we should always do it since that is now a (nice) feature of the core language.
But `newC11Init` also adds the `C11.NONATOMIC` tag which is language specific and should not be part of the core construction.

As a general side note: Boolean parameters, especially one's that practically always take constant values (i.e., the caller passes `true/false`), tend to be hard to read on call sites unless it is an extremely natural parameter (e.g. `newInit(x, y, true)` is unclear but `newBoolConst(true)`  is natural). 
Fortunately, static factory methods have an advantage over constructors due to the fact that you can just provide two methods, one for `false` version and one for the `true` version, that tell you by name what is different.
From a different point of view, you can also say that `true/false` are just magic constants and those should be avoided. You either give them names on the call site like `static final bool INIT_WITH_C11_TAGS = true` or, well, you use two different functions in the factory.

cat/opencl.cat
ThomasHaas: `[X];[Y] == [X & Y] ~ X & Y` is true, because you compose identity relations. What Haining does is correct.

cat/opencl.cat
```diff
@@ -0,0 +1,133 @@
...
+flag ~empty dr as data_race
+
+(* unsequenced_races *)
+let symm(r) = r | r^-1
+let ur = (wi & cnf & ~symm(po)) \ id
+flag ~empty ur as unsequenced_race
+
+(* Barrier divergence *)
+let bsw = bar_sw(dynamic_tag(GLOBAL)) | bar_sw(dynamic_tag(LOCAL))
+let bd = [EF] & ((~wi & swg); unv) \ (bsw ; unv)
```
ThomasHaas, line 132: This looks wild to me. Shouldn't `r;unv` just be `[domain(r)]` here? I mean, the current definition computes huge relations that are then intersected with an identity `[EF]` so all non-identity edges are unnecessary to compute. Thus when right-composing with `unv`, the only relevant edges you get are self-loops which you can directly express using `domain(...)`.

```
let bd = [EF] & [domain(~wi & swg)] \ [domain(bsw)]
```

cat/opencl.cat
```diff
@@ -0,0 +1,133 @@
...
+
+let Acq = (ACQ | SC | ACQ_REL) & (R | F)
+let Rel = (REL | SC | ACQ_REL) & (W | F)
+
+(* Fences sequenced before or after *)
+let Fsb = [F]; sb
+let sbF = sb; [F]
+
+(* Release sequence *)
+let rs_prime = wi | (unv; [R & W])
```
ThomasHaas, line 51: You can use `wi | _ * (R & W)`. If you prefer the `unv` version you can keep it. My suggestion is just optimized for internal computation (RA is better on non-compositions).
Btw. we do not have events that are `R & W` so this is always empty, no? What's the purpose then?

cat/opencl.cat
```diff
@@ -0,0 +1,133 @@
...
+(***************)
+
+(* data_races *)
+let cnf = ((W * W) | (W * R) | (R * W)) & loc
+let dr = cnf & ~(ghb | lhb) & ~(ghb | lhb)^-1 & ~wi & ~incl
+flag ~empty dr as data_race
+
+(* unsequenced_races *)
+let symm(r) = r | r^-1
+let ur = (wi & cnf & ~symm(po)) \ id
```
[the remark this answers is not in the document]
tonghaining: Updated

cat/opencl.cat
tonghaining: I see, updated

cat/opencl.cat
tonghaining: Updated, it was for RMW, I forgot to change it when porting from the herd version.

dartagnan/src/test/java/com/dat3m/dartagnan/comparison/C11AndOpenCLTest.java
tonghaining: Added

cat/opencl.cat
hernanponcedeleon: Does this mean that we need `unv`, otherwise the result relation would be a subset of `id`?

hernanponcedeleon:
Unless @ThomasHaas has more comments, I can merge once the CI passes

cat/opencl.cat
ThomasHaas: Yes. Composing identities gives you only identities. If the intention is to have a cartesian-product-like relation, then you need to have a proper binary relation somewhere (either `F*F` as in the original or `unv` like now).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -147,6 +147,15 @@ public static Init newInit(MemoryObject base, int offset) {
         return new Init(base, offset, address);
     }
 
+    public static Init newC11Init(MemoryObject base, int offset) {
+        final Expression address = offset == 0 ? base :
+                expressions.makeAdd(base, expressions.makeValue(offset, (IntegerType) base.getType()));
+        Init init = new Init(base, offset, address);
```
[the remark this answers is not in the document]
ThomasHaas: Why did you resolve this point?  I think it is a valid point to just call `newInit`. Just the fact that we duplicate the awkward address simplification is reason enough.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/ControlBarrier.java
```diff
@@ -20,7 +20,7 @@ public ControlBarrier(String name, Expression id) {
         this.id = id;
     }
 
-    private ControlBarrier(ControlBarrier other) {
+    protected ControlBarrier(ControlBarrier other) {
```
ThomasHaas, line 23: You can make this private again

tonghaining: Updated

Issue #746, opened 2024-09-30
Full alias analysis produces non-deterministic results


natgavrilenko:
Full alias analysis produces different results when executed multiple times. Field sensitive and field insensitive methods work as expected.
```Java
public class AliasTest {

    private final String testPath = getTestResourcePath("libvsync/bounded_mpmc_check_empty-opt.ll");
    private final String modelPath = "cat/c11.cat";
    private final Arch target = Arch.C11;

    @Test
    public void test() throws Exception {
        Program program = new ProgramParser().parse(new File(testPath));
        Wmm wmm = new ParserCat().parse(new File(getRootPath(modelPath)));
        Configuration baseConfig = Configuration.builder().build();
        VerificationTask baseTask = createTask(program, wmm, baseConfig);
        wmm.configureAll(baseTask.getConfig());
        preprocessProgram(baseTask, baseTask.getConfig());
        preprocessMemoryModel(baseTask, baseTask.getConfig());

        Relation loc = wmm.getRelations().stream().filter(r -> r.getNameOrTerm().equals("loc")).findAny().orElseThrow();
        Configuration config = Configuration.builder()
                //.setOption(ALIAS_METHOD, FIELD_SENSITIVE.asStringOption())
                //.setOption(ALIAS_METHOD, FIELD_INSENSITIVE.asStringOption())
                .setOption(ENABLE_EXTENDED_RELATION_ANALYSIS, "false")
                .build();

        List<Integer> sizes = new LinkedList<>();
        for (int i = 0; i < 10; i++) {
            Context nativeContext = Context.create();
            VerificationTask nativeTask = createTask(program, wmm, config);
            performStaticProgramAnalyses(nativeTask, nativeContext, nativeTask.getConfig());
            performStaticWmmAnalyses(nativeTask, nativeContext, nativeTask.getConfig());
            RelationAnalysis nativeRa = nativeContext.get(RelationAnalysis.class);
            System.out.println(nativeRa.getKnowledge(loc).getMaySet().size());
            sizes.add(nativeRa.getKnowledge(loc).getMaySet().size());
        }
        for (int i = 1; i < 10; i++) {
            assertEquals(sizes.get(i - 1), sizes.get(i));
        }
    }

    private VerificationTask createTask(Program program, Wmm wmm, Configuration config)
            throws InvalidConfigurationException {
        return VerificationTask.builder()
                .withConfig(config)
                .withBound(2)
                .withTarget(target)
[...]

PR #750, opened 2024-10-09
Add support for variable-sized and variable-aligned allocations

ThomasHaas:
This PR generalizes `MemoryObject` to be of variable size given by `Expression size`.
The encoding supports this now.
~~The reason the PR is marked as [DRAFT] is simply because it needs some minor cleanup and maybe more tests, but the functionality should be there.~~

**Points of interest & possible pitfalls:**
- Memory objects get two variables: address and size. The encoding enforces a deterministic memory layout as a function of the sizes of each object.
- We do not check if sizes/addresses are negative or if they overflow. This can cause different memory objects to get the same address.
- Dynamically allocated memory objects whose allocation sites were not executed still get a well-aligned address but receive size 0. As a consequence, these objects can actually overlap naturally but that should not matter I think.
- Somewhat in contrast to the above point, the current encoding always adds some non-zero offset to each address to ensure alignment even if the object happens to be aligned already (I don't encode a check to see if the alignment is already satisfied and unconditionally add some offset).

**EDIT:** I added a new feature, namely custom alignments for allocations. This even works for non-deterministic alignments.
The intrinsic `aligned_alloc` is now supported. However, we do not make any checks such as alignments must be positive and powers of two.


ThomasHaas:
My newly added tests only work with the new alias analysis but `RelationAnalysisTest` seems to invoke the older one's, which causes an exception on those tests.
How should we go about this? Wait until #746 is fixed so that we can use the new analysis, or just exclude the tests?
If we exclude them, it would be easiest to just exclude the whole `miscellaneous` folder (e.g., by commenting out those test cases temporarily).

hernanponcedeleon:
> If we exclude them, it would be easiest to just exclude the whole `miscellaneous` folder (e.g., by commenting out those test cases temporarily).

You should rather create a new class for the new 2 tests and skip those in the RA test rather than (unnecessarily) removing all other miscellaneous tests.

ThomasHaas:
It depends on what unnecessarily means. I could argue that `RelationAnalysisTest` already works around bug #746 but this workaround has become insufficient now and should thus be updated. And IMO workarounds should be as simple as possible.
I mean, do we really want to separate the two new tests into a separate folder and split the `MiscellanousTest` over two classes just to work around a bug? And then we undo this split once #746 is fixed? Or do we want to permanently establish separate tests for non-deterministic/variable-sized allocations?

hernanponcedeleon:
Your point is fair. Let keep the RA test as it is.

Here are the results on the weaver benchmarks with a 15 min timeout, for reachability category, on commit d9abdfa8280f04534c958e30cc52832321cc296d
```
Statistics:            171 Files
  correct:              86
    correct true:       82
    correct false:       4
  incorrect:             0
    incorrect true:      0
    incorrect false:     0
  unknown:              85
  Score:               168 (max: 335)
```
Last year we solved 61 of the tasks.

Hopefully none of the latest commits changes things drastically in terms of what we can do with these benchmarks.

Is this still in DRAFT status or can it already be reviewed?

ThomasHaas:
I think you can review it. I have changed the memory encoding as a whole so there might be regression on other benchmarks, though I kinda doubt it.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -277,65 +284,70 @@ private BooleanFormula encodeForwardProgress(Program program, ProgressModel prog
...
-        for (final MemoryObject memObj : memory.getObjects()) {
-            // For all objects, their 'final' value fetched here represents their constant value.
-            final Formula addressVariable = context.encodeFinalExpression(memObj);
-            final BigInteger addressInteger = memObj2Addr.get(memObj);
-
-            if (addressVariable instanceof BitvectorFormula bitvectorVariable) {
-                final BitvectorFormulaManager bvmgr = fmgr.getBitvectorFormulaManager();
-                final int length = bvmgr.getLength(bitvectorVariable);
-                enc.add(bvmgr.equal(bitvectorVariable, bvmgr.makeBitvector(length, addressInteger)));
+        return encodeMemoryLayout(context.getTask().getProgram().getMemory());
```
hernanponcedeleon, line 292: What is the point of creating a new method `encodeMemoryLayout` when `encodeMemory` is a 2-liner that simply logs one things and calls it.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -277,65 +284,70 @@ private BooleanFormula encodeForwardProgress(Program program, ProgressModel prog
...
+
+            // Encode size & compute alignment
+            if (cur.isStaticallyAllocated()) {
+                enc.add(helper.equals(size, context.encodeFinalExpression(cur.size())));
+                alignment = context.encodeFinalExpression(cur.alignment());
             } else {
-                assert addressVariable instanceof IntegerFormula;
-                final IntegerFormulaManager imgr = fmgr.getIntegerFormulaManager();
-                enc.add(imgr.equal((IntegerFormula) addressVariable, imgr.makeNumber(addressInteger)));
+                // Non-allocated objects get size 0
```
hernanponcedeleon, line 316: Please add "i.e. objects whose corresponding alloc event was not executed", It took me a while to understand what "non-allocated mean"

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -277,65 +284,70 @@ private BooleanFormula encodeForwardProgress(Program program, ProgressModel prog
...
+                enc.add(helper.equals(size,
+                                bmgr.ifThenElse(context.execution(cur.getAllocationSite()),
+                                        context.encodeExpressionAt(cur.size(), cur.getAllocationSite()),
+                                        helper.value(BigInteger.ZERO, helper.typeOf(size)))
+                        )
+                );
+                // Non-allocated objects with variable alignment get alignment 1
```
hernanponcedeleon, line 323: Why not zero? do we have any constraint that alignments must be positive?

EDIT: I guess this is because the first object address matches the alignment, right?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -93,7 +94,15 @@ public static List<Event> eventSequence(Object... events) {
...
     public static Alloc newAlloc(Register register, Type allocType, Expression arraySize,
                                  boolean isHeapAlloc, boolean doesZeroOutMemory) {
-        return new Alloc(register, allocType, arraySize, isHeapAlloc, doesZeroOutMemory);
+        final Expression defaultAlignment = expressions.makeValue(8, types.getArchType());
+        return newAlignedAlloc(register, allocType, arraySize, defaultAlignment, isHeapAlloc, doesZeroOutMemory);
+    }
+
+    public static Alloc newAlignedAlloc(Register register, Type allocType, Expression arraySize, Expression alignment,
+                                 boolean isHeapAlloc, boolean doesZeroOutMemory) {
+        arraySize = expressions.makeCast(arraySize, types.getArchType(), false);
```
hernanponcedeleon, line 103: Why do we now enforce `arraySize` to be 64 bits, but before it was not the case?

benchmarks/miscellaneous/nondet_alloc.c
```diff
@@ -0,0 +1,27 @@
...
+
+#define MAX_SIZE 10
+
+int main()
+{
+    int size = __VERIFIER_nondet_int();
+    __VERIFIER_assume(size > 0 && size <= MAX_SIZE);
+    int *array = malloc(size * sizeof(int));
+
+    __VERIFIER_loop_bound(MAX_SIZE + 1);
```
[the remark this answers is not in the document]
ThomasHaas: No, cause the loop bound is dynamic but still bounded (the loop iterates until `size` not until `MAX_SIZE`)

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: Because there were temporarily two different ways to encode memory, depending on whether it is static or not. The question is whether we need to bring back the static version due to some regressions and if we do that, whether we want to split the functions agains.
For now, I decided to keep only one variant but still have the split. I also think it is not too bad to have the public interface `encodeMemory` be separated from the internal implementation `encodeMemoryLayout` but I guess that is a matter of personal taste.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: Alignment of 0 would mean you have to divide by 0... `x % 0` is ill-defined.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
ThomasHaas: We didn't enforce this because the size was assumed to be static and the encoding used the static value, totally ignoring its type. This made type-mismatches irrelevant. Now we use the proper expression and thus rely on correct types.
There is an argument to be made what type a size should be, but I think it is most canonical if its type matches with pointers (i.e., if pointers are 64 bit then sizes are 64 bit).

benchmarks/miscellaneous/nondet_aligned_alloc.c
```diff
@@ -0,0 +1,24 @@
...
+    size_t size = (size_t)size_int;
+    size_t align = (size_t)align_int;
+    int *array = malloc(42 * sizeof(int));
+    int *array_2 = aligned_alloc(align, size * sizeof(int));
+    // NOTE: making the first allocation non-det-sized makes the solver extremely slow to verify the
+    // custom aligned allocation. Even when fixing the alignment to a constant it gets super slow if that constant
+    // is not a power of two. In other words, aligning variable-sized addresses is very expensive currently, especially
+    // for non-power-of-two alignments (those don't exist in practice though).
+
+    assert (array != array_2);
```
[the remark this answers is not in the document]
ThomasHaas: You want to assert a certain distance between the arrays? I can add that. Overflows do not matter in this example because I generate positive 32-bit sizes but we have a 64-bit address space so even MAX_INT does not overflow.
But this conversion is sublte because the compiler sign-extends but I want a zero-extension, which I get by using positive values (for those the sign bit is 0 so sign and zero-extension do the same).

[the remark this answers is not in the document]
hernanponcedeleon: I want to assert that the arrays do not overlap.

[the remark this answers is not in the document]
ThomasHaas: Done. The assertion assumes a certain ordering between the addresses though. I didn't bother looking for the abs function and somehow I have the feeling that it would involve unsupported intrinsics :P

hernanponcedeleon (approved):

PR #751, opened 2024-10-10
Introduce stdlib.cat

hernanponcedeleon:
In herd, most of what we defined in `basic.cat` is defined in `stdlib.cat` which is included by default in all models (see section 16.7 in https://diy.inria.fr/doc/herd.html). 

This PR mimics this behavior.


ThomasHaas:
We did not always place the stdlib include at the beginning, for example in `vmm.cat`. The default include might cause issue in that regard. Well, I guess you can re-include the stdlib.cat after modifying definitions to overwrite the already created one's.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -62,6 +63,13 @@ public String toString() {
     VisitorCat(Path includePath) {
         this.includePath = includePath;
         this.wmm = new Wmm();
+        try {
+            // The standard library is a cat file stdlib.cat which all models include by default
+            final CatParser parser = getParser(CharStreams.fromPath(Path.of(GlobalSettings.getCatDirectory() + "/stdlib.cat")));
+            parser.mcm().accept(this);
+        } catch (IOException e) {
+            throw new ParsingException("Error parsing stdlib.cat file", e);
+        }
```
ThomasHaas, line 72: Can you put this into a separate function like "includeStdlib". I'm not a big fan of having complex logic sitting around in the constructor.

hernanponcedeleon:
> The failed checks even include the `RelationAnalysisTest`. That means you just exposed some more bugs in RA or AA I think.

Maybe it is the non-det from #746 and this was just an unlucky run.

ThomasHaas:
But the non-det was supposed to only affect the new alias analysis which is disabled in`RelationAnalysisTest`. So if it is non-det behaviour, then you exposed it on the older versions of AA.

hernanponcedeleon:
> I thought about that, and I still decided to remove it to see if this already causes problems. I assume the `vmm.cat` from the libvsync repo should not depend on another cat files so this probably will need to be fixed in the official model. I will discuss that with jonas.

The [latest version](https://github.com/open-s4c/libvsync/blob/main/vmm/vmm.cat) of vmm already solves most of the issues, it only need to properly define `po-loc`

ThomasHaas (approved):
LGTM

PR #752, opened 2024-10-11
Configuration files and management

xeren:
This PR adds the following front-end features:

- The CLI accepts `.properties` files to read options.
- The UI gets a new button, showing a responsive dialog with all available options.  This should be an improvement to one of my earlier PRs.  It now does the crawling in a background thread instead of blocking the interface from the beginning.  You can now use the UI to inspect the options of this project.
- The UI can also load and save `.properties` files.
- Name option `wmm.analysis.assumeAtomicity`.
- Rename option `encoding.locallyConsistent` to `wmm.analysis.assumeLocalConsistency`.


hernanponcedeleon:
> * The CLI accepts `.properties` files to read options.

Amazing! Finally getting closer to #79 

Do you think it would be easy to convert the remaining "options" in `GlobalSettings.java` to proper options and get rid of that class all together?

hernanponcedeleon:
Based on some quick testing, the import `.properties` file from the UI does not seem to work ... even if I change some options there, I cannot see the change to have an effect in the log (however, it seems to work properly from the CLI)

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -75,13 +76,30 @@ public class Dartagnan extends BaseOptions {
...
+                    builder.loadFromFile(argument);
+                    logger.info("Loaded properties from {}", argument);
+                } catch (IOException e) {
+                    logger.warn("Could not load file {}: {}", argument, e.getMessage());
+                }
+            }
+        }
+        return builder.build();
+    }
+
```
ThomasHaas, line 102: I feel like this is reinventing the wheel :)
Your methods is like a mixture of `Configuration.fromCmdLineArguments` and `Configuration.builder().loadFromSource`, the latter of which supports `#include` directives.
I have the feeling that one of the following is possible:
- Generate a single `CharSource` from the arguments and use `loadFromSource`
- Split the args into loads of `.properties` files (not sure what syntax you want to use) and standard arguments. Then use `fromCmdLineArguments` on the former and `loadFromSource/File` on the latter (`loadFromSource` would also support nested loadings with `#include`)

xeren: The facility of `loadFromSource` looks very useful.  Sadly, `fromCmdLineArguments` is not a method of `ConfigurationBuilder`, but with `CharSource`, there is little left to code.

ThomasHaas: While `fromCmdLineArguments` is not a builder method, you can still generate a full `Configuration` and then create a builder starting from that config via `ConfigurationBuilder.copyFrom(Configuration)`. With that builder, you can then import whatever extra files you want.
Though I guess it would be better to first load the file and then the remaining args, so that they overwrite whatever the file says. I mean, if the user explictily specifies something that is also in a file, then I guess the wants to overwrite it.

xeren:
> Do you think it would be easy to convert the remaining "options" in GlobalSettings.java to proper options and get rid of that class all together?

The more recent parts look difficult to change.  Also, that class manages the `DAT3M_HOME` and `DAT3M_OUTPUT` environment variables.  Would you want to replace them, too?

ThomasHaas:
I don't think we can get rid of that class precisely because of the directory informations. But the other options can likely be moved elsewhere.

hernanponcedeleon:
Yes, what I meant if getting rid of the "options" that are still hardcoded there. We can either keep the class for the static methods, or move them somewhere else if there is a better place

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionNames.java
```diff
@@ -64,6 +65,8 @@ public class OptionNames {
 
     // Refinement Options
     public static final String BASELINE = "refinement.baseline";
+    public static final String GRAPHVIZ_DEBUG_FILES = "refinement.generageGraphvizDebugFiles";
+    public static final String SYMMETRY_LEARNING = "refinement.symmetryLearning";
```
ThomasHaas, line 69: The option was named `symmetricLearning` and not `symmetryLearning` (it used to have the latter name).
I changed it a while ago because the latter sounds like we learn the symmetries, but that is not true: we use the symmetries to learn symmetric clauses.

hernanponcedeleon:
The import feature form the ui is still not working. Also, option `modeling.progress` is missing in the large panel (there might be other missing options, I did not really check)

xeren:
I marked `modeling.progress` as 'basic' / non-extra, because it already appears in the main window.  It could as well be both.

ui/src/main/java/com/dat3m/ui/options/OptionsPane.java
```diff
@@ -179,9 +218,321 @@ private void mkGrid() {
...
+        configurationFileChooser.showOpenDialog(this);
+        final File file = configurationFileChooser.getSelectedFile();
+        if (file == null) {
+            return;
+        }
+        final var properties = new HashMap<String, String>();
+        try {
+            final Configuration config = Configuration.builder().loadFromFile(file.toPath()).build();
+            for (String key : List.copyOf(config.getUnusedProperties())) {
+                properties.put(key, config.getProperty(key));
```
xeren, line 301: By using `Configuration` for importing the file, the UI also supports `#include` directives in the format, just like `Dartagnan`.

The method `getProperty` is deprecated, though.  The best alternative I have:
```java
Properties props = new Properties();
props.load(new StringReader(config.asPropertiesString()));
```

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
xeren: I got an `IllegalStateException` from `copyFrom`, because it is not the first operation on a builder.  Instead, I added the file arguments as `#include` directives on top of the `CharSource`.

hernanponcedeleon:
The new options from the larger panel work from the UI. 

However, the import of `*.properties` files is still not working for me. Importing
```
> cat vsyncer.properties 
solver=yices2
```
still uses the default solver (z3).

ui/src/main/java/com/dat3m/ui/options/OptionsPane.java
ThomasHaas: Just use `getProperty()` even if it is deprecated.

xeren:
Options like `solver = yices2` should work now.  I made boolean and enum options case-insensitive for the UI.

hernanponcedeleon:
> Options like `solver = yices2` should work now. I made boolean and enum options case-insensitive for the UI.

Now it seems to be working as I would expect (including changing the displayed value in the ui if the corresponding property is set in the `.properties` file).

I will take a look to the code tonight.

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
```diff
@@ -1,52 +1,22 @@
...
-import org.sosy_lab.common.configuration.Configuration;
-import org.sosy_lab.common.configuration.InvalidConfigurationException;
-import org.sosy_lab.common.configuration.Options;
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 
-@Options
 public class GlobalSettings {
```
hernanponcedeleon, line 10: What about moving the static methods to utils/Utils.java and getting rid of this class?

hernanponcedeleon (commented):
Other than my small comment above, I think we can merge

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
ThomasHaas: I don't think I'm a fan of that. The content is not some random utility but indeed a sort of global configuration/setting. For example, if we ever move the `cat` or `output` directories (or rename them), we would need to update it in this class. I think a class that contains (all) functionality/information related to our environment variables (`DAT3M_HOME/DAT3M_OUTPUT`) and internal project structure should not be in some obscure utility location.
I would rather suggest to move the content to the main class `Dartagnan` (or not move it at all).

hernanponcedeleon: The lets simply keep it as it is.

PR #753, opened 2024-10-11
Fix overlap queries in Alias Analysis

xeren:
This PR solves issue #746.  The example showcased a bug in `InclusionBasedPointerAnalysis`:  Negative offsets are misinterpreted in `overlaps` queries.
On top of that, `Modifier` is normalized and the analysis' non-determinism originating from hash sets is removed.  This should ease the reproduction of errors.


dartagnan/src/test/java/com/dat3m/dartagnan/program/analysis/alias/AliasTest.java
```diff
@@ -0,0 +1,74 @@
...
+    public void test() throws Exception {
+        Program program = new ProgramParser().parse(new File(testPath));
+        Wmm wmm = new ParserCat().parse(new File(getRootPath(modelPath)));
+        Configuration baseConfig = Configuration.builder().build();
+        VerificationTask baseTask = createTask(program, wmm, baseConfig);
+        wmm.configureAll(baseTask.getConfig());
+        preprocessProgram(baseTask, baseTask.getConfig());
+        preprocessMemoryModel(baseTask, baseTask.getConfig());
+
+        Relation loc = wmm.getRelations().stream().filter(r -> r.getNameOrTerm().equals("loc")).findAny().orElseThrow();
```
hernanponcedeleon, line 44:
```suggestion
        Relation loc = wmm.getRelation(RelationNameRepository.LOC);
```

dartagnan/src/test/java/com/dat3m/dartagnan/program/analysis/alias/AliasTest.java
```diff
@@ -0,0 +1,74 @@
...
+        Configuration config = Configuration.builder()
+                //.setOption(ALIAS_METHOD, FIELD_SENSITIVE.asStringOption())
+                //.setOption(ALIAS_METHOD, FIELD_INSENSITIVE.asStringOption())
```
hernanponcedeleon, line 47: Remove

dartagnan/src/test/java/com/dat3m/dartagnan/program/analysis/alias/AliasTest.java
```diff
@@ -0,0 +1,74 @@
...
+                .build();
+
+        List<Integer> sizes = new LinkedList<>();
+        for (int i = 0; i < 10; i++) {
+            Context nativeContext = Context.create();
+            VerificationTask nativeTask = createTask(program, wmm, config);
+            performStaticProgramAnalyses(nativeTask, nativeContext, nativeTask.getConfig());
+            performStaticWmmAnalyses(nativeTask, nativeContext, nativeTask.getConfig());
+            RelationAnalysis nativeRa = nativeContext.get(RelationAnalysis.class);
+            System.out.println(nativeRa.getKnowledge(loc).getMaySet().size());
```
hernanponcedeleon, line 58: Remove

hernanponcedeleon (commented):
If the problem is really fixed, you should be able to use the default alias analysis in `RelationAnalysisTest` and enabled the tests that were commented out in #750

dartagnan/src/test/java/com/dat3m/dartagnan/program/analysis/alias/AliasTest.java
xeren: Actually, I'd like to remove this entire test class again, because it cannot prove determinism.  Should it fail in a future PR, it might be caused by some already merged change.  It was helpful for this PR, though.

hernanponcedeleon: While it cannot prove determinism, it can find bugs when the implementation is not (and it might catch bug if the analysis is modified again in the future). Thus, I see zero advantage of getting rid of the test class.

xeren:
I changed an expected value in `AnalysisTest`.  It seems like the bug caused a tiny "precision" boost for this particular instance, at the expense of soundness.  The new value (meaning `(&x * 1) + 1` may alias with `&x`) is what I expect from `InclusionBasedPointerAnalysis`, since it does not special-treat the case, where a pointer is multiplied by 1.  Such an expression would normally be pre-processed with Constant Folding.

dartagnan/src/test/java/com/dat3m/dartagnan/program/analysis/alias/AliasTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static com.dat3m.dartagnan.configuration.OptionNames.ENABLE_EXTENDED_RELATION_ANALYSIS;
+import static com.dat3m.dartagnan.configuration.Property.PROGRAM_SPEC;
+import static com.dat3m.dartagnan.utils.ResourceHelper.getRootPath;
+import static com.dat3m.dartagnan.utils.ResourceHelper.getTestResourcePath;
+import static com.dat3m.dartagnan.verification.solving.ModelChecker.*;
+import static com.dat3m.dartagnan.wmm.RelationNameRepository.LOC;
+import static org.junit.Assert.assertEquals;
+
+public class AliasTest {
```
ThomasHaas, line 29: Since this is an awkward test, can we add a comment about it? Maybe even rename it to `AliasDeterminismTest` or something along those lines.

ThomasHaas (commented):
When you update the test, I think we can merge.
While I think the new code is confusing (what the hell is "leftDirectedTowardsRight"?), the old code was also confusing, so nothing has changed in that regard :P.

ThomasHaas (approved):
LGTM

PR #756, opened 2024-10-17
Svcomp config

hernanponcedeleon:


svcomp.properties
```diff
@@ -0,0 +1,4 @@
+method=eager
+encoding.integers=true
```
ThomasHaas, line 2: I thought you intend to use BVs and Yices2, no? Or is this different to the settings you will actually submit?

ThomasHaas (approved):
LGTM

PR #759, opened 2024-10-24
Allow bounds to be saved to and loaded from files 

hernanponcedeleon:
This feature needs to be used with caution since the information depends on event ids which may not match if the program is changed. However, if used as intended (just run dartagnan several times in a row and let it automatically adapt the bounds until getting PASS), it can boost performance. I have seen improvements up to 3x.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -413,9 +413,10 @@ public class Dartagnan extends BaseOptions {
     }
 
     private static void increaseBoundAndDump(List<Event> boundEvents, Configuration config) throws IOException {
-        final File boundsFile = new File(config.hasProperty(BOUNDS_SAVE_PATH) ?
-                config.getProperty(BOUNDS_SAVE_PATH) :
-                GlobalSettings.getBoundsFile());
+        if(!config.hasProperty(BOUNDS_SAVE_PATH)) {
+            return;
+        }
+        final File boundsFile = new File(config.getProperty(BOUNDS_SAVE_PATH));
 
         // Parse old entries
         final List<CSVRecord> entries;
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -67,6 +67,7 @@ import java.util.*;
 
 import static com.dat3m.dartagnan.GlobalSettings.getOrCreateOutputDirectory;
 import static com.dat3m.dartagnan.configuration.OptionInfo.collectOptions;
+import static com.dat3m.dartagnan.configuration.OptionNames.BOUNDS_SAVE_PATH;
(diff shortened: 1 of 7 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -398,6 +409,45 @@ public static String generateResultSummary(VerificationTask task, ProverEnvironm
...
+
+        // We read from and write to the same CSV file,
+        // thus we need to split this in two loops
+        List<String[]> modifiedRecords = new ArrayList<>();
+        // We read the file written by the LoopUnrolling pass,
+        // thus we use BOUNDS_SAVE_PATH also for the reader
+        try (Reader reader = new FileReader(config.hasProperty(BOUNDS_SAVE_PATH) ?
+                                            config.getProperty(BOUNDS_SAVE_PATH) :
+                                            GlobalSettings.getBoundsFile())) {
+            for (CSVRecord record : CSVFormat.DEFAULT.parse(reader)) {
```
hernanponcedeleon, line 425: I don't like using the deprecated methods here, but I could not find a cleaner alternative. Any ideas?

hernanponcedeleon:
Thanks @ThomasHaas for the clean-up. 

Unless the CI complains due to the bounds file generation, or times drastically increase, I will merge the code as it.

ThomasHaas:
I'm still not a fan of all the `printStackTrace` calls. I think that needs some cleanup and maybe more warnings.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
```diff
@@ -289,7 +289,7 @@ private Map<Function, Map<CondJump, Integer>> loadLoopBoundsMapFromFile(Program
                         .put(loopJump, bound);
             }
         } catch (IOException e) {
-            e.printStackTrace();
+            // Nothing to be done, filePath is guaranteed to exists when the FileReader is called.
```
ThomasHaas, line 292: Not sure if this is true. `IOExceptions` not only happen if the file is missing I think.

hernanponcedeleon: You are right, I will handle this case as the other ones.

dartagnan/src/main/java/com/dat3m/dartagnan/GlobalSettings.java
```diff
@@ -38,6 +38,10 @@ public static String getCatDirectory() {
         return env + "/cat";
     }
 
+    public static String getBoundsFile() {
+        return getOutputDirectory() + "/bounds.csv";
+    }
+
```
ThomasHaas, line 44: When doing unit tests, thsi should probably return the empty string.
Btw., why do we even need a default location for saving bounds anyways?

hernanponcedeleon: Initially I was not planning to implement an option to allow the specify the saving path, but one of my colleagues requested this. 

You are right that we don't need this anymore.

ThomasHaas (approved):
LGTM

PR #761, opened 2024-10-25
Support of Showing Local Events Location in Spirv

CapZTr:
In #760 `Local` events are added to witness graph. This PR supports showing source-level location information for `Locals` in Spirv code, since they are often related to control flows (`Phi` instructions) so that need to be treated specifically.


hernanponcedeleon:
@CapZTr I thought this PR would need some changes based on the refactoring you are doing for #760, but maybe I am wrong.

Are they related or is this one completely independent?

CapZTr:
> @CapZTr I thought this PR would need some changes based on the refactoring you are doing for #760, but maybe I am wrong.
> 
> Are they related or is this one completely independent?

@hernanponcedeleon Changes of this PR are related to showing source location of `Locals` in Spir-V programs. Functions of the added code work independently with #760 , which extracts the information from IR and record it only. But this information will never be used, because the code for extracting `Locals` and adding them to the graph is not here.

hernanponcedeleon:
If there are independent then we should move forward (sorry for the delay, I thought there was a dependency).

@CapZTr can you please rebase?

@natgavrilenko can you please review?

PR #762, opened 2024-10-26
Improve PPC grammar and visitor

hernanponcedeleon:
The added litmus tests (automatically generated by diy7) shows several limitations of our PPC parser
- missing support for exclusive instructions
- not allowing locations to be referred as `[x]`
- enforcing very specific format for labels
- requiring branch instructions to be preceded by a `cmp`
- throwing an exception if a register is read before being initialized instead of giving it a non-det value.

This PR fixes all these problems.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
@@ -8,6 +8,7 @@ import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.event.Event;
 import com.dat3m.dartagnan.program.event.Tag;
 import com.dat3m.dartagnan.program.event.Tag.C11;
+import com.dat3m.dartagnan.program.event.arch.StoreExclusive;
 import com.dat3m.dartagnan.program.event.core.*;
 import com.dat3m.dartagnan.program.event.lang.catomic.*;
 import com.dat3m.dartagnan.program.event.lang.linux.*;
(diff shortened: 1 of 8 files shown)


hernanponcedeleon:
I just noticed I am using the wrong events for `lwarx` and `stwcx.`, I should use `ExecutionStatus`.

ThomasHaas:
You should use `StoreExclusive` in our `events.arch` package. That one compiles correctly.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -40,6 +41,20 @@ protected VisitorPower(boolean useRC11Scheme, PowerScheme cToPowerScheme) {
...
+
+    @Override
+    public List<Event> visitStoreExclusive(StoreExclusive e) {
+        RMWStoreExclusive store = newRMWStoreExclusiveWithMo(e.getAddress(), e.getMemValue(), true, e.getMo());
+
+        return eventSequence(
+                store,
+                newExecutionStatus(e.getResultRegister(), store)
+        );
+    }
```
ThomasHaas, line 56: IIRC, power has dependency tracking for failing store conditionals.

hernanponcedeleon: This was about `RegReaders` of the `regStatus` register depending on the store, right?

Do you have a reference? The only thing I found is in herd7's repo saying that [there is no dependency between the `lwarx` and its matching `stwcx.`](https://github.com/herd/herdtools7/blob/8e7e63cd400e3eae8eac22dfdbd9dae0f243319b/CHANGES.txt#L45).

ThomasHaas: Ok, if the documentation says so, then it is correct I guess :)

ThomasHaas (approved):
LGTM

PR #765, opened 2024-11-02
Improvements for SVCOMP 2025

hernanponcedeleon:
This PR makes some improvements for SVCOMP 2025:
- removes the llvm optimizations (I did some dryrun a few weeks ago and it did not seem to help much using them)
- it enables the generation of witnesses for data-race violations (which were not required until the 2024 edition)
- it skips the "trying to release an already released lock" assertion when `AssertionType.USER` is configured to be skipped.
- moves some hardcoded options from the svcomp runner to the option file


ThomasHaas (approved):
LGTM

PR #766, opened 2024-11-02
Check if events may alias when deriving RA knowledge from witnesses (…

hernanponcedeleon:
…as a safe guard)


ThomasHaas:
This is really a band-aid fix. The code is still wrong and you should put at least a FIXME.

PR #768, opened 2024-11-04
Fix execution encoding of control barriers

hernanponcedeleon:
The added test (which should not hang) shows that our current encoding of execution barriers is wrong.
The current encoding is 
```
exec(e) = cf(e) /\ forall e' in group-barriers (cf(e') \/ id(e) != id(e'))
```
which allows the barrier in thread 0 not to be executed (since the other barrier id is the same and its control flow is false) even if its control flow is true, thus getting stuck and leading to a liveness violation.

The correct encoding should be
```
exec(e) = cf(e) /\ forall e' in group-barriers ((cf(e') => exec(e')) \/ id(e) != id(e'))
```

### DIFF
--- a/benchmarks/opencl/NonUniformBarrier.cl
+++ b/benchmarks/opencl/NonUniformBarrier.cl
@@ -1,0 +1,7 @@
+__kernel void non_uniform_barrier() {
+
+    if(get_global_id(0) == 0) {
+        barrier(CLK_GLOBAL_MEM_FENCE);
+    }
+
+}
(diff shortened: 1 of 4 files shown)


ThomasHaas:
Isn't the new encoding saying that control barriers never get stuck?
I mean, the formula becomes trivially true if nobody else reaches their control barriers because the `cf => exec` implication is true then.

I also don't get why the test should not hang. You have multiple threads with the same barrier, but some do not reach theirs so the other one's are waiting forever... Seems correct to me.
What is the argument that the barriers of the different threads are not synchronizing? I think this more of an issue of modelling rather than encoding.

ThomasHaas:
But how can you tie the definition of `active` to the control-flow variable? If those concepts are identical, then control barriers cannot possible block at all and only synchronize at best, no?
Cause whenever they would block, it means an active invocation must not reach its barrier. But by definition this cannot happen, because the invocation is active only if it reaches that barrier...

Btw. isn't "invocation" talking about the kernel/the threads? Doesn't this mean all actively running threads must reach the barrier and in your dummy example this is not the case, i.e., it does not terminate because you have two active threads?
Well, unless you say that the thread that does not reach the barrier actually terminates, i.e., becomes inactive in which case the waiting thread may indeed proceed.
If so, you need to tie activeness to the non-termination of the thread.

hernanponcedeleon:
I got confirmation that the shader in this PR has UB because the workgroup has non-uniform control flow, thus I am closing.

That being said, this encoding is probably what we will need to use when adding support for [non-uniform-instructions](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_non_uniform_instructions).

PR #770, opened 2024-11-05
Support explicitly defined offsets in aggregate type

natgavrilenko:


dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/AggregateType.java
```diff
@@ -2,33 +2,61 @@
...
+            maxAlignment = types.getAlignment(fields.get(0));
+            maxOffset = types.getMemorySizeInBytes(fields.get(0));
+            offsets.add(0);
+        }
+        for (int i = 1; i < fields.size(); i++) {
+            maxAlignment = Integer.max(types.getAlignment(fields.get(i)), maxAlignment);
+            maxOffset = paddedSize(maxOffset, maxAlignment);
+            offsets.add(maxOffset);
+            maxOffset += types.getMemorySizeInBytes(fields.get(i));
+        }
```
ThomasHaas, line 40: This computation looks wrong to me. E.g. `struct { int32, int8, int8, int8 }` should have a size of 8 bytes with offsets 0, 4, 5, and 6. It seems to me that your computation will force 4-byte alignment on the `int8` members because of the 4-byte member in the beginning.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/TypeFactory.java
```diff
@@ -92,7 +109,63 @@ public IntegerType getByteType() {
...
+            return -1;
+        }
+        if (type instanceof AggregateType aType) {
+            List<TypeOffset> typeOffsets = aType.getTypeOffsets();
+            if (typeOffsets.isEmpty()) {
+                return 0;
+            }
+            if (aType.getTypeOffsets().stream().anyMatch(o -> getMemorySizeInBytes(o.type()) < 0)) {
+                return -1;
+            }
```
hernanponcedeleon, line 139: Can this if ever really happen? When would we return -1 for the size in bits of a type?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsType.java
```diff
@@ -7,20 +7,24 @@
...
 public class VisitorOpsType extends SpirvBaseVisitor<Type> {
 
     private static final TypeFactory types = TypeFactory.getInstance();
 
     private final ProgramBuilder builder;
+    private final Offset offset;
 
     public VisitorOpsType(ProgramBuilder builder) {
         this.builder = builder;
+        this.offset = (Offset) builder.getDecorationsBuilder().getDecoration(DecorationType.OFFSET);
```
hernanponcedeleon, line 27: Is this assuming that in SPIRV code the OFFSET decoration is always given? is this true?

hernanponcedeleon (approved):

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/AggregateType.java
```diff
@@ -2,33 +2,58 @@
...
 
-    AggregateType(List<Type> directFields) {
-        this.fields = List.copyOf(directFields);
+    private static final TypeFactory types = TypeFactory.getInstance();
+
+    private final List<TypeOffset> directFields;
+
+    AggregateType(List<Type> fields) {
+        this(fields, computeDefaultOffsets(fields));
+    }
```
ThomasHaas, line 20: I think its nicer to get rid of this constructor and instead have a corresponding overload in `TypeFactory.makeAggregateType` (the used helper functions would also move to `TypeFactory` of course). This should also remove the reference to `TypeFactory` in this class altogether.

Issue #767, opened 2024-11-03
Problem with XRA


hernanponcedeleon:
The attached violation witness is generated by dartagnan for [this task](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/pthread-theta/unwind2-nondet.c?ref_type=heads), but when trying to validate it, we return PASS. Disabling the XRA option results in the validation working as expected, suggesting that we have a bug in the XRA. Also, this entry from the log looks suspicious ` #must-edges added (extended): -6`. 

[witness.zip](https://github.com/user-attachments/files/17610220/witness.zip)

PR #771, opened 2024-11-05
Fix witnesses excluding the rf edges they enforce.

xeren:
Should fix issue #767.  When validating a witness, the Relation Analysis extracts `rf` edges, that must occur.  Because `rf` is (usually) functional, any other candidate in-edge for the participating load event is disabled by the witness.  The extraction code disabled even the extracted `rf` edge.  The Extended Relation Analysis then inferred mutual exclusiveness between those events.  Even without XRA, such loads always read 0, because of uninitialized memory.
The bug should affect all violation witnesses with at least one `rf` edge.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
@@ -875,9 +875,8 @@ public class NativeRelationAnalysis implements RelationAnalysis {
                 EventGraph g = witness.getReadFromKnowledge(program, alias);
                 must.addAll(g);
                 for (Event r : g.getRange()) {
-                    may.removeIf((e1, e2) -> e2 == r);
+                    may.removeIf((e1, e2) -> e2 == r && !g.contains(e1, e2));
                 }
-                may.addAll(g);
             }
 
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -877,6 +877,7 @@ public MutableKnowledge visitReadFrom(ReadFrom rf) {
                 for (Event r : g.getRange()) {
                     may.removeIf((e1, e2) -> e2 == r);
                 }
+                may.addAll(g);
```
ThomasHaas, line 880: I would change this code to do `may.removeIf((e1, e2) -> e2 == r && !g.contains(e1, e2))`.
The reason is that this way, if we actually do add an rf-edge between events that may not read from each other, we get the desired mutual exclusion.
Your solution, in some sense, introduces back imprecision in such cases (must-not may suddently become may).

hernanponcedeleon:
> The bug should affect all violation witnesses with at least one `rf` edge.

Based on the large number of witnesses we manage to validate, I was initially surprised about this claim. However, I checked a few cases and the number of must edges in `rf` was always `#threads+1` (i.e. those coming from `pthread_create`) meaning we were not getting a single `rf`-edge from the witness

PR #773, opened 2024-11-08
Remove svcomp specific options

hernanponcedeleon:


ThomasHaas (approved):
Are we sure that we do not want to introduce larger steps eventually?

Otherwise, LGTM.

hernanponcedeleon:
> Are we sure that we do not want to introduce larger steps eventually?

For next year, I am hoping to even get rid of the SVCOMPRunner class and dump/loads from files.
If we eventually want to give larger steps, that will probably also be useful for when we increase the loops of non fully unrolled loops.
In summary, if we add that it wont be svcomp specific

PR #775, opened 2024-11-09
Add support for encoding tuples (Tuple theory)

ThomasHaas:
For #772 I played around with encoding tuples and figured out it was quite easy to support.
So here is a simple implementation.

None of our processing pipeline generates tuple expressions (i.e. structs) and rather tries to eliminate them, so the new encoding is not exercised at all. I have only tested this with some experimental code so far.


hernanponcedeleon:
Does this mean we can handle vector registers? If this is the case, the simplest way to test this would be removing the register decomposition pass and check if we still handle llvm CASes (IIRC the only instruction we have that generates vector registers) correctly

ThomasHaas:
Yes, we should be able to handle struct registers. That is what I used in my experimental code to check local side effects: I made snapshots of the local state of a loop at entry/exit by storing the state in struct-typed register and comparing them in the end (if they are equal, we are local side-effect free).

I can check what happens if RegisterDecomposition is disabled. I still wouldn't disable it in geneal, because we are likely to do better optimizations on primitive types.

ThomasHaas:
Follow-up: I think SCCP now subsumes `RegisterDecomposition` for the most part, at least for LLVM CAS. But if I disable both, we get code like this (DGLM) which get encoded just fine:
```
// CAS
102:     { 0: bv64, 8: bv1 } 2:r19 <- { bv64 2:DUMMY_REG_0, bool 2:DUMMY_REG_1 ? bv1(1) : bv1(0) }
// Extract success bit (second member)
103:     bv1 2:r21 <- { 0: bv64, 8: bv1 } 2:r19[1]
// Extract old value (first member) if CAS failed
104:     bv64 2:r.013 <- (bv1 2:r21 != bv1(0)) ? bv64 2:r3 : ({ 0: bv64, 8: bv1 } 2:r19[0])
```
SCCP would propagate the struct assignment to the extract operators and then apply those statically. Then the struct register becomes dead and we would eliminate it.

ThomasHaas:
I tried to get rid of the special handling of `LlvmCmpXchg` and its `getStructRegister` methods by using proper struct types, however, it does not really result in nice code.
While the special-casing goes away, the need to wrap registers into structs for the caller to just unwrap them is annoying.
Also, we would need to allow `SCCP` or `AssignmentInlining` to propagate `ConstructExpr` to simplify the resulting code (`RegisterDecomposition` cannot even handle it).
Lastly, the amount of code changes are larger than just supporting `MultiRegWriter`s as in #571, and I think those are considerably nicer.
In short, I won't "fix" `LlvmCmpXchg`' special handling in this PR.

hernanponcedeleon:
@tonghaining and @StefanoDalMas keep an eye on this PR since I suspect it might make your life easier.

hernanponcedeleon:
> While the special-casing goes away, the need to wrap registers into structs for the caller to just unwrap them is annoying.

Then what is the point of this PR? 

Pretty much every code I have seen making using of vector registers (`LlvmCAS`, inline assembly, SPIRV code) ends up unwrapping the structure.

ThomasHaas:
Most generated code unwraps structs and that is because you cannot do operations on them, except extract and insert elements.
And precisely because of this, internally wrapping multiple return values into a struct just for the caller to unwrap them is a hassle.
I have a local branch with LLVM-CAS replaced by
But this only applies to the compiler-generated code.

There are various reasons for this PR:
(1) We can handle (=encode) code with structs without relying on `RegisterDecomposition` or similar passes. Note that even if structs are always getting unpacked eventually, until that point they exist as proper structs and thus need to be handled somehow.
(2) It allows the possibility to store/load whole aggregates. This can be useful to construct atomic memory operations that transfer complex types. Notice that in this case we cannot unpack the aggregate without duplicating stores/loads (one for each element) which results in different semantics.
(3) I need it for #772 to instrument side-effect tracking more easily. Here I want to capture the whole loop state in a single aggregate value for later comparison. We can directly encode the equality of the structs in this PR, so no wrapping/unwrapping needed.
(4) A student of mine tries to implement Pointer types with provenance. In a provenance-based encoding pointers become tuples. Note that this encoding is fully orthogonal to the program code: even if the program uses no structs, we still can use tuples in the encoding.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -249,6 +261,9 @@ public BooleanFormula equal(Formula left, Formula right) {
         if (left instanceof BooleanFormula l && right instanceof BooleanFormula r) {
             return booleanFormulaManager.equivalence(l, r);
         }
+        if (left instanceof TupleFormula l && right instanceof TupleFormula r) {
+            return new TupleFormulaManager(this).equal(l, r);
```
hernanponcedeleon, line 265: with do we need to create a new instance and not simply use `tupleFormulaManager`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -282,6 +288,33 @@ public Formula visitITEExpression(ITEExpr iteExpr) {
...
+        for (Expression inner : construct.getOperands()) {
+            elements.add(encode(inner));
+        }
+        return context.getTupleFormulaManager().makeTuple(elements);
+    }
+
+    @Override
+    public Formula visitAggregateCmpExpression(AggregateCmpExpr expr) {
+        final Formula left = encode(expr.getLeft());
+        final Formula right = encode(expr.getLeft());
```
hernanponcedeleon, line 303: `final Formula right = encode(expr.getRight());`

PR #776, opened 2024-11-09
Fix handling of empty type

ThomasHaas:
This fixes a bug introduced in #770. 

EDIT: I renamed the PR and added more fixes, mostly related to handling the empty type (+ one fix for alignments computation of arrays).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/TypeFactory.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/TypeFactory.java
@@ -5,7 +5,10 @@ import com.dat3m.dartagnan.utils.Normalizer;
 import com.google.common.math.IntMath;
 
 import java.math.RoundingMode;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.stream.IntStream;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -170,7 +173,7 @@ public final class TypeFactory {
             return getMemorySizeInBytes(arrayType.getElementType());
         }
         if (type instanceof AggregateType aType) {
-            return aType.getTypeOffsets().stream().map(o -> getAlignment(o.type())).max(Integer::compare).orElseThrow();
+            return aType.getTypeOffsets().stream().map(o -> getAlignment(o.type())).max(Integer::compare).orElse(1);
         }
         throw new UnsupportedOperationException("Cannot compute memory layout of type " + type);
     }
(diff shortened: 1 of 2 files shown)


hernanponcedeleon:
I noticed this regression when running the svcomp benchmarks. The fix restores the previous behavior.

@natgavrilenko can you please review?

ThomasHaas:
There is nothing to review really. The empty struct is 1-aligned but #776 changed the behaviour to throw an error instead.

hernanponcedeleon:
Try [this](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/ldv-races/race-2_2-container_of.c) file. It should be reproducible with all memory models.

ThomasHaas:
I just noticed that the commit message of this PR is wrong (it seems I accidentally took the one from #775). Not sure how that happened.

ThomasHaas:
Btw. I think this part of the alignment computation is wrong (*) as well:
```
        if (type instanceof ArrayType arrayType) {
            return getMemorySizeInBytes(arrayType.getElementType());
        }
```
The array should be aligned according to the alignment of its elements, not their size.

(*) This might technically not wrong because we simply overalign. But it does not match the usually expected alignment.

ThomasHaas:
Actually, there are more issues.

(1) The behaviour of `TypeFactory.getAggregateType(List<Type>)` and `TypeFactory.getAggregateType(List<Type>, List<Integer>)` differ in that the former allows for empty members but the latter does not.

(2) In C, empty structs are undefined behavior, at least in theory!!! The SVCOMP benchmark is thus problematic. GCC allows empty structs as a custom extension and gives them size 0, and since Clang tries to be compatible with GCC, it shows the same behaviour.
Interestingly, in C++ empty structs are allowed but their size is 1 as to disallow different objects occupying the same memory location.
This begs the question: what should we do? Since we do only C for now, the choice of size 0 for the empty struct is probably the best (maybe we need to support a flag in the future like `input.language=c++`). However, we then inherit the awkward situation of having different objects at the same memory location, i.e., unions.
So the preconditions on `TypeFactory.getAggregateType` need to get relaxed.

hernanponcedeleon:
> the choice of size 0 for the empty struct is probably the best

If empty structures are undefined behavior, we should report this and throw an exception when verifying C code.

@natgavrilenko what does the spirv standard say about empty structures? Can we have "This is UB" as a general solution?

ThomasHaas:
> If empty structures are undefined behavior, we should report this and throw an exception when verifying C code.

I think this solution is not practically viable because its a feature that is expected to work since GCC/Clang support it. One might say it is more implementation-defined in practice rather than undefined. On top of that you have the issue that we are technically parsing LLVM and not C directly, and so it makes sense to take LLVM semantics into account. For LLVM, empty structs are not an issue it seems.

ThomasHaas:
> [...] @ThomasHaas could you add this case to AggregateTypeTest?

will do.

PR #778, opened 2024-11-11
Fix/Change llvm parsing order to support metadata referencing globals

ThomasHaas:


ThomasHaas:
Btw. the issue is related to the fact that LLVM Metadata can reference program constants but we parsed the metadata before we parsed the declarations. I just reordered the code, but I cannot tell if this is a general fix or if metadata parsing needs to be done differently altogether.

ThomasHaas:
@hernanponcedeleon I guess you do not care about this change anymore, right? IIRC, it was necessary to parse some UBSAN instrumentation.

PR #779, opened 2024-11-15
Removed SpecId decoration

natgavrilenko:
Removed SpecId decoration. It is more convenient to overwrite SpecConstant via @Input annotation.


dartagnan/src/test/resources/spirv/basic/spec-id-boolean.spv.dis
```diff
@@ -0,0 +1,54 @@
...
+; Generator: Google Clspv; 0
+; Bound: 77
+; Schema: 0
+               OpCapability Shader
+               OpCapability VulkanMemoryModel
+               OpMemoryModel Logical Vulkan
+          %1 = OpExtInstImport "NonSemantic.ClspvReflection.5"
+               OpEntryPoint GLCompute %25 "foo" %8 %gl_LocalInvocationID %14 %18 %22
+               OpDecorate %spec_0 SpecId 0
+               OpDecorate %spec_1 SpecId 1
```
hernanponcedeleon, line 16: Do I understand correctly that these two lines have no effect and the values are actually taken from
```
; @Input: %spec_0=0
; @Input: %spec_1=1
```
?

PR #784, opened 2024-11-19
Fix CI tests

CapZTr:
Fix two CI tests as @hernanponcedeleon commented.


hernanponcedeleon:
For those test classes using yioces2, you also need to override this method
```
    @Override
    protected Configuration getConfiguration(){
        return Configuration.defaultConfiguration();
    }
```
The "default" configuration in `AbstractCTest` sets the option `OptionNames.USE_INTEGERS` which does not work with the yices2 solver.

dartagnan/src/test/java/com/dat3m/dartagnan/c/C11LFDSTest.java
```diff
@@ -55,11 +62,11 @@ protected Provider<Solvers> getSolverProvider() {
     public static Iterable<Object[]> data() throws IOException {
         // Commented ones take too long ATM
```
hernanponcedeleon, line 63: This comment can be removed

hernanponcedeleon:
The problem with the CI is that `C11OrigLFDSTest` uses option `INIT_DYNAMIC_ALLOCATIONS` while `C11LFDSTest` does not. I tried some. This option was added in #624 and at that point the [tests were already commented out](https://github.com/hernanponcedeleon/Dat3M/blob/2d816d07fa1b596395c65e6543fee4dede19bfdf/dartagnan/src/test/java/com/dat3m/dartagnan/c/C11LFDSTest.java). 

@ThomasHaas do you remember why set the option for `c11-orig.cat`, but not for `c11.cat`? My understanding from [this](https://github.com/hernanponcedeleon/Dat3M/pull/624/commits/071212e6363acd8770eee979434348bef1d60661) commit is that it should not be required to set that options for `c11.cat` to work correctly.

ThomasHaas:
The changes related to dynamic tagging are not compatible with the lack of init events:
```
let dynamic_tag(X) = [range([IW & X]; loc)]
```
This relation will be empty for addresses without init events.
Now you have an empty `mo` for writes to dynamically allocated locations:
```
let mo = dynamic_tag(~NAL); co; dynamic_tag(~NAL)
```
This will surely break all sorts of axioms (e.g. SC-per-loc on such addresses would not hold).

Btw. I raised this point in the relevant [PR](https://github.com/hernanponcedeleon/Dat3M/pull/744#discussion_r1780121417) that changed it. My impression there was that the changed model is only applied to litmus code!

hernanponcedeleon:
Ok, I guess that for now the best we can do is to use that option when verifying real code wrt `c11.cat` or `c11-orig.cat`.

PR #786, opened 2024-11-22
Improve spinloop detection

ThomasHaas:
Using the new tuple support from #775, this PR improves the detection of local side effects in loops.
Rather than treating any assignment to live variables as a side-effect, we now check that the register's values have actually changed. In other words, the check is more semantic now.
This PR also fixes a bug where we failed to detect certain side-effects in cases where the loop only writes to but does not read from a register (only happened in Libvsync benchmarks AFAIK).

I know of at least one verdict that has changed (`clh_mutex`) but I haven't updated the unit tests yet.
After the CI runs through, I will update the unit tests with the new expectation.


ThomasHaas:
FYI, this is a small precursor of #772.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicSpinLoopDetection.java
```diff
@@ -72,167 +70,108 @@ private List<LoopData> computeLoopData(Function func, LoopAnalysis loopAnalysis)
...
-            }
-        }
+        // Every live register that is written to is a potential local side effect.
+        loop.writtenLiveRegisters.addAll(Sets.intersection(
+                writtenRegisters,
+                liveRegsAna.getLiveRegistersAt(loop.getStart())
+        ));
     }
 
     private void instrumentLoop(LoopData loop) {
```
hernanponcedeleon, line 96: I find it hard to keep track in my head on where all the new events are added. Please add some documentation on how the IR will look like after the instrumentation, something along the lines of [this](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveLoopBoundAnnotation.java#L32-L43).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicSpinLoopDetection.java
```diff
@@ -72,167 +70,108 @@ private List<LoopData> computeLoopData(Function func, LoopAnalysis loopAnalysis)
...
     }
 
-    private Event newSpinTerminator(Expression guard, Function func) {
+    private Event newSpinTerminator(Expression guard, LoopData loop) {
+        final Function func = loop.getStart().getFunction();
         final Event terminator = func instanceof Thread thread ?
                 EventFactory.newJump(guard, (Label) thread.getExit())
                 : EventFactory.newAbortIf(guard);
-        terminator.addTags(Tag.SPINLOOP, Tag.NONTERMINATION, Tag.NOOPT);
+        terminator.addTags(Tag.SPINLOOP, Tag.NONTERMINATION);
```
hernanponcedeleon, line 154: Why did you remove the `Tag.NOOPT` tag?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicSpinLoopDetection.java
```diff
@@ -72,167 +70,108 @@ private List<LoopData> computeLoopData(Function func, LoopAnalysis loopAnalysis)
...
-            if (!loop.isAlwaysSideEffectful) {
-                numPotentialSpinLoops++;
-                if (loop.sideEffects.isEmpty()) {
-                    numStaticSpinLoops++;
-                }
-            }
-        }
-        return new AnalysisStats(numPotentialSpinLoops, numStaticSpinLoops);
+        int numStaticSpinLoops = Math.toIntExact(loops.stream().filter(LoopData::isSideEffectFree).count());
+        return new AnalysisStats( numStaticSpinLoops);
```
hernanponcedeleon, line 161: Remove extra space

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicSpinLoopDetection.java
ThomasHaas: It's a good question and I don't know the exact answer to this. The first version of Spinloop detection didn't have it and it is not needed in principle. I only vaguely remember that I added the NOOPT a while ago to work around some issue, maybe related to `SyntacticContextAnalysis`?

hernanponcedeleon: I tracked the history of this pass. The pass has been renamed / moved several times, but I think [this](https://github.com/hernanponcedeleon/Dat3M/commit/05c5d18626227f45765a0dc5bbcb1faa8b742935#diff-43ac8b13a92d3b7c3af2ac0ef91c6582b30c172ed8a157708f3123eb7b3059dd) is the commit where the tag was added. It is not really clear if the tag was needed specifically for spinloops.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/DynamicSpinLoopDetection.java
hernanponcedeleon: Thanks. This made reviewing much easier.

hernanponcedeleon (approved):

PR #792, opened 2025-01-15
Add support for OpenCL SPIR-V

tonghaining:


ThomasHaas:
Oh, this is a big one, but I immediately have two questions:

(1) What is the purpose of `ScopedPointerVariable`? It seems to me that in the whole processing pipeline and the final encoding, it is always treated like a simple address/memory object. If so, a processing pass that just eliminates them altogether would be a lot better.

(2) Why does `FunctionCall` need to remember expressions transformers that are used during `Inlining`? Is this somehow related to the scoped pointers?

tonghaining:
(1) Thanks for pointing this out! Yes `ScopedPointerVariable` visitors are not necessary.  I think I initially added them while working on accessing function parameters but forgot to remove them afterward. I’ve now removed all of them.

(2) It was not relevant to the scoped pointers. We now have new cases where one test has multiple functions.

ThomasHaas:
> (2) It was not relevant to the scoped pointers. We now have new cases where one test has multiple functions

I don't understand this point. C-code also has multiple functions, yet we do not need special transformers when inlining.
What is different for SPIRV/OpenCL code? Do you need to translate pointers to thread_local memory objects or something?

tonghaining:
Yes for pointers inside a function call, we sometimes need to translate them into thread-local/workgroup-local memory objects. Does C-code has similar cases?

ThomasHaas:
Ok, I guessed so. But then you are facing a different issue.
In C-code, we first do inlining and then create the threads, so the thread creation pass actually resolves/translates thread-local memory accesses.
In your case, I believe the parser already creates threads during parsing time, i.e., before inlining. 

I think there are two canonical solutions to this:
(1) Do not create threads during parsing time. Instead, we could mark a function as "entry point" (in C-code this is just `main`, for litmus-like code it would be each process) and let the thread creation pass generate the threads. In this case you would actually need to attach GPU-specific scope information already to functions instead of threads.
(2) Let `Inlining` do its usual job and have an extra pass that resolves/translates thread-local accesses if possible. This translation should not make use of custom expression transformers attached to function calls. If anything, I think the necessary information should be attached to the threads (or even better: be computable from the threads).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -233,6 +234,20 @@ public static CondJump newFakeCtrlDep(Register reg, Label target) {
...
+    public static List<CondJump> newSwitch(Expression selector, Label defaultTarget, Map<Expression, Label> cases) {
+        List<CondJump> jumps = new ArrayList<>();
+        Expression defaultCond = expressions.makeFalse();
+        for (Expression caseExpr : cases.keySet()) {
+            Expression cond = expressions.makeEQ(selector, caseExpr);
+            defaultCond = expressions.makeOr(defaultCond, cond);
+            CondJump jump = newJump(cond, cases.get(caseExpr));
+            jumps.add(jump);
+        }
+        CondJump defaultJump = newJump(defaultCond, defaultTarget);
```
hernanponcedeleon, line 246: shouldn't this be `not defaultCond`? i.e. if none of the other cases matched

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Local.java
```diff
@@ -20,7 +20,7 @@ public class Local extends AbstractEvent implements RegWriter, RegReader {
     protected Expression expr;
 
     public Local(Register register, Expression expr) {
-        Preconditions.checkArgument(register.getType().equals(expr.getType()));
+        Preconditions.checkArgument(expr.getType().equals(register.getType()));
```
hernanponcedeleon, line 23: why do you need to reverse the order here?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,238 @@
...
+        store.removeTags(store.getTags());
+        store.addTags(toOpenCLTags(e.getTags()));
+        return eventSequence(store);
+    }
+
+    @Override
+    public List<Event> visitSpirvLoad(SpirvLoad e) {
+        String mo = moToOpenCLTag(Tag.Spirv.getMoTag(e.getTags()));
+        Load load = newLoadWithMo(e.getResultRegister(), e.getAddress(), mo);
+        load.setFunction(e.getFunction());
```
hernanponcedeleon, line 44: why do spirv events need to propagate the function from the original event?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,238 @@
...
+        fence.removeTags(fence.getTags());
+        fence.addTags(toOpenCLTags(e.getTags()));
+        replaceAcqRelTag(fence, Tag.C11.MO_ACQUIRE, Tag.C11.MO_RELEASE);
+        return eventSequence(fence);
+    }
+
+    @Override
+    public List<Event> visitControlBarrier(ControlBarrier e) {
+        Event entryFence = EventFactory.newControlBarrier(e.getName() + "_entry", e.getId());
+        entryFence.addTags(Tag.OpenCL.ENTRY_FENCE, Tag.C11.MO_RELEASE);
```
hernanponcedeleon, line 133: I think we discussed about getting rid of this fence tags and treat barriers in OpenCL the same we do for the other languages (IIRC this was only needed for barrier divergence; we can figure out later how to handle that property)

hernanponcedeleon (commented):
I just saw Natalia stated her review too, so some comments might be repeated

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        Type type = builder.getType(typeId);
+        List<Integer> indexes = idxContexts.stream()
+                .map(SpirvParser.IndexesLiteralIntegerContext::getText)
+                .map(Integer::parseInt)
+                .toList();
+        Expression element = compositeExpression;
+        for (Integer index : indexes) {
+            element = element.getOperands().get(index);
+        }
```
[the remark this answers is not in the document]
ThomasHaas: Each iteration goes down a nesting level. It's not a simple a loop over all elements.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -142,6 +179,40 @@ public Event visitOpReturnValue(SpirvParser.OpReturnValueContext ctx) {
...
+        String pointerId = ctx.pointer().getText();
+        int size = Integer.parseInt(ctx.sizeLiteralInteger().getText());
+        Expression pointerExp = builder.getExpression(pointerId);
+        if (!(pointerExp instanceof ScopedPointerVariable pointerVariable)
+                || !pointerVariable.getScopeId().equals(Tag.Spirv.SC_FUNCTION) ) {
+            throw new ParsingException("Lifetime start can only be applied to a pointer with Function storage class: '%s'", pointerId);
+        }
+        Register register = builder.addRegister(pointerId + "_alloc", pointerVariable.getType());
+        IntegerType pointerIntegerType = TypeFactory.getInstance().getArchType();
+        Expression sizeExpression = new IntLiteral(pointerIntegerType, new BigInteger(Long.toString(size)));
```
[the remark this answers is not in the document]
ThomasHaas: The code shouldn't call `new IntLiteral` in the first place. Expressions should be constructed via `ExpressionFactory`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -58,15 +58,37 @@ public Event visitOpStore(SpirvParser.OpStoreContext ctx) {
...
+                String elementId = resultId + "_" + i;
+                Register register = builder.addRegister(elementId, arrayType.getElementType());
+                registers.add(register);
+                List<Expression> index = List.of(new IntLiteral(types.getArchType(), new BigInteger(Long.toString(i))));
+                Expression elementPointer = HelperTypes.getMemberAddress(pointerId, pointer, arrayType, index);
+                Event load = EventFactory.newLoad(register, elementPointer);
+                events.add(load);
+            }
+            Expression arrayRegister = expressions.makeArray(arrayType.getElementType(), registers, true);
+            builder.addExpression(resultId, arrayRegister);
```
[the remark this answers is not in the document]
ThomasHaas: Also replace the `new IntLiteral` constructor call.
Btw. tuple formulas do cause some issues in model extraction right now. I think we should consider implementing them in JavaSMT to avoid workarounds/special-casing for tuple formulas.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
ThomasHaas: This should definitely be avoided because it causes inconsistent state: the event thinks it belongs to a function, but it is not inserted into the function yet. Basically this function should never be called directly (maybe we can make it private, not sure though).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -157,10 +228,6 @@ private Event visitLoopBranch(String labelId) {
 
     private Event visitConditionalJump(Expression guard, String trueLabelId, String falseLabelId) {
         if (cfBuilder.isBlockStarted(trueLabelId)) {
-            if (cfBuilder.isBlockStarted(falseLabelId)) {
```
natgavrilenko, line 160: Do you have an example of spirv with this structure? I don't see any in gpuverify tests. I want to check how dartagnan unrolls such a weird loop.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -17,6 +17,7 @@
 import com.dat3m.dartagnan.program.event.Event;
 import com.dat3m.dartagnan.program.memory.FinalMemoryValue;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
+import com.dat3m.dartagnan.program.memory.ScopedPointerVariable;
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsAnnotation.java
```diff
@@ -33,7 +33,8 @@ public Void visitOpDecorate(SpirvParser.OpDecorateContext ctx) {
             case BINDING, SPEC_ID -> {
                 // Skip, defines the order of pointers passed by the host
             }
-            case ARRAY_STRIDE, BLOCK, BUFFER_BLOCK, COHERENT, DESCRIPTOR_SET, NO_CONTRACTION, NO_PERSPECTIVE, NON_WRITABLE -> {
+            case ARRAY_STRIDE, BLOCK, BUFFER_BLOCK, COHERENT, DESCRIPTOR_SET, OFFSET, NO_CONTRACTION, NO_PERSPECTIVE, NON_WRITABLE,
```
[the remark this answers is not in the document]
tonghaining: Removed.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,56 @@
...
+        extractCompositeElement(ctx.idResult().getText(), ctx.idResultType().getText(),
+                ctx.composite().getText(), ctx.indexesLiteralInteger());
+        return null;
+    }
+
+    private void extractCompositeElement(String id, String typeId, String compositeId,
+                                    List<SpirvParser.IndexesLiteralIntegerContext> idxContexts) {
+        Expression compositeExpression = builder.getExpression(compositeId);
+        if (!(compositeExpression instanceof ConstructExpr)) {
+            throw new UnsupportedOperationException("Composite extraction is only supported for ConstructExpr");
```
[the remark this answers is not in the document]
tonghaining: Changed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsConstant.java
```diff
@@ -222,10 +222,10 @@ public Set<String> getSupportedOps() {
                 "OpConstant",
                 "OpConstantComposite",
                 "OpConstantNull",
+                "OpSpecConstantComposite",
```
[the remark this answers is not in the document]
tonghaining: I was trying to order them, but I think the new order doesn't make sense, I'll change it back

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsExtension.java
```diff
@@ -17,11 +18,14 @@ public class VisitorOpsExtension extends SpirvBaseVisitor<Void> {
...
     private final Map<String, VisitorExtension<?>> availableVisitors = new HashMap<>();
     private final Map<String, String> visitorIds = new HashMap<>();
+    private final ProgramBuilder builder;
 
     public VisitorOpsExtension(ProgramBuilder builder) {
+        this.builder = builder;
         VisitorExtensionClspvReflection clspv = new VisitorExtensionClspvReflection(builder);
         this.availableVisitors.put("NonSemantic.ClspvReflection.5", clspv);
         this.availableVisitors.put("NonSemantic.ClspvReflection.6", clspv);
+        this.availableVisitors.put("OpenCL.std", clspv); // TODO: add OpenCL Reflection?
```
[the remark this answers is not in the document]
tonghaining: Created a new visitorExtension for opencl std.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsExtension.java
```diff
@@ -39,13 +43,18 @@ public Void visitOpExtInstImport(SpirvParser.OpExtInstImportContext ctx) {
...
     public Void visitOpExtInst(SpirvParser.OpExtInstContext ctx) {
+        String result = ctx.idResult().getText();
         String id = ctx.set().getText();
         String name = visitorIds.get(id);
         if (name != null) {
             VisitorExtension<?> visitor = availableVisitors.get(name);
             String instruction = getFirstTokenText(ctx.instruction());
             if (visitor.getSupportedInstructions().contains(instruction)) {
-                ctx.accept(visitor);
+                Object object = ctx.accept(visitor);
```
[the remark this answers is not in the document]
tonghaining: moved to VisitorExtensionOpenclStd.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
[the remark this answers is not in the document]
tonghaining: Removed.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/Expression.java
```diff
@@ -4,6 +4,7 @@
 import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.memory.FinalMemoryValue;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
+import com.dat3m.dartagnan.program.memory.ScopedPointerVariable;
```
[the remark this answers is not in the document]
tonghaining: removed.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionVisitor.java
```diff
@@ -14,6 +14,8 @@
 import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.memory.FinalMemoryValue;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
+import com.dat3m.dartagnan.program.memory.ScopedPointer;
+import com.dat3m.dartagnan.program.memory.ScopedPointerVariable;
```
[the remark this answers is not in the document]
tonghaining: removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -134,6 +166,11 @@ public Event visitOpReturnValue(SpirvParser.OpReturnValueContext ctx) {
         if (!types.getVoidType().equals(returnType)) {
             String valueId = ctx.valueIdRef().getText();
             Expression expression = builder.getExpression(valueId);
+            if (expression instanceof ScopedPointerVariable) {
+                Register register = builder.addRegister(valueId + "_res", returnType);
+                builder.addEvent(EventFactory.newLocal(register, expression));
+                expression = register;
+            }
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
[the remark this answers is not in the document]
tonghaining: updated

dartagnan/src/main/java/com/dat3m/dartagnan/expression/Expression.java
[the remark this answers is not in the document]
tonghaining: removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
tonghaining: I think these missing examples are from tests I generated beforehand(with using clang's default optimization).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
tonghaining: For single event they are not useful, but for RMW events which are translated to C11 event sequences, functions are needed for creating dummy registers.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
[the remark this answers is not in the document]
tonghaining: sorry I missed that, changed now

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Local.java
```diff
@@ -20,6 +20,9 @@ public class Local extends AbstractEvent implements RegWriter, RegReader {
     protected Expression expr;
 
     public Local(Register register, Expression expr) {
+        if (!register.getType().equals(expr.getType())) {
+            throw new IllegalArgumentException("Type mismatch between register and expression");
+        }
```
ThomasHaas, line 25: This check already exists in the line below (`Preconditions.checkArgument(...)`)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -169,6 +169,9 @@ private static Event inlineBody(FunctionCall call, Snapshot callTarget, int scop
         var substitution = new ExprTransformer() {
             @Override
             public Expression visitRegister(Register register) {
+                if (!registerMap.containsKey(register)) {
+                    return register;
+                }
                 return checkNotNull(registerMap.get(register));
```
ThomasHaas, line 175: Why this change? This looks dangerous

dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
```diff
@@ -123,7 +123,8 @@ public Register newRegister(String name, Type type) {
             final String error = String.format("Register %s already exists in function %s", name, this);
             throw new MalformedProgramException(error);
         }
-        Register register = new Register(name, this, type);
+        Register register;
+        register = new Register(name, this, type);
```
ThomasHaas, line 127: Why split into two lines?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/RMWExtremum.java
```diff
@@ -0,0 +1,36 @@
+package com.dat3m.dartagnan.program.event.common;
+
+import com.dat3m.dartagnan.expression.Expression;
+import com.dat3m.dartagnan.expression.integers.IntCmpOp;
+import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.event.EventVisitor;
+import com.dat3m.dartagnan.program.event.Tag;
+
+public class RMWExtremum extends RMWXchgBase {
+
```
ThomasHaas, line 10: Call it `RMWExtremumBase` and put the `@NoInterface` annotation to be consistent with the other classes.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Local.java
tonghaining: I forgot to drop them after using it for debug, removed now, thanks!

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/RMWExtremum.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/Thread.java
```diff
@@ -8,6 +9,7 @@
 import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.google.common.base.Preconditions;
 
+import java.util.HashSet;
```
ThomasHaas, line 12: Unused imports

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -64,7 +64,7 @@ public void run(Program program) {
     }
 
     private record Snapshot(String name, List<Register> parameters, List<Event> events, List<Register> registers,
-            boolean isVarArgs) {}
+                            boolean isVarArgs) {}
```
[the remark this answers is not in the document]
tonghaining: Changed back

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -13,6 +13,7 @@
 import com.dat3m.dartagnan.program.event.core.*;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadArgument;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
+import com.dat3m.dartagnan.program.memory.ScopedPointerVariable;
```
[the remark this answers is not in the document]
tonghaining: Dropped for all

dartagnan/src/main/java/com/dat3m/dartagnan/program/Thread.java
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -142,6 +179,40 @@ public Event visitOpReturnValue(SpirvParser.OpReturnValueContext ctx) {
...
+        Expression pointerExp = builder.getExpression(pointerId);
+        if (!(pointerExp instanceof ScopedPointerVariable pointerVariable)
+                || !pointerVariable.getScopeId().equals(Tag.Spirv.SC_FUNCTION)) {
+            throw new ParsingException("Lifetime stop can only be applied to a pointer with Function storage class: '%s'", pointerId);
+        }
+        int pointerSize = pointerVariable.getAddress().getKnownSize();
+        if (pointerSize != size) {
+            throw new ParsingException("Lifetime stop size does not match the size of the pointer: '%s'", pointerId);
+        }
+        builder.removeExpression(pointerId);
```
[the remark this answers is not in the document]
tonghaining: I removed these Op from this PR since there are no usage of them in the current tests set.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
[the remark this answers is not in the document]
tonghaining: `new IntLiteral` replaced.
As we talked with @hernanponcedeleon I'll keep the multi-load for now to be consistent with `BuiltIn` since it also uses separate inits there (and Vulkan-flavor SPIRV only access specific element of array with single load).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -135,6 +154,80 @@ private void validateVariableStorageClass(ScopedPointerVariable pointer, String
...
+                }
+                store.addTags(sourceTags);
+            }
+        }
+        builder.addEvent(store);
+        return null;
+    }
+
+    @Override
+    public Event visitOpCopyMemorySized(SpirvParser.OpCopyMemorySizedContext ctx) {
```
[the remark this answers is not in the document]
tonghaining: I removed this OP and commented out the corresponding tests

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -75,6 +84,29 @@ public Event visitOpSelectionMerge(SpirvParser.OpSelectionMergeContext ctx) {
         throw new ParsingException("End label must be null");
     }
 
+    @Override
+    public Event visitOpSwitch(SpirvParser.OpSwitchContext ctx) {
```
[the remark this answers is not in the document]
tonghaining: I removed these Op from this PR since they were only used in the -O1 optimization test set I generated at first.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -142,6 +179,40 @@ public Event visitOpReturnValue(SpirvParser.OpReturnValueContext ctx) {
                 builder.getCurrentFunctionName());
     }
 
+    @Override
+    public Event visitOpLifetimeStart(SpirvParser.OpLifetimeStartContext ctx) {
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
```diff
@@ -134,6 +166,11 @@ public Event visitOpReturnValue(SpirvParser.OpReturnValueContext ctx) {
         if (!types.getVoidType().equals(returnType)) {
             String valueId = ctx.valueIdRef().getText();
             Expression expression = builder.getExpression(valueId);
+            if (expression instanceof ScopedPointerVariable) {
```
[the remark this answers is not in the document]
tonghaining: Removed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/vulkan/VulkanRMWExtremumBase.java
```diff
@@ -5,25 +5,17 @@
 import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.event.EventVisitor;
 import com.dat3m.dartagnan.program.event.Tag;
-import com.dat3m.dartagnan.program.event.common.RMWXchgBase;
+import com.dat3m.dartagnan.program.event.common.RMWExtremumBase;
 
-public class VulkanRMWExtremum extends RMWXchgBase {
+public class VulkanRMWExtremumBase extends RMWExtremumBase {
```
ThomasHaas, line 10: This event should not have the "Base" suffix. 
I used "Base" as a suffix for all classes that only act as an implementation helper but otherwise provide no interface.
What this means is that (1) the base class should never be used as a type anywhere (neither as parameter type, return type, nor type in an instanceof check) and consequently (2) the inheritance from a base class can always be replaced by copying code from the base class into the inheriting class without breaking anything in Dartagnan.
In other words, no class needs to inherit from base classes ever (but they may if it makes the implementation easier).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/opencl/OpenCLRMWExtremumBase.java
```diff
@@ -0,0 +1,35 @@
+package com.dat3m.dartagnan.program.event.arch.opencl;
+
+import com.dat3m.dartagnan.expression.Expression;
+import com.dat3m.dartagnan.expression.integers.IntCmpOp;
+import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.event.EventVisitor;
+import com.dat3m.dartagnan.program.event.Tag;
+import com.dat3m.dartagnan.program.event.common.RMWExtremumBase;
+
+public class OpenCLRMWExtremumBase extends RMWExtremumBase {
```
ThomasHaas, line 10: Remove "Base" from name (see `VulkanRMWExtremumBase` for explanation).

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/functions/FunctionCall.java
```diff
@@ -3,6 +3,7 @@
 import com.dat3m.dartagnan.expression.Expression;
 import com.dat3m.dartagnan.expression.ExpressionVisitor;
 import com.dat3m.dartagnan.expression.Type;
+import com.dat3m.dartagnan.expression.processing.ExprTransformer;
```
ThomasHaas, line 6: Unused import (didn't you remove them?)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/vulkan/VulkanRMWExtremumBase.java
tonghaining: I see, renamed

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/opencl/OpenCLRMWExtremumBase.java
tonghaining: updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/functions/FunctionCall.java
tonghaining: Removed, I used code inspector this time to find unused imports

dartagnan/src/main/antlr4/Spirv.g4
```diff
@@ -1741,6 +1741,12 @@ glsl
     |   unpackSnorm4x8
     |   unpackUnorm2x16
     |   unpackUnorm4x8
+    |   sAddSat
```
natgavrilenko: This is a block for glsl extension. Let's create a separate block for the instructions from opencl extension and add them here 
// Extensions
literalExtInstInteger
    :   clspvReflection
    |   glsl
    ;

dartagnan/src/main/antlr4/SpirvLexer.g4
```diff
@@ -1844,6 +1845,12 @@ ModeExt_InterpolateAtOffset : 'InterpolateAtOffset';
 ModeExt_NMin : 'NMin';
 ModeExt_NMax : 'NMax';
 ModeExt_NClamp : 'NClamp';
+ModeExt_SAddSat : 's_add_sat';
+ModeExt_SSubSat : 's_sub_sat';
+ModeExt_UAddSat : 'u_add_sat';
+ModeExt_USubSat : 'u_sub_sat';
+ModeExt_Shuffle : 'shuffle';
+ModeExt_Shuffle2 : 'shuffle2';
```
natgavrilenko: Could you separate opencl instructions from glsl instructions, just an empty line should be fine

dartagnan/src/main/antlr4/Spirv.g4
```diff
@@ -51,7 +51,7 @@ assertionCompare
     ;
 
 assertionValue
-    :   varName indexValue*
+    :   ModeHeader_And? varName indexValue*
```
natgavrilenko, line 54: I would drop this ModeHeader_And and keep the format of annotations similar for vulkan and opencl. Then, in the output visitor it can be something like this:
String name = ctx.varName().getText();
        Expression expression = builder.getExpression(name);
        if (expression instanceof Register && expression.getType() instanceof ScopedPointerType) {
            expression = builder.getExpression(HelperInputs.castPointerId(name));
        }
        if (expression instanceof ScopedPointerVariable base) {
        ...
        } else {
        // throw error
        }

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final ProgramBuilder builder;
+
+    public VisitorOpsComposite(ProgramBuilder builder) {
+        this.builder = builder;
+    }
+
+    @Override
+    public Event visitOpCompositeExtract(SpirvParser.OpCompositeExtractContext ctx) {
+        extractCompositeElement(ctx.idResult().getText(), ctx.idResultType().getText(),
```
natgavrilenko, line 27: Why not to have the body here instead of calling the private method?

ThomasHaas:
> [...] there will be a weird loop with two backward jumps, and I wonder how it is unrolled.

We normalize such loops first:
- Only the last backward jump remains (if it is unconditional, otherwise we introduce a new one).
- Every other backward jump is transformed to a forward jump that jumps to the last backward jump.

dartagnan/src/main/antlr4/Spirv.g4
```diff
@@ -19,7 +19,7 @@ outputHeader : ModeHeader_Output ModeHeader_Colon assertionList;
 filterHeader : ModeHeader_Filter ModeHeader_Colon assertion;
 configHeader : ModeHeader_Config ModeHeader_Colon literanHeaderUnsignedInteger ModeHeader_Comma literanHeaderUnsignedInteger ModeHeader_Comma literanHeaderUnsignedInteger;
 initList : init (ModeHeader_Comma init)*;
-init : varName ModeHeader_Equal initValue;
+init : varName ModeHeader_Equal initValue indexValue*;
```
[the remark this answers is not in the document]
tonghaining: Dropped

dartagnan/src/main/antlr4/Spirv.g4
tonghaining: This is under the opencl extension in the new generated grammar

dartagnan/src/main/antlr4/SpirvLexer.g4
tonghaining: Moved under the opencl extension in the new generated grammar

dartagnan/src/main/antlr4/Spirv.g4
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,66 @@
...
+import com.dat3m.dartagnan.expression.type.TypeFactory;
+import com.dat3m.dartagnan.parsers.SpirvBaseVisitor;
+import com.dat3m.dartagnan.parsers.SpirvParser;
+import com.dat3m.dartagnan.parsers.program.visitors.spirv.builders.ProgramBuilder;
+import com.dat3m.dartagnan.program.event.Event;
+
+import java.util.List;
+import java.util.Set;
+
+public class VisitorOpsComposite extends SpirvBaseVisitor<Event> {
```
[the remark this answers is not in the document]
tonghaining: Updated to `SpirvBaseVisitor<Void>`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsConversion.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public Void visitOpBitcast(SpirvParser.OpBitcastContext ctx) {
+        String id = ctx.idResult().getText();
+        String typeId = ctx.idResultType().getText();
+        if (builder.getType(typeId) instanceof ScopedPointerType pointerType) {
+            String operand = ctx.operand().getText();
+            Expression operandExpr = builder.getExpression(operand);
+            if (!(operandExpr instanceof ScopedPointerVariable scopedPointerVariable)) {
+                throw new ParsingException("Type '%s' is not a pointer type", operand);
```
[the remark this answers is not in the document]
tonghaining: Error msg updated

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsConversion.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+            if (!pointerType.getScopeId().equals(scopedPointerVariable.getScopeId())) {
+                throw new ParsingException("Storage class mismatch in OpBitcast between '%s' and '%s'", typeId, operand);
```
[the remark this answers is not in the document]
tonghaining: Added id to all the error mesages.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,66 @@
...
+            throw new ParsingException("Composite extraction is only supported for ConstructExpr");
+        }
+        Type type = builder.getType(typeId);
+        List<Integer> indexes = idxContexts.stream()
+                .map(SpirvParser.IndexesLiteralIntegerContext::getText)
+                .map(Integer::parseInt)
+                .toList();
+        Expression element = compositeExpression;
+        for (Integer index : indexes) {
+            element = element.getOperands().get(index);
```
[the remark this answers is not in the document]
tonghaining: Check and tests added

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsControlFlow.java
tonghaining: Added back

hernanponcedeleon:
@tonghaining please rebase and fix the conflicts

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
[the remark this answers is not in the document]
natgavrilenko: Looks good to me, but could you make error messages more informative (at least including id). It will not be very helpful if a user sees something "Index out of bounds: 3".

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
```diff
@@ -86,13 +91,21 @@ public void setEntryPointId(String id) {
...
+    }
+
     public void setArch(Arch arch) {
         if (this.arch != null) {
             throw new ParsingException("Illegal attempt to override memory model");
         }
         this.arch = arch;
     }
 
+    public Arch getArch() {
```
natgavrilenko, line 105: unused?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
```diff
@@ -26,6 +25,7 @@ public class ProgramBuilder {
 
     protected final Map<String, Type> types = new HashMap<>();
     protected final Map<String, Expression> expressions = new HashMap<>();
+    protected final Map<String, Expression> parameterValues = new HashMap<>();
```
natgavrilenko, line 28: Any reason why we cannot use the existing "expressions" map for this?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
```diff
@@ -174,26 +187,36 @@ public MemoryObject allocateVariable(String id, int bytes) {
...
     }
 
+    public void addParameterValue(String id, Expression pointer) {
+        if (parameterValues.containsKey(id)) {
+            throw new ParsingException("Duplicated register pointer definition '%s'", id);
+        }
+        parameterValues.put(id, pointer);
+    }
+
     public Register addRegister(String id, String typeId) {
```
natgavrilenko, line 214: Can we keep only one addRegister method (whichever you prefer)?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
```diff
@@ -36,6 +36,7 @@ public class ProgramBuilder {
     protected String entryPointId;
     protected Arch arch;
     protected Set<String> nextOps;
+    private int nextFunctionId = 0;
```
natgavrilenko, line 39: It is a matter of taste, but I would try to keep ProgramBuilder simple and move nextFunctionId into VisitorOpsFunction. For ThreadCreator, we can pass the whole program + entryPointId, then it can itself find the largest used function Id and pick the next one.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,66 @@
...
+                .map(Integer::parseInt)
+                .toList();
+        Expression element = compositeExpression;
+        for (Integer index : indexes) {
+            if (!(element instanceof ConstructExpr)) {
+                throw new ParsingException("Element is not a ConstructExpr at index: " + index);
+            }
+            List<Expression> operands = element.getOperands();
+            if (index >= operands.size()) {
+                throw new ParsingException("Index out of bounds: " + index);
```
natgavrilenko, line 43: Could you add ids to the error messages? I think a user will be confused if they see just "Index out of bounds: N"

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
natgavrilenko: The exception is correct, arch is set according to OpMemoryModel (vulkan / opencl), which should be defined only once

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/utils/ThreadCreator.java
```diff
@@ -17,43 +18,75 @@
...
         ScopeHierarchy scope = grid.getScoreHierarchy(tid);
         Thread thread = new Thread(name, type, args, tid, start, scope, Set.of());
-        thread.copyDummyCountFrom(function);
-        copyThreadEvents(thread);
+        thread.copyDummyCountFrom(entryFunction);
+        transformer.setTransferFunction(thread);
+        copyFunctionEvents(entryFunction, thread);
         transformReturnEvents(thread);
+        // Create thread-local sub-functions
+        for (Function subFunction : subFunctions) {
```
natgavrilenko, line 63: I see that we cannot use dartagnan's "normal" inlining because it needs to be done before thead creation. But now it does inlining for each individual thread. I think it would be more efficient to first inline functions into the entry point and then create threads.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunction.java
```diff
@@ -87,15 +88,67 @@ public Void visitOpFunctionParameter(SpirvParser.OpFunctionParameterContext ctx)
...
+    }
+
+    private Expression castInput(String id, ScopedPointerType pType) {
+        if (builder.hasInput(id)) {
+            Expression input = builder.getInput(id);
+            Type inputType = input.getType();
+            Type runtimeType = HelperInputs.castInputType(id, pType, inputType);
+            return HelperInputs.castInput(id, runtimeType, input);
+        }
+        // TODO: This is a quick default struct with 10 undefined elements.
```
natgavrilenko, line 138: Could you remove the comment and add a static constant for the default size? Also, since you have a proper case now, we can remove the agType and use the aType directly.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
ThomasHaas: Ah, it is the `ProgramBuilder` inside the `spirv/` directory. Then it makes more sense.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,253 @@
...
+        rmw.addTags(toOpenCLTags(e.getTags()));
+        return visitOpenCLRMWExtremum(rmw);
+    }
+
+    @Override
+    public List<Event> visitGenericVisibleEvent(GenericVisibleEvent e) {
+        Event fence = new GenericVisibleEvent(e.getName(), Tag.FENCE);
+        fence.removeTags(fence.getTags());
+        fence.addTags(toOpenCLTags(e.getTags()));
+        replaceAcqRelTag(fence, Tag.C11.MO_ACQUIRE, Tag.C11.MO_RELEASE);
```
natgavrilenko, line 149: Why replacing with separate tags? Doesn't opencl model have release-acquire memory order?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
tonghaining: We are using it to hold the external parameter pointers, which will be used by each thread to create the parameter Local events. I think we can also add some prefix to the id and use expressions instead (like we did in HelperInputs.castPointerId()), which do you think would be better?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
tonghaining: Updated, I kept the original one

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/helpers/HelperTags.java
```diff
@@ -134,6 +136,30 @@ public static String parseStorageClass(String cls) {
         };
     }
 
+    public static void addFeatureTags(MemoryObject memObj, String storageClass, Arch arch) {
```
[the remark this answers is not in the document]
tonghaining: Moved to VisitorSpirvOpenCL.

Since in OpenCL model we need to add storage class to Init event, and in current ProcessingManager the memory allocation happens after compilation, I added an extra function addTagToMemObj() in programBuilder  to map the SPV tags to OpenCL ones. This is partially duplicate with the map in VisitorSpirvOpenCL.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/utils/ThreadCreator.java
tonghaining: Added SPV thread creator into ThreadCreation pass.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunction.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -237,6 +238,20 @@ public static CondJump newFakeCtrlDep(Register reg, Label target) {
         return jump;
     }
 
+    public static List<CondJump> newSwitch(Expression selector, Label defaultTarget, Map<Expression, Label> cases) {
```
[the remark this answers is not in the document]
tonghaining: Removed

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import static com.dat3m.dartagnan.program.event.EventFactory.*;
+
+public class VisitorSpirvOpenCL extends VisitorC11 {
+
+    @Override
+    public List<Event> visitInit(Init e) {
```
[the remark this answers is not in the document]
tonghaining: Removed.

OpenCL needs to tag SC to Init but the memory allocation takes place after this compilation so this is not useful.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,253 @@
...
+        Event init = EventFactory.newInit(e.getBase(), e.getOffset());
+        init.removeTags(init.getTags());
+        init.addTags(toOpenCLTags(e.getTags()));
+        return eventSequence(init);
+    }
+
+    @Override
+    public List<Event> visitLoad(Load e) {
+        Event load = EventFactory.newLoad(e.getResultRegister(), e.getAddress());
+        load.removeTags(load.getTags());
```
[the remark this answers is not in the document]
tonghaining: Yes the SPV tags need to be removed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    @Override
+    public List<Event> visitSpirvLoad(SpirvLoad e) {
+        String mo = moToOpenCLTag(Tag.Spirv.getMoTag(e.getTags()));
+        Load load = newLoadWithMo(e.getResultRegister(), e.getAddress(), mo);
+        load.setFunction(e.getFunction());
+        load.addTags(Tag.C11.ATOMIC);
+        load.addTags(toOpenCLTags(e.getTags()));
+        replaceAcqRelTag(load, Tag.C11.MO_ACQUIRE);
```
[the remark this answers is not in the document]
tonghaining: Dropped

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,253 @@
...
+        return eventSequence(store);
+    }
+
+    @Override
+    public List<Event> visitSpirvXchg(SpirvXchg e) {
+        String mo = moToOpenCLTag(Tag.Spirv.getMoTag(e.getTags()));
+        String scope = toOpenCLTag(Tag.Spirv.getScopeTag(e.getTags()));
+        AtomicXchg rmw = Atomic.newExchange(e.getResultRegister(), e.getAddress(),
+                e.getValue(), mo);
+        rmw.addTags(scope);
```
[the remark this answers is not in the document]
tonghaining: Yes they seem to be redundant, updated.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
tonghaining: removed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ProcessingManager.java
```diff
@@ -89,6 +89,7 @@ private ProcessingManager(Configuration config) throws InvalidConfigurationExcep
                 GEPToAddition.newInstance(),
                 NaiveDevirtualisation.newInstance(),
                 Inlining.fromConfig(config),
+                ThreadCreation.fromConfig(config),
```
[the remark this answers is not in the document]
tonghaining: I see, thanks!
I have changed it back to their original order. I followed LLVM in ThreadCreation and replaced AbortIf that were added from loop unrolling.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
natgavrilenko: I would avoid "prefix magic", unless really necessary. As far as I see, these parameterValues are used only for input values in entryPointer functions. Can we move them VisitorOpsFunction and append local events after an entry pointer has been created?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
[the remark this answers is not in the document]
natgavrilenko: but "load" is a new event that just created with EventFactory.newLoad(). I don't think it has any spirv tags

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
```diff
@@ -0,0 +1,165 @@
...
+                store,
+                casEnd
+        ));
+    }
+
+    @Override
+    public List<Event> visitSpirvRmwExtremum(SpirvRmwExtremum e) {
+        String mo = moToOpenCLTag(Tag.Spirv.getMoTag(e.getTags()));
+        String scope = Tag.Spirv.toOpenCLTag(Tag.Spirv.getScopeTag(e.getTags()));
+        OpenCLRMWExtremum rmw = Atomic.newRMWExtremum(e.getResultRegister(), e.getAddress(),
```
natgavrilenko, line 122: Why do we need to pass scope as a separate parameter? Isn't it just a tag that should be converted from spirv to opencl?

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
```diff
@@ -0,0 +1,4 @@
+package com.dat3m.dartagnan.program;public interface ThreadGrid {
```
natgavrilenko, line 1: Can we just use SpirvThreadGrid here directly? I don't think it has anything spirv-specific and can be considered a "generic" gpu thread grid.

dartagnan/src/test/java/com/dat3m/dartagnan/spirv/benchmarks/ClangSpirvAssertionsTest.java
```diff
@@ -0,0 +1,120 @@
+package com.dat3m.dartagnan.spirv.benchmarks;
```
natgavrilenko, line 1: How about using the same hierarchy for test classes as for the test resources, i.e. spirv/opencl/benchmarks/... and spirv/vulkan/benchmarks/..?

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunctionTest.java
```diff
@@ -482,9 +480,12 @@ public void testCreatingMultipleFunctions() {
         assertTrue(visitor.forwardCalls.isEmpty());
     }
 
+    // TODO: Entry point tests (Resolving parameters for OpenCL)
```
natgavrilenko, line 483: Could you remove this TODO? I guess we can survive without entry point tests

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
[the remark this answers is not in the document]
tonghaining: Ah right.. dropped.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
tonghaining: Parameter removed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
tonghaining: Updated

dartagnan/src/test/java/com/dat3m/dartagnan/spirv/benchmarks/ClangSpirvAssertionsTest.java
tonghaining: relocated

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunctionTest.java
tonghaining: Test added:D

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
natgavrilenko: I agree with Hernan that we should treat control barriers in the same way for all models, unless there is a really good reason to do otherwise. But I suggest to merge this commit first and change barriers after that in a separate commit.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
```diff
@@ -223,12 +242,12 @@ public List<Event> visitLlvmFence(LlvmFence e) {
         ));
     }
 
-    private List<Event> tagList(List<Event> in) {
+    public List<Event> tagList(List<Event> in) {
```
natgavrilenko, line 245: protected?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
```diff
@@ -223,12 +242,12 @@ public List<Event> visitLlvmFence(LlvmFence e) {
...
         in.forEach(e -> tagEvent(null, e));
         return in;
     }
 
-    private List<Event> tagList(Event originalEvent, List<Event> in) {
+    public List<Event> tagList(Event originalEvent, List<Event> in) {
```
natgavrilenko, line 250: protected?

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
```diff
@@ -35,7 +34,7 @@ public int qfSize() {
         return sg * wg * qf;
     }
 
-    public int dvSize() {
+    public int threadPoolSize() {
```
natgavrilenko, line 37: Any reason to rename this method? I think it was consistent sbSize/wgSize/qfSize/dvSize

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -385,4 +395,76 @@ private List<Event> newAcquireLoad(Register resultRegister, Expression address)
...
+        program.getFunctionByName(program.getEntryPoint()).ifPresent(entryFunction -> {
+            for (int tid = 0; tid < grid.threadPoolSize(); tid++) {
+                final Thread thread = createSPVThreadFromFunction(entryFunction, tid, grid, transformers);
+                program.addThread(thread);
+            }
+            // Remove unused memory objects of the entry function
+            for (ExprTransformer transformer : transformers) {
+                if (transformer instanceof MemoryTransformer memoryTransformer) {
+                    Memory memory = entryFunction.getProgram().getMemory();
+                    for (MemoryObject memoryObject : memoryTransformer.getThreadLocalMemoryObjects()) {
```
natgavrilenko, line 413: Not a big deal, but to me it looks a bit weird that we call here a method getThreadLocalMemoryObject from spirv.parser here.  How about moving MemoryTransformer under program/processinfg/transofrmers?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
tonghaining: protected

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
tonghaining: updated.

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
tonghaining: Changed back.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
tonghaining: Moved to program/processing/transformers.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunction.java
```diff
@@ -87,15 +92,71 @@ public Void visitOpFunctionParameter(SpirvParser.OpFunctionParameterContext ctx)
...
+            Type elementType = HelperTypes.getMemberType(id, aType, List.of(0));
+            return 1 + getFirstElementDepth(id, elementType);
+        } else if (type instanceof AggregateType agType) {
+            Type elementType = HelperTypes.getMemberType(id, agType, List.of(0));
+            return 1 + getFirstElementDepth(id, elementType);
+        }
+        return 0;
+    }
+
+    private void createExternalVariable(String id, Type type) {
```
hernanponcedeleon, line 115: I cannot really follow what is going on with `createExternalVariable`, `addExternalValue`, `createPointedExternalValue`, `addExternalParameterLocalEvent` and `addParameterRegisters`. Why do we need all this?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
```diff
@@ -191,8 +210,9 @@ public String getPointerStorageClass(String id) {
 
     public Register addRegister(String id, String typeId) {
         Type type = getType(typeId);
-        if (type instanceof ScopedPointerType) {
-            throw new ParsingException("Register cannot be a pointer");
+        // TODO: Remove this check when tuple registers are supported
```
hernanponcedeleon, line 213: Tuple registers are supported, it is just that we do not create single memory events for them.

hernanponcedeleon (commented):
First pass up to `VisitorExtensionOpenClStd `

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -385,4 +395,76 @@ private List<Event> newAcquireLoad(Register resultRegister, Expression address)
...
+    // =============================================================================================
+    private void createSPVThreads(Program program) {
+        ThreadGrid grid = program.getGrid();
+        List<ExprTransformer> transformers = program.getTransformers();
+        program.getFunctionByName(program.getEntryPoint()).ifPresent(entryFunction -> {
+            for (int tid = 0; tid < grid.dvSize(); tid++) {
+                final Thread thread = createSPVThreadFromFunction(entryFunction, tid, grid, transformers);
+                program.addThread(thread);
+            }
+            // Remove unused memory objects of the entry function
```
hernanponcedeleon, line 409: By "unused" you mean that a given object `X` was repladed by e.g., `WX` and from now one the program will use `WX`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -7,11 +7,11 @@
 import com.dat3m.dartagnan.expression.ExpressionVisitor;
 import com.dat3m.dartagnan.expression.integers.IntLiteral;
 import com.dat3m.dartagnan.expression.processing.ExprTransformer;
+import com.dat3m.dartagnan.expression.type.FunctionType;
 import com.dat3m.dartagnan.expression.type.IntegerType;
 import com.dat3m.dartagnan.expression.type.TypeFactory;
-import com.dat3m.dartagnan.program.Function;
-import com.dat3m.dartagnan.program.Program;
-import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.processing.transofrmers.MemoryTransformer;
```
hernanponcedeleon, line 13: typo in `transformer`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
hernanponcedeleon: `VisitorSpirvVulkan` is translated to Vulkan event sequences and those also need to create dummy registers for RMW. Why is this not a problem in that case?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
```diff
@@ -198,7 +198,13 @@ private void unrollLoop(CondJump loopBackJump, int bound) {
                 final List<Event> copies = copyPath(loopBegin, loopBackJump, copyCtx, loopId);
 
                 // Insert copy of the loop
-                loopBegin.getPredecessor().insertAfter(copies);
+                if (loopBegin.getPredecessor() == null) {
+                    Event newEntry = copies.get(0);
+                    loopBegin.getFunction().updateEntry(newEntry);
+                    newEntry.insertAfter(copies.subList(1, copies.size()));
+                } else {
```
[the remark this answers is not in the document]
ThomasHaas: Due to the lack of `insertBefore` I think. I think its best to merge #803  and use the new functions rather than working around the limitations.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
tonghaining: We are also doing it for vulkan https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvVulkan.java#L83

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -0,0 +1,66 @@
...
+        Type type = builder.getType(ctx.idResultType().getText());
+        List<Integer> indexes = ctx.indexesLiteralInteger().stream()
+                .map(SpirvParser.IndexesLiteralIntegerContext::getText)
+                .map(Integer::parseInt)
+                .toList();
+        Expression element = compositeExpression;
+        for (Integer index : indexes) {
+            if (!(element instanceof ConstructExpr)) {
+                throw new ParsingException(String.format("Element is not a ConstructExpr at index: %d for: %s", index, id));
+            }
```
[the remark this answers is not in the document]
tonghaining: Yes, line 29 removed.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunction.java
```diff
@@ -24,14 +28,16 @@
 public class VisitorOpsFunction extends SpirvBaseVisitor<Void> {
 
     private static final TypeFactory types = TypeFactory.getInstance();
+    private static final int DEFAULT_SIZE = 10;
+    final Map<String, Function> forwardFunctions = new HashMap<>();
+    final Map<String, Set<FunctionCall>> forwardCalls = new HashMap<>();
```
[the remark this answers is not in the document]
tonghaining: Updated to private.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunction.java
```diff
@@ -87,15 +92,71 @@ public Void visitOpFunctionParameter(SpirvParser.OpFunctionParameterContext ctx)
...
     }
 
+    private int getFirstElementDepth(String id, Type type) {
+        if (type instanceof ArrayType aType) {
+            Type elementType = HelperTypes.getMemberType(id, aType, List.of(0));
+            return 1 + getFirstElementDepth(id, elementType);
+        } else if (type instanceof AggregateType agType) {
+            Type elementType = HelperTypes.getMemberType(id, agType, List.of(0));
+            return 1 + getFirstElementDepth(id, elementType);
+        }
```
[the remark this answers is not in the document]
tonghaining: Merged

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsFunction.java
tonghaining: They are used to create external variables which entry function pointers may point to (which are not declared by OpVariable as our earlier tests).
I update their name as:

`createExternalVariable` -> `createParameterVariable`
`addExternalValue` -> `addParameterVariable`
`createPointedExternalValue` -> `createExternalVariable`
`addExternalParameterLocalEvent` -> `addEntryParameterLocalEvent`

Would it be better as this?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/builders/ProgramBuilder.java
tonghaining: Comment updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/LoopUnrolling.java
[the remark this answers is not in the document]
tonghaining: Yeah it causes issue when loop from the first event.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
tonghaining: Given object `X` in the entry function, it will be replaceds by `X0`  in thread0, `X1` in thread1.., and the entry function objects will not be used anymore.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
tonghaining: Thanks! updated

natgavrilenko (approved):

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorSpirvOpenCL.java
hernanponcedeleon: I was wrong about this. As discussed, let get rid of the two barriers in a different PR

Issue #793, opened 2025-01-16
Support for out-of-bound memory accesses


natgavrilenko:
We should revise the way we handle out-of-bound memory accesses.

Ideally, we should be able to find candidate executions with out-of-bound accesses. Or at least we should treat such accesses consistently (e.g., may alias with any other memory access) and document the limitations in the README. Currently, verification results vary a lot depending on the selected solver and alias analysis method.

There is a summary of verification results for variants of a simple test. The malloc variants are added for future testing (alias analysis for malloc is currently work in progress).

<img width="448" alt="Image" src="https://github.com/user-attachments/assets/1a8db882-362d-4b7e-bb70-3f19500b9ec8" />

[tests.zip](https://github.com/user-attachments/files/18440230/tests.zip)

ThomasHaas:
The issue is a mismatch between the provenance model assumed by the alias analysis vs. the provenance model we encode. 
The different alias analyses only expose parts of this issue but themselves do not really cause it.
Btw. the older alias analyses (especially `alias-insensitive`) should be treated as deprecated I think.

The summary is this:
- Our alias analysis assumes pointers adhere to a provenance model: each pointer has provenance data attached to it which designates what region of memory it can access. Importantly, doing pointer computation on such a pointer does not change it's provenance.
- Our assumed provenance model is that a pointer has shape `(base, offset)` (C-like provenance) where base is a memory object that functions as provenance data. Accesses outside of the `base` object are UB (i.e. `offset < 0` or `offset > sizeof(base))`
- Our encoding, however, does not encode the provenance model (each pointer is a plain integer address). This is why the encoding is incapable of detecting OOB/UB.
- The solution(*) is to implement a proper `PointerType` and appropriate pointer expressions. The encoding then has to encode pointers as tuples `(base, offset)`. The only major difficulty here is that `int2ptr` casts become hard to implement (and some annoying issues related to modelling `pthread`).

I think it is reasonable to assume that all code we support adheres to (at least) a `(base, offset)` provenance model, i.e., where pointers are restricted to within objects. Though we could provide configurable provenance models in principle.

I do have a BA student who is working on this and he has implemented pointer types all the way to the encoding stage.
That being said, my experience with student code is that it does not satisfy standard requirements, so I don't mind if a more experienced person wants to give it a try.

[...]

PR #794, opened 2025-01-21
Support alloc-free dev

CapZTr:


dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -34,6 +33,24 @@ public interface AliasAnalysis {
...
+
+    boolean mayAlias(Alloc a, MemoryCoreEvent e);
+
+    boolean mustAlias(Alloc a, MemFree f);
+
+    boolean mayAlias(Alloc a, MemFree f);
+
+    boolean mustAlias(MemFree a, MemFree b);
+
+    boolean mayAlias(MemFree a, MemFree b);
```
ThomasHaas, line 46: I know this is a draft, but you don't really want to go with so many overloads, do you?

ThomasHaas (commented):
On a more general note about handling `Free`/`Alloc`: I'm not convinced that changing aliasing is the right concept for Alloc/Free. I think it is a misleading road.
What you really care about is not the individual addresses that the Alloc/Free targets, but only the memory objects as a whole. Similarly, you care about the objects a memory access can address if you want to talk about bugs like use-after-free.
So a `canAccessSameObject(x, y)` method would be more appropriate and would also elevate the issue of requiring allocations of known size (the size does not matter!).
If you then introduce a corresponding `sameObj` relation in `.cat`, you should be able to treat `Free` and `Alloc` more-or-less as single-address events (i.e., simple memory events).
A use-after-free (or racy free) would then simply be `~empty ([Free];sameObj;[M] \ hb)`.

natgavrilenko:
> On a more general note about handling `Free`/`Alloc`: I'm not convinced that changing aliasing is the right concept for Alloc/Free. I think it is a misleading road. What you really care about is not the individual addresses that the Alloc/Free targets, but only the memory objects as a whole. Similarly, you care about the objects a memory access can address if you want to talk about bugs like use-after-free. So a `canAccessSameObject(x, y)` method would be more appropriate and would also elevate the issue of requiring allocations of known size (the size does not matter!). If you then introduce a corresponding `sameObj` relation in `.cat`, you should be able to treat `Free` and `Alloc` more-or-less as single-address events (i.e., simple memory events). A use-after-free (or racy free) would then simply be `~empty ([Free];sameObj;[M] \ hb)`.

I think you didn't really check what the code is doing. We need both, individual addresses (i.e. the pointer returned by alloc) and the full allocated memory region to compare with addresses of memory accesses.

ThomasHaas:
Why do you need the individual addresses in the full region if you instead had a `accessesSameObject(x, y)` method? This would hold true for an `Alloc` that is considered accessing the memory object it allocates and a memory access to anywhere inside that object.
Btw. the most compact representation of the memory region of a memory object is just the memory object itself, rather than the individual addresses inside. This makes the approach viable even for objects of unknown/unbounded size.

natgavrilenko:
> Why do you need the individual addresses in the full region if you instead had a `accessesSameObject(x, y)` method? This would hold true for an `Alloc` that is considered accessing the memory object it allocates and a memory access to anywhere inside that object. Btw. the most compact representation of the memory region of a memory object is just the memory object itself, rather than the individual addresses inside. This makes the approach viable even for objects of unknown/unbounded size.

please read the code first

ThomasHaas:
Huh, I read the code?! I can see that all the checks about bounded size allocations and `mayAlias(Alloc x , MemoryCoreEvent y)` checking that `y` accesses something inside `x`. What else should I read in this PR?

hernanponcedeleon:
> I'm not convinced that changing aliasing is the right concept for Alloc/Free. I think it is a misleading road ... So a `canAccessSameObject(x, y)` method would be more appropriate 

Isn't the later some kind of alias (maybe not as fine grained as per address, but at least per object)?

> and would also elevate the issue of requiring allocations of known size (the size does not matter!). 

This is assuming no OOB, right?

> If you then introduce a corresponding `sameObj` relation in `.cat`, you should be able to treat `Free` and `Alloc` more-or-less as single-address events (i.e., simple memory events). A use-after-free (or racy free) would then simply be `~empty ([Free];sameObj;[M] \ hb)`.

Isn't this kind of what the two new relations `allocptr` and `allocmem` are doing? Those relation are not visible in the diff of this PR (which target the initial draft from Natalia), but maybe this is the source of the misunderstanding.

ThomasHaas:
> > I'm not convinced that changing aliasing is the right concept for Alloc/Free. I think it is a misleading road ... So a `canAccessSameObject(x, y)` method would be more appropriate
> 
> Isn't the later some kind of alias (maybe not as fine grained as per address, but at least per object)?

Yes, it is a kind of aliasing. But I think it is worth to differentiate between the concepts of "same address", "overlapping address" and "same object". Covering them all under the term "aliasing" is bound to cause confusion.
Importantly, `sameObject` is actually easier to compute/reason about because we only ever have finitely many objects but possibly unboundedly many addresses. 
For example, our newest alias analysis was not updated in this PR which I think is only partly because of its difficulty, but also partly because it does not compute explicit addresses like the other alias analyses do. However, the analysis does have information about which memory objects a memory event may access, and this is sufficient to implement the desired feature.
That being said, the implementation of `canAccessSameObject` for the other alias analyses would pretty much coincide with what this PR does.

> 
> > and would also elevate the issue of requiring allocations of known size (the size does not matter!).
> 
> This is assuming no OOB, right?

All alias analyses assume no OOB anyways. So nothing changes there.

> 
> > If you then introduce a corresponding `sameObj` relation in `.cat`, you should be able to treat `Free` and `Alloc` more-or-less as single-address events (i.e., simple memory events). A use-after-free (or racy free) would then simply be `~empty ([Free];sameObj;[M] \ hb)`.
> 
> Isn't this kind of what the two new relations `allocptr` and `allocmem` are doing? Those relation are not visible in the diff of this PR (which target the initial draft from Natalia), but maybe this is the source of the misunderstanding.

Yes, I know about those relations. I'm saying that you can likely replace them by a general `sameObj` relation. At least concept-wise, this is worth considering even if the multiple new base relations are preferred for some other reason. Either way, the underlying idea of those relations is still based more on object-based reasoning rather than address-based reasoning.
Lastly, encoding `sameObj` can be easier than `sameAddress`, especially if we eventually use a provenance-based pointer model like mentioned in #793 .

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -34,6 +33,24 @@ public interface AliasAnalysis {
...
+
+    default void checkHeapAlloc(Alloc alloc) {
+        if (!alloc.isHeapAllocation()) {
+            throw new IllegalArgumentException("Alias analysis can be performed only for heap allocation");
+        }
+    }
```
ThomasHaas, line 52: Why do you need specific checks for heap allocations? Alias analyses should work just fine with stack mempry.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/FieldSensitiveAndersen.java
```diff
@@ -77,17 +81,74 @@ public boolean mayAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+        return getMaxAddressSet(e).stream().anyMatch(
+                l -> l.base.equals(a.getAllocatedObject()) && l.offset < getAllocatedSize(a)
+        );
+    }
+
+    @Override
+    public boolean mustAlias(Alloc a, MemoryCoreEvent e) {
+        checkHeapAlloc(a);
+        Set<Location> locs = getMaxAddressSet(e);
+        return locs.isEmpty() ? false : locs.stream().allMatch(
```
natgavrilenko, line 100: Can be simplified as "!locs.isEmpty() && ....".

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -77,15 +78,68 @@ public boolean mustAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
         return getMaxAddressSet(x).size() == 1 && getMaxAddressSet(x).containsAll(getMaxAddressSet(y));
     }
 
+    @Override
+    public boolean mayAlias(Alloc a, MemoryCoreEvent e) {
```
natgavrilenko, line 82: We can combine them in one function to keep the interface clear, something like this

public boolean mayAlias(Event x, Event y) {
        if (x instanceof Alloc alloc) {
            if (y instanceof MemFree free) {
                // Alloc-Free
            }
            // Alloc-MemEvent
        }
        // MemEvent-MemEvent or Free-Free
    }

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -77,15 +78,68 @@ public boolean mustAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+        return getMaxAddressSet(e).stream().allMatch(
+                l -> l.base.equals(a.getAllocatedObject()) && l.offset < getAllocatedSize(a)
+        );
+    }
+
+    @Override
+    public boolean mayAlias(Alloc a, MemFree f) {
+        checkHeapAlloc(a);
+        return getFreedAddresses(f).stream().anyMatch(
+                l -> l.base.equals(a.getAllocatedObject()) && l.offset < getAllocatedSize(a)
```
natgavrilenko, line 101: Contrary to MemoryEvents, which should match any offset in the allocated memory, Free should match address at offset 0 (for must it uses correct offset).

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -77,15 +78,68 @@ public boolean mustAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+        );
+    }
+
+    @Override
+    public boolean mustAlias(Alloc a, MemFree f) {
+        checkHeapAlloc(a);
+        Set<Location> freedLocs = getFreedAddresses(f);
+        return freedLocs.size() == 1
+                && freedLocs.iterator().next().equals(new Location(a.getAllocatedObject(), 0));
```
natgavrilenko, line 110: Would be nice to avoid creating a new Location object each time. Maybe create it during processLocs and put into eventAddressSpaceMap?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
natgavrilenko: I think Thomas is right, and we can keep it for stack memory too.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
ThomasHaas: I still suggest to use two kinds of methods to make it more explicit what you querying. If you don't, you will get a lot of hidden pitfalls like breaking transitivity of aliasing (in the absence of control-flow): `must-alias(mem, free)` + `must-alias(mem, alloc)` + `no-alias(alloc, free)` will be a totally valid combination.
This is bound to cause confusion and errors in the future.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -795,8 +805,13 @@ public MutableKnowledge visitAllocMem(AllocMem aloc) {
             MutableEventGraph must = new MapEventGraph();
             for (Alloc e1 : program.getThreadEvents(Alloc.class)) {
                 if (e1.isHeapAllocation()) {
-                    for (Event e2 : program.getThreadEvents(MemoryEvent.class)) {
-                        may.add(e1, e2);
+                    for (MemoryCoreEvent e2 : program.getThreadEvents(MemoryCoreEvent.class)) {
```
natgavrilenko, line 808: Could you create a list of MemoryCoreEvent before the first loop, so that we don't need to call getThreadEvents each time?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -795,8 +805,13 @@ public MutableKnowledge visitAllocMem(AllocMem aloc) {
             MutableEventGraph must = new MapEventGraph();
             for (Alloc e1 : program.getThreadEvents(Alloc.class)) {
                 if (e1.isHeapAllocation()) {
```
natgavrilenko, line 807: I think it should be easy to extend this approach to stack events, but I guess you want to get the heap part ready first.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -236,25 +334,24 @@ private void processResults(Local e) {
...
 
-    private void processResults(MemoryCoreEvent e) {
-        Expression address = e.getAddress();
+    private Set<Location> getAddressSpace(Expression addrExpr) {
         Set<Location> addresses;
-        if (address instanceof Register) {
-            Set<Location> target = targets.get(address);
-            addresses = target != null ? target : getAddresses(address);
+        if (addrExpr instanceof Register register) {
+            Set<Location> target = targets.get(register);
```
natgavrilenko, line 340: Can be the same addrExpr (without cast)

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -69,69 +69,98 @@ public static AndersenAliasAnalysis fromConfig(Program program, Configuration co
...
-        return getMaxAddressSet(x).size() == 1 && getMaxAddressSet(x).containsAll(getMaxAddressSet(y));
-    }
-
-    @Override
-    public boolean mayAlias(Alloc a, MemoryCoreEvent e) {
-        checkHeapAlloc(a);
-        return getMaxAddressSet(e).stream().anyMatch(
-                l -> l.base.equals(a.getAllocatedObject()) && l.offset < getAllocatedSize(a)
-        );
+    public boolean mayAlias(Event a, Event b) {
```
natgavrilenko, line 72: I think this typecheck is a bit on overkill. We have two cases: 
1) A pair of alloc and memory event -> we need to check if memory event may/must access any address from the allocated region.
2) Everything else, including alloc-free and free-free pairs -> we need to check the "normal" may/must alias for a single pointer.

How about something like this?

@Override
    public boolean mayAlias(Event a, Event b) {
        if (a instanceof Alloc alloc && b instanceof MemoryCoreEvent mem) {
            return mayAccessAllocatedBy(alloc, mem);
        }
        if (b instanceof Alloc alloc && a instanceof MemoryCoreEvent mem) {
            // This case shouldn't be called because alloc->mem relation always starts at alloc, we can keep both to be on the safe side.
            return mayAccessAllocatedBy(alloc, mem);
        }
        return mayAccessSameAddress(a, b);
    }

And the same for must sets and the other analysis classes.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -69,69 +69,98 @@ public static AndersenAliasAnalysis fromConfig(Program program, Configuration co
...
-        return freedLocs.size() == 1
-                && freedLocs.iterator().next().equals(new Location(a.getAllocatedObject(), 0));
+    private boolean mustAccessSameAddress(Event x, Event y) {
+        return getMaxAddressSet(x).size() == 1 && getMaxAddressSet(x).containsAll(getMaxAddressSet(y));
     }
 
-    @Override
-    public boolean mayAlias(MemFree a, MemFree b) {
-        return !Sets.intersection(getFreedAddresses(a), getFreedAddresses(b)).isEmpty();
+    private boolean mayAccessAllocatedBy(Alloc a, Event e) {
```
natgavrilenko, line 127: I would keep only the first part and do alloc-free pair in may/mustAccessSameAddress.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -155,14 +184,11 @@ private void run(Program program) {
...
+            eventAddressSpaceMap.put(f, ImmutableSet.copyOf(getAddressSpace(f.getAddress())));
         }
     }
 
     private void processAllocs(Alloc a) {
-        if (!a.isHeapAllocation()) {
-            return;
-        }
         Register r = a.getResultRegister();
         if (a.getAllocationSize() instanceof IntLiteral i) {
```
natgavrilenko, line 193: We can add alloc to eventAddressSpaceMap and then check alloc-free pair via getMaxAddressSet:

eventAddressSpaceMap.put(a, Set.of(new Location(a.getAllocatedObject(), 0)));

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
```diff
@@ -31,7 +29,54 @@ public static EqualityAliasAnalysis fromConfig(Program program, Configuration co
...
+            if (b instanceof MemFree fb) {
+                return true;
+            } else if (b instanceof Alloc ab) {
+                return true;
+            }
+        }
+        throw new IllegalArgumentException("Unsupported event types for EqualityAliasAnalysis");
+    }
+
+    private boolean mustAccessSameAddress(MemoryCoreEvent a, MemoryCoreEvent b) {
```
natgavrilenko, line 79: I think we don't need all these complex reasoning in mayAlias/mustAlias methods. In the old comment, I meant that we can implement mustAccessAllocatedBy(Alloc a, MemoryCoreEvent e) following the same logic as for mustAccessSameAddress, i.e. must access is true if 1) the memory event uses the register of the alloc event and 2) the register value has not been overwritten.

We can also try reasoning about register + index accesses, but it will be a bit more complex, because we also need to consider sizes of the member elements and offsets.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -169,14 +217,96 @@ public boolean mayAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
-    @Override
-    public boolean mustAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
+    private boolean mustAccessSameAddress(Event x, Event y) {
         final DerivedVariable vx = addressVariables.get(x);
         final DerivedVariable vy = addressVariables.get(y);
         return vx != null && vy != null && vx.base == vy.base && vx.modifier.offset == vy.modifier.offset &&
                 isConstant(vx.modifier) && isConstant(vy.modifier);
     }
 
+    private boolean mayAccessAllocatedBy(Alloc a, Event e) {
```
natgavrilenko, line 227: I will carefully check it and add comments a bit later.

xeren:
You may add the following methods to `AliasAnalysis` and use them instead:
```java
boolean mayObjectAlias(Event a, Event b);
boolean mustObjectAlias(Event a, Event b);
```
Implementing it in `InclusionBasedPointerAnalysis`:
```java
@Override
public boolean mayObjectAlias(Event a, Event b) {
    DerivedVariable addressA = addressVariables.get(a);
    DerivedVariable addressB = addressVariables.get(b);
    return addressA == null || addressB == null ||
            !Collections.disjoint(getAccessibleObjects(addressA), getAccessibleObjects(addressB));
}
@Override
public boolean mustObjectAlias(Event a, Event b) {
    DerivedVariable addressA = addressVariables.get(a);
    DerivedVariable addressB = addressVariables.get(b);
    if (addressA == null | addressB == null) {
        return false;
    }
    if (addressA.base == addressB.base) {
        return true;
    }
    Set<MemoryObject> objectsA = getAccessibleObjects(addressA);
    return objectsA.size() == 1 && objectsA.equals(getAccessibleObjects(addressB));
}
private Set<MemoryObject> getAccessibleObjects(DerivedVariable address) {
    var objects = new HashSet<MemoryObject>();
    objects.add(address.base.object);
    for (IncludeEdge edge : address.base.includes) {
        objects.add(edge.source.object);
    }
    objects.remove(null);
    return objects;
}
private void run(Program program, AliasAnalysis.Config configuration) {
    ...
    for (Alloc alloc : program.getThreadEvents(Alloc.class)) {
        addressVariables.put(alloc, derive(objectVariables.get(alloc.getAllocatedObject())));
    }
    ...
}
```
Implementing it in `EqualityAliasAnalysis`:
```java
@Override
public boolean mayObjectAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
    return true;
}
@Override
public boolean mustObjectAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
    return mustAlias(a, b);
}
```
Implementation for `CombinedAliasAnalysis`:
```java
@Override
public boolean mayObjectAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
    return a1.mayObjectAlias(a, b) && a2.mayObjectAlias(a, b);
}
@Override
public boolean mustObjectAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
    return a1.mustObjectAlias(a, b) || a2.mustObjectAlias(a, b);
}
```
Implementing it in `AndersenAliasAnalysis` and `FieldSensitiveAndersen`:
```java
@Override
public boolean mayObjectAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
    return !Collections.disjoint(getAccessibleObjects(a), getAccessibleObjects(b));
}
@Override
public boolean mustObjectAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
    Set<MemoryObject> objects = getAccessibleObjects(a);
    return objects.size() == 1 && objects.containsAll(getAccessibleObjects(b));
}
private Set<MemoryObject> getAccessibleObjects(MemoryCoreEvent event) {
    var objects = new HashSet<MemoryObject>();
    for (Location location : getMaxAddressSet(event)) {
        objects.add(location.base);
    }
    return objects;
}
```

CapZTr:
Are you suggesting that we only care about the `base` object and don't consider `offset` at all? Won't this result in a loss of precision for analysis when the size of object is integer?

ThomasHaas:
Not really. You would only get precision in the presence of out-of-bounds accesses (UB), which we don't handle correctly either way.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -169,14 +217,96 @@ public boolean mayAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+                return v.modifier.offset < size && v.modifier.offset >= 0;
+            }
+            final List<IncludeEdge> includes = new ArrayList<>(v.base.includes);
+            includes.add(new IncludeEdge(v.base, TRIVIAL_MODIFIER));
+            for (final IncludeEdge i : includes) {
+                if (i.source.object != base) {
+                    continue;
+                }
+                final Modifier m = compose(i.modifier, v.modifier);
+                final boolean may = isConstant(m) ? m.offset < size && m.offset >= 0
```
natgavrilenko, line 248: I would use Rene's suggestion and compare only base objects when checking the whole memory region.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -170,18 +172,47 @@ public boolean mayAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+        final Set<MemoryObject> objsA = getAccessibleObjects(va);
+        return objsA.size() == 1 && objsA.equals(getAccessibleObjects(vb));
+    }
+
     @Override
     public Graphviz getGraphVisualization() {
         return graphviz;
     }
 
+    private Set<MemoryObject> getAccessibleObjects(DerivedVariable address) {
```
natgavrilenko, line 208: Should we precompute the sets of accessible objects instead of generating a new set each time may/mustObjectAlias is called?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -170,18 +172,47 @@ public boolean mayAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+        final Set<MemoryObject> objs = new HashSet<>();
+        objs.add(address.base.object);
+        address.base.includes.stream().forEach(i -> objs.add(i.source.object));
```
natgavrilenko, line 211: "forEach" can be used without stream()

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
```diff
@@ -29,45 +29,60 @@ public static EqualityAliasAnalysis fromConfig(Program program, Configuration co
     }
 
     @Override
-    public boolean mustAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
```
natgavrilenko, line 32: I think this can be implemented in a similar way for Alloc and MemFree, you just need to getAddress in the right way, depending on the type of event, for example

private Expression getAddress(Event e) {
        if (e instanceof MemoryCoreEvent memEvent) {
            return memEvent.getAddress();
        }
        if (e instanceof MemFree memFree) {
            return memFree.getAddress();
        }
        if (e instanceof Alloc alloc) {
            return alloc.getAllocatedObject();
        }
        throw new UnsupportedOperationException("Event type has no address: " + e.getClass().getSimpleName());
    }

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -236,25 +271,24 @@ private void processResults(Local e) {
         }
     }
 
-    private void processResults(MemoryCoreEvent e) {
```
natgavrilenko, line 239: NIT: to keep it similar with processResults(Local), you can pass the whole event here and get address depending on the event type:

private Expression getAddress(Event e) {
        if (e instanceof MemoryCoreEvent memEvent) {
            return memEvent.getAddress();
        }
        if (e instanceof MemFree memFree) {
            return memFree.getAddress();
        }
        if (e instanceof Alloc alloc) {
            return alloc.getAllocatedObject();
        }
        throw new UnsupportedOperationException("Event type has no address: " + e.getClass().getSimpleName());
    }

PR #796, opened 2025-01-27
Use alignment information in VisitorLlvm.visitAllocaInst()

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -539,7 +539,11 @@ public Expression visitAllocaInst(AllocaInstContext ctx) {
             final Type sizeType = parseType(ctx.typeValue().firstClassType());
             sizeExpression = checkExpression(sizeType, ctx.typeValue().value());
         }
-        //final int alignment = parseAlignment(ctx.align());
+        if(ctx.align() != null) {
```
ThomasHaas, line 542: isn't alignment guaranteed to exist?
Either way, you should write the code like this:
```Java
final Event alloc;
if (ctx.align() != null) {
     final Expression alignment = ...;
     alloc = ...
} else {
     alloc = ...
}
block.events.add(alloc);
return register;
```
No need to repeat the common parts.

hernanponcedeleon: > isn't alignment guaranteed to exist? 

At least in our grammar, [the alignment is optional](https://github.com/hernanponcedeleon/Dat3M/blob/4fc81b1d00e2a4cf1c95c6d2bb34a36b89ddf37c/dartagnan/src/main/antlr4/LLVMIR.g4#L590)

ThomasHaas: Ok, I guess this is due to backward compatibility with older LLVM versions.

ThomasHaas (approved):
LGTM

PR #798, opened 2025-02-11
Support sql program

MajorDaxx:
This is just to have a nice comparison to what I have changed. 
But maybe one day it will also be pulled :D

### DIFF
--- a/cat/sql/default.cat
+++ b/cat/sql/default.cat
@@ -1,5 +1,3 @@
-RA
-
 (*
 Base Relations:
 po -> program order relates two events within the same transaction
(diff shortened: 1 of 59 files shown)


ThomasHaas:
You could have added some context to your PR. I'm the only one here knowing what you do.
Btw. to have this diff, you could have made the PR on your own fork as well :P.

**Context:** @MajorDaxx tries to apply Dartagnan to database programs with database consistency modelled in CAT.
The anarchic semantics works over tables both as memory domain (the memory is a set of tables) as well as data domain (communicated values are tables). This requires the SMT solver to support the theory of tables/bags (currently, only CVC5 can do this).
Furthermore, database consistency replaces `rf` by a visibility relation `vis` and `co` by an arbitration relation `ar`.
The PR tries to implement these changes + a new input format for SQL programs.

hernanponcedeleon:
can this be closed?

PR #802, opened 2025-02-21
Add support for inline asm 

StefanoDalMas:
This PR adds the ability to understand inline asm code in dat3m.


cat/arm.cat
```diff
@@ -0,0 +1,67 @@
...
+let dsb.st = po;[DSB.ST];po
+let dmb = po;[DMB];po
+let dmb.ish = po;[DMB.ISH];po
+let dsb = po;[DSB];po
+let isb = po;[ISB];po
+
+(* Initial value *)
+let ci0 = (ctrl; [ISB]) | detour
+let ii0 = dd | rfi | rdw
+let cc0 = dd | ctrl | addrpo (* po-loc deleted *)
```
ThomasHaas, line 33: Why is `po-loc` removed?

cat/arm.cat
```diff
@@ -0,0 +1,67 @@
...
+
+let ppo = (R*W & ic) | (R*R & ii)
+
+(* ARM *)
+let WW = W * W
+let dmb.st=dmb.st & WW
+let dsb.st=dsb.st & WW
+
+(* Common, all arm barriers are strong *)
+let strong = dmb|dsb|dmb.st|dsb.st|dmb.ish
```
ThomasHaas, line 50: Please add spacing. Also for the remaining definitions below.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/AggregateType.java
```diff
@@ -1,23 +1,29 @@
...
+        this.fields = fields;
     }
 
     public List<TypeOffset> getTypeOffsets() {
         return directFields;
     }
 
+    public List<Type> getFields() {
+        return fields;
+    }
```
ThomasHaas, line 25: There is no need to store the types separately.
Just use `Lists.transform(directFields, x -> x.type())`.

cat/arm.cat
hernanponcedeleon: I copied the cat file [from herd7's repo](https://github.com/herd/herdtools7/blob/master/herd/libdir/arm.cat)

ThomasHaas: I see.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+                // returnRegister is not an aggregate, we just get that type
+                registerType = returnRegister.getType();
+            }
+        } else {
+            // in this case we have an index which is higher than the size of the return Register
+            // this means that, like in visitAsmMetadataEntries, we have to "shift" the value by the size of returnRegister type's size
+            // Same example as visitAsmMetadataEntries:
+            // e.g. returnRegisterType = { i32, i32 } "... $2", "=&r, =&r, r" (i32 r5)
+            // we have registerID 2, we have returnRegisterType size = 2, so we have to perform 2-2 = 0, which is the right index
+            // to get argsRegisters[2-2] = argsRegisters[0] = r5.
```
hernanponcedeleon, line 199:
```suggestion
            // registerID is mapped to a register in args. To get the correct position in args we need to shift the id by the size of the return register
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        return registerType;
+    }
+
+    private String makeRegisterName(int registerID) {
+        return "asm_" + registerID;
+    }
+
+    // this function is the entrypoint of the visitor.
+    // the events that are going to be generated are
+    // 1) inputAssignments -- how we map llvm registers to asm ones
```
hernanponcedeleon, line 211:
```suggestion
    // 1) inputAssignments -- how we map llvm registers in asm to asm ones
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    // 2) asmInstructions -- the events representing the asm instructions
+    // 3) outputAssignments -- how we map the asm registers to the return Register
+    @Override
+    public List<Event> visitAsm(InlineAsmParser.AsmContext ctx) {
+        // Given this format 
+        //      Register = Type call asm sideeffect 'asm code', 'constraints, clobbers' ('args')
+        // the visitor pattern is going to take care of first visiting all the 'asm code'
+        // which is going to populate all of the 'asmInstructions' and creating the asmRegisters.
+        // Then, by reading the constraints, we are going to be able to understand which asmRegister
+        // maps to which llvmRegister.
```
hernanponcedeleon, line 221: The visitor will first visit the asm code (which will create the events and asm registers) and then the constraints. The latter will take care of creating input and output assignments.

I think the text above is enough and the rest of comments in this method can be removed.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        events.addAll(inputAssignments);
+        // this is going to generate all of the events related to the asm code (load, store, cmp, jmp,...)
+        events.addAll(asmInstructions);
+        // this has to be the last thing that we do, and it is going to map the asm registers to the returnRegister
+        // it is put as last so that we are sure that we are reading the latest data that was written to the register
+        events.addAll(outputAssignments);
+        return events;
+    }
+
+    // tells if constraint is a numberic constraint e.g. '3'
```
hernanponcedeleon, line 236:
```suggestion
    // Tells if a constraint is a numeric one, e.g. '3'
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    private boolean isConstraintNumeric(InlineAsmParser.ConstraintContext constraint) {
+        return constraint.overlapInOutRegister() != null;
+    }
+
+    // tells us is the constraint is a memory location '=*m'
```
hernanponcedeleon, line 241:
```suggestion
    // Tells if the constraint is a memory location '=*m'
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    private boolean isConstraintMemoryLocation(InlineAsmParser.ConstraintContext constraint) {
+        return constraint.pointerToMemoryLocation() != null;
+    }
+
+    // tells us if the constraint is an output constraint e.g. '=r' or '=&r'
```
hernanponcedeleon, line 246:
```suggestion
    // Tells if the constraint is an output one, e.g. '=r' or '=&r'
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    private boolean isConstraintOutputConstraint(InlineAsmParser.ConstraintContext constraint) {
+        return constraint.outputOpAssign() != null;
+    }
+
+    // tells us if the constraint is an input constraint e.g. 'Q' or '*Q' or 'r'
```
hernanponcedeleon, line 251:
```suggestion
    // Tells if the constraint is an input one, e.g. 'Q' or '*Q' or 'r' 
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        String labelID = ctx.NumbersInline().getText();
+        Label label = getOrNewLabel(labelID);
+        asmInstructions.add(label);
+        return null;
+    }
+
+    @Override
+    public Object visitExpr(InlineAsmParser.ExprContext ctx) {
+        return ctx.getChild(0).accept(this);
+    }
```
hernanponcedeleon, line 416: Isn't this the default implementation? If so, it can be removed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+
+    // If the register with that ID was already defined, we simply return it
+    // otherwise, we create and return the new register.
+    // each time we call this visitor, it picks up the ID of the register -- e.g. $3 -> ID = 3.
```
hernanponcedeleon, line 420:
```suggestion
    // Each time we call this visitor, it picks up the ID of the register -- e.g. $3 -> ID = 3.
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    // if we created a register which refers to the return Register, we have to add to "pendingRegisters",
```
hernanponcedeleon, line 421:
```suggestion
    // if we created a register which will be mapped to the return Register, we have to add to "pendingRegisters", 
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    @Override
+    public Object visitExpr(InlineAsmParser.ExprContext ctx) {
+        return ctx.getChild(0).accept(this);
+    }
+
+    // If the register with that ID was already defined, we simply return it
+    // otherwise, we create and return the new register.
+    // each time we call this visitor, it picks up the ID of the register -- e.g. $3 -> ID = 3.
+    // if we created a register which refers to the return Register, we have to add to "pendingRegisters", 
+    // as we are going to need it during visitMetadata part
```
hernanponcedeleon, line 422:
```suggestion
    // as we are going to need it while visiting the metadata to create the output assignment
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    @Override
+    public Object visitRegister(InlineAsmParser.RegisterContext ctx) {
+        String registerName = ctx.NumbersInline().getText();
+        int registerID = Integer.parseInt(registerName);
+        // if w.h. the register, return it
```
hernanponcedeleon, line 427: Remove

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        if (asmRegisters.containsKey(registerID)) {
+            return asmRegisters.get(registerID);
+        } else {
+            // otherwise, pick up the correct type and create the new Register
```
hernanponcedeleon, line 431: Remove
```suggestion
            // Pick up the correct type and create the new Register
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+            Type registerType = getLlvmRegisterTypeGivenAsmRegisterID(registerID);
+            Register newRegister = this.llvmFunction.getOrNewRegister(makeRegisterName(registerID), registerType);
+            // if the register is part of the return register, we have to register it for the final assignment for the output
+            // as we are going to do it later when we have enough information in visitAsmMetadata when visiting the constraints.
```
hernanponcedeleon, line 435: This is already explained in the doc of the method. Remove.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+            if (isPartOfReturnRegister(registerID) && isReturnRegisterAggregate()) {
+                this.pendingRegisters.add(newRegister);
+            }
+            asmRegisters.put(registerID, newRegister);
+            return newRegister;
+        }
+    }
+
+    // This visitor generates the last two kind of events:
```
hernanponcedeleon, line 444:
```suggestion
    // This visitor generates two sets of events: 
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+            if (isPartOfReturnRegister(registerID) && isReturnRegisterAggregate()) {
+                this.pendingRegisters.add(newRegister);
+            }
+            asmRegisters.put(registerID, newRegister);
+            return newRegister;
+        }
+    }
+
+    // This visitor generates the last two kind of events: 
+    // 1) inputAssignments -> how to map llvm registers (args) to asm ones used in the instructions
```
hernanponcedeleon, line 445:
```suggestion
    // 1) inputAssignments -> how to map llvm registers from args to asm registers used in the asm instructions
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    // 2) outputAssignments -> how to map asm registers to the return register
+    // when we reach this point we have already defined all of the asmInstructions events
+    // which means that we already have created all of the asm registers
+    // we just have to read the constraints, and based on their type, understand if they are going to map
+    // to the (args) llvm registers or to the llvm return Register.
```
hernanponcedeleon, line 450:
```suggestion
    // When this visitor is called, we have already created all of the asm registers.
    // We just have to read the constraints, and based on their type, understand if they are going to be mapped
    // to the args registers or to the return register.
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    @Override
+    public Object visitAsmMetadataEntries(InlineAsmParser.AsmMetadataEntriesContext ctx) {
+        List<InlineAsmParser.ConstraintContext> constraints = ctx.constraint();
+        // we already have all of the registers from LHS of the context call (asmInstrEntries)
+        // now it is time to generate the input and output assignments to let Dat3M understand the llvm to asm registers mapping
```
hernanponcedeleon, line 455: Remove

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        boolean outputRegistersInitialized = returnRegister == null;
+        for (int i = 0; i < constraints.size(); i++) {
```
hernanponcedeleon, line 457: Add a description along the lines "We iterate until we find the first non-output constraint. Then we immediately initialize the return register (the right-hand side of the assignment will be either a single register or an aggregate type deepening on how many output constraints we processed). We then map args registers to asm registers (we need to shift the register ID to find the corresponding args position of the matching register). Numeric constraint just map the registerID with the corresponding numeric position."

All other comments in the method could be removed.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+                inputAssignments.add(EventFactory.newLocal(asmRegisters.get(constraintValue), argsRegisters.get(i - getSizeOfReturnRegister())));
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public Object visitValue(InlineAsmParser.ValueContext ctx) {
+        String valueString = ctx.ConstantValue().getText().substring(1);
+        BigInteger value = new BigInteger(valueString, 10);
```
hernanponcedeleon, line 521: AFAIK you dont need to specify the base

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        checkState(expectedType instanceof IntegerType, "Expected type is not an integer type");
+        return expressions.makeValue(value, (IntegerType) expectedType);
+    }
+
+    @Override
+    public Object visitArmFence(InlineAsmParser.ArmFenceContext ctx) {
+        // check which type of fence it is : DataMemoryBarrier or DataSynchronizationBarrier
+        String type = ctx.DataMemoryBarrier() == null ? ctx.DataSynchronizationBarrier().getText() : ctx.DataMemoryBarrier().getText();
+        String option = ctx.FenceArmOpt().getText();
+        String barrier = type + " " + option;
```
hernanponcedeleon, line 531: can't you just do ctx.getText() and match with e.g. `dmb.ish`? i.e. keep the dots

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    public Object visitRiscvFence(InlineAsmParser.RiscvFenceContext ctx) {
+        String type = ctx.RISCVFence().getText();
+        String firstOption = ctx.FenceRISCVOpt(0).getText();
+        String secondOption = ctx.FenceRISCVOpt(1) == null ? "" : ctx.FenceRISCVOpt(1).getText();
+        String barrier = type + " " + firstOption + " " + secondOption;
```
hernanponcedeleon, line 562: Same as with arm fences. Why we cannot match `fence.r.r`?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    private final Function llvmFunction;
+    private final Register returnRegister;
+    private final ExpressionFactory expressions = ExpressionFactory.getInstance();
+    private CmpInstruction comparator;
+    // keeps track of all the labels defined in the the asm code
+    private final HashMap<String, Label> labelsDefined = new HashMap<>();
+    // used to keep track of which asm register should map to the llvm return register if it is an aggregate type
+    private final List<Expression> pendingRegisters = new ArrayList<>();
+    // holds the LLVM registers that are passed as (args) to the the asm -- asm"..." (args)
+    private final List<Expression> argsRegisters;
```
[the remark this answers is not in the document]
StefanoDalMas: llvmVisitor is going to pass a List<Expression> and when I save them in my class I am keeping them as Expression, shall I cast them to Registers?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+    // expected type of RHS of a comparison.
+    private Type expectedType;
```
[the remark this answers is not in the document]
StefanoDalMas: we do not use the 'w' or 'x' to get type hints anymore, as we rely on clobbers to get the type. This is used whenever we have a comparison with a constant like cmp $1, #0. In the visitValue we have to create a '0' with the correct type ($1 could be a bv64 or a bv32 for example).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
StefanoDalMas: since the parser eats the WhiteSpaces, ctx.getText() would return "dmbish" and I wanted to keep the instruction from the "argument" separated. I could make barrier = type + "." + option;

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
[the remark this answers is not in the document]
hernanponcedeleon: Ok, that one makes sense. What about pendingRegisters?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,619 @@
...
+// =r or =&r means we need to map an asm register to the return register. These are called Output constraints
+// Q, *Q or r means we need to map an LLVM register from 'args' to an asm register. These are called Input constraints
+// =*m it means that the register is a memory location, and is not mapped to any register. These are called Indirect constraints.
+// a constant X, it means that a register from 'args' is mapped to the Xth asm register which in turn is mapped to the return register.
+// Here are some examples to understand what is happening.
+// We are going to use ARMV7 names for readability "$0, $1, $2, ...", but other inline asm formats follow the same pattern 
+// BASE CASE
+// a)  
+// asm: r10 = i32 call asm "ldr $0, $1"," =r, *Q"(ptr r9) 
+// Code variables: asmRegisterNames := [$0, $1] ; argsRegisters := [r9] ; constraints := [=r, *Q]
```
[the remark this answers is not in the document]
StefanoDalMas: Yes asmRegistersNames became asmRegisters, argsRegisters remained the same. For constraints, they are only kept in the visitAsmMetadataEntries (they were called clobbers, but the name I gave was wrong)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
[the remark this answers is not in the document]
hernanponcedeleon: Can't you rather encapsulate this as part of the `CmpInstruction`? i.e. whenever you are using `expectedType`, rather use `comparator.left.getType()`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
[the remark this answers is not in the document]
StefanoDalMas: It was left from an older version, removed!

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
hernanponcedeleon: > The visitor will first visit the asm code (which will create the events and asm registers) and then the constraints. The latter will take care of creating input and output assignments.

I meant to add this sentence as a description of the overall method in replacement off all the inlined comments

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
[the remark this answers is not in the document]
StefanoDalMas: Atm in visitCompare I use the expectedType to create the 2nd register, and then I am able with both the registers to create the CmpInstruction. How would I be able to change this to be part of the Cmp in a clean way?

cat/arm.cat
hernanponcedeleon: @StefanoDalMas

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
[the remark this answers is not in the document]
hernanponcedeleon: You are right, `visitValue()` is called between the first registers was created and we actually create the `CmpInstruction` object.

cat/arm.cat
StefanoDalMas: is this spacing ok?

ThomasHaas: It's okay, but inconsistent about `;`. Sometimes you have `a; b` and sometimes `a ; b`.
You should be consistent but can otherwise decide on what spacing you like (even `a;b` is fine).

hernanponcedeleon:
It seems that (contrary to what we would usually expect) the assume solver is much faster than refinement not only on armv8, but even on armv7.  The latter is unexpected due to the use of recursion. 

These are the times for refinement **ALONE**
```
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1265 s -- in com.dat3m.dartagnan.inlineAsm.armv8.libvsync.AsmLibvsyncArmv8Test
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv8.ck.AsmCkArmv8Test
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 110.7 s -- in com.dat3m.dartagnan.inlineAsm.armv8.ck.AsmCkArmv8Test
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv7.libvsync.AsmLibvsyncArmv7Test
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 2695 s -- in com.dat3m.dartagnan.inlineAsm.armv7.libvsync.AsmLibvsyncArmv7Test
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv7.ck.AsmCkArmv7Test
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 365.7 s -- in com.dat3m.dartagnan.inlineAsm.armv7.ck.AsmCkArmv7Test
```
These are the times for **BOTH** assume and refinement
```
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv8.libvsync.AsmLibvsyncArmv8Test
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1862 s -- in com.dat3m.dartagnan.inlineAsm.armv8.libvsync.AsmLibvsyncArmv8Test
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv8.ck.AsmCkArmv8Test
[INFO] Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 307.9 s -- in com.dat3m.dartagnan.inlineAsm.armv8.ck.AsmCkArmv8Test
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv7.libvsync.AsmLibvsyncArmv7Test
[INFO] Tests run: 18, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3040 s -- in com.dat3m.dartagnan.inlineAsm.armv7.libvsync.AsmLibvsyncArmv7Test
[INFO] Running com.dat3m.dartagnan.inlineAsm.armv7.ck.AsmCkArmv7Test
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 406.9 s -- in com.dat3m.dartagnan.inlineAsm.armv7.ck.AsmCkArmv7Test```
The later is usually never the case and we need to investigate why this is happening here. 
```
@StefanoDalMas can you put back the assume solver in the tests, comment out (do not remove) the refinement solver and add a TODO comment saying that this takes long and we need to investigate?

cat/arm.cat
StefanoDalMas: I opted for ```a; b```

dartagnan/src/test/java/com/dat3m/dartagnan/inlineAsm/armv7/libvsync/AsmLibvsyncArmv7Test.java
```diff
@@ -0,0 +1,109 @@
...
+            {"once", 5, PASS}
+        });
+    }
+
+    @Test
+    public void testAllSolvers() throws Exception {
+        try (SolverContext ctx = mkCtx(); ProverWithTracker prover = mkProver(ctx)) {
+            assertEquals(expected, RefinementSolver.run(ctx, prover, mkTask()).getResult());
+        }
+    }
```
hernanponcedeleon, line 85: Didn't we agree the assume solver was much faster than refinement. It seems some changes were lost after some rebase

StefanoDalMas: From what I remembered for armv7 Assume was the slower one.
I tried running only seqcount and seqlock for armv7 with Z3 on the Mac
Refinment 6 s circa
Assume took 162 s

hernanponcedeleon: Yes, I was confused with armv8. I just run armv7 locally with the assume solver and it run OOM for libvsync benchmarks

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserInlineAsm.java
```diff
@@ -0,0 +1,46 @@
...
+import com.dat3m.dartagnan.parsers.InlineAsmLexer;
+import com.dat3m.dartagnan.parsers.InlineAsmParser;
+import com.dat3m.dartagnan.parsers.program.visitors.VisitorInlineAsm;
+import com.dat3m.dartagnan.program.Function;
+import com.dat3m.dartagnan.program.Register;
+import com.dat3m.dartagnan.program.event.Event;
+
+public class ParserInlineAsm {
+
+    private VisitorInlineAsm visitor;
```
ThomasHaas, line 22: This can be made a local variable. Alternatively, you initialize it in the constructor and get rid of the other three members.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+    private class CmpInstruction {
+
+        private final Expression left;
+        private final Expression right;
+
+        private CmpInstruction(Expression left, Expression right) {
+            this.left = left;
+            this.right = right;
+        }
+    }
```
ThomasHaas, line 104: You can replace this by a record:
```Java
private record CmpInstruction(Expression left, Expression right)
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+
+    private final List<Local> inputAssignments = new ArrayList<>();
+    private final List<Event> asmInstructions = new ArrayList<>();
+    private final List<Local> outputAssignments = new ArrayList<>();
+    private final Function llvmFunction;
+    private final Register returnRegister;
+    private final ExpressionFactory expressions = ExpressionFactory.getInstance();
+    private CmpInstruction comparator;
+    // keeps track of all the labels defined in the the asm code
```
ThomasHaas, line 113: Typo: "... the the ..."

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+    private final HashMap<String, Label> labelsDefined = new HashMap<>();
+    // used to keep track of which asm register should map to the llvm return register if it is an aggregate type
+    private final List<Register> pendingRegisters = new ArrayList<>();
+    // holds the LLVM registers that are passed (as args) to the the asm side
```
ThomasHaas, line 117: Typo: "... the the ..."

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+        this.llvmFunction = llvmFunction;
+        this.returnRegister = returnRegister;
+        this.argsRegisters = llvmArguments;
+    }
+
+    // Returns the size of the return register
+    // null / void -> 0
+    // i32 / bool -> 1
+    // aggregateType -> the size of the aggregate
+    private int getSizeOfReturnRegister() {
```
ThomasHaas, line 134: I find this name a bit misleading. It relates to the number of ASM registers that form the return value?
Maybe it would be better to call the function like `getNumASMReturnRegisters`?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+//      2. the first args register maps to the next asm register, i.e. asm_3 <- r9
+// c) MULTIPLE ARGS
+// asm: r10 = i32 call asm "ldr $0, $1 ; ldr $0, $2 "," =r, r, *Q"(i32 r8, ptr r9)
+// Code variables: asmRegisters := [asm_0, asm_1, asm_2] ; argsRegisters := [r8, r9] ; constraints := [=r, r, *Q]
+//    1. the first asm register maps to the output, i.e. r10 <- asm_0
+//    2. the two args registers map to the next two asm registers, i.e.
+//       - asm_1 <- r8
+//       - asm_2 <- r9
+// THERE IS NO RETURN REGISTER
+// d)
```
ThomasHaas, line 68: I think the two lines should be one.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+// asm: r11 = call { i32, i32, i32, i32 } asm "ldr $0, $4 ; add $1, $0, $3 ; add $2, $1, $4 ; ldr $2, $0", "=&r,=&r,=&r,=&r,*Q,3"(ptr r10, i32 r8)
+// Code variables: asmRegisters := [asm_0, asm_1, asm_2, asm_3, asm_4] ; argsRegisters := [r10, r8] ; constraints := [=&r, =&r, =&r, =&r, *Q, 3]
+//    1. we have 4 output constraints, so we have an aggregate type for the return register
+//       - r11[0] <- asm_0
+//       - r11[1] <- asm_1
+//       - r11[2] <- asm_2
+//       - r11[3] <- asm_3
+//    2. map function parameter to next asm register, i.e. asm_4 <- r10
+//    3. the third asm register is related both to the return register (already above in r11[3] <- asm_3) and to an args register, i.e. asm_3 <- r8
+public class VisitorInlineAsm extends InlineAsmBaseVisitor<Object> {
```
ThomasHaas, line 93: What I'm about to say is not related to the highlighted part, but rather the whole introductory text.
It seems to me that inline assembly uses X asm registers, some of which are return registers, some of which are argument/input registers, and some of which are both.
It is possible that there are registers that are none of the above, i.e., could the inline assembly directly reference registers without symbolic placeholder, or is this considered mal-formed inline assembly? I assume it mal-formed, but I want to be sure.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+
+    // This function lets us know which type we need to assign to the created asm register.
+    // In order to do so, we have to understand which llvm register it is going to be mapped to.
+    // Given the registerID of the register e.g. $2 -> registerID = 2
+    // returns the type of the llvm register it is mapped to by the clobbers
+    // if it is referencing the return register, return its type.
+    private Type getLlvmRegisterTypeGivenAsmRegisterID(int registerID) {
+        Type registerType;
+        if (isPartOfReturnRegister(registerID)) {
+            if (returnRegister.getType() instanceof AggregateType at) {
```
ThomasHaas, line 176: You have a `isReturnRegisterAggregate` function. Also, you should be consistent how you check for aggregates: either you use `instanceof` or you check the size, but not both if possible

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -0,0 +1,550 @@
...
+        }
+        return null;
+    }
+
+    @Override
+    public Object visitValue(InlineAsmParser.ValueContext ctx) {
+        String valueString = ctx.ConstantValue().getText().substring(1);
+        BigInteger value = new BigInteger(valueString);
+        checkState(expectedType instanceof IntegerType, "Expected type is not an integer type");
+        return expressions.makeValue(value, (IntegerType) expectedType);
```
ThomasHaas, line 465: You should move `checkState` to the very first line.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
StefanoDalMas: I think it depends on how you think about the composite register -- I thought that if I have a register which has aggregateType { i32, i32 }, that would be a return register with the "Tuple" type, therefore I would have a size of 2. 
So yes I can think of that as "an aggregate type is composed of multiple registers". Does the getNumASMReturnRegisters sound better to you?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
StefanoDalMas: Never heard of this type of class, thanks!

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
StefanoDalMas: exactly, it is malformed!
It is expected that the names of the inline asm refer only to registers passed as (args) after the clobber list and constraints list, or refer to the return Register

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
ThomasHaas: I understand where you are coming from. I shouldn't have said "misleading" but rather "ambiguous". I think the word "size" just has too many interpretations (for example number of bits/bytes).
In the case of aggregates, it was also not clear why you only need to count the top-level and don't need to recurse down (e.g., nested aggregates). Intuitively, I would have said that `{ { int32 a; int32 b } }` has size two, but due to the nesting it is counted as one.
I understand that inline assembly has only primitive registers and so you won't get nested aggregate types in the return register,  but I needed to understand quite a bit of the context to see that the code is correct.
You can keep the name if you want, but encoding a little bit more of the context into the method name might make it easier to understand. Alternatively, you can also expand the comment and add the idea that size is related to the number of ASM registers.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
ThomasHaas: Can you maybe add a comment about this? Something like "all registers in the inline assembly must only refer to [...]. Otherwise, the inline assembly is malformed."

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
ThomasHaas: "AmountOfTypes" is also too ambiguous. To me it sounds like it would count nested types, possibly also the types on intermediate levels, and possibly even only unique types.
Just make the name more specific rather than more general. That's the easiest IMO.
I think `getNumASMReturnRegisters` is probably the best. It does not invoke the idea that the method is general-purpose and so limits the reader from interpreting too much into it. 
It probably also better reflects the usages of this method in your code. For example:

```Java
Expression llvmRegister = argsRegisters.get(i - getSizeOfReturnRegister());
// --------
Expression llvmRegister = argsRegisters.get(i - getNumASMReturnRegisters());
```
The second version more easily suggests that your index shifting is related to the number of ASM return registers. It even makes sense without knowing the implementation of `getNumASMReturnRegisters()` whereas the first version can be confusing.

StefanoDalMas: Do the latest changes make more sense to you?

StefanoDalMas: I added a little bit more context, tell me if you like it more

ThomasHaas: Yeah, it's good now.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorInlineAsm.java
```diff
@@ -161,6 +164,8 @@ private Label getOrNewLabel(String labelName) {
     // Given the registerID of the register e.g. $2 -> registerID = 2
     // returns the type of the llvm register it is mapped to by the clobbers
     // if it is referencing the return register, return its type.
+    // As saif in the introduction, we are sure that an asm register is going to refer to a llvm one
```
hernanponcedeleon, line 167: Typo in "saif"

PR #803, opened 2025-02-23
Simplify IR Manipulation operations & Update existing code

ThomasHaas:
This is a simple PR that adds more utility functions to `IRHelper` and `EventFactory`, and a long missing `Event.insertBefore` function. Also `EventFactory.newInit` now uses `MemoryObject.getFeatureTags` which was previously only used in `EventFactory.newC11Init` for some unknown reason.
Lastly, I removed `EventFactory.newC11Init`.


dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
```diff
@@ -149,14 +152,34 @@ public void append(Event event){
...
 
+    public void updateEntry(Event event) {
+        Preconditions.checkArgument(event.getFunction() == this);
+        entry = event;
+        Event next;
+        while((next = entry.getPredecessor()) != null){
+            entry = next;
+        }
+    }
+
```
hernanponcedeleon, line 182: Isn't this actually updating the entry with the whole chain ending in `event`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/AbstractEvent.java
```diff
@@ -217,20 +218,35 @@ public void insertAfter(List<Event> toBeInserted) {
...
         }
         Preconditions.checkState(currentUsers.isEmpty(), "Cannot replace event that is still in use.");
         this.insertAfter(replacement);
         this.forceDelete();
+        return replacement;
     }
 
     @Override
-    public void replaceBy(List<Event> replacement) {
+    public <TCol extends Iterable<? extends Event>> TCol replaceBy(TCol replacement) {
```
hernanponcedeleon, line 245: What is this `TCol`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/Function.java
ThomasHaas: It is a symmetric version to `updateExit`: it searches for the new entry event starting from `event` and going backwards (`getPredecessor`).
It's not a particularly nice method (just like `updateExit`) and should not really be called directly.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/AbstractEvent.java
ThomasHaas: Short for `TCollection` (thought technically it's an `Iterable`).
The idea was that you can call something like
```
List<Event> replacement = event.replaceBy(eventSequence(...));
```
It just returns the collection of events you have put in with the same type (so if you used `Set<Event>` it would return a set).
That being said, I didn't end up using it anywhere I believe. So I might remove this feature again.

ThomasHaas: I removed the return values again.

hernanponcedeleon (approved):

PR #804, opened 2025-02-24
Move axiom from AddInitReadHandling to stdlib.cat

hernanponcedeleon:


cat/stdlib.cat
```diff
@@ -10,3 +10,5 @@ let coi = co & int
 let fri = fr & int
 
 let fencerel(F) = po;[F];po
+
+empty (([R] \ [range(rf)]);loc;[IW])
```
ThomasHaas, line 14: it might be good to add a comment for this axiom.

ThomasHaas (approved):
LGTM

PR #807, opened 2025-02-24
Fix nasty encoding bug related to XRA

ThomasHaas:
This PR fixes a bug in the computation of encode sets in particular edge-case that is AFAIK not even constructible on development right now.

The edge-case:
- Consider a new base relation `b` with empty must set but non-empty may set (prior to XRA). For example, `b` could be the identity on all events that produce value `42`.
- Suppose we place the axiom `empty b`, then XRA will (correctly) deduce that the may set must be empty (e.g., no event is allowed to produce value `42`).
- The `WmmEncoder` will now (wrongly) deduce that the encode set must be empty and potentially encode nothing (e.g., we fail to enforce that events are not allowed to produce value 42).

We usually capture this case by enlarging the encode sets with "discrepancies": information we could compute top-down but not bottom-up. However, if the discrepancy is directly on a base relation rather than between a derived relation and its dependencies, we fail to do so. In the case of base relations, the "bottom-up information" is just the initial approximation of its may/must sets.

The reason that this issue is not really reproducible by e.g. `empty po` is that although the discrepancy is not correctly encoded, the contradiction between the non-empty must set and the empty may set is.
That's why the counterexample should have an empty must set.


hernanponcedeleon:
Is the possible to programmatically construct the problematic situation? If so, can you add a regression test in `/test/java/com/dat3m/dartagnan/encoding`?

ThomasHaas:
Hmm, I don't think I can do this without introducing a new base relation. I also cannot construct a new base relation just within the test, because the corresponding visitor methods in RA/XRA/WmmEncoder need to get overwritten properly.

The only reason I found this bug is because I helped a colleague with adding speculation into Dartagnan.
Then adding an axiom like `empty (spec)` for this new relation should have prevented speculation, but it did not.

hernanponcedeleon (approved):

PR #808, opened 2025-02-24
Streamline CF encoding of jumps and Cbs

ThomasHaas:
We frequently check if a jump was taken or a control barrier was blocked by construction encodings like `cf(jump) /\ cond(jump)` resp. `cf(cb) /\ not exec(cb)`.
However, most use-cases should not care about what the encoding of, e.g., "a jump was taken" looks like.
So I added `EncodingContext.jumpTaken` and `EncodingContext.blocked/unblocked` as new methods to abstract the details.


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -475,10 +475,9 @@ private BooleanFormula isStuckEncoding(Thread thread, EncodingContext context) {
 
         private BooleanFormula generateBarrierStucknessEncoding(Thread thread, EncodingContext context) {
             final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
-            return bmgr.or(thread.getEvents().stream()
-                    .filter(ControlBarrier.class::isInstance)
-                    .map(e -> bmgr.and(context.controlFlow(e), bmgr.not(context.execution(e))))
-                    .toList());
+            return thread.getEvents(ControlBarrier.class).stream()
+                    .map(context::blocked)
+                    .reduce(bmgr.makeFalse(), bmgr::or);
```
hernanponcedeleon, line 480: This change is not semantically equivalent. For example, if the thread never started (due to e.g. weak forward progress), its control flow if not true, the previous definition would not consider it blocked, but the current would.

Why is this not a problem?

ThomasHaas: Why do you think the current encoding would consider it blocked? The control-flow variable is still checked like usual.

hernanponcedeleon: You are right, I confused it with `unblocked`

hernanponcedeleon (approved):

hernanponcedeleon:
@ThomasHaas can you rebase so I can merge?

PR #810, opened 2025-02-26
Parallel execution of tests in the CI

tonghaining:


hernanponcedeleon:
I was trying to test the weekly job, but [according to the documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) cron jobs will only trigger when they are on the default branch. I will merge and if it does not works as expected when it is on master, I will revise.

PR #811, opened 2025-02-27
Add dartagnan script

db7:
Currently, it is hard to remember the whole command line to start Dartagnan. This commit add a script called `dartagnan` which simplifies that.  It requires `DAT3M_HOME` to be set.  The script can be extended to contain more options in the future.


hernanponcedeleon:
Where I see a script like this bringing the most value is if it would hide the fact that the user needs to match the cat file with the correct target. 

What about doing something similar to vsyncer where the script takes care of these mappings
| cat      | target      | option-name
| ------------- | ------------- | ------------- |
| aarch64.cat | arm8 | arm8 |
| c11.cat | c11 | c11 |
| imm.cat | imm | imm |
| linux-kernel.cat | lkmm| lkmm |
| opencl.cat | opencl | opencl |
| power.cat | power | power |
| ptx-v7.5.cat | ptx | ptx |
| rc11.cat | c11 | rc11 |
| riscv.cat | riscv | riscv |
| sc.cat | c11 | sc |
| spirv.cat | vulkan | vulkan |
| tso.cat | tso | tso |
| vmm.cat | c11 | vmm |

db7:
sure, i had already added that for one case in the script.

db7:
@hernanponcedeleon -- I have added the mapping.

scripts/dartagnan
```diff
@@ -0,0 +1,128 @@
...
+if [ -z "${java_cmd}" ]; then
+    java_cmd="java"
+fi
+
+usage () {
+    echo "${this} [-cat MODEL] [-target TARGET] [-property PROPERTY] <FILE>"
+}
+
+while [ "$#" != "0" ]; do
+    #echo ">> $@"
```
hernanponcedeleon, line 30: Remove

scripts/dartagnan
```diff
@@ -0,0 +1,128 @@
...
+        ;;
+        *)
+            args="${args} $1"
+        ;;
+    esac
+    shift
+done
+
+# cat argument is mandatory. If the argument is not a file, we guess it to be
+# a file in DAT3M_HOME with .cat suffix.
```
hernanponcedeleon, line 56:
```suggestion
# a file in DAT3M_HOME/cat with .cat suffix.
```

scripts/dartagnan
```diff
@@ -0,0 +1,128 @@
...
+if [ -z "${cat}" ]; then
+    echo "error: no cat file specified"
+    exit 1
+elif [ ! -f "${cat}" ]; then
+    cat="${DAT3M_HOME}/cat/$cat".cat
+fi
+
+# if target is not given, we try to guess the target
```
hernanponcedeleon, line 64:
```suggestion
# if target is not given, we try to guess the target from the cat file
```

scripts/dartagnan
```diff
@@ -0,0 +1,128 @@
...
+		*aarch64*)
+        		target=arm8
+                ;;
+		*c11*)
+        		target=c11
+                ;;
+		*imm*)
+        		target=imm
+                ;;
+		*lkmm*)
```
hernanponcedeleon, line 76:
```suggestion
		*linux-kernel*)
```

scripts/dartagnan
```diff
@@ -0,0 +1,128 @@
...
+        		target=lkmm
+                ;;
+		*opencl*)
+        		target=opencl
+                ;;
+		*power*)
+        		target=power
+                ;;
+		*ptx*)
```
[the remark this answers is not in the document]
hernanponcedeleon: Ohhh I did not notice that. Now, this way is even better because it means it also works with ptx-v6.0

scripts/dartagnan
```diff
@@ -0,0 +1,127 @@
...
+        		target=arm8
+                ;;
+		*c11*)
+        		target=c11
+                ;;
+		*imm*)
+        		target=imm
+                ;;
+		*lkmm*)
+        		target=lkmm
```
hernanponcedeleon, line 76:
```suggestion
        		target=linux-kernel
```

db7:
@hernanponcedeleon let me know if you are happy with the script. Also, I'd suggest you to try the script on the different targets before merging, just to be sure.

hernanponcedeleon:
I am starting to believe we should simply hide the whole java cmd to the user, including the `--target` option and place the script in the root rather than a `scripts` folder. @ThomasHaas what do you think? 

We could thus simplify the REAME as 

======================================
...
You can also run Dartagnan from the console:

```
scripts/dartagnan -cat <CAT MODEL> [-property <PROPERTY>] <FILE> [OPTIONS]
```

The properties to be checked are specified via the `-property` option. Possible values are `program_spec` (e.g, safety properties as program assertions), `liveness` (i.e., all spinloops terminate), `cat_spec` (e.g., [data races as specified in the cat file](https://github.com/hernanponcedeleon/Dat3M/blob/master/cat/rc11.cat#L34-L39)). Default is `program_spec,cat_spec,liveness`.

Further options can be specified by adding `--<dat3m-option>=<value>` to `[OPTIONS]`. Common `dat3m-options` include:
- `bound`: unrolling bound for the BMC (default is 1).
- `solver`: specifies which SMT solver to use as a backend. Since we use [JavaSMT](https://github.com/sosy-lab/java-smt), several SMT solvers are supported depending on the OS and the used SMT logic (default is Z3).
- `method`: specifies which solving method to use. Option `lazy` (the default one) uses a customized solver for memory consistency. Option `eager` solves a monolithic formula using SMT solving. 

======================================

Also, we should probably also use the same default property from the script or running java, i.e. keep `program_spec,cat_spec,liveness` as default.

Finally, depending on which PR gets merged first (this or #772), we need to streamline `liveness` vs `termination`.

db7:
I think you should add the `scripts/dartagnan` to a install target and install the software at the right place, eg,
```
/usr/local/bin <- dartagnan script
/usr/local/share/dat3m <- DAT3M_HOME
```

scripts/dartagnan
hernanponcedeleon: This one is still wrong

ThomasHaas:
I think it's good to have a "quick user guide" that just mentions the script and its usage.
However, I would still keep an extended guide with a few more details and common options.
It might be time to write a proper user guide :)

hernanponcedeleon:
My proposal for the README is not satisfying since we would still need to mention `target` for the UI. 

Lets keep the README as it is for now and we can create a better user guide as a different PR.

@db7 there is still a problem with the name of the LKMM cat file.

Also, since we will keep the `scripts` folder, it would be good to also move `Dartagnan-SVCOMP.sh` there.

db7:
> My proposal for the README is not satisfying since we would still need to mention `target` for the UI.
> 
> Lets keep the README as it is for now and we can create a better user guide as a different PR.
> 
> @db7 there is still a problem with the name of the LKMM cat file.
> 
> Also, since we will keep the `scripts` folder, it would be good to also move `Dartagnan-SVCOMP.sh` there.

Shall I move this script in there?

scripts/dartagnan
ThomasHaas: I think Hernán marked the wrong part. The target is correct, but the cat model is `linux-kernel.cat` and not `lkmm.cat`, so the cat-file won't be detected correctly.

hernanponcedeleon: Yes, I was referring to the cat file, i.e. the `lkmm` at line 75

db7: ok, I see it now.

db7:
> > My proposal for the README is not satisfying since we would still need to mention `target` for the UI.
> > Lets keep the README as it is for now and we can create a better user guide as a different PR.
> > @db7 there is still a problem with the name of the LKMM cat file.
> > Also, since we will keep the `scripts` folder, it would be good to also move `Dartagnan-SVCOMP.sh` there.
> 
> Shall I move this script in there?

Done

hernanponcedeleon:
> @hernanponcedeleon let me know if you are happy with the script. Also, I'd suggest you to try the script on the different targets before merging, just to be sure.

All seems good after testing. I will merge once the CI passes.

hernanponcedeleon (approved):

Issue #817, opened 2025-03-06
Bug in SparseConditionalConstantPropagation


hernanponcedeleon:
The attach file shows a bug in our `SparseConditionalConstantPropagation` pass. This is how the IR looks like before the pass
```
[10] function bv64 popper_fn(bv64 r0)
1235: l1:
1236:    #__VERIFIER_loop_bound(bv64(1))
1237: l4.loop/itr_1:
1238:    {  } __localLiveSnapshot#0 <- {  }
1239:    === Calling ck_pr_md_load_int ===
1240:    bv64 1:asm_0 = load(*&pusher_done)
1241:    bv32 1:r7 <- trunc bv64 1:asm_0 to bv32
1242:    === Returning from ck_pr_md_load_int ===
1243:    bv1 r6 <- (bv32 1:r7 != bv32(0)) ? bv1(1) : bv1(0)
1244:    bv1 r7 <- bv1 r6 ^ bv1(1)
1245:    if(bv1 r7 != bv1(0)); then goto l4.l8
1246:    goto l4.l9
```
and this is how it looks after
```
[3] thread bv64 popper_fn(bv64 r0)
494:     ThreadStart by T0:main::ThreadCreate(T3:popper_fn, bv64(0))
495:     bool __startT3 = load(*&__com3__popper_fn, A)
496:     assume(bool __startT3)
497:     bv64 r0 := Argument(0) from ThreadCreate(T3:popper_fn, bv64(0))
498:     #__VERIFIER_loop_bound(bv64(1))
499:  l4.loop/itr_1:
500:     === Calling ck_pr_md_load_int ===
501:     bv64 1:asm_0 = load(*&pusher_done)
502:     === Returning from ck_pr_md_load_int ===
503:     if(!((trunc bv64 1:asm_0 to bv32) != bv32(0))); then goto l4.l8
504:     goto l4.l9
```
The `truncate` instruction was remove which leads to rf edges between `bv64` and `bv32` types like in the attached witness (coming from this cmd `java -jar dartagnan/target/dartagnan.jar cat/aarch64.cat --target=arm8 stack.ll --witness=png --property=liveness`).

[SCCP.zip](https://github.com/user-attachments/files/19106692/SCCP.zip)

ThomasHaas:
The truncation is in line 503. The assignment is just inlined

ThomasHaas:
I would rather assume this is induced by this code which explains what can go wrong when mixing sizes of stores and reads.
[...]

PR #820, opened 2025-03-07
Improved automatic data type conversion along rf-edges (+ related improvements)

ThomasHaas:
`EncodingContext.equal(Formula f, Formula g)` encodes equality modulo type conversions, e.g., if `f` and `g` have different are BV formulas of different bit-width, then one of them is truncated or extended to the other.
Which parameter was type-converted to the other was totally unclear from the callers perspective and caused ugly issues when encoding value equality induced by rf-edges.

I added a third parameter that let's the caller how the type conversion should be done (or if it should be done at all).

I also added the ability to evaluate `TupleFormulas` over a model. This requires using the helper method `EncodingHelper.evaluate(Formula, Model)` rather than invoking `Model.evaluate(Formula)`

@hernan-poncedeleon This will fix the strange values in the witness of the benchmark in #817

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
@@ -257,8 +257,9 @@ public final class EncodingContext {
     public BooleanFormula equal(Formula left, Formula right, ConversionMode cMode) {
         if (cMode == ConversionMode.LEFT_TO_RIGHT) {
             return equal(right, left, ConversionMode.RIGHT_TO_LEFT);
-        } else if (cMode == ConversionMode.NO) {
-            new EncodingHelper(formulaManager).checkEqualTypes(left, right);
+        } else if (cMode == ConversionMode.NO && !new EncodingHelper(formulaManager).hasSameType(left, right)) {
+            final String error = String.format("Mismatching formula types: %s and %s", left, right);
+            throw new IllegalArgumentException(error);
         }
 
         if (left instanceof IntegerFormula l) {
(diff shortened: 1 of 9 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -261,36 +248,40 @@ public Formula lastValue(MemoryObject base, int offset, int size) {
...
+        NO,
+        LEFT_TO_RIGHT,
+        RIGHT_TO_LEFT,
+    }
+
+    public BooleanFormula equal(Formula left, Formula right, ConversionMode cMode) {
+        if (cMode == ConversionMode.LEFT_TO_RIGHT) {
+            return equal(right, left, ConversionMode.RIGHT_TO_LEFT);
+        } else if (cMode == ConversionMode.NO) {
+            new EncodingHelper(formulaManager).checkEqualTypes(left, right);
```
hernanponcedeleon, line 261: Either this just returns a Boolean that gets lost or I am missing something

ThomasHaas: You are right. The checking method used to throw an exception but then I moved it and changed it to return a bool and forgot to change the callsite.

ThomasHaas:
@hernanponcedeleon Can you reproduce the CI error on this branch? Locally, I get correct results. I hope this is not some strange non-deterministic bug on our side.

hernanponcedeleon:
I cannot reproduce locally either. This might be a flaky test. I will retrigger the CI to see what happens

hernanponcedeleon (approved):

PR #822, opened 2025-03-08
Improved expressions package

ThomasHaas:
- Generalized `ExtractExpr` to supported nested indices
- Added `InsertExpr` (+ enabled LLVM parser to parse insertvalue instructions).
- Changed `Expressions.getOperand`s to return `ImmutableList` and overall increased the usage of `ImmutableList`.
- Improved `ExpressionsHelper` + refactored some code around it.

**TODO:**
- ~I want to add a bitvector concatenation operation.~ (I have decided against this for now)
-  **(DONE):** I want to bring `GEPExpr`'s semantics closer to LLVM's GEP. This could also fix https://github.com/hernanponcedeleon/Dat3M/issues/780 if the SPIRV parser is changed to translate `OpAccessChain` into GEPExpr.

**Disclaimer:** `InsertExpr` is untested.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
@@ -39,6 +39,18 @@ public final class TupleFormulaManager {
         return f.elements.get(index);
     }
 
+    public Formula extract(TupleFormula f, Iterable<Integer> indices) {
+        Formula extract = f;
+        for (int i : indices) {
+            if (extract instanceof TupleFormula tuple) {
+                extract = tuple.elements.get(i);
+            } else {
+                throw new IllegalArgumentException("Cannot extract from a non-tuple formula '%s'");
+            }
+        }
+        return extract;
+    }
+
     public TupleFormula makeTuple(List<Formula> elements) {
         return new TupleFormula(elements);
     }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
@@ -55,9 +55,8 @@ public final class TupleFormulaManager {
     }
 
     public TupleFormula insert(TupleFormula f, Formula value, Iterable<Integer> indices) {
-        Preconditions.checkArgument(!Iterables.isEmpty(indices));
-
         final int index = Iterables.getFirst(indices, -1);
+        Preconditions.checkArgument(0 <= index && index < f.elements.size());
         final Iterable<Integer> subindices = Iterables.skip(indices, 1);
         final boolean recurse = !Iterables.isEmpty(subindices);
         final List<Formula> newElements = IntStream.range(0, f.elements.size())
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
@@ -55,9 +55,8 @@ public final class TupleFormulaManager {
     }
 
     public TupleFormula insert(TupleFormula f, Formula value, Iterable<Integer> indices) {
-        Preconditions.checkArgument(!Iterables.isEmpty(indices));
-
         final int index = Iterables.getFirst(indices, -1);
+        Preconditions.checkArgument(0 < index && index < f.elements.size());
         final Iterable<Integer> subindices = Iterables.skip(indices, 1);
         final boolean recurse = !Iterables.isEmpty(subindices);
         final List<Formula> newElements = IntStream.range(0, f.elements.size())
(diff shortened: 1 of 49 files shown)


ThomasHaas:
The build seems to fail because of `Mockito`. I don't really know why.

hernanponcedeleon:
> The build seems to fail because of `Mockito`. I don't really know why.

The (failing) tests create a function by mocking the function type, e.g.,
```
Function function = new Function("mock", mock(FunctionType.class), List.of(), 0, null);
```
At some point in the code (it is not clear exactly where from the callstack), your changes need to access the correct type which does not really exists.

Since the tests mock the function type, you probably can change this for whatever real function type you want.

ThomasHaas:
> Since the tests mock the function type, you probably can change this for whatever real function type you want.

Let's see if it works. On my side, `Mockito` also fails to mock `Register` and `MemoryObject`, but those don't seem to be an issue in the CI.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
```diff
@@ -39,6 +42,33 @@ public Formula extract(TupleFormula f, int index) {
...
+        Formula extract = f;
+        for (int i : indices) {
+            if (extract instanceof TupleFormula tuple) {
+                extract = tuple.elements.get(i);
+            } else {
+                throw new IllegalArgumentException("Cannot extract from a non-tuple formula '%s'");
+            }
+        }
+        return extract;
+    }
```
hernanponcedeleon, line 55: What is the purpose of passing an iterable if you will always return the element matching the last position in the iterable?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionKind.java
```diff
@@ -21,7 +21,7 @@ enum Other implements ExpressionKind {
         GEP,
         CONSTRUCT,
         ITE,
-        EXTRACT;
+        EXTRACT, INSERT;
```
hernanponcedeleon, line 24: new line

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/ExpressionHelper.java
```diff
@@ -13,15 +17,56 @@ public static void checkSameType(Expression x, Expression y) {
...
+
+    public static Type extractType(Type type, Iterable<Integer> indices) {
+        for (int index : indices) {
+            checkInbounds(type, index);
+            if (type instanceof AggregateType aggregateType) {
+                type = aggregateType.getFields().get(index).type();
+            } else if (type instanceof ArrayType arrayType) {
+                type = arrayType.getElementType();
+            } else {
+                assert false : "unreachable";
```
hernanponcedeleon, line 67: We should avoid this kind of assertions. As I mentioned recently in a previous PR, assertions are disabled once we build the artifact and thus, they do not get triggered. If the check is important, we should throw an exception

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -671,12 +669,22 @@ public Expression visitSDivInst(SDivInstContext ctx) {
...
+        final ImmutableList<Integer> indices =
+                ctx.IntLit().stream().map(n -> Integer.parseInt(n.getText())).collect(ImmutableList.toImmutableList());
+        return assignToRegister(expressions.makeExtract(aggregate, indices));
+    }
+
+    @Override
+    public Expression visitInsertValueInst(InsertValueInstContext ctx) {
+        final Type typeAgg = parseType(ctx.typeValue(0));
+        final Type insertType = parseType(ctx.typeValue(1));
+        check(typeAgg instanceof AggregateType, "Non-aggregate type in %s.", ctx);
```
hernanponcedeleon, line 682: what about arrays? shouldn't we use `isAggregateLike`?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/utils/ExpressionHelper.java
ThomasHaas: And what if the check is unimportant? The code is literally unreachable, but I didn't wanna leave out the else case without documentation. The assertion is just a better version of a "// unreachable" comment.
Anyways, I can upgrade the assertion to a proper exception, even if it cannot ever trigger.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
ThomasHaas: I think you misread the code :). The `TupleFormula` can be nested and each index goes one nesting level deeper. Notice that both variables `extract` and `tuple` are modified in each iteration.

hernanponcedeleon:
> * **(DONE):** I want to bring `GEPExpr`'s semantics closer to LLVM's GEP. This could also fix [Types of address computations & overflows #780](https://github.com/hernanponcedeleon/Dat3M/issues/780) if the SPIRV parser is changed to translate `OpAccessChain` into GEPExpr.

I talked with @natgavrilenko and she will take a look to the SPIRV parser

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/formulas/TupleFormulaManager.java
hernanponcedeleon: Ok, now it makes more sense

PR #823, opened 2025-03-10
New client for ck's stack_empty benchmark

StefanoDalMas:
This PR introduces the new client for benchmarks/ck/stack_empty.c and its new test case for ck armv8


hernanponcedeleon:
@ThomasHaas the new liveness check is reporting a violation for [this](https://github.com/hernanponcedeleon/Dat3M/blob/068ad3290aea4acabfd7cb34c91db406dcb388e6/dartagnan/src/test/resources/inlineasm/armv8/ck/stack_empty_assume.ll) program (we change the while [here](https://github.com/hernanponcedeleon/Dat3M/blob/068ad3290aea4acabfd7cb34c91db406dcb388e6/benchmarks/ck/stack_empty_assume.c#L36-L37) to be sure the problem is not actually coming from the client code)
```
> java -jar dartagnan/target/dartagnan.jar cat/aarch64.cat --target=arm8 

dartagnan/src/test/resources/inlineasm/armv8/ck/stack_empty_assume.ll --bound=2 --witness=png
============ Termination violation found ============
        E624:   ck_stack_pop_upmc @stack_empty.c#35 -> @ck_stack.h#106
FAIL
```
From the attached witness, this is my understanding. Thread `T3` repeats its state when observing twice (`E525` and `E579`) value 56 as the next pointer. The argument is that we could repeat the suffix of that loop iteration infinitely and thus create a non-termination witness. However, the second read gets its value from `E385` in `T2`. Being able to infinitely often repeat the suffix requires that we could infinitely often generate such writes, and I don't see how that is the case since T2 finished.

ThomasHaas:
You have not attached a witness :)

hernanponcedeleon:
I cannot upload the `*.png` file (neither directly, nor as a `*.zip` file). 

[Here](https://github.com/user-attachments/files/19196621/stack_empty_assume.zip) is the `*.dot` file compressed. You can either convert it to a `*.png` locally or using [this editor online](https://dreampuf.github.io/GraphvizOnline/?engine=dot).

ThomasHaas:
I found the bug. It was a slight mistake in the encoding that allowed two iterations to get infix-suffix matched although they could never be equivalent. Basically, there is an optimization that avoids encoding anything between such iterations but this optimization was not applied everywhere: the matching relation was still encoded but the equivalence constraints between the iterations were not.
I will open a PR tomorrow.

dartagnan/src/test/java/com/dat3m/dartagnan/inlineAsm/armv8/ck/AsmCkArmv8Test.java
```diff
@@ -56,7 +56,7 @@ public static Iterable<Object[]> data() throws IOException {
             {"mcslock", 2, Result.PASS},
             {"ticketlock", 1, Result.PASS},
             {"spsc_queue", 1, Result.PASS},
-            // {"stack_empty", 1, Result.PASS}, // https://github.com/hernanponcedeleon/Dat3M/issues/817
+            {"stack_empty", 2, Result.UNKNOWN},
```
ThomasHaas, line 59: Is this version a lot slower than the `XYZ_assume` one? IIRC, the assume version returned PASS in like 30 seconds with B=3 using Z3, so I'm wondering why you keep the bound at 2.

hernanponcedeleon: The assume version was not checking the case where the call to pop fails. Using a while (as one should) make it much slower. It took 28 min to get UNKNOWN with B=3 on my machine. 

Due to the side-effectful spinloop (causing the problem fixed in #824), isn't the case that the new liveness theory would never return PASS?

ThomasHaas: The counterexample found before fixing #824 was not globally side-effectful: the non-terminating loop failed all of its conditional stores, so it had only local side-effects by updating some status register. We can detect repetition of local states.
I also started a run locally and I see that it does indeed take a lot longer, so the assume statement was quite powerful.

.github/workflows/maven.yml
```diff
@@ -19,8 +19,8 @@ on:
 
 jobs:
 
-  Asm-tests:
-    name: Asm tests
+  Ck-armv7-tests:
+    name: Ck armv7 tests
```
hernanponcedeleon, line 23: Lets use `Concurrency Kit` for the name description (you can keep `ck` for the previous line). Same applied to CK-armv8

dartagnan/src/test/java/com/dat3m/dartagnan/asm/armv8/libvsync/AsmLibvsyncArmv8Test.java
```diff
@@ -94,7 +94,7 @@ private SolverContext mkCtx() throws InvalidConfigurationException {
                 cfg,
                 BasicLogManager.create(cfg),
                 ShutdownManager.create().getNotifier(),
-                SolverContextFactory.Solvers.YICES2);
+                SolverContextFactory.Solvers.Z3);
```
hernanponcedeleon, line 97: Use yices

PR #824, opened 2025-03-12
Fix nontermination detection

ThomasHaas:


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/NonTerminationEncoder.java
```diff
@@ -384,11 +385,10 @@ private BooleanFormula areEquivalent(Event x, Event y) {
             }
         }
         if (x instanceof MemoryCoreEvent e && y instanceof MemoryCoreEvent f) {
-            equality = bmgr.and(
-                    equality,
-                    context.equal(context.value(e), context.value(f)),
-                    context.equal(context.address(e), context.address(f))
-            );
+            equality = bmgr.and(equality, context.equal(context.address(e), context.address(f)));
+            if (x instanceof Store || x instanceof Load) {
```
hernanponcedeleon, line 389: Why do we need this check if both `x, y, in MemoryCoreEvent`? Which class of events if this check supposed to filter out?

ThomasHaas: `GenericMemoryEvent` like srcu-sync. I just generalized this code to be more in-line with our interfaces.
Actually, I was surprised that the previous version did not cause any issues in the presence of `GenericMemoryEvent`, but I guess they only appear in litmus code and only outside of loops, so this code was never executed.

hernanponcedeleon: Why there is no need to check that `y` is not e.g., `srcu-sync`?  Does ` x in Store (resp. Load) => y in Store (resp. Load)`?

ThomasHaas: Yes, the implication holds because one of the first lines of this method checks `x.getClass() == y.getClass()`, so if I know the class of one, I know the class of the other. The `instanceof` checks still often check both but simply because I use its built-in casting mechanism.

hernanponcedeleon (approved):

PR #825, opened 2025-03-12
Use LEFT_TO_RIGHT when enforcing same value in encoding of final valu…

hernanponcedeleon:
…es of addresses


hernanponcedeleon:
A remaining of #820

ThomasHaas (approved):
LGTM.

Did you run into a test where this was a problem?

hernanponcedeleon:
> LGTM.
> 
> Did you run into a test where this was a problem?

Yes, I was playing around with some SPIRV code from llama.cpp while I noticed this. But the code still has quite some issue to be able to provide a proper test for the CI.

ThomasHaas:
> Yes, I was playing around with some SPIRV code from llama.cpp while I noticed this. But the code still has quite some issue to be able to provide a proper test for the CI.

I was just curious. I totally believe that this is an issue even without a test :).

PR #828, opened 2025-03-19
RISCV and PPC visitors

StefanoDalMas:
Following [#802 ](https://github.com/hernanponcedeleon/Dat3M/pull/802), this PR adds support to reason about ppc and riscv inline asm.
The common functions across visitors which are independent from the grammar have been moved into parsers/program/utils/AsmUtils.java file


dartagnan/src/test/java/com/dat3m/dartagnan/asm/ppc64/ck/AsmCkPpc64Test.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Test
+    public void testAllSolvers() throws Exception {// TODO : RefinementSolver takes too long to run, we have to investigate this
+        // try (SolverContext ctx = mkCtx(); ProverWithTracker prover = mkProver(ctx)) {
+        //     assertEquals(expected, RefinementSolver.run(ctx, prover, mkTask()).getResult());
+        // }
+        try (SolverContext ctx = mkCtx(); ProverWithTracker prover = mkProver(ctx)) {
+            assertEquals(expected, AssumeSolver.run(ctx, prover, mkTask()).getResult());
+        }
+    }
```
hernanponcedeleon, line 71: This seems to be copied from the other tests. In general, I would expect refinement to work much better than assume for PPC. This might indeed be the reason why the CI run in a OOM problem

StefanoDalMas: Apparently the ebr makes the test loop forever. I have tried with z3 locally with either Assume and Refinement. I have pushed one which only uses Refinement to see if the CI with Yices breaks down

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -376,18 +381,59 @@ public Expression visitCallInst(CallInstContext ctx) {
...
+                                events.add(EventFactory.newLocal(resultRegister, program.newConstant(resultRegister.getType())));
+                            }
+                        } catch (ParsingException x86ParsingException){
+                            logger.warn("Found unrecognized token for inline assembly : '{}'. Setting non deterministic value ", x86ParsingException.getMessage());
+                            if(resultRegister != null){
+                                events.add(EventFactory.newLocal(resultRegister, program.newConstant(resultRegister.getType())));
+                            }
+                        }
+                    } 
+                }
```
ThomasHaas, line 436: This code is horrendous :O.
First of all, don't nest try-catch blocks. You could do this in a loop over a list containing all the parsers.
Secondly, the code you execute per parser is basically the same: you could have this inside a helper function that takes the parser as input and returns a (optional) set of events (optional might be useful for saying "I cannot handle this code").

Pseudo code
```
final List<AsmParser> parsers = List.of(new ParserX86(...), new ParserPPC(...), ...);
bool success = false;
for (var parser : parsers) {
   Optional<List<Event>> events = tryParse(parser, ...);
   if (events.hasValue()) {
      success = true;
      ...
   }
}
if (!success) {
   throw new ParsingException(...);
}
```

StefanoDalMas: I completely forgot about this part before opening the PR, thank you!

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -376,21 +382,40 @@ public Expression visitCallInst(CallInstContext ctx) {
...
+            Optional<List<Event>> events = Optional.empty();
+            for(ParserAsm parser : parsers){
+                // we have to generate the stream each time as the parser consumes it
+                CharStream charStream = CharStreams.fromString(asmCode);
+                try {
+                    events = tryParse(parser,charStream);    
+                } catch (ProgramProcessingException e) {
+                    logger.warn("Support for inline assembly instruction '{}' is not available for parser '{}'. Setting non deterministic value ", e.getMessage(), parser.getClass().getSimpleName());
+                    if(resultRegister != null){
+                        Event nonDeterministicValue = EventFactory.newLocal(resultRegister, program.newConstant(resultRegister.getType()));
```
ThomasHaas, line 403: You should use `EventFactory.Svcomp.newNonDetChoice(Register)`.
I know the previous code didn't do it either, but it was also wrong :).

PS. I need to move this event into the standard namespace; it has nothing to do with Svcomp (anymore).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -1487,5 +1513,15 @@ public <T extends MdNode> Optional<T> getField(String fieldName) {
             return Optional.empty();
         }
     }
+    
+    // ----------------------------------------------------------------------------------------------------------------
+    // Helper to parse inline asm code
+    private Optional<List<Event>> tryParse(ParserAsm parser, CharStream asmCode) throws ProgramProcessingException{
```
ThomasHaas, line 1519: Why the `ProgramProcessingException`? I think in the parser we should only use `ParsingException`.

dartagnan/src/test/java/com/dat3m/dartagnan/asm/ppc64/ck/AsmCkPpc64Test.java
hernanponcedeleon: Have you tried on your mac or on the work laptop? How long does it take?

StefanoDalMas: I have tried with Z3 MacBook and if I run the CI after 30 min it is still going(whereas running Java -jar …. it takes about 10 minutes).
I have not tried on the work laptop as I think that the CI machines are more powerful and they are running Yices2, so I just pushed and saw the results

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
ThomasHaas: I think you can add a new exception type or use `UnsupportedOperationException` for unimplemented cases. That's the canonical choice at least

hernanponcedeleon: The current implementation is wrong. Unless the code uses armv8 asm (which is the first parser in your list), you will throw and exception and simply treat the whole call as a non-det value. This is the reason why all asm tests except armv8 are failing

hernanponcedeleon (commented):
Rather than `ParserX` and `VisitorX`, use `ParserAsmX` and `VisitorAsmX`. This is more inlined with e.g., `VisitorLitmusX`

hernanponcedeleon (commented):
Now that you removed ebr, the sum of all asm tests is still smaller than others tests in the CI (except for arm7, most of them are done in 1 min). 

Thus, I would merge all asm test again into a single test

dartagnan/src/main/antlr4/RISCVLexer.g4
```diff
@@ -0,0 +1,57 @@
...
+
+import BaseLexer;
+
+
+// instructions 
+Load                        : 'ld' | 'lw' | 'lwu';
+LoadImmediate               : 'li';
+LoadExclusive               : 'lr.d' | 'lr.w';
+LoadAcquireExclusive        : 'lr.d.aq';
+LoadAcquireReleaseExclusive : 'lr.d.aqrl';
```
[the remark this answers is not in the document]
StefanoDalMas: This is an instruction generated by the "handwritten" / "generated" code made by Sebastian, so I am not 100% confident this instruction is used / useful / implemented atm in riscv

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserAsm.java
```diff
@@ -0,0 +1,18 @@
...
+import com.dat3m.dartagnan.program.event.Event;
+
+
+public class ParserAsm {
+
+    public ParserAsm() {
+    }
+
+    public List<Event> parse(CharStream charStream) {
+        return null;
```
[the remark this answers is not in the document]
StefanoDalMas: I think I can make the ParserAsm abstract, so that I cannot instantiate it, and only their children can override the parse() method

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorAsmPPC.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    // This visitor generates two sets of events: 
+    // 1) inputAssignments -> how to map llvm registers from args to asm registers used in the asm instructions
+    // 2) outputAssignments -> how to map asm registers to the return register
+    // When this visitor is called, we have already created all of the asm registers.
+    // We just have to read the constraints, and based on their type, understand if they are going to be mapped
+    // to the args registers or to the return register.
+    @Override
+    public Object visitAsmMetadataEntries(PPCParser.AsmMetadataEntriesContext ctx) {
```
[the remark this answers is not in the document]
StefanoDalMas: I think in some way it could be done -- I have tried several times to do this without ever being able to properly define some AsmGrammar as father and then having all of the children as various visitors... I can check some more how to do it but I don't know how it's going to end up

StefanoDalMas:
In this run asm took 60 mins, so maybe splitting ck and libvsync could reduce the time for the CI to 40 min like Litmus ones

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorAsmPPC.java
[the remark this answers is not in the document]
StefanoDalMas: I am double checking and the rules for AsmMetadataEntries might have some small differences(e.g. the '*m' case in arm was not mapping to any of the (args) part passed to the inline asm, whereas for PPC they are used as if they were standard input registers e.g. 'r'. 
I think due to these small differences in parsing the metadata we could not refactor this part of the code (maybe only the isNumericConstraint or something like this), what do you think?

dartagnan/src/main/antlr4/RISCVLexer.g4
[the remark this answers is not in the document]
hernanponcedeleon: This probably means if is never used. Then remove

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorAsmPPC.java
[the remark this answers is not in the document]
hernanponcedeleon: As far as I can see, the only difference is that arm grammar has
```
constraint : outputOpAssign | memoryAddress | inputOpGeneralReg | overlapInOutRegister | pointerToMemoryLocation;
```
while ppc has
```
constraint : outputOpAssign | inputOpGeneralReg | overlapInOutRegister;
```
Since these are disjunctions, I do not see why having a constraint rule independent of the specific grammar would not be possible.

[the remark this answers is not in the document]
StefanoDalMas: For example the parser rule inputOpGeneralReg for arm is RLiteral, and for PPC is RLiteral | Equals? Ast MLiteral. Doesn't this cause a problem? Also, since metadata for arm is skipping the "*m", whereas PPC uses them as input constraints, doesn't this also mean that the code itself is going to be different?
Or do you want me to refactor all of the function signature in InlineUtils? I don't fully get it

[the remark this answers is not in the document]
hernanponcedeleon: As discussed today, let keep it as it is

dartagnan/src/main/antlr4/Arm.g4
```diff
@@ -1,17 +1,17 @@
-grammar InlineAsm;
+grammar Arm;
```
hernanponcedeleon, line 1: As discussed today, rename the grammar files as `asmX`

hernanponcedeleon (approved):

Issue #789, opened 2024-12-11
Modeling of pthread_join


hernanponcedeleon:
I was discussing with @natgavrilenko about our modeling of `pthread_joins`, and we found it to be wrong.

Currently we report a liveness violation for this program
```
#include <pthread.h>
#include <stdatomic.h>

atomic_int x;

void *thread_0(void *unused)
{
    // Anything can go here, we just need the thread to exists
}

void *thread_1(void *unused)
{
    while(atomic_load_explicit(&x, memory_order_seq_cst) == 0);
}

int main()
{
    pthread_t t0, t1;
    atomic_init(&x, 0);

    pthread_create(&t0, NULL, thread_0, NULL);
    pthread_create(&t1, NULL, thread_1, NULL);

    pthread_join(t0, NULL);

    atomic_store_explicit(&x, 1, memory_order_seq_cst);

    pthread_join(t1, NULL);

    return 0;
}

```

ThomasHaas:
Yes this is wrong, because the we do not construct a proper spinloop that is waiting for the thread. This also means we do not detect non-termination of `pthread_join` itself as liveness violation.

Doing it the proper way is not super straightforward, because `pthread_join` gets processed after all the loop unrolling, so either we can somehow generate the spinloop already before that (maybe by partially processing `pthread_join`) or we perform another loop unrolling/spinloop detection pass after that (maybe applying them directly to the newly generated loops).

Short-term solutions are not reporting liveness violations when joining fails by marking the terminating jump as `Tag.EXCEPTIONAL_TERMINATION`.

PR #829, opened 2025-03-19
Model pthread_join() as spinloop

CapZTr:
Fix of #789


hernanponcedeleon:
Please add some unit tests. Besides the one listed in the issue which targets liveness, it would be good to have some test having an assertion.

ThomasHaas:
You should also add a `TODO` or `FIXME`. We certainly don't want to keep the code around like this.
Btw. `LoopUnrolling` contains public static members to construct the names for loop labels. You should reuse those.

PR #830, opened 2025-03-20
Tag AbortIf events as EXCEPTIONAL_TERMINATION

hernanponcedeleon:
Currently, we do not tag events coming from an explicit call to `abort()`, resulting on (wrongly) reporting a non-termination violation for this pattern (assuming at least two threads).
```
    pthread_mutex_lock(&m);
    if(__VERIFIER_nondet_bool()) {
        abort();
    }
    pthread_mutex_unlock(&m);
```


ThomasHaas (approved):
LGTM

PR #831, opened 2025-03-20
Fix arm visitor for ck/ebr client

StefanoDalMas:
I noticed that ebr was not properly parsed (the visitor did not create the events for sub $0, $0, #1). 
I have also removed the armv7 as it is using the "itt" instruction which is not supported, and I have made so the extraction of the value relies more on the grammar.


hernanponcedeleon:
There are still exceptions for `dartagnan/src/test/resources/asm/armv8/ck/ebr.ll` resulting on setting non-det values for the whole asm call
```
[21.03.2025] 16:06:55 [33m[WARN][m VisitorLlvm.visitCallInst - Found unrecognized token for inline assembly : '#1'. Setting non deterministic value 
```

StefanoDalMas:
Also by cloning this branch on a new folder and re building with mvn I do not get this error, did you rebuild it?

hernanponcedeleon:
My bad ... I was using `scripts/dartagnan` and that one picks the `*.jar` file based on `DAT3M_HOME`, thus I was running my local version rather than your fork.

Indeed, there are no more `Found unrecognized token for inline assembly`.

hernanponcedeleon (approved):

PR #832, opened 2025-03-20
Cleanup spirv tests

natgavrilenko:


dartagnan/src/test/java/com/dat3m/dartagnan/spirv/opencl/patterns/SpirvRacesTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Parameterized.Parameters(name = "{index}: {0}, {1}, {2}")
+    public static Iterable<Object[]> data() throws IOException {
+        return Arrays.asList(new Object[][]{
+                {"corr.spv.dis", false, PASS},
+                {"iriw.spv.dis", false, PASS},
+                {"sb.spv.dis", false, PASS},
+
+                {"mp.spv.dis", false, FAIL},
+                {"mp.spv.dis", true, PASS},
+                {"mp-acq2rx.spv.dis", false, FAIL},
```
hernanponcedeleon, line 58: if we even have races when with the filter, what is the point of having the version ignoring the filter?

natgavrilenko: I think it is good to have a few tests with both variations, they don't have any visible impact on the testing time. If there were hundreds of such tests, then yes.

hernanponcedeleon: Ok (as long as we don't end up doing this for all tests)

hernanponcedeleon (approved):

PR #836, opened 2025-03-25
Add -solver option to dartagnan script

db7:
The scripts uses z3 as default.

### DIFF
--- a/scripts/dartagnan
+++ b/scripts/dartagnan
@@ -6,6 +6,7 @@
 # * -cat      : memory model
 # * -target   : target architecture
 # * -property : property to be checked
+# * ... see usage function for other options
 #
 # Environment variables:
 #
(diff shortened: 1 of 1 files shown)


scripts/dartagnan
```diff
@@ -76,7 +86,7 @@ if [ -z "${target}" ]; then
 		*imm*)
         		target=imm
                 ;;
-		*linux-kernel*)
+		*lkmm*)
```
hernanponcedeleon, line 89: This will not work. The cat file is called `linux-kernel.cat`, not `lkmm.cat`

db7: Ohh, that was a mistake! Sorry

db7: Fixed!

hernanponcedeleon (approved):

PR #837, opened 2025-03-26
Add SVCOMP support for nontermination

hernanponcedeleon:
I added support for the termination category in the SVCOMP wrapper and used the new category proposal from [here](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1363) to evaluate the new termination encoding from #772.

The new encoding allows us to detect some non-termination cases that the old encoding did not

termination-1

but it is also slower for proving some cases of termination (which is expected)

termination-2

I also fixed a bug related to an index in the code related to the `SIMPLE` instrumentation variant.


ThomasHaas (approved):
LGTM

PR #847, opened 2025-04-01
Better support for IR-level threading and better handling of pthread_join

ThomasHaas:
This PR:
- Adds proper core-level `ThreadJoin` and `ThreadReturn` events. The former blocks until the target thread is terminated and once it unblocks, it returns the value of the `ThreadReturn` event executed by the terminating thread. This also solves issue #789  without the workaround of #841.
- Adds high-level `DynamicThreadCreate` and `DynamicThreadJoin` events that get eventually lowered to (static) `ThreadCreate` and `ThreadJoin` events. These high-level events are similar to `pthread_create` and `pthread_join` which are now compiled by the intrinsic pass into the new events.
The idea of the high-level events is that we can hopefully support other languages with dynamic thread spawning and other custom thread spawning extensions such as spawning of interrupt handlers or spawning of threads with custom (GPU) scopes.
- `DynamicThreadJoin` returns an error status when joining on an invalid tid or a thread whose return type does not match with the expected return type. `pthread_join` checks the status code and asserts on failure.
This means we detect the following errors as part of "program_spec": (1) invalid tid provided to `pthread_join` (a tid is invalid if it was not obtained by a `pthread_create` call, e.g., joining on "main" or uninitialized `pthread_t` is wrong) and (2) invalid join on `pthread_self`, i.e., a thread joining with itself. Notice that this is a program_spec error because `pthread_join` defines it as such, but we could also model this as a non-termination error.

**Other changes:** 
- Adds `CallEvent` interface for all call-like events (`FunctionCall` and `DynamicThreadCreate` currently). Adds `CallBase` event as a common implementation base.
- `NaiveDevirtualization` can devirtualize all `CallEvent`s, meaning it also works on `DynamicThreadCreate` with function pointers. Since `pthread_create` gets lowered to `DynamicThreadCreate`, the special-casing for `pthread_create` is not needed anymore. Also, the early intrinsics pass can now run before devirtualization, actually before all other passes.
- Adds benchmark tests for returning values from threads and invalid joining.
- Made `AssignmentInlining` more aggressive w.r.t. `ConstructExpr`, i.e., it always inlines them.
- Fixes bug when encoding nested aggregates.
- Adds `UniqueIdGenerator` class that streamlines the generation of unique names for dummy/temporary registers. So far it mostly is used to generate the usual `DUMMY_REG#XYZ` names, but it can also generate other unique register and even label names.

**Note:** The PR should be fully functional, but there are likely a few pieces of code I'm going to refactor.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/common/CallBase.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    protected CallBase(CallBase other) {
+        super(other);
+        this.funcType = other.funcType;
+        this.callTarget = other.callTarget;
+        this.arguments = new ArrayList<>(other.arguments);
+    }
+
+    public boolean isDirectCall() { return callTarget instanceof Function; }
```
hernanponcedeleon, line 47: Do you really need this one? I think the current default is doing exactly the same (since `getCallTarget() { return callTarget; }`)

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/threading/ThreadJoin.java
```diff
@@ -0,0 +1,50 @@
...
+    public ThreadJoin(Register resultRegister, Thread joinThread) {
+        Preconditions.checkArgument(resultRegister.getType().equals(joinThread.getFunctionType().getReturnType()));
+        this.resultRegister = resultRegister;
+        this.joinThread = joinThread;
+    }
+
+    protected ThreadJoin(ThreadJoin other) {
+        super(other);
+        this.resultRegister = other.resultRegister;
+        this.joinThread = null;
```
hernanponcedeleon, line 24: why not `this.joinThread = other.joinThread`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.math.BigInteger;
+import java.util.HashSet;
+import java.util.Set;
+
+
+/*
+    This event joins with given thread, i.e., it blocks until the target thread terminates properly.
+    - The target thread is specified by a thread id, typically the one returned by DynamicThreadCreate.
+    - The type of the return register must be { bv? status, TRet retValue } where TRet may be void if no value is returned.
```
hernanponcedeleon, line 23: what does `bv?` means here?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
```diff
@@ -0,0 +1,89 @@
...
+        INVALID_TID,            // The provided tid does not match with a joinable thread
+        INVALID_RETURN_TYPE     // The provided tid is valid, but the return type of that thread does not match
+                                // with the expected return type.
+    }
+
+    protected Register resultRegister;
+    protected Expression tid;
+
+    public DynamicThreadJoin(Register resultRegister, Expression tid) {
+        Preconditions.checkArgument(resultRegister.getType() instanceof AggregateType aggType
```
hernanponcedeleon, line 42: why it the return type an aggregate? what are the other two fields used for?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public DynamicThreadJoin getCopy() {
+        return new DynamicThreadJoin(this);
+    }
+
+    @Override
+    public Set<Register.Read> getRegisterReads() {
+        return Register.collectRegisterReads(tid, Register.UsageType.OTHER, new HashSet<>());
```
hernanponcedeleon, line 82: Is `tid` always coming from a local event? Can't it come from a load, and if so, shouldn't the usage be `DATA`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -400,6 +403,84 @@ private List<Event> inlineAtomicEnd(FunctionCall ignored) {
...
+        assert resultRegister.getType() instanceof IntegerType;
+
+        final Register tidReg = call.getFunction().newUniqueRegister("__tid", types.getArchType());
+        final DynamicThreadCreate createEvent = newDynamicThreadCreate(tidReg, PTHREAD_THREAD_TYPE, targetFunction, List.of(argument));
+
+        return eventSequence(
+                createEvent,
+                newStore(pidResultAddress, tidReg),
+                // TODO: Allow to return failure value (!= 0)
+                newLocal(resultRegister, expressions.makeZero((IntegerType) resultRegister.getType()))
```
hernanponcedeleon, line 428: This completely breaks `THREAD_CREATE_ALWAYS_SUCCEEDS` and needs to be fixed

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -400,6 +403,84 @@ private List<Event> inlineAtomicEnd(FunctionCall ignored) {
...
+        final Register resultRegister = getResultRegister(call);
+        assert resultRegister.getType() instanceof IntegerType;
+
+        final Type joinType = types.getAggregateType(List.of(types.getIntegerType(8), PTHREAD_THREAD_TYPE.getReturnType()));
+        final Register joinReg = call.getFunction().newUniqueRegister("__joinReg", joinType);
+
+        final Expression status = expressions.makeExtract(joinReg, 0);
+        final Expression retVal = expressions.makeExtract(joinReg, 1);
+
+        final Expression statusSuccess = expressions.makeValue(SUCCESS.ordinal(), (IntegerType) status.getType());
```
hernanponcedeleon, line 448: This means `SUCCESS` need always to be first in the enum, right? If so, this should be documented in the enum itself.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -400,6 +403,84 @@ private List<Event> inlineAtomicEnd(FunctionCall ignored) {
...
+        final Expression statusInvalidTId = expressions.makeValue(INVALID_TID.ordinal(), (IntegerType) status.getType());
```
hernanponcedeleon, line 449: What about the 3rd possible enum value?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -98,156 +103,121 @@ public void run(Program program) {
...
+                            newLocal(statusRegister, invalidRetType),
+                            EventFactory.newGoto(joinEnd)
+                    );
+                } else {
+                    // Successful join
+                    caseBody = eventSequence(
+                            joinCase,
+                            newThreadJoin(retValRegister, data.thread()),
+                            newAcquireLoad(syncRegister, data.comAddress),
+                            newAssume(expressions.makeNot(syncRegister)),
```
hernanponcedeleon, line 215: I don't understand why we need this assume

hernanponcedeleon:
>   The idea of the high-level events is that we can hopefully support other languages with dynamic thread spawning and other custom thread spawning extensions such as spawning of interrupt handlers or spawning of threads with custom (GPU) scopes.

The only current use case of this is in your interrupts brach, right?

NOTE: the changes in this PR solve some exceptions due to a comparison of expression with different integer types (e.g. bits) in several SVCOMP benchmarks.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -486,15 +488,20 @@ Formula makeVariable(String name, Type type) {
...
+            for (TypeOffset field : aggType.getFields()) {
+                fields.add(makeVariable(name + "@" + field.offset(), field.type()));
+            }
+            return tupleFormulaManager.makeTuple(fields);
+        }
+        if (type instanceof ArrayType arrType) {
+            Preconditions.checkArgument(arrType.hasKnownNumElements(), "Cannot encode array of unknown size.");
+            final List<Formula> elements = new ArrayList<>(arrType.getNumElements());
+            for (int i = 0; i < arrType.getNumElements(); i++) {
+                elements.add(makeVariable(name + "[" + i + "]", arrType.getElementType()));
```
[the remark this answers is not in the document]
ThomasHaas: I don't see much benefit. The getter is trivial (and probably gets inlined anyway) and the line of code is not so long that I need to shorten it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
ThomasHaas: It should indicate that different sizes are valid. The next lines explain that the type must be large enough to contain all error values. 
I can change it to `bvX` if you think it is clearer.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
ThomasHaas: You quoted the comment that explains it above :P.
There are only two fields: the first is the status and the second is value returned by the thread.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
ThomasHaas: It would probably even be `ADDR`. But this event gets compiled down and so its dependencies do not matter much. The way we implement it, namely by a switch of all possible tids, will give you `CTRL` in the end.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: I'm not sure what you mean. Did you encounter an issue?
It would surprise me, because the semantics is identical to what it used to be: we always return success, even if the thread creation failed. Not only is the semantics identical, but the code we generate is literally the same.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: The third value is no pthread error code: all threads in pthread have a fixed signature, namely `void *f (void*)`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
ThomasHaas: To establish synchronization without a spinloop. `ThreadJoin` will block until the target thread terminates, but it will not give us synchronization edges (e.g. `rf/hb`).
Notice that this is basically the code we had before we used a spinloop, but back then the code tried to block & synchronize, and now the two parts are done by separate events.

ThomasHaas:
> > The idea of the high-level events is that we can hopefully support other languages with dynamic thread spawning and other custom thread spawning extensions such as spawning of interrupt handlers or spawning of threads with custom (GPU) scopes.
> 
> The only current use case of this is in your interrupts brach, right?

Yes, because I add a `ThreadType` to `DynamicThreadCreate` which allows me to control what kind of thread is spawned (`STANDARD` or `INTERRUPT_HANDLER`). For hosts that dynamically invoke GPU kernels, `DynamicThreadCreate` could also get reused.
Besides this, the high-level events enable lowering of `pthread_join/create` as early as possible. In fact, I would eventually want to get rid of the whole "late intrinsics" pass.

> NOTE: the changes in this PR solve some exceptions due to a comparison of expression with different integer types (e.g. bits) in several SVCOMP benchmarks.

Interesting. I don't recall changing anything related to that.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/threading/ThreadJoin.java
hernanponcedeleon: Throwing an exception sounds like the best option

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: I did not encounter any problem, but I was worried about this https://github.com/hernanponcedeleon/Dat3M/blob/b3db2b01607b54a17c2fcd4e97cee5528eda96fb/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java#L515

However, this seems to be dead code because we AFAIK we do not have events tagged by `Tag.STARTLOAD`. Shall we get rid of this and the option `THREAD_CREATE_ALWAYS_SUCCEEDS` (at least until we take care of this TODO)?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: Is the value `INVALID_RETURN_TYPE` used anywhere?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: That printing is certainly broken. I fixed it in #869. That being said, even in #869 we would return the success error code for spuriously failing pthread_create calls.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: Yes, the `ThreadCreation` pass can generate this status for mistyped joins. 
Mistyped joins just cannot exist in C-code since all valid tids correspond to threads created via `pthread_create` which enforces a single unique type.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/threading/ThreadCreate.java
```diff
@@ -24,8 +24,7 @@ public ThreadCreate(List<Expression> arguments) {
 
     protected ThreadCreate(ThreadCreate other) {
         super(other);
```
hernanponcedeleon, line 26: why do we need the `super()` ... same for `ThreadJoin`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/threading/ThreadJoin.java
```diff
@@ -42,8 +37,8 @@ protected String defaultString() {
     }
 
     @Override
-    public ThreadJoin getCopy() {
-        return new ThreadJoin(this);
+    public ThreadCreate getCopy() {
```
hernanponcedeleon, line 40: should be `ThreadJoin`

ThomasHaas: copy&paste error :(

hernanponcedeleon (approved):

PR #848, opened 2025-04-05
Hierarchical scheduling

ThomasHaas:
This PR adds hierarchical scheduling (progress), i.e., scheduling can be specified on a per-scope level.
The configuration input is as follows:
```
// Uniform scheduling: all scopes are scheduled with the same progress model
--modeling.progress=<progModel>  

// Per-scope scheduling. Defaults to "fair" for unspecified scopes, 
// unless the special key-value-pair "default=<progModel>" is provided.
--modeling.progress=[<scope>=<progressModel>,...]
```

**How does hierarchical scheduling work?**
(1) Each scope has a progress model that specifies how a thread group on that scope forwards(!) progress to its children.
For example, `--modeling.progress=[QF=obe,default=unfair]` means that whenever a QF-group experiences fair progress, then its contained WGs will experience OBE, meaning that if the WG got scheduled once it will get rescheduled.
Notice, however, that this is only true as long as the QF itself experiences fair progress.
Consider a case with 1 DEV consisting of 2 QFs, each consisting of 2 WGs. WIth the above option, DEV will forward progress unfairly, meaning only one QFs is guaranteed to experience fair progress (due to minimal progress). For that QF, the WGs inside will experience OBE. However, for the other QF, the WGs will experience unfair scheduling, i.e., even if they were scheduled once, they might not get rescheduled (because their QF does not get rescheduled).
**In short:** if a group of subgroups/threads gets scheduled unfairly, then all of its subgroups/threads will get scheduled unfairly no matter what the progress model of the group and its subgroups is.

(2) There is always minimal progress unless all threads are blocked, meaning if something can run then something will run.
In the example of (1), if the fairly scheduled QF were to block, meaning all its contained subgroups/threads are blocked, then the unfairly scheduled QF would have to get scheduled. Since that QF then experiences fair scheduling (assuming the first QF remains blocked), the contained WGs will experience OBE.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
@@ -5,15 +5,16 @@ import com.dat3m.dartagnan.expression.Expression;
 import com.dat3m.dartagnan.expression.integers.IntCmpOp;
 import com.dat3m.dartagnan.expression.integers.IntLiteral;
 import com.dat3m.dartagnan.expression.type.IntegerType;
-import com.dat3m.dartagnan.program.Program;
-import com.dat3m.dartagnan.program.Register;
-import com.dat3m.dartagnan.program.ScopeHierarchy;
 import com.dat3m.dartagnan.program.Thread;
+import com.dat3m.dartagnan.program.*;
 import com.dat3m.dartagnan.program.analysis.BranchEquivalence;
 import com.dat3m.dartagnan.program.analysis.ExecutionAnalysis;
 import com.dat3m.dartagnan.program.analysis.ReachingDefinitionsAnalysis;
 import com.dat3m.dartagnan.program.event.*;
-import com.dat3m.dartagnan.program.event.core.*;
+import com.dat3m.dartagnan.program.event.core.CondJump;
+import com.dat3m.dartagnan.program.event.core.ControlBarrier;
+import com.dat3m.dartagnan.program.event.core.Label;
+import com.dat3m.dartagnan.program.event.core.NamedBarrier;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.dat3m.dartagnan.program.memory.Memory;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
@@ -112,10 +113,6 @@ public class ProgramEncoder implements Encoder {
 
     // ====================================== Control flow ======================================
 
-    private boolean isInitThread(Thread thread) {
-        return thread.getEntry().getSuccessor() instanceof Init;
-    }
-
     /*
         A thread is enabled if it has no creator or the corresponding ThreadCreate
         event was executed (and didn't fail spuriously).
@@ -180,17 +177,21 @@ public class ProgramEncoder implements Encoder {
     public BooleanFormula encodeControlFlow() {
         logger.info("Encoding program control flow with progress model {}", context.getTask().getProgressModel());
 
+        final Program program = context.getTask().getProgram();
         final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
         final ForwardProgressEncoder progressEncoder = new ForwardProgressEncoder();
         List<BooleanFormula> enc = new ArrayList<>();
-        for(Thread t : context.getTask().getProgram().getThreads()){
+        for(Thread t : program.getThreads()){
             enc.add(encodeConsistentThreadCF(t));
-            if (isInitThread(t)) {
-                // Init threads are always scheduled fairly
+            if (IRHelper.isInitThread(t)) {
+                // Init threads are always progressing
                 enc.add(progressEncoder.encodeFairForwardProgress(t));
             }
         }
-        enc.add(encodeForwardProgress(context.getTask().getProgram(), context.getTask().getProgressModel()));
+
+        // Actual forward progress
+        enc.add(progressEncoder.encodeForwardProgress(program, context.getTask().getProgressModel()));
+
         return bmgr.and(enc);
     }
 
@@ -343,16 +344,6 @@ public class ProgramEncoder implements Encoder {
         enc = bmgr.and(enc, imgr.equal(syncCount, imgr.sum(syncCountMembers)));
 
         return bmgr.and(enc, bmgr.implication(context.sync(e1), context.execution(e1)));
-    }
-
-    private BooleanFormula encodeForwardProgress(Program program, ProgressModel progressModel) {
-        final ForwardProgressEncoder progressEncoder = new ForwardProgressEncoder();
-        return switch (progressModel) {
-            case FAIR -> progressEncoder.encodeFairForwardProgress(program);
-            case HSA -> progressEncoder.encodeHSAForwardProgress(program);
-            case OBE -> progressEncoder.encodeOBEForwardProgress(program);
-            case UNFAIR -> progressEncoder.encodeUnfairForwardProgress(program);
-        };
     }
 
     // ====================================== Memory layout ======================================
@@ -517,7 +508,7 @@ public class ProgramEncoder implements Encoder {
     private class ForwardProgressEncoder {
 
         /*
-            Encodes fair forward progress for a single thread: the thread will eventually get scheduled (if it is enabled).
+            Encodes fair forward progress for a single thread: the thread will eventually get scheduled (if it is schedulable).
             In particular, if the thread is enabled then it will eventually execute.
          */
         private BooleanFormula encodeFairForwardProgress(Thread thread) {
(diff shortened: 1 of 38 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -545,99 +536,119 @@ private BooleanFormula encodeFairForwardProgress(Thread thread) {
...
-            must terminate (unless blocked).
-
-            NOTE: For producing correct verdicts on loop-based nontermination, we do not really require
-                  minimal progress guarantees.
-                  This may change in the presence of control barriers, depending on how they affect scheduling,
-                  i.e., do blocked threads remain eligible for scheduling?
-         */
-        private BooleanFormula encodeMinimalProgress(Program program) {
-            final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
+        // ================================================= New Encoding =================================================
```
hernanponcedeleon: Why "New"? Are you still keeping the legacy one? If this one is working, we should get rid of the other one

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -545,99 +536,119 @@ private BooleanFormula encodeFairForwardProgress(Thread thread) {
...
 
-            return bmgr.or(allThreadsTerminated, aThreadIsStuck);
+        private BooleanFormula isSchedulable(ThreadHierarchy threadHierarchy) {
+            return context.getBooleanFormulaManager().makeVariable("schedulable " + threadHierarchy.toString());
         }
 
-        // ---------------------------------------------------------------------------
+        private BooleanFormula wasScheduledOnce(ThreadHierarchy threadHierarchy) {
+            return context.getBooleanFormulaManager().makeVariable("wasScheduledOnce " + threadHierarchy.toString());
+        }
```
hernanponcedeleon: For consistency, either use `Var` prefix in all or none

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ExecutionGraphVisualizer.java
```diff
@@ -311,10 +305,11 @@ private String eventToNode(EventModel e) {
         }
         final String callStack = makeContextString(
             synContext.getContextInfo(e.getEvent()).getContextOfType(CallContext.class), " -> \\n");
-        final String nodeString = String.format("%s:T%s/E%s\\n%s%s\n%s",
+        final String nodeString = String.format("%s:T%s/E%s%s\\n%s%s\n%s",
```
hernanponcedeleon, line 308: I would rather put the scope hierarchy next to the thread and move the event id to the next line (together with the line of code where it comes from)

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
```diff
@@ -61,8 +61,14 @@ public boolean isImplied(Event start, Event implied) {
...
-        // weakest implication holds but not strongest: progress model decides
-        final boolean implication = switch (progressModel) {
+
+        if (!progressModel.isUniform()) {
+            // For mixed-hierarchy models, we only rely on strongest implication.
+            return strongestImplication; // FALSE
+        }
+
+        // weakest implication holds but not strongest & model is uniform: progress model decides
+        final boolean implication = switch (progressModel.getProgressAtScope("ALL")) {
```
hernanponcedeleon, line 71: Where is this "ALL" coming from? Is it supposed to be `Tag.OpenCL.ALL`? If so, this will not work with other models

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
```diff
@@ -101,7 +101,7 @@ private void reachingDefinitionMustOverride(ReachingDefinitionsAnalysis.Method m
         Configuration config = Configuration.builder().setOption(REACHING_DEFINITIONS_METHOD, method.name()).build();
         Context context = Context.create();
         context.register(BranchEquivalence.class, BranchEquivalence.fromConfig(program, config));
-        context.register(ExecutionAnalysis.class, ExecutionAnalysis.fromConfig(program, ProgressModel.FAIR, context, config));
+        context.register(ExecutionAnalysis.class, ExecutionAnalysis.fromConfig(program, ProgressModel.uniform(ProgressModel.FAIR), context, config));
```
hernanponcedeleon, line 104: Why not `ProgressModel::defaultHierarchy`?

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
```diff
@@ -651,7 +651,7 @@ private AliasAnalysis analyze(Program program, Alias method) throws InvalidConfi
         ProcessingManager.fromConfig(configuration).run(program);
         Context analysisContext = Context.create();
         analysisContext.register(BranchEquivalence.class, BranchEquivalence.fromConfig(program, configuration));
-        analysisContext.register(ExecutionAnalysis.class, ExecutionAnalysis.fromConfig(program, ProgressModel.FAIR, analysisContext, configuration));
+        analysisContext.register(ExecutionAnalysis.class, ExecutionAnalysis.fromConfig(program, ProgressModel.uniform(ProgressModel.FAIR), analysisContext, configuration));
```
hernanponcedeleon, line 654: Why not `ProgressModel::defaultHierarchy`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
ThomasHaas: No, this is in fact arbitrary. Since the hierarchy is uniform, all scopes have the same progress model. I just need to provide any one since I don't have a `getUniformProgress` method or similar.
Though I could see adding a `getDefaultProgressModel` method and using that one here.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
ThomasHaas: The original tests were explicitly stating FAIR. Since technically the default progress model might change in the future (though unlikely to ever happen), it is better to have the tests be independent of the default settings.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
ThomasHaas: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -180,17 +177,21 @@ private int getWorkgroupId(Thread thread) {
...
-                // Init threads are always scheduled fairly
+            if (IRHelper.isInitThread(t)) {
+                // Init threads are always progressing
                 enc.add(progressEncoder.encodeFairForwardProgress(t));
             }
         }
-        enc.add(encodeForwardProgress(context.getTask().getProgram(), context.getTask().getProgressModel()));
+
+        // Actual forward progress
+        enc.add(progressEncoder.encodeForwardProgress(program, context.getTask().getProgressModel()));
```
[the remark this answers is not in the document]
ThomasHaas: It depends on whether we want to put the `Hierarchy` suffix everywhere. I think the only place where I changed the name was in the `BaseOptions` class but I could also change it back. I don't mind either way.
So you prefer to put `...Hierarchy` everywhere?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
ThomasHaas: I added the `getDefaultProgressModel` method.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ExecutionGraphVisualizer.java
ThomasHaas: Done.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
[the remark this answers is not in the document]
ThomasHaas: I removed the `Hierarchy` suffix everywhere.

hernanponcedeleon:
The code LGTM.

I will add some unit test during the week and if there are no surprises, we can merge.

hernanponcedeleon:
> The comment in `mp-wg_obe-t_hsa` talks about `OBE/OBE` but the benchmark name suggests `OBE/HSA`. Is this intended?

Yes. 

The naming is related to what happens in the code: OBE mean we decide the id dynamically (via `atomic_add`) while HSA means we do it statically (using built-ins). Generally, deciding dynamically should make it fail for HSA while deciding statically should making it fail for OBE (and this happens in all other tests). 

Since that benchmark statically decides the id at the lowest level, I was expecting `SQ=OBE` to cause problems. The comment explains why this is not the case.

ThomasHaas:
> @ThomasHaas can you take a look to the examples I added and double check the expected results match your expectations?
> 
> I think the only "non-obvious" result is for `mp-wg_obe-t_hsa`.

All the results make sense to me.

hernanponcedeleon:
I had to manually fix some conflicts after the rebase. @ThomasHaas can you please double check nothing got mess-up in the process?

ThomasHaas:
> I had to manually fix some conflicts after the rebase. @ThomasHaas can you please double check nothing got mess-up in the process?

LGTM

PR #850, opened 2025-04-11
Progress64 benchmarks

StefanoDalMas:
This PR adds benchmarks and tests from Progress64 by ARM. 
At the moment only the clients which do not use mixed size concurrency instructions are inserted.


benchmarks/progress64/rwlock.c
```diff
@@ -0,0 +1,34 @@
...
+
+void *run(void *arg) {
+    p64_rwlock_acquire_wr(&lock);
+    x++;
+    y++;
+    p64_rwlock_release_wr(&lock);
+    return NULL;
+}
+
+int main() {
```
hernanponcedeleon, line 16: This is not a good test for a read-write lock. Take a look to [this](https://github.com/open-s4c/libvsync/blob/f6391a7181258a7a797ee6f9b2ef27a46496a9cc/test/include/test/boilerplate/reader_writer.h)

benchmarks/progress64/README.md
```diff
@@ -0,0 +1,21 @@
+# How to generate LLVM files with Inline Asm with Progress64
```
hernanponcedeleon, line 1: @ThomasHaas can you try this on your arm machine to check the documentation is correct?

benchmarks/progress64/rwlock.c
ThomasHaas: It doesn't even comile, no? Where are `lock, x, y` defined?

benchmarks/progress64/README.md
ThomasHaas: > @ThomasHaas can you try this on your arm machine to check the documentation is correct?

`rwlock.c` doesn't compile because it misses definitions for the shared variables.

For `seqlock.c` I get this error:
```
/Users/thomashaas/ExternalTools/progress64/src/arch/ldxstx.h:223:24: error: controlling expression type 'uintptr_t *' (aka 'unsigned long *') not compatible with any generic association type
    return (void *)ldx((uintptr_t *)var, mm);
                       ^~~~~~~~~~~~~~~~
/Users/thomashaas/ExternalTools/progress64/src/arch/ldxstx.h:203:11: note: expanded from macro 'ldx'
_Generic((var), \
          ^~~
/Users/thomashaas/ExternalTools/progress64/src/arch/ldxstx.h:227:16: error: controlling expression type 'uintptr_t *' (aka 'unsigned long *') not compatible with any generic association type
    return stx((uintptr_t *)var, (uintptr_t)val, mm);
```
And a bunch of warnings that the assembly uses wrong register sizes:
```
/Users/thomashaas/ExternalTools/progress64/src/arch/ldxstx.h:195:27: warning: value size does not match register size specified by the constraint and modifier [-Wasm-operand-widths]
                   : "r" (neu), "r" (var)
                          ^
/Users/thomashaas/ExternalTools/progress64/src/arch/ldxstx.h:193:35: note: use constraint modifier "w"
    __asm volatile("stxp %w0, %1, %H1, [%2]"
                                  ^~~
                                  %wH1
```

These are all errors/warnings inside `progress64`, so I don't know how much we can do about it.

benchmarks/progress64/rwlock.c
StefanoDalMas: ah! I think I added the definition of those two in the src file, adding that now

benchmarks/progress64/README.md
StefanoDalMas: as Clang complains at times when GCC does not , in order to make that work I had to tweak some lines in the file which contains the inline asm

hernanponcedeleon: This is a huge red flag. We should keep changes to a minimal and if changes are indeed needed we need to
- provide a patch so someone else can regenerate the llvm files if needed
- document why the change is indeed needed

PR #851, opened 2025-04-11
Add debug mode to dartagnan script. Off by default

hernanponcedeleon:
@db7 can you test if this matches what you told me a while ago you wanted to have?


db7:
works perfectly! very nice.

PR #853, opened 2025-04-13
Use proper exit codes

hernanponcedeleon:
Vsyncer uses the string we print in the console to detect the type of violation found (`program_spec`, `cat_spec`, `termination`). This was always fragile, and it is completely broken since we rename `liveness` to `termination`.

This PR fixes this by making dartagnan return different exit codes.

I already have a prototype fix for vsyncer, but I will keep this PR as draft until I am sure that it works correctly with libvsync CI.


hernanponcedeleon:
libvsync CI passed, so this PR is ready to be reviewed.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -231,23 +233,27 @@ public static void main(String[] args) throws Exception {
...
                 System.out.println("Total verification time: " + Utils.toTimeString(endTime - startTime));
 
                 // We only generate witnesses if we are not validating one.
                 if (o.getWitnessType().equals(GRAPHML) && !o.runValidator()) {
-                    generateWitnessIfAble(task, prover, modelChecker, summary);
+                    generateWitnessIfAble(task, prover, modelChecker, summary.toString());
+                }
+                int exitCode = summary.code().asInt();
+                if(exitCode > 0) {
+                    System.exit(exitCode);
```
ThomasHaas, line 246: I would add a method `ExitCode.isError()` or even `ResultSummary.isError()` rather than checking the integer value directly.

Also, why is this check even necessary here? Isn't `System.exit(0)` the default if the main function terminates?

hernanponcedeleon: > I would add a method `ExitCode.isError()` or even `ResultSummary.isError()` rather than checking the integer value directly.

Ok

> 
> Also, why is this check even necessary here? Isn't `System.exit(0)` the default if the main function terminates?

The `java` cmd would have exit value 0, but this is not the same as explicitly calling `System.exit(0)`. Explicitly calling this causes issues with `ApplicationTest` since this terminates the JVM and then the unit tests crash. In principle, there should be ways to mock the `System` class to avoid this using `PowerMockito` (mockito alone is not sufficient since we need to mock a static method), but I spent a whole day trying all documentation I found online and nothing worked. 

Thus I went with the option of letting the JVM finish on its own.

ThomasHaas: Hmm, so it is basically an issue related to the fact that you cannot unit test code containing `System.exit()` calls.
It does feel awkward to unit test `main` directly though, especially if you can only unit test the succesful path. Anyways, can you put a comment there that this code is a workaround to unit test `main` (at least successful paths of main)?

hernanponcedeleon: After giving this another though, I would rather remove the check and get rid of `ApplicationTest`. I don't think that class brings much value in terms of testing and keeping it forces us to need workarounds like this one.

ThomasHaas (approved):
LGTM

PR #854, opened 2025-04-15
Always include DAT3M_HOME/include when compiling code

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Compilation.java
```diff
@@ -21,8 +22,7 @@ public static File compileWithClang(File file, String cflags) throws Exception {
         final String outputFileName = getOutputName(file, ".ll");
 
         ArrayList<String> cmd = new ArrayList<>(
-                asList("clang", "-Xclang", "-disable-O0-optnone", "-S",
-                        "-emit-llvm", "-g", "-gcolumn-info", "-o", outputFileName));
+                asList("clang", "-I", getHomeDirectory() + "/include", "-Xclang", "-disable-O0-optnone", "-S", "-emit-llvm", "-g", "-gcolumn-info", "-o", outputFileName));
```
ThomasHaas, line 25: Won't this generate `-I <path>` rather than `-I<path>` (i.e., a difference in spacing)? I don't even know if this matters, but usually you don't have the space, right?

PR #855, opened 2025-04-16
Add OpenCL ThreadGrid

tonghaining:


ThomasHaas:
Recently when adding support for hierarchical scheduling in #848, I also had the feeling that the way scope hierarchies, thread grids, etc. work are kinda messy.
I'm not sure how much of this is related to this PR, but let me give my two cents and you can decide if there is anything valuable in there.

I think a `Program` should have a `ScopeHierarchy` that tells us which scopes do actually exist. This is basically a `List<String>` where the order of the list gives the hierarchy and the entries are the names of the scope levels.
A `ThreadGrid` is then just a mapping of scopes to a number of threads `Map<String, Integer>` or more simpler a `List<Integer>` whose size matches with the size of the scope hierarchy. It's only purpose is to act as a generalized entry point for the program.
Lastly, each `Thread` has a thread id that is a vector describing its position in the hierarchy, i.e., also a `List<Integer>` with one entry per scope level of the `Program`'s `ScopeHierarchy` (ignoring init threads for now, because they conceptually live in an independent scope hierarchy).
Actually, I would assume there always exist (possibly implicitly) at least two scopes, namely `GLOBAL` and `THREAD`, even for programs that are independent of GPUs. GPU scopes would just sit inbetween those two "natural" scopes.
On one hand, this would get rid of the ugly `Optional<ScopeHierarchy>` members, because we can make them non-optional.
One the other hand, it would allow us to generalize the whole concept of thread-local memory to scope-local memory where `scope=THREAD` means the `MemoryObject` is duplicated per thread (i.e., standard thread-local memory) and `scope=GLOBAL` means the `MemoryObject` is not duplicated at all (because all threads are part of the same global scope). It would naturally also work for all intermediate scopes and likely streamline the way `ThreadCreation` works (or in fact every kind of memory allocation). Again, rather than using the actual scope names, it might suffice to just give the scope level as an integer because at the and of the day, the names of scopes are (or should be) quite irrelevant in our IR.
Lastly, `ControlBarriers` could easily be extended to take the scope/scope level as parameter, generalizing them quite nicely. Of course, the SPIRV frontend would still generate `ControlBarriers` with the appropriate scope (e.g., WG for Vulkan IIRC), but at least our encoding would work uniformly without needing to check what architecture we are on.

ThomasHaas:
@tonghaining I see that you try to implement what I said on your fork, but I think you took my words too literally.
If I say that a `ThreadGrid` is basically a `List<Integer>` then I don't mean that we should use the raw `List<Integer>` directly.
It is still a good idea to have proper wrapper types like for example `public record ThreadId(List<Integer> ids)` or `public record ThreadGrid(List<Integer> sizes)`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeHierarchy.java
```diff
@@ -0,0 +1,34 @@
...
+import com.dat3m.dartagnan.configuration.Arch;
+import com.dat3m.dartagnan.program.event.Tag;
+
+import java.util.List;
+
+public class ScopeHierarchy {
+
+    private final List<String> scopes;
+
+    public ScopeHierarchy(Arch arch) {
```
ThomasHaas, line 12: `ScopeHierarchy` should be more general: you should have a constructor that takes an arbitrary list of scopes.
You can still keep this constructor (though I prefer a static factory method like `ScopeHierarchy.fromArch(Arch)`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeIds.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ScopeIds {
+
+    private final ScopeHierarchy scopeHierarchy;
+    private final List<Integer> ids;
+
+    public ScopeIds(Arch arch, List<Integer> ids) {
+        this.scopeHierarchy = new ScopeHierarchy(arch);
```
ThomasHaas, line 15: Taking the `Arch` and constructing a new `ScopeHierarchy` instance is not nice and unnecessarily specific.
The `ScopeHierarchy` should be part of the arguments.

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeSizes.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collections;
+import java.util.List;
+
+public class ScopeSizes {
+
+    private final ScopeHierarchy scopeHierarchy;
+    private final List<Integer> sizes;
+
+    public ScopeSizes(Arch arch, List<Integer> sizes) {
+        this.scopeHierarchy = new ScopeHierarchy(arch);
```
ThomasHaas, line 15: See comment about `ScopeIds`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeSizes.java
```diff
@@ -0,0 +1,54 @@
...
+        if (sizes.size() != scopeHierarchy.getScopes().size()) {
+            throw new IllegalArgumentException("Scope sizes must be of length " + scopeHierarchy.getScopes().size());
+        }
+        this.sizes = new ArrayList<>(sizes);
+        Collections.reverse(this.sizes); // Reverse the list to match the order of scopes in the hierarchy (top-down)
```
ThomasHaas, line 20: Why does the caller give a wrongly ordered list in the first place?

ThomasHaas (commented):
Can you revert the change where you moved `Thread` into a new package?
It results in annoyingly many changes in the diff.
I would do this kind of refactor either in a separate PR or as a last change in this PR (after the rest has been approved).

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeSizes.java
tonghaining: It was that in all existing SPIR-V test, the thread grid configuration is bottom-up e.g. from subgroup to device.

tonghaining: Yeah that makes sense—I'll move it to the parser.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/extenstions/VisitorExtensionClspvReflection.java
```diff
@@ -118,11 +118,12 @@ private Void setPushConstantValue(String argument, String offsetId, String sizeI
     }
 
     private List<Integer> computePushConstantValue(String command) {
-        ThreadGrid grid = builder.getThreadGrid();
+        ScopeSizes grid = builder.getScopeSizes();
         return switch (command) {
-            case "PushConstantGlobalSize" -> List.of(grid.dvSize(), 1, 1);
-            case "PushConstantEnqueuedLocalSize" -> List.of(grid.wgSize(), 1, 1);
-            case "PushConstantNumWorkgroups" -> List.of(grid.qfSize() / grid.wgSize(), 1, 1);
+            case "PushConstantGlobalSize" -> List.of(grid.getSize(Tag.Vulkan.DEVICE), 1, 1);
```
natgavrilenko, line 123: ScopeSize is generic, it shouldn't use Vulkan tags. And in general, no Vulkan tags should be used before compilation.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
```diff
@@ -73,21 +83,44 @@ public Expression getDecoration(String id, Type type) {
...
-            case "DeviceIndex" -> makeScalar(id, type, 0);
-            case "SubgroupId" -> makeScalar(id, type, grid.sgId(tid));
-            case "WorkgroupId" -> makeArray(id, type, grid.wgId(tid), 0, 0);
-            case "SubgroupSize" -> makeScalar(id, type, grid.sgSize());
-            case "WorkgroupSize" -> makeArray(id, type, grid.wgSize(), 1, 1);
-            case "GlobalSize" -> makeArray(id, type, grid.dvSize(), 1, 1);
-            case "NumWorkgroups" -> makeArray(id, type, grid.dvSize() / grid.wgSize(), 1, 1);
-            default -> throw new ParsingException("Unsupported decoration '%s'", mapping.get(id));
-        };
+        if (arch == Arch.VULKAN) {
```
natgavrilenko, line 86: Same here, no arch-specific tags before compilation. Grid/Scope size are the same for all architectures, just the names are different.

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeHierarchy.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeIds.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/thread/ScopeSizes.java
tonghaining: Updated

ThomasHaas:
> Overall, we might not need ScopeHierachy at all. The grid is conceptually the same for all cases:
> 
>     * thread
> 
>     * subgroup / warp / cta
> 
>     * workgroup
> 
>     * queue family (only for vulkan)
> 
>     * device.
> 
> 
> Why not to use the same thread grid directly?

While I agree that most concepts like `ThreadGrid` and position of a thread in a hierarchy are describable purely in terms of scope levels (this is also what I advocate for), I still think you need an object that describes the set of available levels. If you also attach names to those levels, you get precisely `ScopeHierarchy`.
Letting `ThreadGrid` implicitly define those levels doesn't seem like a good idea to me. I can easily imagine programs that have scopes but no notion of thread grids. I can also imagine programs that have multiple scope hierarchies(*) because, say, parts of the program run on different devices.
To me, a `ThreadGrid` is a generalized entry point(**) for the program: it should contain a function and a number of threads per scope level that execute that function (these can be independent of a `ScopeHierarchy`).
`ThreadCreation` should spawn threads according to the `ThreadGrid` and place them inside a `ScopeHierarchy` that is either already explicitly given as part of the `ThreadGrid` or implicitly given by the target `Arch`.
In the latter case, the `ScopeHierarchy` would only get created during `ThreadCreation` and doesn't need to exist beforehand. In the former case, the parser would already define the `ScopeHierarchy`.

(*) Actually, in this case I would probably generalize `ScopeHierarchy` to be a tree rather than a list. Then every program has precisely one `ScopeHierarchy` and every thread lives in a leaf of that tree. A thread position (or generalized id) is then a mapping from scopes along the unique path to ids.

(**) Even for LLVM code, we could have a `ThreadGrid` describing a single thread for the main function. If we also generalize `ScopeHierarchy` to be non-optional, we can streamline the whole thread creation. For non-GPU code we would just use a trivial `ScopeHierarchy`. 
I don't think this needs to be done in this PR, but this is something that I will eventually do if I find the time.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/transformers/MemoryTransformer.java
```diff
@@ -45,48 +45,13 @@ public MemoryTransformer(ScopeSizes grid, Function function, BuiltIn builtIn, Se
...
-                    i -> i / grid.getSize(Tag.OpenCL.ALL));
-        }
-        throw new UnsupportedOperationException("Thread grid not supported for architecture: " + program.getArch());
+        this.scopeIdProvider = List.of(grid::thId, grid::sgId, grid::wgId, grid::qfId, grid::dvId);
+        this.namePrefixIdxProvider = List.of(
+                i -> i,
+                i -> i / grid.sgSize(),
+                i -> i / grid.wgSize(),
+                i -> i / grid.qfSize(),
+                i -> i / grid.dvSize());
```
ThomasHaas, line 54: Can't you just use the levels `3, 2, 1, 0` instead? Then the whole `ScopeSize` subclasses become obsolete.
And if you cannot rely on levels for some reason, your previous solution that checked the individual architectures was still better than the current one.

Btw. how does `ScopeSize` differ from `ThreadGrid`? Did you remove `ThreadGrid`?

tonghaining:
I'm feeling that the arch-specific computations in `threadGrid`, `BuiltIn`, and `MemoryTransformer` are from the way SPIR-V annotations define memory scopes: `subgroup, workgroup, queuefamily` for Vulkan, and `subgroup, workgroup, device` for OpenCL. To generalize this, I’m thinking of redefining the annotations using a unified hierarchy — for example: `thread`, `subgroup/cta/warp`, `workgroup`, `queuefamily/cluster`, and `device`. This would allow threadGrid to be mapped consistently across architectures.

natgavrilenko:
> I'm feeling that the arch-specific computations in `threadGrid`, `BuiltIn`, and `MemoryTransformer` are from the way SPIR-V annotations define memory scopes: `subgroup, workgroup, queuefamily` for Vulkan, and `subgroup, workgroup, device` for OpenCL. To generalize this, I’m thinking of redefining the annotations using a unified hierarchy — for example: `thread`, `subgroup/cta/warp`, `workgroup`, `queuefamily/cluster`, and `device`. This would allow threadGrid to be mapped consistently across architectures.

I totally agree that spirv should use a generic hierarchy, and only after compilation we can use arch-specific names is we really want (maybe we don't). But a small correction. In general, spirv can have the following scopes:
- Invocation (single thread)
- Subgroup (aka warp, aka cta)
- Workgroup
- QueueFamily (only Vulkan, other should always have a single queue family)
- Device (aka gpu)
- CrossDevice (aka cluster)

ThomasHaas:
The `ScopeHierarchy` needs to use the correct scope names for each architecture. Not because we internally care about them, but because the user cares about them when interacting with the tool. For example, when specifying hierarchical scheduling, he specifies the scopes by name. Similarly, if we print an execution witness, we print the scope names attached to the threads. If we eventually extend CAT with a construct like `let swg = sameScope("wg")`, we also need the scope names.
Therefore, we need to preserve the scope names provided by the frontend/the architecture.

We should just avoid making special case distinctions on those scopes internally, or at least reduce the special casing to as few places as possible.
If SPIRV already uses different scopes on the language level, I think the parser should probably preserve them, but the compilation pass might be allowed to change them.
So basically only the parser and/or the compilation pass should really be the places where reasoning about the concrete scopes is happening. I would consider the `MemoryTransformer` to be part of the parser (since it is defined at parsing time), so it also allowed to do so (though ideally I would get rid of this class entirely).

ThomasHaas:
It seems we went full circle and are back to the original code more or less.
That's fine with me. We can do generalizations in a different PR.

tonghaining:
Yeah, I'm feeling the same.. The threadGrid's hierarchy seems to be static in SPIR-V, and since we're only using SPIR-V on single device, I think this should work fine for now. But thanks a lot for the help!

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/decorations/BuiltIn.java
```diff
@@ -78,14 +78,14 @@ private Expression getDecorationExpressions(String id, Type type) {
             case "SubgroupLocalInvocationId" -> makeScalar(id, type, tid % grid.sgSize());
             case "LocalInvocationId" -> makeArray(id, type, tid % grid.wgSize(), 0, 0);
             case "LocalInvocationIndex" -> makeScalar(id, type, tid % grid.wgSize()); // scalar of LocalInvocationId
-            case "GlobalInvocationId" -> makeArray(id, type, tid % grid.dvSize(), 0, 0);
-            case "DeviceIndex" -> makeScalar(id, type, 0);
+            case "GlobalInvocationId" -> makeArray(id, type, tid % grid.cdvSize(), 0, 0);
```
natgavrilenko, line 81: Are you sure that global id and global size are computed across devices and not within a single device? Since it is a 3D value, I would expect then within a single device, but please check the documentation.

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
```diff
@@ -6,36 +6,42 @@
 
 public class ThreadGrid {
 
-    private final int sg;
-    private final int wg;
-    private final int qf;
-    private final int dv;
+    private final int th; // Number of threads per subgroup
```
natgavrilenko, line 9: Can we have thCount, .. dvCount? Then no comments needed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
```diff
@@ -6,36 +6,42 @@
...
-        return sg * wg * qf;
+        return th * sg * wg;
     }
 
     public int dvSize() {
-        return sg * wg * qf * dv;
+        return th * sg * wg * qf;
+    }
+
+    public int cdvSize() { // Number of cross-device threads
```
natgavrilenko, line 43: Maybe clSize (for clusterSize)?

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
```diff
@@ -55,10 +61,6 @@ public int qfId(int tid) {
     }
 
     public int dvId(int tid) {
-        return tid / dvSize();
-    }
-
-    public ScopeHierarchy getScoreHierarchy(int tid) {
-        return ScopeHierarchy.ScopeHierarchyForVulkan(qfId(tid), wgId(tid), sgId(tid));
+        return (tid / cdvSize()) / dvSize();
```
natgavrilenko, line 64: dvSize can be passed to the constructor. For cdvSize, we should throw exception if it is not 1.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
```diff
@@ -529,10 +530,9 @@ public static String toOpenCLTag(String tag) {
                 case SEQ_CST -> C11.MO_SC;
 
                 // Scope
-                // TODO: OpenCL Kernel supports sub_group, but it's not mentioned in the model
```
natgavrilenko, line 532: Shouldn't we then throw exception for subgroup?

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
tonghaining: Renamed

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
tonghaining: The name cluster seems to be used as `A partition of invocations in a subgroup` in their documentation.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
tonghaining: Updated

dartagnan/src/main/java/com/dat3m/dartagnan/program/ThreadGrid.java
natgavrilenko: How about system?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
natgavrilenko: Good. It is the same exception as for QUEUE_FAMILY and SHADER_CALL, so they can be merged together

natgavrilenko (approved):

PR #856, opened 2025-04-16
Add support for Barrier Inlining

tonghaining:


benchmarks/opencl/barrier/inlining/barrier-inlining-1.cl
```diff
@@ -0,0 +1,17 @@
+// clang -x cl -cl-std=CL2.0 -target spir-unknown-unknown -cl-opt-disable -emit-llvm -c barrier-inlining-1.cl -o a.bc
```
natgavrilenko, line 1: Nice tests! I see you moved loop benchmarks under "barrier". Could you restructure spirv loop tests in the same way, so that directory tree is the same for opencl benchmarks and the corresponding spirv tests?

Otherwise, looks good to me.

tonghaining: Thanks, do you mean that we should add the inlining+loop tests for both Vulkan and OpenCL? Or I just add an extra barrier folder layer to the spirv/vulkan and spirv/opencl tests?

natgavrilenko: I think so far all loop tests are vulkan and all inlining tests are opencl, but this is fine. So the directories would be like:
spirv/vulkan/barrier/loop/,,,
spirv/opencl/barrier/inlining/...

natgavrilenko (approved):

PR #859, opened 2025-04-18
Deprecate *.spv.dis. Use *.spvasm instead

hernanponcedeleon:
The [convention seems to be](https://github.com/KhronosGroup/SPIRV-Tools?tab=readme-ov-file#command-line-tools) using `*.spvasm` as the SPIR-V assembly file extension.


ThomasHaas:
The link is broken

hernanponcedeleon:
> The link is broken

Fixed

dartagnan/src/main/resources/warn.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
natgavrilenko, line 1: Why do we need a new file instead of using log4j2.xml?

hernanponcedeleon: File `log4j2.xml` was replaced in #851. We now have different files for the different debug levels and which level to use is handled by `scripts/dartagnan`. However, when testing this PR, I realized the warning level should be the default.

natgavrilenko: Ohh, I see there are now two almost identical log configuration files. The conventional way would be to have a single config file and change log level something like Configurator.setRootLevel(Level.getLevel(System.getProperty("logLevel"))).
Then the desired log level can be passes as a property, e.g. "-DlogLevel=OFF".

But anyway this is not related to the filename extensions, those look fine to me.

PR #860, opened 2025-04-22
Fix parsing of default value for progress models

ThomasHaas:


dartagnan/src/main/java/com/dat3m/dartagnan/configuration/ProgressModel.java
```diff
@@ -115,8 +115,9 @@ public String toString() {
                             .withKeyValueSeparator("=")
                             .split(value);
                     final Map<String, ProgressModel> parsed = Maps.transformValues(split, ProgressModel::valueOf);
-                    final ProgressModel defaultProg = parsed.getOrDefault("default", ProgressModel.getDefault());
-                    final Map<String, ProgressModel> result = ImmutableMap.copyOf(Maps.filterKeys(parsed, k -> !k.equals("default")));
+                    final String defaultOpt = "DEFAULT";
```
hernanponcedeleon, line 118: Where is the all caps coming from? I could not find the source in any of your code, so I assume it comes from the `TypeConverter/Congifuration` in `sosylab.common`. If that is the case we should add a comment to document the limitation is not on our side.

hernanponcedeleon: what do you mean by "other cases"? 

BTW, I noticed that the current implementation does not even require to precisely match to scopes tags, i.e. no caps work too. This is already better than I originally thought.

ThomasHaas: > what do you mean by "other cases"?

I mean that the converter could also work for lower case "default" or even mixed case "deFAuLT" in case the option annotation has not set `UpperCase=true`.

> BTW, I noticed that the current implementation does not even require to precisely match to scopes tags, i.e. no caps work too. This is already better than I originally thought.

This has nothing to do with the converter. It is the upper case option that we use everywhere.
The same is true for, e.g., `--property=...` and `--method=...` where you can also put lower/mixed/upper case.

hernanponcedeleon (approved):

PR #864, opened 2025-04-25
OpCompositeConstruct and vector and aggregate values on memory operations

hernanponcedeleon:


ThomasHaas:
Just FYI, the support we have for loading/storing aggregate values is limited.
Concretely, if you have load `r` of type A and a store `w` of type B with either of them being an aggregate, and the alias analysis cannot exclude that they won't alias, then our encoding will try to encode an rf-edge `rf(w, r)` and therefore also `value(w) == value(r)`.
This last part will result in an exception, because we cannot convert type B to type A.
If both A and B are aggregate types of the same shape but with different layout (say one of them uses non-default offsets), then the encoding won't throw an exception but is semantically incorrect (assuming the rf-edge can actually happen, but this is likely UB or at least a form of MSA).

dartagnan/src/test/resources/spirv/vulkan/basic/v3.spvasm
```diff
@@ -0,0 +1,161 @@
...
+%gl_GlobalInvocationID = OpVariable %_ptr_Input_v3uint Input
+         %13 = OpVariable %_ptr_Private_v3uint Private %gl_WorkGroupSize
+         %17 = OpVariable %_ptr_StorageBuffer__struct_15 StorageBuffer
+         %21 = OpVariable %_ptr_StorageBuffer__struct_19 StorageBuffer
+         %22 = OpVariable %_ptr_StorageBuffer__struct_19 StorageBuffer
+         %23 = OpVariable %_ptr_StorageBuffer__struct_19 StorageBuffer
+     %uint_5 = OpConstant %uint 5
+         %26 = OpFunction %void None %25
+         %27 = OpLabel
+               OpNoLine
```
natgavrilenko, line 88: Why are there all these OpNoLine? Can we remove them?

dartagnan/src/test/java/com/dat3m/dartagnan/spirv/vulkan/basic/SpirvAssertionsTest.java
```diff
@@ -118,7 +118,10 @@ public static Iterable<Object[]> data() throws IOException {
                 {"push-constant-mixed.spvasm", 1, PASS},
                 {"bitwise-scalar.spvasm", 1, PASS},
                 {"bitwise-vector.spvasm", 1, PASS},
-                {"idx-overflow.spvasm", 1, PASS}
+                {"idx-overflow.spvasm", 1, PASS},
+                {"composite-construct.spvasm", 1, PASS},
```
natgavrilenko, line 122: Can you move composite-construct next to the other composite tests?
For v3 and monoid, maybe we can have more descriptive names, e.g. read-write-struct and read-write-vector.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -38,6 +41,32 @@ public Void visitOpCompositeExtract(SpirvParser.OpCompositeExtractContext ctx) {
...
+        if (type instanceof AggregateType aggregateType) {
+            final List<Expression> elements = new ArrayList<>(aggregateType.getFields().size());
+            for (SpirvParser.ConstituentsContext vCtx : ctx.constituents()) {
+                String idCtx = vCtx.idRef().getText();
+                elements.add(builder.getExpression(idCtx));
+            }
+            builder.addExpression(id, expressions.makeConstruct(aggregateType, elements));
+        }
+        if (type instanceof ArrayType arrayType) {
+            if (arrayType.getElementType() instanceof ArrayType) {
```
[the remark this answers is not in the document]
hernanponcedeleon: This is not checking if the array type matches the return type, it is actually checking if the object is a vector of vectors. As we discussed, the spec if not too clear about what should be done here, thus currently we do not support that case.

dartagnan/src/test/resources/spirv/vulkan/basic/v3.spvasm
hernanponcedeleon: Because I generated the SPIRV code with debug information. I just regenerate it again without it (since anyway the original file will be lost) and rename it.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -64,6 +93,9 @@ public Event visitOpLoad(SpirvParser.OpLoadContext ctx) {
         List<Event> events = new ArrayList<>();
         if (builder.getType(resultType) instanceof ArrayType arrayType) {
             Type elType = arrayType.getElementType();
+            if(!(elType instanceof BooleanType || elType instanceof IntegerType || elType instanceof FloatType)) {
```
ThomasHaas, line 96: This kind of type check happens in at least 3 places. I think you should refactor this into a `isScalar` function

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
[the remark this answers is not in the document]
natgavrilenko: Then we misunderstood each other. The standard says "a contiguous subset of the scalars consumed can be represented by a vector operand instead". I though that we agreed to support only the "normal" case where the types of elements match the members of the result types, but omit the case where we need to flatten a vector. And the "normal" should be also handled by expressions.makeArray()

But actually when checking this vector case, I noticed that we don't throw any exception, because makeArray doesn't check types of the elements.

@Test
    public void testCompositeConstructMismatchingTypeArray() {
        // given
        String input = "%result = OpCompositeConstruct %composite %member1 %member2";
        builder.mockIntType("%uint", 32);
        builder.mockVectorType("%composite", "%uint", 2);
        builder.mockConstant("%member1", "%uint", 1);
        builder.mockConstant("%member2", "%composite", List.of(2, 3));

        try {
            visit(input);
            fail("Should throw exception");
        } catch (Exception e) {
            assertEquals("Some error text", e.getMessage());
        }
    }

Also, we should throw error an error is result type is neither array or structure. But now there is no error for this either:

@Test
    public void testCompositeConstructResultNotComposite() {
        // given
        String input = "%result = OpCompositeConstruct %uint %member1 %member2";
        builder.mockIntType("%uint", 32);
        builder.mockConstant("%member1", "%uint", 1);
        builder.mockConstant("%member2", "%uint", 2);

        try {
            visit(input);
            fail("Should throw exception");
        } catch (Exception e) {
            assertEquals("Some error text", e.getMessage());
        }
    }

I think we should add unit tests for corner cases, they really catch errors.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -38,6 +41,40 @@ public Void visitOpCompositeExtract(SpirvParser.OpCompositeExtractContext ctx) {
...
+                throw new ParsingException("Unsupported OpCompositeConstruct for vector of vectors");
+            }
+            final List<Expression> elements = new ArrayList<>(arrayType.getNumElements());
+            for (SpirvParser.ConstituentsContext vCtx : ctx.constituents()) {
+                String idCtx = vCtx.idRef().getText();
+                elements.add(builder.getExpression(idCtx));
+            }
+            try {
```
natgavrilenko, line 69: Minor thing, could you make error text more similar to Extract and Insert. Also, Extract and Insert have try-catch in a different method. Both (same or different method) are file, but it would be nice to have them symmetric.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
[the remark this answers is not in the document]
hernanponcedeleon: Done

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -44,13 +44,42 @@ public VisitorOpsMemory(ProgramBuilder builder) {
     public Event visitOpStore(SpirvParser.OpStoreContext ctx) {
         Expression pointer = builder.getExpression(ctx.pointer().getText());
         Expression value = builder.getExpression(ctx.object().getText());
-        Event event = EventFactory.newStore(pointer, value);
+        List<Event> events = new ArrayList<>();
```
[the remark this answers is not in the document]
hernanponcedeleon: Done

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsCompositeTest.java
```diff
@@ -557,6 +558,100 @@ public void testVectorShuffle() {
...
+
+        // when
+        visit(input);
+
+        // then
+        ConstructExpr shuffle = (ConstructExpr) builder.getExpression("%result");
+        assertEquals(builder.getExpression("%bv32(1)"), shuffle.getOperands().get(0));
+        assertEquals(builder.getExpression("%bv32(2)"), shuffle.getOperands().get(1));
+        assertEquals(builder.getExpression("%bv32(3)"), shuffle.getOperands().get(2));
+    }
```
natgavrilenko, line 653: I would also add positive tests to construct array and struct without "flattening", and tests with RuntimeArray,

Minor thing, could you keep the convention of test cases order, first the positive OpCompositeContruct tests and then negative OpCompositeContruct tests. Now they are the other way around.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -110,15 +112,63 @@ private Expression getInsertion(Expression compositeExpr, Expression objectExpr,
         try {
             return expressions.makeInsert(compositeExpr, objectExpr, indexes);
         } catch (Exception e) {
-            throw new ParsingException("Element type mismatch or index out of bounds in OpCompositeInsert for '%s'", id);
+            throw new ParsingException(String.format("%s Offending id: '%s'", e.getMessage(), id));
+        }
+    }
+
+    @Override
```
natgavrilenko, line 119: Minor thing. Could you keep the order of methods, first public and then private.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -43,54 +47,103 @@ public VisitorOpsMemory(ProgramBuilder builder) {
     @Override
     public Event visitOpStore(SpirvParser.OpStoreContext ctx) {
         Expression pointer = builder.getExpression(ctx.pointer().getText());
-        Expression value = builder.getExpression(ctx.object().getText());
-        Event event = EventFactory.newStore(pointer, value);
+        String valueId = ctx.object().getText();
+        Expression value = builder.getExpression(valueId);
+        Type type = value.getType();
+        List<Event> events = visitMemoryAccess(valueId, type, pointer, (i, exp) -> {
```
natgavrilenko, line 53: Explicit return can be dropped:

List<Event> events = visitMemoryAccess(valueId, type, pointer, (i, exp) ->
                i == -1 ? EventFactory.newStore(exp, value) 
                        : EventFactory.newStore(exp, expressions.makeExtract(value, i)));

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -110,15 +112,63 @@ private Expression getInsertion(Expression compositeExpr, Expression objectExpr,
...
+                elements.add(builder.getExpression(idCtx));
+            }
+            builder.addExpression(id, getConstruct(type, elements, id));
+        }
+        if (type instanceof ArrayType arrayType) {
+            final List<Expression> elements = new ArrayList<>(arrayType.getNumElements());
+            for (SpirvParser.ConstituentsContext vCtx : ctx.constituents()) {
+                String idCtx = vCtx.idRef().getText();
+                Expression elem = builder.getExpression(idCtx);
+                if (elem.getType().equals(type)) {
```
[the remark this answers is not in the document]
natgavrilenko: You may want to take a look at TypeFactory.isStaticTypeOf(), to compare array types with known and unknown number of elements.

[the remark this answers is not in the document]
hernanponcedeleon: I am not convinced the spec allows arrays of unknown size for this op.
> Result Type must be a composite type,

and composite type allows vectors, but not arrays
> Composite: An aggregate, a matrix, or a vector.

My understanding from this
> Vector: An ordered homogeneous collection of two or more scalars. Vector sizes are quite restrictive and dependent on the execution model.

is that we should not allow vectors to have unkown size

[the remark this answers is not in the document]
natgavrilenko: I think it should be possible for a return type to be a runtime array. According to the spirv spec for OpCompositeConstruct "Result Type must be a composite type". The definition of composite "Composite: An aggregate, a matrix, or a vector.", the an aggregate is "Aggregate: A structure or an array".

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsCompositeTest.java
```diff
@@ -153,7 +154,7 @@ public void testCompositeExtractElementNotConstructExpr() {
             fail("Should throw exception");
         } catch (ParsingException e) {
             // then
-            assertEquals("Index out of bounds in OpCompositeExtract for '%extract'", e.getMessage());
+            assertEquals("Non-aggregate type bv32. Offending id: '%extract'", e.getMessage());
```
natgavrilenko, line 157: I think "out of bounds" was more informative, but I live with the new message too.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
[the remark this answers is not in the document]
hernanponcedeleon: You are right, I confused the spirv definition of aggregate with the one in our IR. Will fix this tomorrow

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -95,9 +90,9 @@ public Void visitOpVectorShuffle(SpirvParser.OpVectorShuffleContext ctx) {
         List<Expression> concat = new ArrayList<>();
         for (Integer index : components) {
             if (index >= 0 && index < s1) {
-                concat.add(v1.getOperands().get(index));
+                concat.add(v1 instanceof ConstructExpr ? v1.getOperands().get(index) : expressions.makeExtract(v1, index));
             } else if (index >= s1 && index < s1 + s2) {
-                concat.add(v2.getOperands().get(index - s1));
+                concat.add(v2 instanceof ConstructExpr ? v2.getOperands().get(index - s1) : expressions.makeExtract(v2, index - s1));
```
ThomasHaas, line 95: Why do you need to do the case distinction? Don't you trust the expression simplifier? :)

hernanponcedeleon: This has nothing to do with the expression simplifier. It was rather causing problems already at parsing time (the one I told you yesterday night).

My previous understanding of `InsertExpr.getOperand()` was wrong. Calling `insert([old_0,...], new_0).getOperands()` returns `[old_0,...,new_0]` and not `[new_0,...]`.

ThomasHaas: I understand that the previous code was wrong, but why don't you just do the extract at all times?
Now, both cases of the ternary operator are semantically identical, the former is just a step the simplifier would otherwise do.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsComposite.java
```diff
@@ -8,11 +8,14 @@
 import com.dat3m.dartagnan.parsers.SpirvBaseVisitor;
 import com.dat3m.dartagnan.parsers.SpirvParser;
 import com.dat3m.dartagnan.parsers.program.visitors.spirv.builders.ProgramBuilder;
+import com.dat3m.dartagnan.expression.aggregates.ConstructExpr;
```
natgavrilenko, line 11: Unused import

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsCompositeTest.java
```diff
@@ -2,6 +2,7 @@
 
 import com.dat3m.dartagnan.exception.ParsingException;
 import com.dat3m.dartagnan.expression.aggregates.ExtractExpr;
+import com.dat3m.dartagnan.expression.aggregates.ConstructExpr;
```
natgavrilenko, line 5: Unused import

natgavrilenko (approved):

PR #868, opened 2025-04-26
Add leading zeros when formatting milliseconds

tcherrou:
Formatting only seconds and milliseconds would produce an incorrect string due to missing leading zeros.

Example:  The line `return String.format("%d.%d secs", seconds, milliseconds);` would return the string `"0.18 secs"`  with the call `toTimeString(18)`  .

Fix: Force milliseconds to be at least 3 digits, padding with leading zeros when necessary.


ThomasHaas (approved):
LGTM

PR #869, opened 2025-04-26
Unifying SMT Formulas & Expressions: TypedFormula & more

ThomasHaas:
~This PR is WIP.~

**Issues:**
1. We replicate a lot of computation on SMT expressions that we could express with our IR expressions. This results in plenty of redundancies and possibly even inconsistencies.
2. Our IR Types are more expressive than SMT types and so we should express most/all computation using our IR Types.
Consider, for example,  two types `Agg1 {0: bv64, 8: bv64 }` and `Agg2 {0: bv64, 16: bv64 }` that only differ in their layout.
When translating values of those types into SMT, they end up looking identical because a `TupleFormula` loses layout information. Therefore, on the SMT level, instances of both types may compare equal although in our IR they would not!
Encoding the layout information somehow into the SMT formula (e.g., making padding bits explicit) seems like a possibility, but it could be unnecessarily inefficient and wouldn't really solve the underlying issue.
3. We are implicity assuming that there is a 1-to-1 correspondence between our IR Types and SMT Types. This is generally not the case as point (2) shows.
4. When evaluating a formula, we directly query the SMT Model and therefore again lose IR type information. Generally, we should avoid querying the SMT model directly.

**Solution:**
- Add a `TypedFormula` class that carries the type represented by the formula. This class is also an `Expression` which allows us to embed SMT formulas into our IR (at least during encoding time).
We can even wrap variables like `BooleanFormula context.execution(event)` into a `BooleanType` expression and perform computation over it. 
Then we should be able to avoid making explicit computations on SMT formulas (at least when encoding IR expressions/types) and ideally lose the IR type information only on the boundary to the SMT prover, i.e., right before asserting formulas.
The above points solve (1) and (2).
- `TypedFormula` also decouples the SMT type used for representing values from the Type they represent.
Indeed, we could encode every IR type into SMT bitvectors (or even SAT!) while still retaining the full power and type safety of our IR.
For example, an IR tuple `(bv64 a, bv64 b)` could be encoded into SMT formula `bv128 a:b` and it would still be distinguishable from an IR integer `bv128 c` encoded into the same SMT type.
More importantly, the size of the resulting encoding can be independent from the memory size of the represented type.
Consider a PointerType represented as two values `(bv64 base, bv64 offset)`. The encoding uses 128 bits but represents a 64 bit value. This is also important for correct semantics. Consider a 64-bit load that reads such a pointer, then it does read the value `base + offset` and not just the first 64 bits of the encoding (which would just be `base`).
This addresses point (3).
- `TypedValue` is the counterpart of `TypedFormula` that can represent IR-typed models of SMT formulas.
A new class `IREvaluator` is used to evaluate the SMT model. It simplifies the interaction with the SMT model and retains typing information, solving issue (4).

**Main changes:**
- Added `TypedFormula` and `TypedValue` (models of `TypedFormula`). `TypedFormula` implements `Expression` and therefore we can use it just like other expressions. This allows us to basically build SMT formulas using `ExpressionFactory`.
- Most encoders that performed computation over integers/bitvectors using the JavaSMT formula managers now instead use `ExpressionFactory`. E.g., `ControlBarriers` encoding does not use mathematical integers anymore and instead uses whatever encoding `IntegerType` has (which translates to integers or bitvectors on the SMT level, depending on chosen options).
- `ExpressionEncoder` and many `EncodingContext` methods return `TypedFormulas` (e.g. `context.address(MemoryObject)` is now a typed formula).
- Added `IREvaluator` to evaluate `TypedFormulas` (or any `Expression`) over an SMT model returning `TypedValues`.
Furthermore, it simplifies checking whether events were executed, jumps were taken, relation edges are present etc.
Basically, there is no need to manually inspect the SMT model directly anymore.

**Other changes:**
- `ExpressionEncoder` now is responsible for (almost) all encoding related to IR Types and IR Expressions and generates `TypedFormula`s.
- `EncodingContext` got cleaned up: it had a random assortment of functionality that made the class very convoluted.
- `FormulaManagerExt(ended)` is a new wrapper around `FormulaManager` that enhances JavaSMT's with custom functionality.
It now hosts the `TupleFormulaManager`. It provides `ifThenElse` and `evaluate` extensions that work over `TupleFormula` making it appear as if `TupleFormulas` are natively supported.
Similarly, `ModelExt` wraps a `Model` (actually an `Evaluator`) and extends the Model with evaluation of `TupleFormula`s.
- We apply fewer implicit conversion (maybe none at all?). For example,  `Assert` and `Assume` now require `BooleanType` expressions rather than any expression that is then implicitly cast to a boolean one during encoding.

**Possible TODOs:**
- `TypedValue` should probably replace all usages of `ValueModel`.
- The usage/encoding of `FinalMemoryValue` is a little awkward right now.


ThomasHaas:
I made some more changes that also streamline of we evaluate the SMT model over our IR.
I think this PR cleans up a lot of technical mess we had in our backend. There is surely more that can be done, but I don't wanna overbloat this PR.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/TypedFormula.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public ImmutableList<Expression> getOperands() {
+        return ImmutableList.of();
+    }
+
+    @Override
+    public ExpressionKind getKind() {
+        return () -> "FORMULA";
```
hernanponcedeleon, line 84: Why a string an not a member of `ExpressionKind.Other`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/TypedValue.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.common.base.Preconditions;
+
+/*
+    TODO: This is a temporary solution and code should not be too dependent on this class.
+     The issue is as follows: while TypedFormulas allow us to construct SMT encodings agnostic to the actual
+     SMT types used in the encoding, the corresponding TypedValues generated from an SMT model
+     exposes the underlying SMT types.
+     What we rather want to have are Literal expressions, e.g., rather
+     than TypedValue<IntegerType, BigInteger>, we should return IntLiteral.
+     Similarly, rather than TypedValue<BooleanType, Boolean>, we should return BoolLiteral, etc.
```
hernanponcedeleon, line 13: Why would we want the literal rather than the typed version? I understand that for literals the type information is obvious (because there is a unique SMT representation), but I guess the whole point is to have a unified interface

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -343,6 +344,10 @@ public Expression makeGeneralZero(Type type) {
     }
 
     public Expression makeCast(Expression expression, Type type, boolean signed) {
+        if (expression.getType().equals(type)) {
+            return expression;
+        }
+
```
hernanponcedeleon, line 350: This should probably be done by ExprSimplified` and not the factory (we should not have any magic in here)

dartagnan/src/main/java/com/dat3m/dartagnan/smt/FormulaManagerExt.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private final FormulaManager fmgr;
+    private final TupleFormulaManager tmgr;
+
+    public FormulaManagerExt(FormulaManager fmgr) {
+        this.fmgr = fmgr;
+        this.tmgr = new TupleFormulaManager(this);
+    }
+
+    public FormulaManager getUnderlyingFormulaManager() { return fmgr; }
```
hernanponcedeleon, line 25: All other methods (the one following) use the same method name as JavaSMT (i.e. without the "underlying"). I don't see why this one should be a special case.

dartagnan/src/main/java/com/dat3m/dartagnan/smt/ModelExt.java
```diff
@@ -0,0 +1,44 @@
...
+import java.math.BigInteger;
+
+public class ModelExt implements AutoCloseable {
+    private final Evaluator evaluator;
+
+    public ModelExt(Evaluator evaluator) {
+        this.evaluator = Preconditions.checkNotNull(evaluator);
+    }
+
+    public Evaluator getUnderlyingEvaluator() { return evaluator; }
```
hernanponcedeleon, line 15: I would remove "underlying"

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphml/WitnessBuilder.java
```diff
@@ -188,20 +189,20 @@ public WitnessGraph build() {
...
+                .filter(executedCEvents)
                 .toList();
         Map<Integer, List<Event>> map = new HashMap<>();
         for (Event e : execEvents) {
             // TODO improve this: these events correspond to return statements
             if (e instanceof Store store && store.getMemValue() instanceof BoolLiteral bVal && !bVal.getValue()) {
                 continue;
             }
-            BigInteger var = model.evaluate(context.clockVariable("hb", e));
+            BigInteger var = model.getSMTModel().evaluate(context.clockVariable("hb", e));
```
hernanponcedeleon, line 205: Create "IREvaluator.happensBeforeClock()" method similar to `IREvaluator..memoryOrderClock()`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -1,346 +1,529 @@
...
+        }
+
+        @Override
+        public TypedFormula<BooleanType, BooleanFormula> visitBoolBinaryExpression(BoolBinaryExpr bBin) {
+            final TypedFormula<BooleanType, BooleanFormula> lhs = encodeBooleanExpr(bBin.getLeft());
+            final TypedFormula<BooleanType, BooleanFormula> rhs = encodeBooleanExpr(bBin.getRight());
+            final BooleanFormula result = switch (bBin.getKind()) {
+                case AND -> bmgr.and(lhs.formula(), rhs.formula());
+                case OR -> bmgr.or(lhs.formula(), rhs.formula());
+                case IFF -> bmgr.equivalence(lhs.formula(), rhs.formula());
```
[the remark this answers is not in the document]
ThomasHaas: There is no default case because the switch is exhaustive. Without the default case, the program won't even compile if you extend the operation enum without handling the new case.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -1,346 +1,529 @@
...
+                            BitvectorFormula bvbit = bvmgr.extract(bv, i, i);
+                            cttz = fmgr.ifThenElse(bvmgr.equal(bvbit, bv1), bvi, cttz);
+                        }
+                        yield cttz;
                     }
-                    return enc;
-                }
+                };
+
+                return new TypedFormula<IntegerType, Formula>(iUn.getType(), result);
```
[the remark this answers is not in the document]
ThomasHaas: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/TypedFormula.java
ThomasHaas: It's technically not a string, but an anonymous class. I would rather get rid of `ExpressionKind.Other` and change all the use-cases to anonymous classes.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/TypedValue.java
ThomasHaas: The comment refers to literals in our IR, i.e., well-typed expressions. And we want to use them precisely because there is NO unique SMT representation even for literals: an IntLiteral may be represented by a BV formula or an Int formula and the user should not need to know this. TypedValue makes this SMT representation explicit, but since models of int formulas and bv formulas are both given by BigInteger, the fact that the SMT representation is leaked is not so obvious here.
However, suppose we encode aggregates as a single bit-string rather than a tuple, then we probably still want the IR model to make it look like it was a tuple all along.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
ThomasHaas: 1. I don't think this is magic, because the user cannot (and should not) rely on the syntactic representation of the returned expression (a reason why all methods should return `Expression` and nothing more specific really). For example, a cast to BooleanType is represented by an ITE expression and is not even a `CastExpression`.
2. ExprSimplifier cannot do this unless we introduce cast expressions for each and every type, in particular unnecessary identity casts like Ptr2Ptr. Therefore, I consider this check to not be an optimization (which ExprSimplifier can do) but rather a choice on how to represent the cast in the first place.

dartagnan/src/main/java/com/dat3m/dartagnan/smt/FormulaManagerExt.java
ThomasHaas: For two reasons. First, it IS a special case, because `FormulaManager` is not just another manager like `BooleanFormulaManager` or `IntegerFormulaManager`. `FormulaManager` is the container of those two.
Secondly, the whole point of `FormulaManagerExt` is to wrap `FormulaManager` in such a way that we should ideally never need to access the underlying `FormulaManager` at all. Therefore, giving its getter method a distinct name is deliberate.

dartagnan/src/main/java/com/dat3m/dartagnan/smt/ModelExt.java
ThomasHaas: Same argument as above.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphml/WitnessBuilder.java
ThomasHaas: No, IREvaluator should not have such specific functionality that only makes sense for one particular use-case related to SVCOMP. It could have a general `clockValue(Relation rel, Event e)` method though (but it's also a bit questionable).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/TypedFormula.java
hernanponcedeleon: Either way we need to be consistent about what we use

ThomasHaas: Then I will change the other usages.

ThomasHaas: Changing the other usages is was harder than it seemed, so I changed this particular case to follow the other's instead.

hernanponcedeleon (approved):

PR #871, opened 2025-04-29
Treat calls to unknown functions as assert(0)

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -952,8 +954,10 @@ private List<Event> inlineUnknownFunction(FunctionCall call) {
             replacement.addAll(inlineCallAsNonDet(call));
             opt = "Replacing return value by non-deterministic value. ";
         }
-        logger.warn(String.format("Found unknown function %s. %s" +
-            "Detecting calls to unknown functions requires --property=program_spec", funcName, opt));
+        if (unknown_functions.add(funcName)) {
+            logger.warn(String.format("Found unknown function %s. %s" +
+                "Detecting calls to unknown functions requires --property=program_spec", funcName, opt));
+        }
```
ThomasHaas, line 960: Why don't you do this in the function that marks the intrinsics, pretty much the way it was done before. But instead of throwing the exception, you print the set of functions? It generates a single print mentioning each function only once.

ThomasHaas (approved):
LGTM

PR #875, opened 2025-05-07
Fix direct data and addr dependency extraction for witness

CapZTr:
The methods for direct data and addr dependency extraction in `ExecutionModelManager` is problematic since in some cases it lacks of edges of these two relations and the derived ones based on them in the witness graph. The easiest fix is to delete these two methods and let them extracted by the fallback method. Now the graph shows identical relation information as what the solver has.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModelManager.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModelManager.java
@@ -455,56 +455,6 @@ public class ExecutionModelManager {
         }
 
         @Override
-        public Void visitAddressDependency(DirectAddressDependency addrDirect) {
-            SimpleGraph rg = (SimpleGraph) relGraphCache.get(addrDirect.getDefinedRelation());
-            for (ThreadModel tm : executionModel.getThreadModels()) {
-                Set<RegWriterModel> writes = new HashSet<>();
-                for (EventModel em : tm.getEventModels()) {
-                    if (em instanceof RegWriterModel rwm) {
-                        writes.add(rwm);
-                        continue;
-                    }
-                    if (em instanceof RegReaderModel rrm) {
-                        for (RegWriterModel write : writes) {
-                            for (Register.Read read : rrm.getRegisterReads()) {
-                                if (read.register() == write.getResultRegister()
-                                    && read.usageType() == Register.UsageType.ADDR) {
-                                    rg.add(new Edge(write.getId(), rrm.getId()));
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-            return null;
-        }
-
-        @Override
-        public Void visitInternalDataDependency(DirectDataDependency idd) {
-            SimpleGraph rg = (SimpleGraph) relGraphCache.get(idd.getDefinedRelation());
-            for (ThreadModel tm : executionModel.getThreadModels()) {
-                Set<RegWriterModel> writes = new HashSet<>();
-                for (EventModel em : tm.getEventModels()) {
-                    if (em instanceof RegWriterModel rwm) {
-                        writes.add(rwm);
-                        continue;
-                    }
-                    if (em instanceof RegReaderModel rrm) {
-                        for (RegWriterModel write : writes) {
-                            for (Register.Read read : rrm.getRegisterReads()) {
-                                if (read.register() == write.getResultRegister()
-                                    && read.usageType() == Register.UsageType.DATA) {
-                                    rg.add(new Edge(write.getId(), rrm.getId()));
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-            return null;
-        }
-
-        @Override
         public Void visitEmpty(Empty empty) {
             return null;
         }


dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModelManager.java
```diff
@@ -454,56 +454,6 @@ public Void visitControlDependency(DirectControlDependency ctrlDirect) {
             return null;
         }
 
-        @Override
```
natgavrilenko, line 457: This seems to work correctly, but why do we need this RelationGraphPopulator at all? Isn't it doing the same as relation analysis?

CapZTr: Yes, it basically works in the same way. But here we only care about the events that are supposed to be visualized in witness. And we need this visitor to extract and compute the relation information from the execution avoiding querying from the solver every time.

ThomasHaas: To be more precise: the code evaluates the CAT model on the single given execution (~ interpretation of base relations). Relation analysis evaluates the CAT model abstractly on ALL possible executions.

The major reason why the computation needs to be done is because the encoding is optimized and may not have variables that represent some derived edges. The extreme example is the refinement/lazy solver, which does not encode derived relations at all. But even for the eager solver, some edges might be inactive (outside of the encode set) and therefore are not encoded, but neither is their value statically known.

ThomasHaas: Dependencies are generally not static, unfortunately.
For some relations, you can reuse the must set by projecting it down to executed events.
However, what I personally dislike is that we rely too heavily on the exactness of a static analysis. It is sound to have `must(po)=emptyset` in principle (and we even have an option to get this behaviour).

I think there is no way around having an RA-independent computation somewhere. That being said, parts of this functionality are already implemented in `RefinementSolver/WMMSolver` and could be reused better.
But because this requires some adapations to `WMMSolver` that I didn't want @CapZTr to touch, so we ended up having partially duplicated code. If I ever find the time, I will streamline this code better.

natgavrilenko (approved):

dartagnan/src/main/java/com/dat3m/dartagnan/verification/model/ExecutionModelManager.java
ThomasHaas: I don't understand what you mean by "you need to check the may-sets anyways"? As I said, we should not rely on the precision of a static analysis here. It is totally valid to have `must(r) = empty` and `may(r) = full` and it should not in the slightest affect the generated witness. Similarly, we can have `encodeSet(r) = full` or even `encodeSet(r) = empty` and neither of those should affect the witness.

The only piece of computation you can really share is the propagation mechanism (the fixed point engine if you will), but the domain of computation is different for each use-case: concrete domain vs. abstract interval domain vs. abstract activeness domain.

natgavrilenko: I mean that a visualization graphs should show all edges, not only those in encode/active sets. May set is an over-approximation of all edges that can be present in a graph, even if they don't affect the final result. Therefore, a graph util should check for each edge in may set whether it is present in the candidate execution. In some cases, it can be done more efficiently (e.g., using must sets).

The problem with native alias analysis is that it modifies the original may sets when generating encode/active sets, and therefore there is no more good over-approximation of edges that we should draw. What we ideally should do is to draw each edge in must set and query the solver for each edge that is in a may set but not in the must set.

ThomasHaas: > I mean that a visualization graphs should show all edges, not only those in encode/active sets. 

I agree.

> May set is an over-approximation of all edges that can be present in a graph, even if they don't affect the final result. Therefore, a graph util should check for each edge in may set whether it is present in the candidate execution. In some cases, it can be done more efficiently (e.g., using must sets).

Instead of checking for every edge in the may-set, you can check for every edge (as if the may-set was full).
Apart from that, how do you check if an edge is present in the graph, if it was not encoded? An approximate answer is to require that the values of all base relations need to be known (either statically, or from the encoding). For derived relations, you can then do a recomputation from the base relations.

> The problem with native alias analysis is that it modifies the original may sets when generating encode/active sets, and therefore there is no more good over-approximation of edges that we should draw. 

The generation of active sets does not change the result of RA: this would be wrong. Are you saying this because it invokes parts of RA again?

> What we ideally should do is to draw each edge in must set and query the solver for each edge that is in a may set but not in the must set.

This is precisely what you cannot do because for edges inside the may-set (and outside the must-set) the solver may not have encoded what value it has. That is the whole point of the encode/active sets. 
That being said, we currently work under the assumption that at least for `rf` and `co` (and possibly some more relations), the encoding won't be missing any edges (Indeed, the encoding  of `rf/co` does ignore the active sets entirely). In general, this is not true for `po/ext/int` etc. in case you disable the must-analysis because then RA is too imprecise to know the edges statically but the active set computation is precise enough to avoid encoding them.
So to recover some relations, you cannot rely on may/must because those may be too imprecise and you cannot rely on encode/active sets because they don't give you the whole picture. Therefore, the only fundamental way to recover those edges must somehow avoid both RA and the encoding altogether (or at least query the SMT cleverly).
Fortunately, most relations are computable solely from the set of events in the execution graph and this is what is done here. I see no way around having this computation somewhere in Dartagnan (we shouldn't duplicate it though).

PR #879, opened 2025-05-08
LLVM Debug Records

xeren:
This PR modifies the LLVM grammar to recognize files with [Debug Records](https://llvm.org/docs/SourceLevelDebugging.html).  These entities replace the debug intrinsic calls generated by Clang 19+.

An alternative would be to execute `clang` with `--write-experimental-debuginfo=false`[[1]](https://discourse.llvm.org/t/psa-ir-output-changing-from-debug-intrinsics-to-debug-records/79578).

### DIFF
--- a/dartagnan/src/main/antlr4/LLVMIR.g4
+++ b/dartagnan/src/main/antlr4/LLVMIR.g4
@@ -115,6 +115,8 @@ instruction: // Instructions producing values.
 	// Instructions not producing values.
 	| storeInst
 	| fenceInst
+	// Debug Records, which are not actually instructions, but appear interleaved.
+	| debugRecord
 	)
 	(',' metadataAttachment)*;
 terminator:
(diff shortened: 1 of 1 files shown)


ThomasHaas (approved):
LGTM

hernanponcedeleon:
How do debug records relate to instructions? While it is nice that with this change we will stop failing parsing llvm files generated by newer clang versions (I still need to test if this is the case), it would also be nice to make use proper use of the debug information (we definitely need this for e.g. SVCOMP)

ThomasHaas:
Debug records replace debug instructions but appear in the LLVM at roughly the same place as those instructions used to.
We have never used metadata from debug instructions, so this PR is no regression in that regard.

hernanponcedeleon:
ok ... I will then merge after I test this on a few examples generated by newer clang versions

hernanponcedeleon:
I just installed clang19 on my system, tried a few of our benchmarks and all seems good.

PR #882, opened 2025-05-11
[DRAFT] Generalized ThreadHierarchy & removed ScopeHierarchy

ThomasHaas:


ThomasHaas:
**Description**: `ThreadHierarchy` is class that already existed and was used to define hierarchical scheduling. It was computed from the `ScopeHierarchy` associated to threads.
The idea now is to make `ThreadHierarchy` the canonical container for all threads in the program, i.e., `Program` directly manages a `ThreadHierarchy`. Whenever a `Thread` is added to the program, it will have to get inserted into the hierarchy.
The hierarchy has, by default, a root node with scope `GLOBAL` and a `INIT` node containing all init threads, even for CPU programs (LLVM/Litmus).

PR #885, opened 2025-05-13
Configure log level from console

hernanponcedeleon:
As proposed by @natgavrilenko in #859, this PR restores a single log4j2 file and allows to configure the different log levels from the console


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -131,6 +133,9 @@ private static SolverContext createSolverContext(Configuration config, ShutdownN
 
     public static void main(String[] args) throws Exception {
 
+        final Level logLevel = Level.getLevel(System.getProperty("logLevel", "off"));
```
natgavrilenko, line 136: Doesn't log4j have own configuration variable for this? Did you test that overwriting config file path works. Should be -Dlog4j2.configurationFile=/path/to/log4j2.xml

hernanponcedeleon: Apparently, there are no builtins that allow to change the log level for the root logger, only to specific loggers (*). The two ways of doing this are (1) programmatically (as in my previous commit), or (2) directly in the `xml` file which allows us to change the behavior from the command line with `-DlogLevel=X`. My latest commit follows (2).

(*) Using `-Dlog4j2.logger.com.dat3m.dartagnan.verification.solving.RefinementSolver=info` should at least change the behavior of that particular logger, but (1) I could not make this work, (2) most of our logging comes at the root level, so this would not even be a solution for us.

natgavrilenko (approved):

PR #889, opened 2025-05-16
Generic LKMM Intrinsics

xeren:
Replaces the verifier's Linux-specific intrinsics described in `lkmm.h`.  The intrinsics feature an additional parameter for the intended access size.  Formerly, the access size was fixed to 32 bits.  It can now support at least 8, 16, 32 and 64 bits.  This could ease mocking linux kernel code.

The existing benchmarks were recompiled with Clang 19.1.7.


ThomasHaas:
This seems to be an improvement over what we have. The up/down casts could be a bit dangerous, for example:
```
long v = ...;
int x;
WRITE_ONCE(x, v)
```
I think this would compile just fine but implicitly truncate v down to 32 bits when storing. I guess in the real kernel, this would give a compilation error.

ThomasHaas:
It's fine. At the end of the day, it is most important to retain the correct access types(*) for mocked code, and not to type-check it.

(*) In terms of access size at least. Retaining pointer-typed accesses is probably a lot harder and even unmocked code might happily cast between integers and pointers here.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1312,61 +1312,85 @@ private List<Event> handleLKMMIntrinsic(FunctionCall call) {
...
-                mo = Tag.Linux.intToMo(((IntLiteral) p2).getValueAsInt());
-                checkArgument(p3 instanceof IntLiteral, "No support for variable operator.");
-                op = IntBinaryOp.intToOp(((IntLiteral) p3).getValueAsInt());
-                result.add(EventFactory.Linux.newRMWFetchOp(p0, reg, p1, op, mo));
+            case "__LKMM_atomic_fetch_op" -> {
+                checkArguments(5, call);
+                final Type bytes = toLKMMAccessSize(p1);
+                final String mo = toLKMMMemoryOrder(p3);
+                final IntBinaryOp op = toLKMMOperation(p4);
+                final Register dummy = call.getFunction().newRegister(bytes);
```
ThomasHaas, line 1361: I think you can use `newUniqueRegister` with a base name like "__lkmm_temp/dummy".
Generally, we should reduce the usages of the generic dummy name, ideally getting rid of it altogether (eventually).

ThomasHaas:
> Why are 3 benchmarks being changed?

The benchmarks seemed to be ill-typed before. They use `atomic_t` which is defined as a struct in our headers. `atomic_set` accesses the struct member correctly, while `WRITE_ONCE` just overwrites the whole struct.

xeren:
> > Why are 3 benchmarks being changed?
> 
> The benchmarks seemed to be ill-typed before. They use `atomic_t` which is defined as a struct in our headers. `atomic_set` accesses the struct member correctly, while `WRITE_ONCE` just overwrites the whole struct.

Yes, on those benchmarks, clang complained about being unable to cast `int` to `atomic_t` and vice versa.  Formerly, the cast types were `void*` and `atomic_t*`, respectively.  I decided to modify the accesses, rather than the variable types.

ThomasHaas:
Have you tried your changes on [qspinlock](https://github.com/huawei-drc/cna-verification)?
That repo currently requires Clang 14, because it only issues warnings due to type mismatches. Newer Clang versions make them errors, but there is probably a setting to get the warnings back.
In either case, we probably want `qspinlock` to remain compilable with Clang 14 and our headers and ideally get rid of all the warnings.

include/lkmm.h
```diff
@@ -1,45 +1,59 @@
...
+
+typedef enum __LKMM_operation {
+  __LKMM_op_add,
+  __LKMM_op_sub,
+  __LKMM_op_and,
+  __LKMM_op_or,
+} __LKMM_operation;
+
+typedef typeof(sizeof(int)) __LKMM_size_t;
+typedef long long __LKMM_int_t;
```
ThomasHaas, line 28: Why has this become a 64 bit int?

xeren: It is supposed to cover the signed and unsigned value space of the maximal supported value type.  It could be `intmax_t`.

xeren:
> Have you tried your changes on [qspinlock](https://github.com/huawei-drc/cna-verification)?

Compilation works on clang 19.1.7.  I reproduced the errors you mentioned with the old `lkmm.h`.  The new header fixes all of them.  Only one warning remains, which is not related to the header.
```
$ clang -Xclang -disable-O0-optnone -S -emit-llvm -g -gcolumn-info -Iinclude -I../Dat3M/include client-code.c -o client-code.ll

client-code.c:111:39: warning: cast to 'void *' from smaller integer type 'int' [-Wint-to-void-pointer-cast]
  111 |         pthread_create(&t[i], 0, run, (void *)i);
      |                                       ^~~~~~~~~
1 warning generated.
```

ThomasHaas:
> Anyway, seeing an improvement in the analyses is a welcome side effect of this PR :)

I have to take this back. With this PR we give seemingly wrong verdicts on qspinlock: even without interrupts and using the standard LKMM, we find termination violations.
There are two possibilities:
- This PR does something wrong
- The old lkmm.h was too imprecise and hid a bug we can find now.

hernanponcedeleon:
Do you consistently get violations for all models where the code was supposedly correct (lkmm-v01, armv8, riscv, power)?

What is the minimum number of threads you need to find a violation? I would find `T=3` suspicious

ThomasHaas:
Found it. The inlining of the `cmpxchg` intrinsic is wrong: it swaps expected value with new value which breaks a lot of stuff.

include/lkmm.h
```diff
@@ -1,45 +1,59 @@
...
-extern int __LKMM_LOAD(void*, memory_order);
-extern void __LKMM_STORE(void*, int, memory_order);
-extern void __LKMM_FENCE(int);
-extern int __LKMM_XCHG(int*, int, memory_order);
-extern int __LKMM_CMPXCHG(int*, int, int, memory_order, memory_order);
-extern void __LKMM_ATOMIC_OP(int*, int, operation);
-extern int __LKMM_ATOMIC_FETCH_OP(int*, int, memory_order, operation);
-extern int __LKMM_ATOMIC_OP_RETURN(int*, int, memory_order, operation);
+typedef enum __LKMM_memory_order {
+  __LKMM_relaxed,
```
hernanponcedeleon, line 4: LKMM does not have a `relaxed` mo. Based on [this](https://github.com/torvalds/linux/blob/4a95bc121ccdaee04c4d72f84dbfa6b880a514b6/tools/memory-model/linux-kernel.def#L36), all references to `relaxed` below needs to be changed to `once`.

Be aware of [this](https://github.com/hernanponcedeleon/Dat3M/blob/d8e6066d65d3ddbb3299b933b6caa6ac5a8cbc39/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java#L207) comment when `relaxed` is removed.

ThomasHaas: I removed it.

hernanponcedeleon: I will make a run with the benchmarks of the cna-verification repo and if there are no surprises we can merge

ThomasHaas: Oh damn. We need to recompile all existing LKMM tests with the updated header. Annoying...

ThomasHaas:
With the new header, we accurately track the size of atomic accesses, right?
If so, could we combine this with MSA to check the original qspinlock?

hernanponcedeleon (approved):
There were no surprised with qspinlock so from my side we can merge.

Issue #880, opened 2025-05-09
Mismatch between full and field insensitive alias analysis


hernanponcedeleon:
With the full alias analysis, we return PASS for [this](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/ldv-linux-3.14-races/linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i?ref_type=heads) SVCOMP benchmark (the expected result is FALSE according to the competition).

However, there are a bunch of warnings coming from the analysis
```
[09.05.2025] 19:53:25 [33m[WARN][m InclusionBasedPointerAnalysis.postProcess - empty pointer set for bv64 101:r12 = load(*bv64 101:r11) @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#5791
Context: T1:ldv_insmod_4 -> ldv_insmod_w83977af_init_4_6 @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#7807 -> w83977af_init @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#7840 -> Iter#2 @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#6249 -> w83977af_open @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#6249 -> dma_zalloc_coherent @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#6369 -> dma_alloc_attrs @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#5933 -> is_device_dma_capable @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#5865
[09.05.2025] 19:53:25 [33m[WARN][m InclusionBasedPointerAnalysis.postProcess - empty pointer set for bv64 191:r47 = load(*bv64 192:r21 + bv64(8)) @linux-3.14--drivers--net--irda--w83977af_ir.ko.cil.i#5914
[...]

PR #890, opened 2025-05-16
Add warnings for empty pointer sets in Alias Analysis

xeren:
Adds a check at the end of `FieldSensitiveAndersen` and `AndersenAliasAnalysis`.  The methods can be better compared wrt. null pointer dereferences, like in #880.


hernanponcedeleon (approved):
@xeren can you please rebase?

PR #891, opened 2025-05-17
Fixes to cat grammar

hernanponcedeleon:
Currently we cannot parse [this](https://github.com/herd/herdtools7/blob/55d31bd75001fbcaf720253ad3504b658b95cd7d/herd/libdir/rc11.cat#L4) memory model.

`herd7` reports `*undef*` independent of the name given to the axiom, thus I copy the same output.

### DIFF
--- a/dartagnan/src/main/antlr4/Cat.g4
+++ b/dartagnan/src/main/antlr4/Cat.g4
@@ -5,7 +5,7 @@ import com.dat3m.dartagnan.wmm.axiom.*;
 }
 
 mcm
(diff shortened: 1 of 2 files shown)


ThomasHaas:
Is it really a good idea to mimic the bad output of herd?

ThomasHaas:
If the default is `flag name`, I would just go with `flag name (*undef*)` or similar. Using "as undef" is confusing IMO

ThomasHaas (approved):
LGTM

PR #892, opened 2025-05-19
Better allocation of thread-local memory & SCCP across threads

ThomasHaas:
- Thread-local variables can be replaced by local stack allocations. This allows `Mem2Reg` to replace them, possibly getting rid of all memory accesses of thread-local variables.
- SCCP now propagates constant arguments across thread creations.

For CNA with B=2, this reduces the number of memory accesses by quite a bit:
```
// Old
	#Stores: 69
	#Loads: 106
	#Inits: 21
	#Others: 0
	#must|may rf edges: 3|1147
	#must|may co edges: 36|783

// New
	#Stores: 66
	#Loads: 94
	#Inits: 18
	#Others: 0
	#must|may rf edges: 3|787
	#must|may co edges: 57|660
```
The verification time is not improved by that much.

It should have impact on all code that makes use of thread local variables.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
@@ -271,7 +271,7 @@ public class ThreadCreation implements ProgramProcessor {
         thread.getEntry().insertAfter(body);
 
         // ------------------- Create thread-local variables -------------------
-        replaceGlobalsByThreadLocals(function.getProgram().getMemory(), thread);
+        replaceThreadLocalsWithStackalloc(function.getProgram().getMemory(), thread);
 
         // ------------------- Add end & return label -------------------
         final Register returnRegister = function.hasReturnValue() ?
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -334,25 +334,60 @@ private ThreadData createLLVMThreadFromFunction(Function function, int tid, Thre
...
+        final TypeFactory types = TypeFactory.getInstance();
+        final ExpressionFactory exprs = ExpressionFactory.getInstance();
+
+        // Translate thread-local memory object to local stack allocation
+        Map<MemoryObject, Register> toLocalRegister = new HashMap<>();
+        for (MemoryObject memoryObject : memory.getObjects()) {
+            if (!memoryObject.isThreadLocal()) {
+                continue;
+            }
+            Preconditions.checkState(memoryObject.hasKnownSize());
```
hernanponcedeleon, line 348: Do all thread local objects have known size? Can't we have a thread local array of unknown size?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -334,25 +334,60 @@ private ThreadData createLLVMThreadFromFunction(Function function, int tid, Thre
...
+            // Compute type of memory object based on initial values
+            final List<Type> contentTypes = new ArrayList<>();
+            final List<Integer> offsets = new ArrayList<>();
+            for (Integer initOffset : memoryObject.getInitializedFields()) {
+                contentTypes.add(memoryObject.getInitialValue(initOffset).getType());
+                offsets.add(initOffset);
+            }
+            final Type memoryType = types.getAggregateType(contentTypes, offsets);
+
+            // Allocate single object of memory type
```
hernanponcedeleon, line 359: This suggests we cannot have thread local arrays, which is think it is not accurate

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
ThomasHaas: Thread-locals are declared like global variables. Global variables always have known size (how could they not?).
Variable-sized arrays can only be mimicked by thread-local pointers (known size) to variable-sized arrays.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
ThomasHaas: Well, we currently lose content-type information during parsing, so I cannot use the original type here.
An array `__thread int array[4];` will get treated like `__thread struct {int a; int b; int c; int d} array;` Similarly, nested structs will get flattened into a single struct. I think it doesn't matter, because they all have the same memory layout.

In short, we support all types here because we only rely care about the memory layout.

PR #893, opened 2025-05-20
Update LKMM model and compilation

hernanponcedeleon:
Upgrade the model (and the compilation) with the changes from [this](https://lkml.org/lkml/2024/9/19/521) patch series.

### DIFF
--- a/cat/linux-kernel.cat
+++ b/cat/linux-kernel.cat
@@ -50,6 +50,16 @@ let R4rmb = R \ Noreturn	(* Reads for which rmb works *)
 let rmb = [R4rmb] ; fencerel(Rmb) ; [R4rmb]
 let wmb = [W] ; fencerel(Wmb) ; [W]
 let mb = ([M] ; fencerel(Mb) ; [M]) |
+	(*
+	 * full-barrier RMWs (successful cmpxchg(), xchg(), etc.) act as
+	 * though there were enclosed by smp_mb().
+	 * The effect of these virtual smp_mb() is formalized by adding
+	 * Mb tags to the read and write of the operation, and providing
+	 * the same ordering as though there were additional po edges
+	 * between the Mb tag and the read resp. write.
+	 *)
+	([M] ; po ; [Mb & R]) |
+	([Mb & W] ; po ; [M]) |
 	([M] ; fencerel(Before-atomic) ; [RMW] ; po? ; [M]) |
 	([M] ; po? ; [RMW] ; fencerel(After-atomic) ; [M]) |
 	([M] ; po? ; [LKW] ; fencerel(After-spinlock) ; [M]) |
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Tag.java
@@ -197,11 +197,19 @@ public final class Tag {
         public static final String AFTER_UNLOCK_LOCK        = "After-unlock-lock";
 
         public static String loadMO(String mo) {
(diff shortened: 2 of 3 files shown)


ThomasHaas:
Is this really the whole change? I thought the patch was meant to get rid of all these hidden tag-conversions including the hidden removal of acquire tags from writes.

For example, part of the patch is this here:
```
+let Acquire = ACQUIRE \ W \ FailedRMW
+let Release = RELEASE \ R \ FailedRMW
```
With these changes, the `loadMo` and `storeMo` become obsolete as every tag can be retained as is.

hernanponcedeleon:
You are right ... those changes where in the `.bell` file and I missed them (we should start properly using the `.bell` file; it is annoying to have to move things from there to the `.cat` file every time there is a change in lkmm).

The new tags definitions are causing problems since we do not properly support unary expression like
```
let FailedRMW = RMW \ (domain(rmw) | range(rmw))
```

ThomasHaas:
You could include the .bell file directly and comment out all the unsupported/unnecessary stuff.

hernanponcedeleon:
Now that we have support for unary predicates, I was trying to update this PR, but I found some issues.

I do not think any of the changes in the code of this PR could introduce non-determinism. However, I get different results by running the same benchmarks twice
```
> java -jar dartagnan/target/dartagnan.jar cat/linux-kernel.cat --target=lkmm --method=eager --property=program_spec litmus/LKMM/lkml/unlock-lock-write-ordering-2.litmus
Test: litmus/LKMM/lkml/unlock-lock-write-ordering-2.litmus
Result: FAIL
Reason: Program specification violation found
Condition: exists (1:r1 == bv64(1)) && (2:r2 == bv64(1)) && (2:r3 == bv64(0))
Time: 0.521 secs

> java -jar dartagnan/target/dartagnan.jar cat/linux-kernel.cat --target=lkmm --method=eager --property=program_spec litmus/LKMM/lkml/unlock-lock-write-ordering-2.litmus
Test: litmus/LKMM/lkml/unlock-lock-write-ordering-2.litmus
Result: PASS
Time: 0.495 secs
```
I somehow suspect the problem comes from the difference between sets that are part of the change to `linux-kernel.cat`. @xeren do we need to "register" every base set in `RelationNameRepositoy`/`Wmm`, similar to how you did for `VISIBLE` and `MEMORY`?

hernanponcedeleon:
I do get different sizes of may/must sets, so the problem is indeed in the RA
```
	#Relations: 356
	#Axioms: 10
	#may-edges removed (extended): 342
	total #must|may|exclusive edges: 3213|6941|0
	#must|may rf edges: 0|12
	#must|may co edges: 8|16
	#Relations: 356
	#Axioms: 10
	#may-edges removed (extended): 302
	total #must|may|exclusive edges: 2999|6892|0
	#must|may rf edges: 0|12
	#must|may co edges: 8|16

```

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/processing/MergeEquivalentRelations.java
```diff
@@ -77,6 +77,10 @@ private boolean areEquivalent(Relation r1, Relation r2) {
             return set1.getTag().equals(set2.getTag());
         }
 
+        if (def1 instanceof Projection p1 && def2 instanceof Projection p2) {
+            return p1.getDimension().equals(p2.getDimension());
+        }
+
```
ThomasHaas: This is no valid fix: the dimension AND the projected relation must be equal. You could do
```Java
if (def1 instanceof Projection p1 && def2 instanceof Projection p2 && p1.getDimension() != p2.getDimension()) {
       return false;
}
// ... equality of the inner relations gets checked later
```

PR #896, opened 2025-05-30
Update Vulkan model avvis semantics

natgavrilenko:
Applied update of the official model:
https://github.com/KhronosGroup/Vulkan-MemoryModel/pull/40

### DIFF
--- a/cat/spirv-check.cat
+++ b/cat/spirv-check.cat
@@ -44,15 +44,14 @@ let inscope = (DV * DV) |
               (swg & ((DV | QF | WG) * (DV | QF | WG))) |
               (ssg & ((DV | QF | WG | SG) * (DV | QF | WG | SG)))
 
- // mutually ordered atomics
- let mutordatom = (loc & vloc & inscope & (ATOM * ATOM)) \ id
+// mutually ordered atomics
+let mutordatom = (loc & vloc & inscope & (ATOM * ATOM)) \ id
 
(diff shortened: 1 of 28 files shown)


cat/spirv-check.cat
```diff
@@ -44,15 +44,14 @@ let inscope = (DV * DV) |
...
-let avvisSet = [AV | VIS]; (vloc & loc)
-let avvisinc = ((avvisSem | avvisSem^-1) \ id) |
-               (avvisSet | avvisSet^-1)
+// mutually ordered atomics
+let mutordatom = (loc & vloc & inscope & (ATOM * ATOM)) \ id
+
+let avvisSem = ((SC0 | SC1) * AVDEVICE) | (VISDEVICE * (SC0 | SC1)) |
+               (SC0 * (SEMSC0 & SEMAV)) | (SC1 * (SEMSC1 & SEMAV)) |
+               ((SEMSC0 & SEMVIS) * SC0) | ((SEMSC1 & SEMVIS) * SC1)
+let avvisSet = ((vloc & loc) ; [AV]) | ([VIS] ; (vloc & loc))
```
hernanponcedeleon: I don't think this is equivalent to the alloy one. IIUC that one contains both `(vloc & loc) ; [VIS]` and `[AV] ; (vloc & loc)`

natgavrilenko: You are right, the avvisSet change was not a part of this patch, updated

hernanponcedeleon (approved):

PR #894, opened 2025-05-22
Add batch execution mode

hernanponcedeleon:
If one wants to run many tests from the command line, we need to start the JVM for each test which is very slow for litmus tests.

This PR adds a mode that allows to run tests in a batch and normalize the output to make it more machine readable.

Here is an example of how the output looks for the lock benchmarks (setting `--timeout=10`)
```
================ Configuration ==================
cat = cat/vmm.cat
batch = benchmarks/locks/
timeout = 10
=================================================

Test: benchmarks/locks/clh_mutex.c
Result: UNKNOWN
Reason: Not fully unrolled loops
Details:
	E116: 	T1:thread_n -> clh_mutex_lock @clh_mutex.c#17 -> @clh_mutex.h#143
Time: 1.846 secs

Test: benchmarks/locks/cna.c
Result: ERROR
Reason: NullPointerException
Details:
	Unknown error occurred

Test: benchmarks/locks/deadlock.c
Result: FAIL
Reason: Termination violation found
Details:
	E28:	@deadlock.c#8
Time: 0.090 secs

Test: benchmarks/locks/linuxrwlock.c
Result: FAIL
Reason: CAT specification violation found
Details:
	Flag r-data-race
	E127 / E127	@linuxrwlock.c#25 / @linuxrwlock.c#25
Time: 4.730 secs

Test: benchmarks/locks/mutex.c
Result: FAIL
Reason: CAT specification violation found
Details:
	Flag w-data-race
	E305 / E305	@mutex.c#21 / @mutex.c#21
	E174 / E174	@mutex.c#21 / @mutex.c#21
Time: 0.838 secs

Test: benchmarks/locks/mutex_musl.c
Result: FAIL
Reason: Program specification violation found
Details:
	E70:	@mutex_musl.c#38: user assertion
Time: 0.732 secs

Test: benchmarks/locks/pthread_mutex.c
Result: PASS
Time: 0.113 secs

Test: benchmarks/locks/seqlock.c
Result: PASS
Time: 1.236 secs

Test: benchmarks/locks/spinlock.c
Result: FAIL
Reason: CAT specification violation found
Details:
	Flag w-data-race
	E132 / E132	@spinlock.c#21 / @spinlock.c#21
	E193 / E193	@spinlock.c#21 / @spinlock.c#21
Time: 0.181 secs

Test: benchmarks/locks/ticket_awnsb_mutex.c
Result: TIMEDOUT
Time: 10.000 secs

Test: benchmarks/locks/ticketlock.c
[...]

PR #899, opened 2025-06-05
Benchexec integration + skip assertions

hernanponcedeleon:
The integration with `benchexec` (the execution infrastructure for SVCOMP) is broken after #894.

Also, when we skip some assertions, we should not report if they are triggered, but we should still abort and tag the corresponding event with `EXCEPTIONAL_TERMINATION` (I noticed this while testing the termination benchmarks from SVCOMP).


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -996,24 +996,18 @@ private List<Event> inlineAlignedAlloc(FunctionCall call) {
...
     private List<Event> inlineAssert(FunctionCall call, AssertionType skip, String errorMsg) {
-        if(notToInline.contains(skip)) {
-            return List.of();
-        }
         final Expression condition = expressions.makeFalse();
         final Event assertion = EventFactory.newAssert(condition, errorMsg);
         final Event abort = EventFactory.newAbortIf(expressions.makeTrue());
         abort.addTags(Tag.EXCEPTIONAL_TERMINATION);
-        return List.of(assertion, abort);
+        return notToInline.contains(skip) ? List.of(abort) : List.of(assertion, abort);
```
ThomasHaas: You could define `assertion = notToInline.contains(skip) ? null : EventFactory.newAssert(...)` and then use `return eventSequence(assertions, abort)`. It is a little bit nicer because then you do not construct the assertion event in the first place.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -996,24 +996,18 @@ private List<Event> inlineAlignedAlloc(FunctionCall call) {
...
 
     private List<Event> inlineVerifierAssert(FunctionCall call, AssertionType skip, String errorMsg) {
-        if(notToInline.contains(skip)) {
-            return List.of();
-        }
         assert call.getArguments().size() == 1;
         final Expression condition = call.getArguments().get(0);
         final Event assertion = EventFactory.newAssert(expressions.makeBooleanCast(condition), errorMsg);
-        return List.of(assertion);
+        return notToInline.contains(skip) ? List.of() : List.of(assertion);
```
ThomasHaas: The change here is pointless.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: You missed the `return eventSequence(...)`. I think `List.of` will throw an exception with null entries, and even if it does not, the caller won't expect null values.

PR #901, opened 2025-06-08
Align cat definition with the alloy model

natgavrilenko:


litmus/VULKAN/Manual/asmo-mixed-atomicity-write.litmus
```diff
@@ -8,7 +9,7 @@ P3:r3=0;
 x=0;
 }
  P0@sg 0,wg 0, qf 0 | P1@sg 0,wg 1, qf 0  | P2@sg 0,wg 2, qf 0   | P3@sg 0,wg 3, qf 0   ;
- st.av.dv.sc0 x, 1  | st.av.dv.sc0 x, 2   | ld.atom.dv.sc0 r0, x | ld.atom.dv.sc0 r2, x ;
+ st.av.dv.sc0 x, 1  | st.atom.dv.sc0 x, 2 | ld.atom.dv.sc0 r0, x | ld.atom.dv.sc0 r2, x ;
```
hernanponcedeleon, line 12: `asmo-mixed-atomicity-read` makes **all** reads to be `vis`. Shouldn't both stores by `av` for consistency?

hernanponcedeleon (commented):
Can you also rename the cat file to `vulkan.cat` and remove `spirv-check` as wee agreed?

litmus/VULKAN/Manual/asmo-mixed-atomicity-write.litmus
natgavrilenko: It is done this way intentionally. According to the current version of the model, relaxing one write makes this behavior possible. But the model says that the behavior shouldn't be possible even if all the reads are relaxed.

PR #904, opened 2025-06-10
Fix MemToReg

xeren:
Fixes a bug, where thread-local allocations were *partially* promoted to registers, while some usages were kept unchanged.

Consider the following minimal example.  The buggy `MemToReg` removed the allocation of `a`, but kept the store.  Expected behavior is to keep the allocation.
```c
int a[2];        // r1 := alloc(2 * sizeof(int))
int *i;
if(*) i = &a[0]; // r2 := r1
else i = &a[1];  // r2 := r1 + sizeof(int)
*i = 1;          // store(*r2, 1)
```
In comparison, the following two programs were already handled as expected before this fix.
```c
int a[2];    // r1 := alloc(2 * sizeof(int))
int i;
if(*) i = 0; // r2 := 0
else i = 1;  // r2 := 1
a[i] = 1;    // store(*(r1 + r2), 1)

int a[2];                     // r1 := alloc(2 * sizeof(int))
int *i = (*) ? &a[0] : &a[1]; // r2 := ite(*, r1, r1 + sizeof(int))
*i = 1;                       // store(*r2, 1)
```

### DIFF
--- a/benchmarks/miscellaneous/memtoreg_merging_pointers.c
+++ b/benchmarks/miscellaneous/memtoreg_merging_pointers.c
@@ -7,12 +7,12 @@
 int main()
 {
     int a[2];
-    int *i;
+    int *i, *j;
     if (__VERIFIER_nondet_int())
-        i = &a[0];
+        i = &a[0], j = &a[0];
     else
-        i = &a[1];
+        i = &a[1], j = &a[1];
     *i = 1;
-    assert(*i == 1);
+    assert(*j == 1);
     return 0;
 }
--- a/benchmarks/miscellaneous/memtoreg_merging_pointers.c
+++ b/benchmarks/miscellaneous/memtoreg_merging_pointers.c
@@ -1,0 +1,18 @@
+#include <assert.h>
+#include <dat3m.h>
+
+// Issue: If MemToReg does not handle dataflow correctly, it may remove the allocation of an array, while keeping its usages.
+// Expected: PASS
+
+int main()
+{
+    int a[2];
+    int *i;
+    if (__VERIFIER_nondet_int())
+        i = &a[0];
+    else
+        i = &a[1];
+    *i = 1;
+    assert(*i == 1);
+    return 0;
+}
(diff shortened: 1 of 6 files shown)


hernanponcedeleon:
> Consider the following minimal example. The buggy `MemToReg` removed the allocation of `a`, but kept the store. Expected behavior is to keep the allocation.
> 
> ```c
> int a[2];        // r1 := alloc(2 * sizeof(int))
> int *i;
> if(*) i = &a[0]; // r2 := r1
> else i = &a[1];  // r2 := r1 + sizeof(int)
> *i = 1;          // store(*r2, 1)
> ```

Can you convert this into a C program and add it as a regression?

ThomasHaas:
> Can you convert this into a C program and add it as a regression?

I extracted a test from libvsync for René to debug this. We should use that test probably. However, having a C-program as test seems a bit awkward. The bug in `Mem2Reg` did not result in wrong verdicts that we can check for, but rather in wrong IR.
But maybe we can take the C-test, invoke our processing pipeline(*),  and check that all loads and stores got eliminated.


The fixes in this PR seem to work. The CI failure was just unlucky.


(*) For the bug to trigger, I think we need to invoke at least `Inlining` + `BranchReordering` before doing `Mem2Reg`.

xeren:
> Can you convert this into a C program and add it as a regression?

The exact program I used is the following.  I expect PASS.  The eager solver returned FAIL because the .  The lazy solver threw a `NullPointerException` while fetching the address of the store (and would also fail at the additional load).
I could add this into `benchmarks/miscellaneous`.
```c
void assert(int);
int __VERIFIER_nondet_int(void);
int main()
{
    int a[2]; // was removed by MemToReg
    int *i;
    if(__VERIFIER_nondet_int()) i = &a[0];
    else i = &a[1];
    *i = 1; // was kept
    assert(*i == 1); // was kept
    return 0;
}
```

> What about having something like [this](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java)?

The new unit tests proved quite useful.  The `memToReg3` test captures my minimal example from above, while `memToReg1` and `memToReg2` capture the two other variants.

> For the bug to trigger, I think we need to invoke at least Inlining + BranchReordering before doing Mem2Reg.

In your example, `BranchReordering` created a situation just like here, where multiple branches merge with different values on a promotable variable pointing into a non-promotable variable.  I think my example accurately describes the nature of this bug even without `BranchReordering` or `Inlining`.

xeren:
I must have misinterpreted my results on the example.  With the eager solver, the test passed even on the former branch, but it was not supposed to.  Instead, `InclusionBasedPointerAnalysis` still admitted `mustAlias` for the two events, since the expressions were equal.  This is intended for IBPA, so I modified the example (and fixed a bug in IBPA with uninitialized address registers).
From my part, this should now be ready to merge.

PR #905, opened 2025-06-12
Add Intrinsics pthread_detach

xeren:
This PR implements `pthread_detach` as a built-in.  Any thread can set the 'detached' flag of a POSIX thread.  Attempts to join a detached thread will fail.  However, the detached thread still shares the memory.

This implementation gives no synchronization guarantees:
- The 'detached' flag has a different memory location than the 'running' flag.
- The 'detached' flag is accessed non-atomically.

Consequently, concurrently detaching and joining the same thread can be racy.


hernanponcedeleon:
1. I haven't yet checked the code, but I test it on the SVCOMP benchmarks we were previous throwing an exception and we get correct results on those.

benchmarks/miscellaneous/pthread.c
```diff
@@ -412,10 +412,59 @@ void key_test()
...
 int main()
 {
-    mutex_test();
+    //mutex_test();
     cond_test();
-    rwlock_test();
-    key_test();
+    //rwlock_test();
+    //key_test();
+    //detach_test();
```
hernanponcedeleon: Please re-enable (rrobably some left over from testing)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -203,9 +207,13 @@ private void resolveDynamicThreadJoin(Program program, List<ThreadData> threadDa
                             EventFactory.newGoto(joinEnd)
                     );
                 } else {
+                    final Expression detachedOffset = expressions.makeOne((IntegerType) data.comAddress.getType());
                     // Successful join
                     caseBody = eventSequence(
                             joinCase,
+                            newLoad(detachedRegister, expressions.makeAdd(data.comAddress, detachedOffset)),
```
hernanponcedeleon: I do not understand how `comAddress+1` may not clash with other addresses.

EDIT: after reading the whole code, I got it

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -239,6 +247,73 @@ private void resolveDynamicThreadJoin(Program program, List<ThreadData> threadDa
...
+            final Register detachedRegister = caller.newRegister("__detachDetached#" + joinCounter, types.getBooleanType());
+
+            // ----- Construct a switch case for each possible tid -----
+            final Label joinEnd = EventFactory.newLabel("__detachEnd#" + joinCounter);
+            final Map<Expression, List<Event>> tid2joinCases = new LinkedHashMap<>();
+            for (ThreadData data : threadData) {
+                if (!data.isJoinable() || data.thread() == caller) {
+                    // NOTE: We treat self-joins as an invalid tid id error (~ in alignment with pthread_join semantics)
+                    // We could alternatively make this a non-termination case
+                    continue;
```
hernanponcedeleon: I guess this is now about "self-detaching". Is this really supposed to be an error?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -239,6 +247,73 @@ private void resolveDynamicThreadJoin(Program program, List<ThreadData> threadDa
...
+                }
+
+                final int tid = data.thread().getId();
+                if (tidExpr instanceof IntLiteral iConst && iConst.getValueAsInt() != tid) {
+                    // Little optimization if we join with a constant address
```
hernanponcedeleon: join -> detach

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -239,6 +247,73 @@ private void resolveDynamicThreadJoin(Program program, List<ThreadData> threadDa
...
+            for (Expression tid : tid2joinCases.keySet()) {
+                switchJumpTable.add(EventFactory.newJump(
+                        expressions.makeEQ(tidExpr, tid), (Label)tid2joinCases.get(tid).get(0))
+                );
+            }
+            // In the case where no tid matches, we return an error status.
+            switchJumpTable.add(EventFactory.newLocal(statusRegister, invalidTidValue));
+            switchJumpTable.add(EventFactory.newGoto(joinEnd));
+
+            // ----- Generate actual replacement for the DynamicJoinEvent -----
```
hernanponcedeleon: join -> detach

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -239,6 +247,73 @@ private void resolveDynamicThreadJoin(Program program, List<ThreadData> threadDa
...
+            final List<Event> replacement = eventSequence(
+                    switchJumpTable,
+                    tid2joinCases.values(),
+                    joinEnd
```
hernanponcedeleon: join -> detach

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -313,7 +388,7 @@ private ThreadData createLLVMThreadFromFunction(Function function, int tid, Thre
             }
 
             // We use accesses to a common memory object to synchronize creator and thread.
-            final MemoryObject comAddress = function.getProgram().getMemory().allocate(1);
+            final MemoryObject comAddress = function.getProgram().getMemory().allocate(2);
```
hernanponcedeleon: Ok, now my previous comment was answered. Please add a comment here about why we need 2 entries

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
xeren: No, considering that [this documentation](https://www.man7.org/linux/man-pages/man3/pthread_detach.3.html) names `pthread_detach(pthread_self())` as an example.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
hernanponcedeleon: I was thinking again about this one. Would it not be better to use a single (non-boolean) location to store the state of the thread: NOT_RUNNING, RUNNING, TERMINATED/JOINABLE, JOINED (maybe it can be merged with NOT_RUNNING, but maybe having it explicitly allows us to reason about UB), DETACHED (this way we could detect UB due to trying to detach a detached thread)?

hernanponcedeleon: @xeren have you seen my latest comment here?

xeren: Sorry, I haven't seen this.  I had two arguments for my implementation choice:
If we merged both locations, then calls to `pthread_detach` would have to be at least relaxed atomic, to be safely racy with `pthread_exit`.  Then the race with `pthread_join` would also be safe.  Since both `pthread_detach` and `pthread_join` may return EINVAL, this should be valid.
The state chart would look like below.  NOT_RUNNING and JOINED can be merged, as long as we have an isomorphism between worker threads and `ThreadCreate` events.  `pthread_exit` now needs a load event to properly set the state.

```mermaid
graph TD
i(["not running"])
r(["running"])
d(["detached"])
j(["joinable"])
f((("joined")))
i--create-->r
r--detach-->d
i--create-->d
r--exit-->j
j--join,detach-->f
d--exit-->f
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
```diff
@@ -78,6 +78,7 @@ public Expression visitRegister(Register reg) {
                         // A usage counter may be missing for assignments of the form "r = f(r)"
                         // which do not allow for substituting r for f(r) in later usages.
                         usageCounter.containsKey(lastAssignment)
+                        && preDominatorTree.contains(curEvent)
```
hernanponcedeleon, line 81: Why does this PR affect this pass?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -81,6 +83,12 @@ public class ThreadCreation implements ProgramProcessor {
     private final TypeFactory types = TypeFactory.getInstance();
     private final ExpressionFactory expressions = ExpressionFactory.getInstance();
     private final IntegerType archType = types.getArchType();
+    // The thread state consists of two flags: ALIVE and JOINABLE.
+    private final IntegerType threadStateType = types.getIntegerType(2);
+    private final Expression threadStateStopped = expressions.makeValue(0, threadStateType);
+    private final Expression threadStateRunning = expressions.makeValue(3, threadStateType);
+    private final Expression threadStateNotAlive = expressions.makeValue(2, threadStateType);
+    private final Expression threadStateNotJoinable = expressions.makeValue(1, threadStateType);
```
hernanponcedeleon: Define them in ascending order

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadDetach.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public Expression getTid() {
+        return tid;
+    }
+
+    @Override
+    protected String defaultString() {
+        return String.format("DynamicThreadDetach(%s)", tid);
+    }
```
ThomasHaas: You should also print the register.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadJoin.java
```diff
@@ -31,8 +31,9 @@ public class DynamicThreadJoin extends AbstractEvent implements RegWriter, RegRe
     public enum Status {
         SUCCESS(0),                // The join was successful
         INVALID_TID(1),            // The provided tid does not match with a joinable thread
-        INVALID_RETURN_TYPE(2);    // The provided tid is valid, but the return type of that thread does not match
+        INVALID_RETURN_TYPE(2),    // The provided tid is valid, but the return type of that thread does not match
                                              // with the expected return type.
+        DETACHED_THREAD(3);        // The provided tid belongs to a non-joinable thread.
```
ThomasHaas: You might want to change the comment of `INVALID_TID` to read as "the provided tid does not match with any spawned thread." Originally I wrote it like that,  cause I thought we might reuse `INVALID_TID` for non-joinable threads.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -416,18 +416,29 @@ private List<Event> inlinePthreadCreate(FunctionCall call) {
...
+        final Label skipAttrLabel = newLabel("__pthread_create_skip_attr");
+        final Label skipDetachLabel = newLabel("__pthread_create_skip_detach");
 
         return eventSequence(
                 createEvent,
+                newJump(expressions.makeEQ(attributes, expressions.makeGeneralZero(attributes.getType())), skipAttrLabel),
+                newLoad(attributesRegister, attributes),
+                // if detach
+                newJumpUnless(testPthreadCreateDetached(attributesRegister), skipDetachLabel),
+                newDynamicThreadDetach(resultRegister, tidReg), // detach should always succeed
```
ThomasHaas: the result register is just used as a dummy here, right? I mean, it gets overwritten in the end anyways.
Maybe add a short comment that the result gets overwritten and its just a dummy.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -517,25 +535,82 @@ private List<Event> inlinePthreadAttr(FunctionCall call) {
...
+        checkArgument((getter || suffix.startsWith("set")) && P_THREAD_ATTR.contains(suffix.substring(3)),
+                "Unrecognized intrinsics \"%s\"", call);
+        final Register oldValue = call.getFunction().newRegister(getPthreadAttrType());
+        final Label end = EventFactory.newLabel("__pthread_return");
+        final Label returnEINVAL = EventFactory.newLabel("__pthread_return_EINVAL");
+        final PthreadAttrImplementation impl = switch (suffix.substring(3)) {
+            case "detachstate" -> inlinePthreadAttrDetachState(oldValue, getter ? null : value, returnEINVAL, end);
+            default -> null;
+        };
+        final Expression EINVAL = expressions.makeValue(22, (IntegerType) errorRegister.getType());//TODO this may vary
```
ThomasHaas: I don't like that this magic number is embedded into this function. Maybe we can have an enum of error codes somewhere?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -517,25 +535,82 @@ private List<Event> inlinePthreadAttr(FunctionCall call) {
...
+    }
+
+    private record PthreadAttrImplementation(Expression out, List<Event> errorChecks) {}
+
+    private PthreadAttrImplementation inlinePthreadAttrDetachState(Expression oldValue, Expression detachstate, Label returnEINVAL, Label end) {
+        final int flagIndex = 1;
+        final IntegerType attrType = (IntegerType) oldValue.getType();
+        final IntegerType valueType = types.getIntegerType(32);
+        final Expression createDetached = expressions.makeValue(2, valueType);//TODO may vary by platform
+        final Expression createJoinable = expressions.makeValue(1, valueType);//TODO may vary by platform
```
ThomasHaas: Same as for EINVAL: those number should be in some enum with a descriptive name.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -517,25 +535,82 @@ private List<Event> inlinePthreadAttr(FunctionCall call) {
...
+    private IntegerType getPthreadAttrType() {
+        return types.getIntegerType(2);
+    }
+
+    private record PthreadAttrImplementation(Expression out, List<Event> errorChecks) {}
+
+    private PthreadAttrImplementation inlinePthreadAttrDetachState(Expression oldValue, Expression detachstate, Label returnEINVAL, Label end) {
+        final int flagIndex = 1;
+        final IntegerType attrType = (IntegerType) oldValue.getType();
+        final IntegerType valueType = types.getIntegerType(32);
```
ThomasHaas: Is the type here supposed to be hardcoded into the function? Is the value type for each attribute different?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -48,7 +50,7 @@
  * LLVM:
  * This pass handles (reachable) pthread-related function calls.
  * - each pthread_create call spawns a new Thread object.
- * - pthread_join calls are lowered to appropriate synchronization primitives.
+ * - pthread_join and pthread_detach calls are lowered to appropriate synchronization primitives.
```
ThomasHaas, line 53: Oh, it seems the comments on this class are quite outdated. We can fix this in a different PR though.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -203,12 +213,17 @@ private void resolveDynamicThreadJoin(Program program, List<ThreadData> threadDa
                             EventFactory.newGoto(joinEnd)
                     );
                 } else {
+                    final Expression isAlive = expressions.makeIntExtract(threadStateRegister, 0, 0);
+                    final Expression isJoinable = expressions.makeIntExtract(threadStateRegister, 1, 1);
```
ThomasHaas: Can we make this more streamlined and use an enum or something to define the meaning of each bit?
```
enum ThreadStateBits {
     Alive,         
     Joinable;

     // Maybe with a .getBit()  and a .getInvertedMask() method
}
```
And then even define the thread states like `Expression threadStateNotAlive = makeValue(~Alive.getBit(), ...)` or `Expression threadStateNotAlive = makeValue(Alive.getInvertedMask(), ...)`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
xeren: Formerly, when `AssignmentInlining` is run on a function containing an unreachable inlinable assignment, it threw an unhandled `NullPointerException` in `isDominatedBy` next line.  The reason seems to be, that `DominatorTree` does not collect information about unreachable events.

The new exit code for worker threads, which seems to be dead in some tests, features such an assignment.  This occurred in `LLVM tests` [here](https://github.com/hernanponcedeleon/Dat3M/actions/runs/16169448793/job/45639065695).  Instead of adding another dead code elimination pass, I chose to modify this check here.

An alternative could be to add a `containsKey` test in `isDominatedBy`:
```java
public class DominatorTree<TNode> {
    //...
    public boolean isDominatedBy(TNode node, TNode dominator) {
        return !immDominator.containsKey(node) || getNearestCommonDominator(node, dominator) == dominator;
    }
    //...
}
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
xeren: Ok.  (This is the chronological order in the livetime of a thread.)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
ThomasHaas: Unreachable nodes and dominators interact weirdly. The most correct definition would likely be to say that unreachable nodes are dominated by every node just like you did.
However, I deliberately chose to throw an exception because the dominator tree can be constructed for only a subset of nodes/events (e.g. for only the body of a loop) and then it is unclear if a missing node is really unreachable or if it is just outside of the domain of the construction.
So I think it is fine to keep the dominator tree code as is and let the user of the analysis decide how to handle the edge-case.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: Yes, attributes have individual types.  This attribute in particular is declared as below.
```c
#define PTHREAD_CREATE_JOINABLE /*platform-specific*/
#define PTHREAD_CREATE_DETACHED /*platform-specific*/
int pthread_attr_getdetachstate(const pthread_attr_t *attr, int *detachstate);
int pthread_attr_setdetachstate(pthread_attr_t *attr, int detachstate);
```
~Luckily, both my machine and the CI platform define the constants equally.~

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
ThomasHaas: So you decided to change the DominatorTree implementation. We can keep it for now, but I think it is a little bit dangerous.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/PosixErrorCode.java
```diff
@@ -0,0 +1,144 @@
+package com.dat3m.dartagnan.program.processing;
+
+// This excerpt of the
+enum PosixErrorCode {
```
ThomasHaas: What is that comment?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/PosixErrorCode.java
```diff
@@ -0,0 +1,144 @@
...
+    E0,
```
ThomasHaas: Wouldn't it be fair to name this first entry SUCCESS or NOERR or something? I think the value 0 is always defined as absence of errors because the usual way to check errors is setting `errno = 0`, call the function, and then check `errno != 0` (if the function does not already return the error code directly). Also, on startup the initial value of `errno` should be 0.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
```diff
@@ -317,17 +408,19 @@ private ThreadData createLLVMThreadFromFunction(Function function, int tid, Thre
             // We use accesses to a common memory object to synchronize creator and thread.
             final MemoryObject comAddress = function.getProgram().getMemory().allocate(1);
             comAddress.setName("__com_" + function.getName() + "#" + tid);
-            comAddress.setInitialValue(0, expressions.makeFalse());
+            comAddress.setInitialValue(0, threadState(0));
 
             // Sync
-            final Register startSignal = thread.newRegister("__startT" + tid, types.getBooleanType());
+            final Register threadState = thread.newRegister("__threadStateT" + tid, threadStateType);
+            final Expression isAlive = expressions.makeIntExtract(threadState, 0, 0);
```
ThomasHaas: This should use the `threadStateFlag` method with ALIVE as flag value.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm7.java
```diff
@@ -42,4 +44,25 @@ public List<Event> visitAtomicStore(AtomicStore e) {
...
+    @Override
+    public List<Event> visitLKMMFetchOp(LKMMFetchOp e) {
+        final Expression value = expressions.makeIntBinary(e.getResultRegister(), e.getOperator(), e.getOperand());
+        final Load load = newRMWLoad(e.getResultRegister(), e.getAddress());
+        return eventSequence(
+                load,
+                AArch64.DMB.newISHBarrier(),
+                newRMWStore(load, e.getAddress(), value)
+        );
+    }
```
ThomasHaas: This code looks so dangerous. I know that `VisitorArm7` is "just a hack" but can be maybe at least add preconditions to check that the memory orders are as expected?
In fact, it seems you generate both RLX and ACQ/REL versions of these events, all of which get compiled to the same core events. I expect that the RLX version should not have the barrier and that the barrier must be placed before the load in case of ACQ and after the store in case of REL. 
Put differently, the compilation scheme is likely wrong even for the few cases we care about.

While you are at it, can you add preconditions checks for ACQ and REL to `visitAtomicLoad` and `visitAtomicStore`? At least I think those two should only be used for ACQ/REL operations.

xeren: REL stores guarantee, that everything *before* them gets stored with them.  ACQ loads guarantee, that any load *after* them fetches the newest data.  This is why I think that the fences are already positioned right where they should.

ThomasHaas: Sorry, you are right. I accidentally swapped them in my head. REL is before the store and ACQ is after the load, so it is correct.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/AssignmentInlining.java
hernanponcedeleon: > So I think it is fine to keep the dominator tree code as is and let the user of the analysis decide how to handle the edge-case.

This is what the currently implementation is doing, right? If so, can we mark this one as resolved?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ThreadCreation.java
hernanponcedeleon: This is implemented via `threadStateFlag`, right? Can we resolve?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
hernanponcedeleon: Are both the constants (`PTHREAD_DETACHSTATE_DETACHED` and `PTHREAD_DETACHSTATE_JOINABLE`) and the type (`int32`) platform dependent?

xeren: The type is `int` and the constants are defined by the header deployed with the implementation.  POSIX only specifies an API, but no ABI [[man7](https://www.man7.org/linux/man-pages/man3/pthread_attr_setdetachstate.3.html), [OpenGroup](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/pthread.h.html)].

| Implementation | `PTHREAD_CREATE_JOINABLE` | `PTHREAD_CREATE_DETACHED` |
| - | - | - |
| [Debian Glibc](https://sources.debian.org/src/glibc/2.28-8/sysdeps/htl/bits/pthreadtypes.h/) | 0 | 1 |
| [ArchLinux Glibc](https://archlinux.org/packages/core/x86_64/glibc/) | 0 | 1 |
| [PTHREADS-WIN32](https://github.com/BrianGladman/pthreads) | 0 | 1 |
| [MinGW](https://github.com/mirror/mingw-w64/blob/master/mingw-w64-libraries/winpthreads/include/pthread.h) | 0 | 4 |
| MacOSX | 1 | 2 |

I was hoping these values would coincide.  CI succeeds only because the benchmarks were compiled on the machine I extracted the definitions from.  Eventually, I'd like to target all platforms, but for now we could set for one specific platform, like Linux with Glibc.

ThomasHaas:
I think you can merge

PR #907, opened 2025-06-18
Remove mixed size logging if it is disabled

natgavrilenko:
Creates way too much logs for litmus tests


dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -94,7 +95,9 @@ public static void performStaticProgramAnalyses(VerificationTask task, Context a
                 analysisContext, config));
         final AliasAnalysis alias = AliasAnalysis.fromConfig(program, analysisContext, config, logger.isWarnEnabled());
         analysisContext.register(AliasAnalysis.class, alias);
-        checkForMixedSizeAccesses(program, alias);
+        if (config.hasProperty(MIXED_SIZE) && Boolean.getBoolean(config.getProperty(MIXED_SIZE))) {
+            checkForMixedSizeAccesses(program, alias);
+        }
```
ThomasHaas: I think then you can remove the code altogether. IIRC, the code was meant to check for MSA precisely when the option is not set to notify the user that he should enable it.
Alternatively, you can disable it just for litmus tests and keep it for real code.

EDIT: Btw. I also think for several litmus formats, the parser is actually just wrong and that is why there is MSA. In that case I would rather fix the parser.

natgavrilenko: Removed. You can fix the problem and put it back, if you like. But if the problem is final state, then it will also be an issue for spirv.

ThomasHaas: I think the problem is that the litmus parsers do not allocate sufficiently-sized memory objects (e.g., they default to 1 byte allocations but store 8 byte values). Then every access to that location, whether via an event or via the final state is MSA.
At least I remember that this problem existed in the ARM parser and it was likely only fixed there. I don't know about the other parsers and SPIRV.

Issue #535, opened 2023-10-14
Improved support for CAT features


ThomasHaas:
We need to improve our support for CAT features to handle the new ARM/ASL model. Having those features is generally useful even outside of ARM/ASL. I want to list the features and give some insight in how to implement them.

1. Support for function declarations and function calls. If we do not support recursive functions (though ARM/ASL does have those but I think they are unnecessary), it would be fine to have the parser directly inline all calls during parsing. So there is no need to update the backend in any way.
`herd7`'s support for function calls is quite extensive. For example (from [herd7 enumerations.cat](https://github.com/herd/herdtools7/blob/master/herd/libdir/enumerations.cat))
```
let add_pair p = map (fun r -> (p | r))
```
declares a function `add_pair` that takes a single parameter `p` and returns a partially applied `map` function (curried).
The result is a function that takes a set of relations `R` and maps this set to a new one by adding `p` to each `r in R`.
I don't believe we need such extensive support(\*) but we should aim to support the basics at least.
2. Support for `includes` (`include "other.cat"`). The parser should just inline the included files directly at the location where it is included. Again, there is no need to update our backend. However, we could move common definitions like `fr`, `fre`, `rfi`, etc. into a `basics.cat` and then simplify our parsing of those relations.
3. Support for `ifs` (needed for ARM/ASL). The syntax looks like this `(if COND then r1 else r2)`. Here `COND` is like a compilation flag, meaning that during parsing time the parser will resolve the conditional choice. More concretely, the syntax is `COND ::= "OPTION" | COND && COND | COND || COND | not COND`, i.e., boolean expressions over atomic options flags "OPTION" (just any string). Again, this feature is parser-only and the backend doesn't need to change.
4.  Support for special options that modify how the CAT file is interpreted. ARM/ASL has this line at the top "catdep (* This option says that the cat file computes dependencies *)", which likely means that local operations are visible so that internal data dependencies can be accessed (e.g. `idd` in our case). Such a feature would also allow for options like `partialco` to change how `co` is interpreted, or verification options like `local-consistency` to state helpful information that simplifies reasoning about the CAT file.

Apart form the support for new syntax, we should also add support for proper unary predicates as mentioned in #371.

[...]

PR #908, opened 2025-06-18
Support unary predicates in cat

xeren:
This PR affects the intermediate representation for memory models, such that `Relation` can model sets as well.  This relates to issues #535 and #371.

Features:
- The cat parser accepts `domain(r)` and `range(r)` without brackets.
- The builtin `New` constructs a free set.
- The `empty r` axiom allows a set as operand.

API changes:
- Add `Relation.isUnaryRelation()` and `Relation.isBinaryRelation()`.
- Add *unary* and *binary* type checks to constructors of `Constraint`.
- Add `Wmm.newRelation(boolean)` and `Wmm.newRelation(String,boolean)`.
- Replace `DomainIdentity` and `RangeIdentity` with `Domain` and `Range`.
- Remodel `SetIdentity` as non-static.
- Add `TagSet` as its bottom-level replacement for static `SetIdentity`.
- `RelationAnalysis` also learns `Knowledge` for unary relations, stored as identity graphs.

CAAT changes:
- Replace `ProjectionIdentityGraph` with `ProjectionSet`.
- Add n-ary `IntersectionSet` and `UnionSet`.
- Remove guarantee that `SetPredicate` only depends on `SetPredicate`.
- Add `AbstractWMMSet`, `DynamicDefaultWMMSet` and `MaterializedWMMSet`.
- Build instances of `SetPredicate` for unary relations in `ExecutionGraph`.

Additional:
- Fix a bug in `caat` where `Element.derivLength` was not updated.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
@@ -12,6 +12,7 @@ import com.dat3m.dartagnan.expression.booleans.*;
 import com.dat3m.dartagnan.expression.integers.*;
 import com.dat3m.dartagnan.expression.misc.ITEExpr;
 import com.dat3m.dartagnan.expression.utils.IntegerHelper;
+import com.dat3m.dartagnan.program.memory.MemoryObject;
 import com.google.common.base.VerifyException;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
@@ -163,6 +164,24 @@ public class ExprSimplifier extends ExprTransformer {
                     throw new VerifyException(String.format("Unexpected comparison operator '%s'. Missing normalization?", op));
             };
             return expressions.makeValue(cmpResult);
+        }
+
+        // ------- Operations on memory objects -------
+        if (left instanceof MemoryObject lMem && right instanceof MemoryObject rMem) {
+            final boolean sameObj = lMem.equals(rMem);
+
+            final Boolean cmpResult = switch (op) {
+                case EQ -> sameObj;
+                case NEQ -> !sameObj;
+                case LT, ULT -> sameObj ? false : null;
+                case LTE, ULTE -> sameObj ? true : null;
+                default ->
+                    throw new VerifyException(String.format("Unexpected comparison operator '%s'. Missing normalization?", op));
+            };
+
+            if (cmpResult != null) {
+                return expressions.makeValue(cmpResult);
+            }
         }
 
         return expressions.makeIntCmp(left, op, right);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
@@ -3,11 +3,8 @@ package com.dat3m.dartagnan.program.processing;
 import com.dat3m.dartagnan.expression.Expression;
 import com.dat3m.dartagnan.expression.ExpressionFactory;
 import com.dat3m.dartagnan.expression.ExpressionVisitor;
-import com.dat3m.dartagnan.expression.integers.IntLiteral;
 import com.dat3m.dartagnan.expression.processing.ExprTransformer;
 import com.dat3m.dartagnan.expression.processing.ExpressionInspector;
-import com.dat3m.dartagnan.expression.type.IntegerType;
-import com.dat3m.dartagnan.expression.type.TypeFactory;
 import com.dat3m.dartagnan.program.Function;
 import com.dat3m.dartagnan.program.IRHelper;
 import com.dat3m.dartagnan.program.Program;
@@ -39,8 +36,6 @@ import java.util.stream.Collectors;
 public class NaiveDevirtualisation implements ProgramProcessor {
 
     private static final Logger logger = LogManager.getLogger(NaiveDevirtualisation.class);
-
-    private int nextAvailableFuncAddress = 8; // We use 8-aligned addresses
 
     private NaiveDevirtualisation() {
     }
@@ -97,16 +92,16 @@ public class NaiveDevirtualisation implements ProgramProcessor {
         }
     }
 
-    private boolean assignAddressToFunction(Function func, Map<Function, IntLiteral> func2AddressMap) {
-        final IntegerType ptrType = TypeFactory.getInstance().getArchType();
-        final ExpressionFactory expressions = ExpressionFactory.getInstance();
-        if (!func2AddressMap.containsKey(func)) {
-            logger.debug("Assigned address \"{}\" to function \"{}\"", nextAvailableFuncAddress, func);
-            func2AddressMap.put(func, expressions.makeValue(nextAvailableFuncAddress, ptrType));
-            nextAvailableFuncAddress += 8;
-            return true;
+    private boolean assignAddressToFunction(Function func, Map<Function, MemoryObject> func2AddressMap) {
+        if (func2AddressMap.containsKey(func)) {
+            return false;
         }
-        return false;
+
+        final MemoryObject funcAddr = func.getProgram().getMemory().allocate(1);
+        funcAddr.setName(String.format("__funcAddr_%s", func.getName()));
+        func2AddressMap.put(func, funcAddr);
+        logger.debug("Assigned address to function \"{}\"", func);
+        return true;
     }
 
     private void applyTransformerToEvent(Event e, ExpressionVisitor<Expression> transformer) {
@@ -119,7 +114,7 @@ public class NaiveDevirtualisation implements ProgramProcessor {
         }
     }
 
-    private void devirtualise(Function function, Map<Function, IntLiteral> func2AddressMap) {
+    private void devirtualise(Function function, Map<Function, MemoryObject> func2AddressMap) {
         final ExpressionFactory expressions = ExpressionFactory.getInstance();
 
         int devirtCounter = 0;
@@ -147,8 +142,8 @@ public class NaiveDevirtualisation implements ProgramProcessor {
             final Expression funcPtr = call.getCallTarget();
             // Construct call table
             for (Function possibleTarget : possibleTargets) {
-                final IntLiteral targetAddress = func2AddressMap.get(possibleTarget);
-                final Label caseLabel = EventFactory.newLabel(String.format("__Ldevirt_%s#%s", targetAddress.getValue(), devirtCounter));
+                final MemoryObject targetAddress = func2AddressMap.get(possibleTarget);
+                final Label caseLabel = EventFactory.newLabel(String.format("__Ldevirt_%s#%s", possibleTarget.getName(), devirtCounter));
                 final CondJump caseJump = EventFactory.newJump(expressions.makeEQ(funcPtr, targetAddress), caseLabel);
                 caseLabels.add(caseLabel);
                 caseJumps.add(caseJump);
@@ -178,7 +173,7 @@ public class NaiveDevirtualisation implements ProgramProcessor {
         return !call.isDirectCall();
     }
 
-    private List<Function> getPossibleTargets(CallEvent call, Map<Function, IntLiteral> func2AddressMap) {
+    private List<Function> getPossibleTargets(CallEvent call, Map<Function, MemoryObject> func2AddressMap) {
         Preconditions.checkArgument(needsDevirtualization(call));
         return func2AddressMap.keySet().stream()
                 .filter(f -> f.getFunctionType() == call.getCallType())
@@ -210,7 +205,7 @@ public class NaiveDevirtualisation implements ProgramProcessor {
 
     private static class FunctionToAddressTransformer extends ExprTransformer {
 
-        private final Map<Function, IntLiteral> func2AddressMap = new HashMap<>();
+        private final Map<Function, MemoryObject> func2AddressMap = new HashMap<>();
 
         @Override
         public Expression visitFunction(Function function) {


dartagnan/src/main/antlr4/Cat.g4
```diff
@@ -105,7 +105,7 @@ COMMA   :   ',';
 
 DOMAIN      :   'domain';
 RANGE       :   'range';
-NEW         :   'new';
+NEW         :   '[n|N]ew';
```
hernanponcedeleon: I think the meaning of this (`new` for rels, `New` for set) is quite obscure.  Why not using `new_X()` for `X in {relation,set}`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodeSets.java
```diff
@@ -113,13 +113,39 @@ public Map<Relation, EventGraph> visitRangeIdentity(RangeIdentity rangeId) {
         return Map.of(rangeId.getOperand(), result);
     }
 
+    @Override
+    public Map<Relation, EventGraph> visitSetIdentity(SetIdentity id) {
+        final RelationAnalysis.Knowledge k1 = ra.getKnowledge(id.getDomain());
+        return Map.of(id.getDomain(),
+                news.filter((e1, e2) -> k1.getMaySet().contains(e1, e2) && !k1.getMustSet().contains(e1, e2)));
+    }
```
ThomasHaas: I think you have a `filterUnknowns` helper method that does this kind of filtering.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -416,6 +416,22 @@ public Void visitTransitiveClosure(TransitiveClosure trans) {
             return null;
         }
 
+        @Override
+        public Void visitSetIdentity(SetIdentity id) {
+            final Relation rel = id.getDefinedRelation();
+            final Relation r1 = id.getDomain();
+            EventGraph mustSet = ra.getKnowledge(rel).getMustSet();
+            EncodingContext.EdgeEncoder enc0 = context.edge(rel);
+            EncodingContext.EdgeEncoder enc1 = context.edge(r1);
```
ThomasHaas: Use proper names like "setId"/"dom" and "encSetId"/"encDom"

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -432,6 +448,24 @@ public Void visitInverse(Inverse inv) {
...
 
+        @Override
+        public Void visitProduct(CartesianProduct product) {
+            final Relation rel = product.getDefinedRelation();
+            final Relation r1 = product.getDomain();
+            final Relation r2 = product.getRange();
+            EventGraph mustSet = ra.getKnowledge(rel).getMustSet();
+            EncodingContext.EdgeEncoder enc0 = context.edge(rel);
+            EncodingContext.EdgeEncoder enc1 = context.edge(r1);
+            EncodingContext.EdgeEncoder enc2 = context.edge(r2);
```
ThomasHaas: Same as above: use better names like "dom" and "range" (or "rng").

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -150,13 +149,6 @@ public Void visitLetDefinition(LetDefinitionContext ctx) {
         if (definedPredicate instanceof Relation rel) {
             String alias = createUniqueName(name);
             wmm.addAlias(alias, rel);
```
ThomasHaas: Now we can only define relations, no? I think 
```
Relation definedPredicate = (Relation)ctx.e.accept(this);
```
is feasible now.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -133,7 +132,7 @@ public Void visitAxiomDefinition(AxiomDefinitionContext ctx) {
     }
 
     private String createUniqueName(String name) {
-        if (namespace.containsKey(name) && !nameOccurrenceCounter.containsKey(name)) {
+        if ((namespace.containsKey(name) || wmm.getRelation(name) != null) && !nameOccurrenceCounter.containsKey(name)) {
```
ThomasHaas: Why is this change necessary?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -214,15 +206,30 @@ public Object visitExpr(ExprContext ctx) {
...
+            return wmm.getOrCreatePredefinedRelation(name);
+        }
+        final Relation namedRelation = wmm.getRelation(name);
+        final Relation existingSet = namedRelation != null && namedRelation.getDefinition() instanceof TagSet s && s.getTag().equals(name) ?
+                namedRelation : wmm.getRelations().stream()
+                .filter(r -> r.getDefinition() instanceof TagSet s && s.getTag().equals(name))
+                .findAny().orElse(null);
+        if (existingSet != null) {
+            return existingSet;
+        }
```
ThomasHaas: I feel all this is pointless. Just add the generated sets to the namespace (as the old code did) then this whole piece of code can get deleted. I guess this is also the reason why you needed to change "createUniqueName".

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat/reasoning/Reasoner.java
```diff
@@ -332,5 +311,24 @@ public Conjunction<CAATLiteral> visitSetDifference(SetPredicate set, Element ele
...
+            };
+            for (Edge e : edges) {
+                // We use the first edge we find
+                if (e.getDerivationLength() < ele.getDerivationLength()) {
+                    Conjunction<CAATLiteral> reason = computeReason(inner, e);
+                    assert !reason.isFalse();
+                    return reason;
+                }
+            }
+            throw new IllegalStateException("ProjectionIdentityGraph: No matching edge is found");
```
ThomasHaas: "ProjectionSet: No matching element found."

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/basePredicates/DynamicDefaultWMMSet.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public void repopulate() {
+        // Careful: The wrapped model <getModel> might get closed/disposed while ExecutionModel as a whole is
+        // still in use. The caller should make sure that the underlying model is still alive right now.
+        final EncodingContext ctx = model.getContext();
+        final IREvaluator m = new IREvaluator(ctx, model.getModel());
+        final EncodingContext.EdgeEncoder edge = ctx.edge(relation);
+        final RelationAnalysis.Knowledge k = ctx.getAnalysisContext().get(RelationAnalysis.class).getKnowledge(relation);
+
+        if (k.getMaySet().size() < domain.size() * domain.size()) {
```
ThomasHaas: Should compare against a single "domain.size()" instead of its square.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/basePredicates/DynamicDefaultWMMSet.java
```diff
@@ -0,0 +1,50 @@
...
+                if (e != null) {
+                    simpleSet.add(e);
+                }
+            }
+        }
+    }
+
+    private Element getEdgeFromEventData(EncodingContext.EdgeEncoder edge, EventData e1, IREvaluator m) {
+        return m.hasElement(edge, e1.getEvent()) ? new Element(e1.getId()) : null;
+    }
```
ThomasHaas: Rename to "getElementFromEventData"

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -560,15 +559,15 @@ private static void addBiases(Wmm wmm, EnumSet<Baseline> biases) {
...
         final Relation frStandard = wmm.addDefinition(new Composition(wmm.newRelation(), rfinv, co));
 
-        // ([R] \ [range(rf)]);loc;[W]
-        final Relation reads = wmm.addDefinition(new SetIdentity(wmm.newRelation(), wmm.getFilter(Tag.READ)));
-        final Relation rfRange = wmm.addDefinition(new RangeIdentity(wmm.newRelation(), rf));
-        final Relation writes = wmm.addDefinition(new SetIdentity(wmm.newRelation(), Filter.byTag(Tag.WRITE)));
-        final Relation ur = wmm.addDefinition(new Difference(wmm.newRelation(), reads, rfRange));
-        final Relation urloc = wmm.addDefinition(new Composition(wmm.newRelation(), ur, loc));
-        final Relation urlocwrites = wmm.addDefinition(new Composition(wmm.newRelation(), urloc, writes));
+        // ((R \ range(rf)) * W) & loc
```
ThomasHaas: I'm not convinced that this formulation is better, because the lhs of the intersection is now always large.
I think the original one was fine, and all you need to do is move the difference inside the set identity "[R] \ [range(rf)] --> [R \ range(rf)]".

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -560,15 +559,15 @@ private static void addBiases(Wmm wmm, EnumSet<Baseline> biases) {
...
+        final Relation reads = wmm.addDefinition(new TagSet(wmm.newSet(), Tag.READ));
+        final Relation rfRange = wmm.addDefinition(new Range(wmm.newSet(), rf));
+        final Relation writes = wmm.addDefinition(new TagSet(wmm.newSet(), Tag.WRITE));
+        final Relation ur = wmm.addDefinition(new Difference(wmm.newSet(), reads, rfRange));
+        final Relation urw = wmm.addDefinition(new CartesianProduct(wmm.newRelation(), ur, writes));
+        final Relation urlocwrites = wmm.addDefinition(new Intersection(wmm.newRelation(), urw, loc));
 
-        // let fr = rf^-1;co | ([R] \ [range(rf)]);loc;[W]
+        // let fr = rf^-1;co | (((R \ range(rf)) * W) & loc)
```
ThomasHaas: Same as above.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
```diff
@@ -14,12 +15,37 @@
 public final class Relation implements Dependent<Relation> {
 
     private final Wmm wmm;
+    private final boolean unary;
     Definition definition = new Definition.Undefined(this);
     private boolean isRecursive;
     final List<String> names = new ArrayList<>();
 
-    Relation(Wmm wmm) {
+    Relation(Wmm wmm, boolean unary) {
```
ThomasHaas: Not a big fan of the boolean value here. I think an arity integer or even better, an arity enum (Arity.UNARY, Arity.BINARY), is the better choice.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/TagSet.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public <T> T accept(Visitor<? extends T> visitor) {
+        return visitor.visitTagSet(this);
+    }
+
+    @Override
+    public String getTerm() {
+        return super.getTerm();
+    }
```
ThomasHaas: Why overwrite it, if you just call the super method?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
xeren: I encountered `Already bound name W`, and others like `ext` and figured that I probably should check the global namespace as well when computing unique names.  A better way would be to add relations to the local namespace on every occasion.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
xeren: I agree.  This change's purpose was for debugging.  It should be as you described now.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
ThomasHaas: I think the only case where relations are "secretely" generated (i.e. outside of the parser) is when invoking `wmm.getPredefinedRelation` (or whatever it is called). This one is a little annoying. I think for binary relations, we just opted to never generate named subrelations (except for the hidden one's like `idd`) so that there was never a name conflict.
I guess if you do the same with tag sets generated there, it would should work correctly. At least I don't see why the code that worked for binary relations shouldn't just work out-of-the-box for unary relations anymore.

xeren: Another `Already bound name F` occurred, because in `opencl.cat`, `dynamic_tag(X)` is defined before `F` is declared, which is used in that function.  `dynamic_tag` is called twice, so the namespace lookup misses twice, leading to a redefinition of `F`.
Maybe `createUniqueName` should always check `wmm`, rather than `namespace`, to not be sensitive to local scopes and to guarantee uniqueness in `wmm`.

ThomasHaas: I cannot really follow. You removed the update of the namespace in `visitExprBasic` and that is why you create relations without registering their names in the namespace. If you do this correctly, and assuming that `wmm.getOrCreatePredefinedRelation(k)` does not create named subrelations for tag set, I think it should work just fine.
What breaks if you use the original 1-line code but replace the `Filter.byTag(k)` with `addDefinition(new TagSet(wmm.newSet(k), k))`?

xeren: > What breaks if you use the original 1-line code but replace the `Filter.byTag(k)` with `addDefinition(new TagSet(wmm.newSet(k), k))`?

There are two loosely-dependent namespace systems at work here: `Wmm.getRelation(String)` and `VisitorCat.namespace`.   During parsing, I get collisions in `Wmm.addAlias(String,Relation)`, because the previous call to `VisitorCat.createUniqueName(String)` only checks the other namespace system.

Previously we used `Wmm.addFilter(Filter)` to bind a name to a set.  That is a third namespace system and it does not check for name collisions.  Now that sets and relations have merged, these collisions are detected.  This also happens with the oneliner.

> You removed the update of the namespace in visitExprBasic and that is why you create relations without registering their names in the namespace.

I removed the update in `visitExprBasic`, because I added an additional lookup with `wmm.getRelation(name)`.  With this, binding the name in the local namespace became obsolete.  `let` instructions still modify that namespace.

ThomasHaas: I want to understand how collisions even happen. I understand that there are two namespaces (the parser and the wmm), but from my understanding every name in the Wmm should also appear in the parser's namespace if `visitExprBasic` and `visitLetExpr` register all names they encounter, no? The only exception I can think of is if `wmm.getOrCreatePredefinedRelation(k)` creates named subrelations which then appear in the wmm's namespace but not in the parser's.

Either way, my main concern was that something might go wrong if the parser's namespace is not aligned with the wmm's namespace, especially when functions that capture only the parser's namespace are involved. But after thinking about it for a little bit, I don't think anything breaks there.

xeren:
> I know that in some domains (e.g., mathematics) relations can be also unary, but I think the terms "relation" and "set" are much more common in the domain of WMMs.

> I do not see why that would contradict my proposal. That class could be called Predicate and then methods like isUnaryRelation() and isBinaryRelation() could be renamed to isSet() and isRelation() to match the common terms used in WMMs

With currently 569 usages of `Relation` in `development`, that will cause a lot of conflicts.  I'd rather do the renaming in another PR.  I can rename `isSet()` and `isRelation()` immediately, though.
> A quite annoying problem is that the `Predicate<T>` class exists in the java standard namespace.
That is why I used the name `CAATPredicate` in CAAT to avoid name clashes.

Then I throw `WmmPredicate` and `EventPredicate` into the ring.  The first would be a double naming, as the package already says `wmm`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
```diff
@@ -43,39 +43,39 @@ public Arity getArity() {
...
     }
 
     /**
      * @throws IllegalArgumentException This instance does not describe an event set in executions.
      */
-    public void checkUnaryRelation() {
-        Preconditions.checkArgument(isUnaryRelation(), "Non-unary relation %s.", this);
+    public void checkSet() {
+        Preconditions.checkArgument(isSet(), "Non-unary relation %s.", this);
     }
```
ThomasHaas: I'm not a big fan of public instance methods that throw exceptions like this. I think the implementation lacks too much context to give a proper exception (is the relation really an argument?) with a proper message (what is the context where the relation was expected to be a set?). At the same time, it being an instance method makes it feel like the de facto way of checking the relation and raising an exception (which should not be the case, since the caller knows more about the situation).

I think such methods are fine if they are static (or private), possibly even in a helper class (I used this pattern in `ExpressionHelper`). The helper class is not necessary, but the method being static makes it feel more like a utility rather than the de facto way of checking a relation to me.
It is my subjective opinion though. Maybe someone else should give their opinion as well. @hernanponcedeleon?

Also, I would probably go with the names `checkIfSet` and `checkIfRelation` to make it even more clear what is actually checked for.

Lastly, I would get rid of all the unnecessary Java Docs comments.
Regular comments on methods are already too much most of the time if they just state the obvious. And I think Java Doc comments are even worse in that regard.
I think we should only introduce them once we really care about providing a public API that can be consumed by other developers. But then we also need to think about what makes up the public API in the first place (not every public method is really the API).

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/IREvaluator.java
```diff
@@ -112,7 +112,17 @@ public TypedValue<IntegerType, BigInteger> size(MemoryObject memoryObject) {
     // ====================================================================================
     // Memory Model
 
+    public boolean hasElement(Relation rel, Event a) {
+        Preconditions.checkArgument(rel.isSet(), "Non-unary relation %s", rel);
+        return TRUE.equals(smtModel.evaluate(ctx.edge(rel, a, a)));
```
hernanponcedeleon, line 117: AT the SMT level, why are we encoding set membership as relation membership? Wouldn't it be better to extend the encoding context with something like `ctx.in(rel, a)`?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -214,15 +204,21 @@ public Object visitExpr(ExprContext ctx) {
 
     @Override
     public Object visitExprNew(ExprNewContext ctx) {
-        return addDefinition(new Free(wmm.newRelation()));
+        final boolean unary = ctx.call.getText().equals("New");
+        return addDefinition(new Free(wmm.newRelation(unary ? Relation.Arity.UNARY : Relation.Arity.BINARY)));
```
hernanponcedeleon: As I already wrote in my initial review, this is ugly. Our grammar should properly distinguish between creating a new set or creating a new relation and we should have two different visitors

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -334,17 +321,19 @@ public Relation visitExprTransRef(ExprTransRefContext c) {
     }
 
     @Override
-    public Relation visitExprDomainIdentity(ExprDomainIdentityContext c) {
-        Relation r0 = getAndResetRelationToBeDefined().orElseGet(wmm::newRelation);
+    public Relation visitExprDomain(ExprDomainContext c) {
+        checkNoRecursion(c);
```
hernanponcedeleon: What exactly is the problem in supporting recursion in domain/range operations?

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/basePredicates/DynamicDefaultWMMSet.java
```diff
@@ -0,0 +1,49 @@
+package com.dat3m.dartagnan.solver.caat4wmm.basePredicates;
+
+import com.dat3m.dartagnan.encoding.EncodingContext;
+import com.dat3m.dartagnan.encoding.IREvaluator;
+import com.dat3m.dartagnan.solver.caat.predicates.sets.Element;
+import com.dat3m.dartagnan.verification.model.EventData;
+import com.dat3m.dartagnan.wmm.Relation;
+import com.dat3m.dartagnan.wmm.analysis.RelationAnalysis;
+
+// A default implementation for any encoded relation, e.g. base relations or non-base but cut relations.
```
hernanponcedeleon: relation -> set (we should at least get comments to use the correct term in this PR)

Do we also have cutting for sets?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/SetIdentity.java
```diff
@@ -1,21 +1,24 @@
...
 public class SetIdentity extends Definition {
 
-    private final Filter filter;
+    private final Relation domain;
 
-    public SetIdentity(Relation r0, Filter s1) {
-        super(r0, "[" + s1 + "]");
-        filter = s1;
+    public SetIdentity(Relation r0, Relation r1) {
+        super(Relation.checkIsRelation(r0), "[%s]");
```
hernanponcedeleon, line 13: what is `%s` matched with?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
ThomasHaas: I agree, there shouldn't be a problem (at least in principle).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/SetIdentity.java
ThomasHaas: Should be the inner relation `r1`. IIRC, the string provided there is always a pattern that is filled out with the dependencies.

dartagnan/src/main/antlr4/Cat.g4
ThomasHaas: I agree, but I would probably prefer `newSet` and `newRelation` without the underscore.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/IREvaluator.java
ThomasHaas: It would certainly be cleaner, but I guess it comes with the most changes in the code base if you want to consistently distinguish between sets and relations.
For example, whenever you iterate over a may/must-set (which are always binary) and you want to encode the elements, you have to check if its a set or a relation an call the appropriate method in `EncodingContext`.
But maybe this case-distinction can be put into some helper method that than invokes the correct method on `EncodingContext`.

Either way, I agree that a clean distinction between set/relation (element/edge) would be best.
I also understand that `RelationAnalysis` and consequently may/must-sets do not want to make this distinction.
The question is where is the best place to bridge this gap in the code base? I'm not sure.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
ThomasHaas: You mean two different visitor classes, one for sets, one for relations? Or just two different visitor methods?
If you really mean classes, then I don't think this will work nicely due to the mutual recursion between the two.

hernanponcedeleon: two different methods. If the grammar does something like
```
newRel: NEWREL
newSet: NEWSET
...

NEWREL: `NewRelation`
NEWSET: `NewSet`
```
Then we will have methods `visitNewRel` and `visitNewSet`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
xeren: Our parser created the relations in `let rec` up-front as binary relations.  So, an `IllegalArgumentException` would have been thrown in the constructors of `Domain` and `Range`, instead of a `ParsingException` here.
I have now added a simple visitor to guess the expected arity.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/basePredicates/DynamicDefaultWMMSet.java
xeren: I think so, because `Difference` is agnostic to the arity.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -235,6 +236,7 @@ private final class RelationEncoder implements Constraint.Visitor<Void> {
                 SyncWith.class,
                 SameVirtualLocation.class, // FIXME?!
                 Empty.class,
+                TagSet.class,
                 // Static because the underlying sets are static:
                 SetIdentity.class,
                 CartesianProduct.class
```
ThomasHaas: You can now remove this :)

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/IREvaluator.java
xeren: Currently, this would be the only usage for `EncodingContext.in(Relation,Event)`.  All other passages would require something like `EdgeEncoder.encodeSetMembership(Event)`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/SetIdentity.java
xeren: Yes.  The operation became a unary definition.  With this pattern, the term string can react to changes to the operand's name or term.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
xeren: This is now implemented.  Note that it conflicts with #878.

hernanponcedeleon (commented):
LGTM. 

Unless @ThomasHaas has further comments, we can merge

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
```diff
@@ -96,14 +95,14 @@ private void constructMappings() {
...
+                        predicateToRelationMap.put(relation, graph);
                     }
                 }
                 for (DependencyGraph<Relation>.Node node : component) {
                     Relation relation = node.getContent();
                     if (relation.isRecursive()) {
                         // side effect leads to calculation of children
-                        RecursiveGraph graph = (RecursiveGraph) relationGraphMap.get(relation);
+                        RecursiveGraph graph = (RecursiveGraph) predicateToRelationMap.get(relation);
                         graph.setConcreteGraph(createGraphFromRelation(relation));
```
ThomasHaas, line 109: I guess this will not work for recursive sets. You don't need to implement that feature (who defines recursive sets anyways), but you can double check that we throw an exception here?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/Range.java
```diff
@@ -5,15 +5,13 @@
 
 import java.util.List;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
-public class DomainIdentity extends Definition {
+public class Range extends Definition {
```
ThomasHaas: Do we really need `Range` and `Domain` as two classes? Can't we do it like Refinement does and have one `Projection` class with an enum member deciding between domain and range projection?
I think it could reduce some code duplication.

PR #910, opened 2025-06-30
Execution scope for control barrier

natgavrilenko:


dartagnan/src/test/java/com/dat3m/dartagnan/spirv/opencl/barrier/scope/SpirvAssertionsTest.java
```diff
@@ -0,0 +1,85 @@
...
+        this.programPath = getTestResourcePath("spirv/opencl/barrier/scope/" + file);
+        this.bound = bound;
+        this.expected = expected;
+    }
+
+    @Parameterized.Parameters(name = "{index}: {0}, {1}, {2}")
+    public static Iterable<Object[]> data() throws IOException {
+        return Arrays.asList(new Object[][]{
+                {"barrier-inscope-wg.spvasm", 1, PASS},
+                {"barrier-not-inscope-wg.spvasm", 1, FAIL},
```
hernanponcedeleon, line 52: `opencl.cat` does not refer to `sw` and that is why we do not have the corresponding tests here, right?

litmus/VULKAN/Barrier/barrier-inscope.litmus
```diff
@@ -0,0 +1,9 @@
+Vulkan barrier-inscope
+{
+x=0;
+}
+ P0@sg 0, wg 0, qf 0      | P1@sg 1, wg 0, qf 0      ;
+ st.av.dv.sc0 x, 1        | cbar.acq_rel.dv.semsc0 1 ;
+ cbar.acq_rel.dv.semsc0 1 | ld.vis.dv.sc0 r0, x      ;
```
hernanponcedeleon, line 7: `dv` is the scope for the memory part of the barrier right? Is the control part always `wg` for the litmus format?

litmus/VULKAN/Kronos-Group/scopeaccum.litmus
```diff
@@ -4,8 +4,8 @@ Vulkan scopeaccum
 P2:r0=0;
 x=0;
 }
- P0@sg 0, wg 0, qf 0                       | P1@sg 1,wg 0, qf 0                         | P2@sg 0,wg 1, qf 0                         ;
- st.av.dv.sc0 x, 1                         | cbar.acq_rel.wg.semsc0 1                   | cbar.acq_rel.dv.semsc0 2                   ;
- cbar.acq_rel.wg.semsc0 1                  | cbar.acq_rel.dv.semsc0 2                   | ld.vis.dv.sc0 r0, x                        ;
+ P0@sg 0, wg 0, qf 0      | P1@sg 0, wg 0, qf 0      | P2@sg 1, wg 0, qf 0      ;
+ st.av.wg.sc0 x, 1        | cbar.acq_rel.sg.semsc0 1 | cbar.acq_rel.wg.semsc0 2 ;
+ cbar.acq_rel.sg.semsc0 1 | cbar.acq_rel.wg.semsc0 2 | ld.vis.wg.sc0 r0, x      ;
```
hernanponcedeleon, line 9: The idea of the tests in `Kronos-group` (we actually have a typo in the folder name) and the Data-race folders was to be in sync with the "official" alloy tests. Why are you changing these?

dartagnan/src/test/java/com/dat3m/dartagnan/spirv/opencl/barrier/scope/SpirvAssertionsTest.java
natgavrilenko: Yes, the model has no subgroup scope. In principle, we don't need to have it in the model to use as an execution scope, but I don't see much value to implement subgroup scope for execution only. Probably, Haining can add both at some point.

litmus/VULKAN/Barrier/barrier-inscope.litmus
natgavrilenko: Yes, 'dv' is the memory scope, and in litmus format execution scope is always workgroup. But nothing prevents a user to use different ids in different subgroups an have a de-facto subgroup barrier.

litmus/VULKAN/Kronos-Group/scopeaccum.litmus
natgavrilenko: The problem with the original test is that workgroup barrier synchronizes across different workgroups, which cannot be the case in real programs (according to vulkan spec, execution scope can be only subgroup or workgroup). I think it should be updated in alloy too.

hernanponcedeleon: 1. In a previous comment you said the scope we put in the syntax is the "memory scope" and now you are talking about the "execution scope".
2. I only found this in the spec
> When Execution is Workgroup or larger, behavior is undefined unless all invocations within Execution execute the same dynamic instance of this instruction.

Where exactly does it say that the execution scope can only be subgroup or workgroup?

natgavrilenko: VUID-StandaloneSpirv-None-04636
Scope for execution must be limited to Workgroup or Subgroup

https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#spirvenv-module-validation

hernanponcedeleon: Ok, this clarifies 2. 

However, the changes you did to the syntax are in the memory (not execution) scope.

natgavrilenko: Looks like there is some miscommunication. Check the original tests (scopeaccum and test6). They rely on control barriers synchronizing across workgroups. The last litmus test (test7) is similar to test6, except that it should fail due to mismatching storage class semantics (but it also attempts to use control-barriers across workgroups).

The only change in these tests is to:
- change subgroup memory scope to workgroup memory scope
- change device memory scope to workgroup memory scope
- put communicating threads into the same workgroup, so that control barrier synchronization makes sense

PR #911, opened 2025-06-30
Update storage class semantics in memory-scopes test

natgavrilenko:
Added workgroup storage class semantics to the instructions. Otherwise, mismatching semantics is not sufficient in the version of vulkan model with relaxed semsc-to-semsc ordering.


hernanponcedeleon:
Just to be sure I understand your comment correctly: the idea is that this test is supposed to give different results on two versions of the model, but right not this is not the case because for that we need workgroup storage class semantics.

natgavrilenko:
No, the idea of this test is just to test storage classes. But now accesses to workgroup variables don't have workgroup storage class semantics, which was overlooked in the original version (it is supposed to be a simple test for different storage classes, not for different semantics variations).
That's why this test starts failing if we remove ordering of semsc-to-semsc for memory accesses (workgroup storage class is accesses without workgroup storage class semantics)

PR #912, opened 2025-07-09
Remove unnecessary constraint from vulkan model

natgavrilenko:


hernanponcedeleon:
The alloy model states _"In a race-free program, any reads and writes at the same location must be ordered by locord and/or mutordatom"_, which mean this property is not enforced for racy programs.

Am I right that by having `locord-complete` as part of the consistency check we were potentially missing to detect data races? @natgavrilenko did you encounter any test where this was the case?

natgavrilenko:
Yes, "In a race-free program, any reads and writes at the same location must be ordered by locord and/or mutordatom" is the definition of the data-race constraint. 
The removed locordcomplete constraint checks that mutordatom and locord relations don't have edges that are not in loc, which we cannot have by the definition of locord and mutordatom.

PR #913, opened 2025-07-15
Fix PTX models

hernanponcedeleon:
We made a mistake in the definition of `observation / sync` when we ported the alloy model to cat. The test `rmw-chain-1` shows that our current model is wrong.

I came up with the other tests when I was trying to figure it out if the other usage of `observation` should also be transitively closed. I got confirmation from the NVIDIA folks that the current behavior is the intended one.

### DIFF
--- a/litmus/PTX/Manual/rmw-chain-1.litmus
+++ b/litmus/PTX/Manual/rmw-chain-1.litmus
@@ -1,0 +1,11 @@
+PTX RMW-chain-1
+{
+P1:r0=0;
+P2:r1=0; P2:r2=0;
+x=0; y=0;
+}
+ P0@cta 0,gpu 0       | P1@cta 0,gpu 0                | P2@cta 0,gpu 0       ;
+ st.relaxed.gpu x, 1  | atom.relaxed.gpu.add r0, y, 1 | ld.acquire.gpu r1, y ;
+ st.release.gpu y, 1  |                               | ld.relaxed.gpu r2, x ;
+
+exists (1:r0 = 1 /\ 2:r1 = 2 /\ 2:r2 = 0)
--- a/litmus/PTX/Manual/rmw-chain-2a.litmus
+++ b/litmus/PTX/Manual/rmw-chain-2a.litmus
@@ -1,0 +1,12 @@
+PTX RMW-chain-2a
+{
+P0:r0=0;
+P1:r1=0;
+P2:r2=0;
+x=0; y=0;
+}
+ P0@cta 0,gpu 0       | P1@cta 0,gpu 0                | P2@cta 0,gpu 0       ;
+ ld.acquire.gpu r0, y | atom.relaxed.gpu.add r1, x, 0 | ld.relaxed.gpu r2, x ;
+ st.weak x, 1         |                               | st.release.gpu y, 1  ;
+
+exists (0:r0 = 1 /\ 1:r1 = 1 /\ 2:r2 = 1)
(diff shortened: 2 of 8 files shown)


ThomasHaas:
Does it mean the other usage of `observation` should not be transitively closed, i.e., it would be wrong to do so?
If so, then the original recursive formulation was wrong, right?

hernanponcedeleon:
> Does it mean the other usage of `observation` should not be transitively closed, i.e., it would be wrong to do so? 

It means that if you use the new definition `let observation = (morally-strong & rf) | rmw`, then your should used `observation+` in `sync`, but `observation` in `cause`.

>If so, then the original recursive formulation was wrong, right?

No. AFAIK the old recursive definition is not equivalent to the transitive closure. Defining `let rec observation = (morally-strong & rf) | (observation; rmw; observation)` and using `observation` both in `sync` and `cause` give the same results (at least for these tests).

litmus/PTX/Manual/rmw-chain-1.litmus
```diff
@@ -0,0 +1,11 @@
+PTX RMW-chain-1
+{
+P1:r0=0;
+P2:r1=0; P2:r2=0;
+x=0; y=0;
+}
+ P0@cta 0,gpu 0       | P1@cta 0,gpu 0                | P2@cta 0,gpu 0       ;
```
natgavrilenko: Minor thing: a space is missing after the comma "cta 0,gpu"

litmus/PTX/Manual/rmw-chain-2a.litmus
```diff
@@ -0,0 +1,12 @@
+PTX RMW-chain-2a
```
natgavrilenko: The change to the model and chains1 test look good to me, but I don't get what is the purpose of the chains2-x tests. As far as I see, they are essentially variations of an LB pattern with a suffix:
- with the current assertions, chains2a and chains2c can execute the whole P1 at the end,
- in chains2b P1 executes only a load and no store.

If we want to check the impact of release-acquire semantics, shouldn't the test otherwise be the same as chains1? Or am I missing something?

hernanponcedeleon: You are right. Test `chain-1` was intended to show that our current model is wrong. For `chain-2X`, I started trying to do something (a), then did some changes to try something else (b) and end up with a non-sense test.

(a) The purpose of this test was to check if the change from v6.0 to v7.5 should have used `observation+` also in the definition of `cause`. In fact, the new `SB+RMW-a` gives a different result for `let cause = observation?...` and `let cause = observation+?...`. You are right that it is similar to `chain-1` in the sense that an "intervening" RMW might allow an otherwise forbid behavior. Test `chain-2c` (now called `SB+RMW-c`) just shows that this behavior relies on the store being weak, i.e., relaxed-atomic is enough to forbid the behavior even in the presence of the RMW.

(b) I wanted to make the RMW such that it was "obvious" that it should not play a role in changing the behavior of the test. That is why I went for `+0`, but I missed the fact that this allowed to fully execute P1 at the end to fulfill the assertions.  Having `+0` has lead to [unsound optimizations in compilers](https://github.com/llvm/llvm-project/issues/101551) and this is what I wanted to test with `chain-2b`.

I changed the test now such that `SB+RMW-a` shows that an intervening RMW allows the otherwise forbidden behavior of `SB+RMW-b`, and making the weak-store a relaxed-atomic also forbids the behavior in `SB+RMW-c`.

natgavrilenko: I think the tests are LB+RMW not SB+RMW (SB starts with a store and LB starts with a load), otherwise looks good

natgavrilenko (approved):

PR #914, opened 2025-07-21
Return proper exit code for bounded verification

hernanponcedeleon:
We are currently returning the wrong exit-code when loops are not unrolled. This was causing problems with vsyncer in which it would return `CheckOK` with the initial (sometimes insufficient) unrolling bound.


ThomasHaas (approved):
LGTM

PR #916, opened 2025-07-23
Fix Mem2Reg failing to zero out memory

ThomasHaas:
Mem2Reg didn't respect `Alloc.doesZeroOutMemory`, potentially replacing allocations by uninitialized registers rather than zero-initialized ones.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
@@ -76,7 +76,20 @@ public class MemToReg implements FunctionProcessor {
     private void promoteAll(Function function, Matcher matcher) {
         // Replace every unmarked address.
         final Map<RegWriter, Promotable> promotableObjects = collectPromotableObjects(function, matcher);
-        final Map<Event, List<Event>> updates = new HashMap<>(Maps.toMap(promotableObjects.keySet(), k -> List.of()));
+        final Map<Event, List<Event>> updates = new HashMap<>();
+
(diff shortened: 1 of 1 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
```diff
@@ -76,7 +76,20 @@ private Matcher analyze(Function function) {
...
         final Map<RegWriter, Promotable> promotableObjects = collectPromotableObjects(function, matcher);
-        final Map<Event, List<Event>> updates = new HashMap<>(Maps.toMap(promotableObjects.keySet(), k -> List.of()));
+        final Map<Event, List<Event>> updates = new HashMap<>();
+
+        // Compute replacement of allocation sites:
+        for (final Map.Entry<RegWriter, Promotable> entry : promotableObjects.entrySet()) {
+            final Alloc alloc = (Alloc) entry.getKey();
+            final List<Event> replacement = alloc.doesZeroOutMemory() ?
+                    entry.getValue().replacingRegisters.values().stream()
+                            .map(reg -> (Event) EventFactory.newLocal(reg, expressions.makeGeneralZero(reg.getType())))
```
hernanponcedeleon, line 86: Is the cast really necessary?

ThomasHaas: Yeah, unfortunately. Otherwise, it will return a `List<Local>` which is not compatible with `List<Event>`.
An alternative is to change the type to `List<? extends Events>` everywhere, so that lists of subtypes are allowed.

PR #918, opened 2025-07-24
Refactor + Report reads from uninit registers only for consistent executions

ThomasHaas:


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java
```diff
@@ -174,7 +169,6 @@ private void write(String content) {
             try (FileWriter writer = new FileWriter(file, true);
                     PrintWriter printer = new PrintWriter(writer)) {
                 printer.append(removeDuplicatedDeclarations(content));
-                printer.close();
```
hernanponcedeleon, line 177: Is `PrinterWriter` auto closable?

ThomasHaas: Yes, otherwise you couldn't use it in the `try-with-resources` syntax.

PR #922, opened 2025-08-13
Made GEPToAddition recursive to handle nested GEPExpr

marcrobm:
Previously, the GEPToAddition pass did not account for nested GEPExprs thereby leaving remaining GEPExpr.
Within the linux kernel, the container_of(...) makro actually results in generation of nested expressions.
Below is a minimal example that triggers the GEPToAddition pass to not remove a GEPExpr.
```
#include <stddef.h>

struct member { int x; };
struct parent {
    int x;
    struct member y;
    int z;
};

int main()
{
    struct member m; 
    struct parent *p = (struct parent *)((char *) &m - ((size_t) &((struct parent *)0)->y));
    p->z = 0;
    return 0;
}
```


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/GEPToAddition.java
```diff
@@ -97,7 +97,7 @@ public Expression visitGEPExpression(GEPExpr gep) {
 
             final Expression base = gep.getBase().accept(this);
             final Expression castOffset = expressions.makeCast(totalOffset, base.getType(), true);
-            return expressions.makeAdd(base, castOffset);
+            return expressions.makeAdd(base, castOffset).accept(this);
```
ThomasHaas: That's interesting. We do recursively resolve nested GEPs that occur in the base, but we do not do it when they occur in the offset of another GEP. 

This might be a bit nit-picky, but I would do the recursion inside the loop that traverses the offsets, i.e., do `ìndex = index.accept(this)` as first line of the for loop. That might be a bit cleaner and avoids traversing the same expressions multiple times.

Either way, thanks for the PR!

marcrobm: Thanks for the quick feedback. 
In addition to adding `index = index.accept(this);` as the first line of the for loop traversing the offset, the initial assignment to `totalOffset` also has to be considered (`indices.get(0).accept(this)` instead of `indices.get(0)`). Alternatively we could just visit `totalOffset` itself. 
`final Expression castOffset = expressions.makeCast(totalOffset.accept(this), base.getType(), true);`

ThomasHaas: Right, the first one has to be considered too. Actually, my initial idea was also to use `final Expression castOffset = expressions.makeCast(totalOffset.accept(this), base.getType(), true);`, but I think it is mathematically cleaner to recursively resolve inner GEPs first and then handle the outer one.
Again, this is totally nit-picky and I would accept either solution.
So feel free to pick either one and I will merge :)

PR #923, opened 2025-08-14
Add support alloc-free

CapZTr:
This PR adds support to detect heap memory safety issues in:

1. adding new event `MemFree` with new tag `FREE`
2. modifying event `MemAlloc` (renamed from `Alloc`), which has new tag `ALLOC` when it is a heap allocation
3. enhancing alias analysis to handle the two events above
4. adding new event `Termination` with tag `TERMINATION` which gets executed if and only if the program terminated normally
5. adding new base relation `AllocPtr` and `AllocMem` and derived ones into models `c11` and `rc11`, NOTE: `vmm` is still WIP


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/LazyRelationAnalysis.java
```diff
@@ -287,6 +287,29 @@ public RelationAnalysis.Knowledge visitCASDependency(CASDependency definition) {
...
+            long start = System.currentTimeMillis();
+            RelationAnalysis.Knowledge base = nativeInitializer.visitAllocPtr(definition);
+            EventGraph may = ImmutableMapEventGraph.from(base.getMaySet());
+            EventGraph must = ImmutableMapEventGraph.from(base.getMustSet());
+            time(definition, start, System.currentTimeMillis());
+            return new RelationAnalysis.Knowledge(may, must);
+        }
+
+        @Override
+        // TODO: May and must sets of AllocMem can become very large for some programs.
```
natgavrilenko: I will do a detailed review a bit later this week, but just a quick comment for now.

This should use LazyEventGraph. ImmutableEventGraph is the same explicit representation as in the native relation analysis, just with an immutable collection. It could be more efficient to use it in derived relations, but it may still require a lot of memory to store a may set of allocmem. With a LazyEventGraph, it will compute a set each time on demand. This might be a bit slower, but will use less memory. 

Also, if alias analysis is thread-safe now, could you also change visitSameLocation to use LazyEventGraph? It used to be one of the most problematic relations in terms of the size of a may set. Hopefully, it should get better with a lazy graph.

And could you also remove the TODO here?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemAlloc.java
```diff
@@ -33,8 +33,9 @@ public final class MemAlloc extends AbstractEvent implements RegReader, RegWrite
     // This will be set at the end of the program processing.
     private transient MemoryObject allocatedObject;
 
-    public MemAlloc(Register resultRegister, Type allocType, Expression arraySize, Expression alignment, boolean isHeapAllocation,
-                 boolean doesZeroOutMemory) {
+    public MemAlloc(Register resultRegister, Type allocType, Expression arraySize, Expression alignment,
+            boolean isHeapAllocation, boolean doesZeroOutMemory) {
+        super(resultRegister, "alloc");
```
ThomasHaas: This is not going to work. The address expression (or any expression for that matter) is evaluated before the execution of the event, but at that point in time, `resultRegister` will not contain the correct address (it is most likely even uninitialized).

That is why I said you probably need to add a new event. Either a dummy event that can be added after the alloc so it can refer to the result register:
```
p <- alloc(...) // No memory event. Using "p" as address here is wrong.
MemAlloc(p); // Visible allocation marker. Memory event with address "p".
```
Or, a new `MemAlloc` that replaces `Alloc` after the memory allocation pass ran, i.e., after memory objects have been allocated. Because then `MemAlloc` can use the allocated `MemoryObject` as address.

The first solution is certainly easier to implement for now (probably like 10 lines of code).
For the second solution, I can also imagine `MemAlloc` just inheriting from `Alloc` (but then you need to implement the `MemoryCoreEvent` interface) and adding the new address field. But I guess quite a bit of existing code would need to get updated.

natgavrilenko: Please give a minimal example of a program where it is not going to work

CapZTr: Here I use `resultRegister` in constructor and later `address` will be set to `allocatedObject` when `setAllocatedObject` is called.

ThomasHaas: > Please give a minimal example of a program where it is not going to work

No frontend will generate such a program, but code like this would fail to get optimized
```
int64 p <- ... // alloc or a local assigment
int64 p <- alloc(...)
```
The second alloc appears to make use of `p` which keeps the previous assignment alive and thus it cannot get deleted although it is dead. It will probably also generate some encoding for that, but I don't think it will result in soundness errors. Though I can also imagine it breaking the alias analysis if we ever want to run it before memory allocation (which is reasonable).
Apart from that, the conceptual error is important in itself IMO.

> Here I use resultRegister in constructor and later address will be set to allocatedObject when setAllocatedObject is called.

I know, but that only partly addresses the issue. On construction, the alloc event is essentially broken until the memory allocation pass goes through. Creating a conceptually broken object and then adding a workaround will only cause problems in the long run. At the very least, it deserves comments that document the "hacks" and explain why the hack is not problematic in practice. Though I would advice against it, because then you kinda need to understand what the rest of Dartagnan does and anyone who writes passes in the future can easily run into the trap (e.g., alias analysis running before memory allocation).
In 99% of the cases though, the wrong address will only cause optimization obstacles if anything.

CapZTr: >e.g., alias analysis running before memory allocation

I'm confused about it: how can the two solutions suggested by you help with this case?

ThomasHaas: Currently (and in both of my suggested solutions), the `Alloc` event is no memory event, so no pass/analysis can accidentally treat it as such. This also means that an alias analysis that uses `getAddress()` to determine the accessed address of each memory event is going to do the right thing.
In your version, `getAddress()` gives a wrong value before memory allocation, so the analyses have to special case on `Alloc` and if they don't, they may produce wrong results without knowing.
And there is no way that, a couple month down the line, everyone remembers that `Alloc` needs special treatment... Even worse, because it will likely work in many cases (e.g., whenever an analysis runs after memory allocation), it ends up being a time bomb for a very long time.

Btw. the "cheap" solution with the extra allocation marker is somewhat implemented [here](https://github.com/LightOneM/Dat3M/blob/Pointer%26Free/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/AllocM.java) by a former student of mine. It also uses the `sobj` (same object) relation to formulate memory errors:
```
flag ~empty (([M];sobj;[MFR]) \ hb*) as possible-access-after-free
let alloc-free-matches = [MAL];loc;[MFR]
flag ~empty [MFR] \ [range(alloc-free-matches)] as invalid-free
flag ~empty [MAL] \ [domain(alloc-free-matches)] as memory-leak
```
I have not checked any details of the code base nor the CAT files, so take them with a grain of salt.

Lastly, I think you shouldn't inherit from `GenericMemoryEvent`. If you want to use that class, you can just create instances of it with the `ALLOC/FREE` tag. If you create separate classes (which is totally fine), then they should probably extend `AbstractMemoryCoreEvent` instead.

cat/c11.cat
```diff
@@ -100,4 +100,17 @@ flag ~empty dr as unsequenced_race
...
+ * Base relations:
+ * allocptr - relates (ALLOC | FREE) -> (FREE) when both events use the same pointer
+ * allocmem - relates (ALLOC) -> (M) when the second event accesses the memory allocated by the first event
+ *)
+flag ~empty (((ALLOC * FREE) & allocptr) \ hb) as alloc-race
+flag ~empty (((FREE * FREE) \ id) & allocptr) as double-free
+flag ~empty ([FREE] \ [range(allocptr & (ALLOC * FREE))]) as free-without-alloc
+flag ~empty (allocmem \ hb) as use-before-alloc
+flag ~empty ((allocmem^-1 ; (allocptr & (ALLOC * FREE))) \ hb) as use-after-free
+flag ~empty [domain(ALLOC * TERMINATION)] \ [domain(allocptr)] as alloc-without-free
```
ThomasHaas, line 116: I think you could put this into a separate .cat file and just include it in the other models. I guess it makes sense for any (language-level) memory model that defines a `hb` relation.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -172,6 +177,10 @@ public static Init newInit(MemoryObject base, int offset) {
         return init;
     }
 
+    public static Termination newTerminationEvent() {
+        return terminationEvent;
+    }
```
ThomasHaas, line 182: Don't return an existing instance; just create a new one.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/MemFree.java
```diff
@@ -0,0 +1,21 @@
...
+        addTags(Tag.FREE);
+        removeTags(Tag.MEMORY);
+    }
+
+    @Override
+    public MemFree getCopy() {
+        MemFree other = new MemFree(address);
+        other.setFunction(this.getFunction());
+        other.copyAllMetadataFrom(this);
+        return other;
```
ThomasHaas, line 23: I would prefer a proper copy constructor. Also, don't set the function of the newly created event: this breaks invariants we try to uphold.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/core/Termination.java
```diff
@@ -0,0 +1,40 @@
...
+        final BooleanFormula exitReached = bmgr.and(program.getThreads().stream()
+                .filter(t -> !t.equals(getThread()))
+                .map(t -> bmgr.equivalence(ctx.execution(t.getEntry()), ctx.execution(t.getExit())))
+                .toList());
+        final BooleanFormula terminated = program.getThreadEventsWithAllTags(Tag.NONTERMINATION).stream()
+                .map(CondJump.class::cast)
+                .map(jump -> bmgr.not(ctx.jumpTaken(jump)))
+                .reduce(bmgr.makeTrue(), bmgr::and);
+        return bmgr.implication(ctx.execution(this), bmgr.and(ctx.controlFlow(this), exitReached, terminated));
+    }
```
ThomasHaas, line 39: I would move this encoding into `ProgramEncoder`. To make it accessible from everywhere, I think it would be reasonable to have `EncodingContext` provide a `hasProgramTerminated()` function that returns a fresh boolean variable. `ProgramEncoder` can then encode `hasProgramTerminated <=> "actual encoding"` and `execution(terminatorEvent) <=> hasProgramTerminated`.

EDIT: If `ctx.hasProgramTerminated()` exists, then it is also fine to let `Termination.encodeExec` encode `exec(TermationEvent) <=> ctx.hasProgramTerminated`. Basically, the point is to get the definition of what it means for a program to be terminated outside of the `Termination` event and into a more central place.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NonterminationDetection.java
```diff
@@ -83,6 +83,7 @@ public void run(Program program) {
                     .forEach(this::instrumentLoop);
         }
 
+        program.addTerminationThread();
```
ThomasHaas, line 86: This shouldn't be here. Indeed, the code will not get reached if only spin loop detection is enabled.
If you want to put it into an existing pass, then `ThreadCreation` might be a better option. But that one might not run for Litmus code IIRC.
So a dedicated pass right at the end of the processing pipeline seems reasonable.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Wmm.java
```diff
@@ -205,23 +205,28 @@ private Relation makePredefinedRelation(String name) {
...
             case EMPTY -> new Empty(r);
             case IDDTRANS -> new TransitiveClosure(r, getOrCreatePredefinedRelation(IDD));
-            case DATA -> intersection(r,
-                    getOrCreatePredefinedRelation(IDDTRANS),
-                    addDefinition(product(newRelation(), Tag.MEMORY, Tag.MEMORY))
-            );
+            case DATA -> intersection(r, getOrCreatePredefinedRelation(IDDTRANS),
+                    addDefinition(new CartesianProduct(newRelation(),
+                    Filter.union(Filter.byTag(Tag.MEMORY), Filter.byTag(Tag.ALLOC)),
+                    Filter.byTag(Tag.MEMORY))));
```
ThomasHaas, line 211: I'm really not a big fan of removing the memory tags from `Alloc` and `Free`. It messes with so many things.
Is there any imaginable case where treating Alloc/Free as proper memory events is really wrong?
If so, then it suggests that `malloc` and `calloc` are ought to be treated differently, since the latter zero-initializes the memory and so almost acts as a write (and hence a proper memory access).

If it is hard to imagine, I would go with the simpler solution and add the tags (or rather, not remove them in the first place) and see if we ever encounter an issue with it.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/definition/AllocPtr.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public AllocPtr(Relation r) {
+        super(r, ALLOCPTR);
+    }
+
+    @Override
+    public <T> T accept(Visitor<? extends T> v) {
+        return v.visitAllocPtr(this);
+    }
+}
```
ThomasHaas, line 18: In your current iteration, `Alloc` and `Free` have `loc`-edges (which I think they should), so `let allocPtr = [ALLOC];loc;[FREE]` holds and you can save yourself this whole relation.

PR #924, opened 2025-09-09
Better program printing

ThomasHaas:
- Streamlined IR printing to always use `=` when doing assignments (before we used `<-`, `:=` and `=`).
- `Printer` has more features now, the most important being that it now prints the memory and the entry point of the program. It has some configuration options, but they have to be set in the code right now (we could use a `Configuration` object instead so that the user can configure the printer).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/TSOXchg.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/TSOXchg.java
@@ -21,7 +21,7 @@ public class TSOXchg extends RMWXchgBase {
 
     @Override
     public String defaultString() {
-        return String.format("xchg(*%s, %s)", address, resultRegister);
+        return String.format("xchg(%s, %s)", address, resultRegister);
     }
 
     @Override
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
@@ -4,24 +4,8 @@ import com.dat3m.dartagnan.expression.Expression;
 import com.dat3m.dartagnan.program.event.EventVisitor;
 import com.dat3m.dartagnan.program.event.Tag;
 import com.dat3m.dartagnan.program.event.common.StoreBase;
-import com.dat3m.dartagnan.program.event.core.Store;
-import com.dat3m.dartagnan.program.event.metadata.CustomPrinting;
-import com.dat3m.dartagnan.program.event.metadata.MemoryOrder;
-
-import java.util.Optional;
 
 public class LKMMStore extends StoreBase {
-
-    // A custom printer to make core stores appear like LKMMStore
-    public static final CustomPrinting CUSTOM_CORE_PRINTING = (e -> {
-        assert e instanceof Store;
-        Store store = (Store)e;
-        MemoryOrder memoryOrder = store.getMetadata(MemoryOrder.class);
-        if (memoryOrder != null && memoryOrder.value().equals(Tag.Linux.MO_ONCE)) {
-            return Optional.of( "STORE_ONCE(" + store.getAddress() + ", " + store.getMemValue() + ")\t### LKMM");
-        }
-        return Optional.empty(); // Use default printing
-    });
 
     public LKMMStore(Expression address, Expression value, String mo) {
         super(address, value, mo);
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
@@ -18,7 +18,7 @@ public class LKMMStore extends StoreBase {
     @Override
     public String defaultString() {
         if (mo.equals(Tag.Linux.MO_ONCE)) {
-            return "STORE_ONCE(" + address + ", " + value + ")\t### LKMM";
+            return "STORE_ONCE(" + address + ", " + value + ")";
         }
         return super.defaultString();
     }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
@@ -1,5 +1,6 @@
 package com.dat3m.dartagnan.utils.printer;
 
+import com.dat3m.dartagnan.program.Entrypoint;
 import com.dat3m.dartagnan.program.Function;
 import com.dat3m.dartagnan.program.Program;
 import com.dat3m.dartagnan.program.Thread;
@@ -7,8 +8,10 @@ import com.dat3m.dartagnan.program.event.Event;
 import com.dat3m.dartagnan.program.event.core.Init;
 import com.dat3m.dartagnan.program.event.core.Label;
 import com.dat3m.dartagnan.program.event.core.Skip;
+import com.dat3m.dartagnan.program.memory.MemoryObject;
 
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class Printer {
 
@@ -35,7 +38,19 @@ public class Printer {
         if(name == null){
             name = "program";
         }
-        result.append(name).append("\n");
+        result.append(name);
+        if (!(program.getEntrypoint() instanceof Entrypoint.Resolved)) {
+            result.append(" ").append(program.getEntrypoint());
+        }
+        result.append("\n");
+
+        for (MemoryObject obj : program.getMemory().getObjects()) {
+            if (obj.isStaticallyAllocated()) {
+                appendMemoryObject(obj);
+            }
+        }
+
+        result.append("\n");
 
         if (mode == Mode.THREADS || mode == Mode.ALL) {
             for (Thread thread : program.getThreads()) {
@@ -63,6 +78,17 @@ public class Printer {
         }
 
         return result.toString();
+    }
+
+    private void appendMemoryObject(MemoryObject obj) {
+        result.append("\nstatic ").append(obj.getName())
+                .append(String.format(" [size=%s, align=%s] ", obj.getKnownSize(), obj.getKnownAlignment()));
+
+        final int displayLimit = 10;
+        var fieldStrings = obj.getInitializedFields().stream().sorted().limit(displayLimit)
+                        .map(field -> String.format("%s: %s", field, obj.getInitialValue(field)));
+        fieldStrings = obj.getInitializedFields().size() > displayLimit ? Stream.concat(fieldStrings, Stream.of("...")) : fieldStrings;
+        result.append(fieldStrings.collect(Collectors.joining(", ", "{ ", " }")));
     }
 
     public Printer setShowAuxiliaryEvents(boolean flag) {
@@ -95,6 +121,9 @@ public class Printer {
         result.append(functionSignatureToString(func));
         if (func instanceof Thread t && t.hasScope()) {
            result.append(" ").append(t.getScopeHierarchy());
+        }
+        if (func.getProgram().getEntrypoint().getEntryFunctions().contains(func)) {
+            result.append(" #Entrypoint");
         }
         result.append("\n");
         for (Event e : func.getEvents()) {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
@@ -1,5 +1,8 @@
 package com.dat3m.dartagnan.utils.printer;
 
+import com.dat3m.dartagnan.configuration.OptionNames;
+import com.dat3m.dartagnan.expression.ExpressionPrinter;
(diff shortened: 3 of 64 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
```diff
@@ -4,25 +4,9 @@
...
-    public static final CustomPrinting CUSTOM_CORE_PRINTING = (e -> {
-        assert e instanceof Store;
-        Store store = (Store)e;
-        MemoryOrder memoryOrder = store.getMetadata(MemoryOrder.class);
-        if (memoryOrder != null && memoryOrder.value().equals(Tag.Linux.MO_ONCE)) {
-            return Optional.of( "STORE_ONCE(" + store.getAddress() + ", " + store.getMemValue() + ")\t### LKMM");
-        }
-        return Optional.empty(); // Use default printing
-    });
-
```
hernanponcedeleon, line 25: Why is all this gone?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/TSOXchg.java
```diff
@@ -21,7 +21,7 @@ private TSOXchg(TSOXchg other){
 
     @Override
     public String defaultString() {
-        return String.format("xchg(*%s, %s)", address, resultRegister);
+        return String.format("xchg(%s, %s)", address, resultRegister);
```
hernanponcedeleon, line 24: In tso-xchg, the register used for the stored value is the same as the target register. Does tso-xchg behave exactly as all other xchgs? i.e., does it return the old value? If so, I would use `r = xchg(x, r)` to be consistent with all other IR printing.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
```diff
@@ -1,46 +1,150 @@
...
+import com.dat3m.dartagnan.program.memory.MemoryObject;
+import com.dat3m.dartagnan.program.misc.NonDetValue;
 
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public class Printer {
 
+    // =========================== Configurables ===========================
+    // TODO: Maybe use Configuration with proper @Option tags
```
hernanponcedeleon: Can you please add this to this PR. It is much easier to use than having to rebuild.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
```diff
@@ -1,46 +1,150 @@
...
+            result.append(" ").append(program.getEntrypoint());
         }
-        result.append(name).append("\n");
+    }
+
+    // -------------------------------------------------------------------------------------------
+    // Program constants
+
+    private void appendProgramConstants(Program program, StringBuilder result) {
+        result.append("Constants:");
```
hernanponcedeleon: Maybe `Non-deterministic Constants` to make it clearer what those really are

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
```diff
@@ -1,46 +1,150 @@
...
+        appendMainBody(program, result);
+
+        return result.toString();
+    }
+
+    private void appendHeader(Program program, StringBuilder result) {
+        final String name = program.getName() != null ?  program.getName() : "program";
+        result.append(name);
+        if (!(program.getEntrypoint() instanceof Entrypoint.Resolved)) {
+            result.append(" ").append(program.getEntrypoint());
```
hernanponcedeleon, line 116: Not a big fan of how the header looks currently.
- For litmus it looks like this
```
======== After processing ========
xf-barrier.litmus
```
and it is ok.

- For C code it looks like this
```
======== After processing ========
ttas.ll Entry@main
```
I think I would completely drop the `Entry`.

- For GPU code it looks like this
```
======== After processing ========
xf-barrier-1.1.2.spvasm Entry@%24 with Grid(th=1,sg=1,wg=2,qf=1,dv=1)
```
I would just keep `Grid ...`. Not only it is ugly that functions gets a numerical name, but also, I think the notion of entry only makes sense when you do not know where the "flow" of the program starts. If the flow starts at all threads, I find the notion of entry useless.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
ThomasHaas: It was not used anymore.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/TSOXchg.java
ThomasHaas: It should behave like the other one's. I can change the printing

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
ThomasHaas: If you are convinced that you will make use of that feature, sure :)
I just didn't add it because I felt like it would be quite under-used :P

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
ThomasHaas: Sure

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
hernanponcedeleon: but then the current printing does not look how it should. This is from `benchmarks/lkmm/qspinlock.c`
```
55:      === Calling intrinsic __LKMM_store ===
56:      store(&x, bv32(1), Once)
57:      === Returning from intrinsic __LKMM_store ===
```

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
hernanponcedeleon: I would already make my life easier for testing (e.g., I have not yet checked how the printed constants look like).

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
ThomasHaas: The numerical name is because the SPIRV parser just gives them those names (it also gives such numerical names to memory object IIRC).

Also, before ThreadCreation, you don't have any threads in C/SPIRV code, so you don't know where the "flow" of the program starts. Sure, you can guess that for C it must be "main" and for SPIRV I don't even know how you would tell (is there only one function?). And even if you know the entry point for SPIRV, you still wouldn't know how many threads and in what grid they are aligned: the IR is simply incomplete and information is missing.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
ThomasHaas: I don't understand. Are you saying that it should be `WRITE_ONCE`? We haven't printed stores like that for probably well over a year now.
Apart from that, I think it is probably best to avoid custom printing of core events in the IR.

hernanponcedeleon: IIRC that was the whole point of why we added `CustomPrinting` back in the day. However, the situation is suboptimal no matter what: e.g., `atomic_set` get `ONCE` as memory order so we will print it as `WRITE_ONCE`. Similar with `AtomicRead` vs `READ_ONCE`.

I thought we were currently printing `WRITE_ONCE` in litmus code, but this only happens before the processing, after the processing we still get
```
3:       store(&buf, bv64(1), Once)
```
There is some inconsistency between C code and litmus before processing, i.e.
```
// litmus
3:       STORE_ONCE(bv64 buf, bv64(1))  ### LKMM
```
vs
```
// C
3:       call __LKMM_store(&x, bv64(4), bv64(1), bv32(0))
```
but I do not think there is anything we can do.  For C code only a printer between resolving intrinsics and compilation would actually print `STORE_ONCE`.

ThomasHaas: > IIRC that was the whole point of why we added `CustomPrinting` back in the day. 

Yes, but back in the day many things were not so streamlined. Nowadays, it is an awkward feature IMO, especially since our IR is essentially a standalone language nowadays (with features from assembly code, llvm, and gpu code).

>However, the situation is suboptimal no matter what: e.g., `atomic_set` get `ONCE` as memory order so we will print it as `WRITE_ONCE`. Similar with `AtomicRead` vs `READ_ONCE`.

Well, you can always make the printing distinguish those things (assuming they are distinguishable) or not put the custom printing on certain events.

> There is some inconsistency between C code and litmus before processing, i.e.
> 
> ```
> // litmus
> 3:       STORE_ONCE(bv64 buf, bv64(1))  ### LKMM
> ```
> 
> vs
> 
> ```
> // C
> 3:       call __LKMM_store(&x, bv64(4), bv64(1), bv32(0))
> ```
> 
> but I do not think there is anything we can do. For C code only a printer between resolving intrinsics and compilation would actually print `STORE_ONCE`.

Yeah, during processing, you cannot have any uniformness really. Different frontends simply generate different code. Forcing the frontends to generate the same kind of code is just unreasonble (e.g., we could make the Litmus parser generate a call to __LKMM_store instead...)

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
hernanponcedeleon: I know the numerical name for the function/memory comes from the compiler, I just say that it looks ugly in the printer.

Before thread creation I would know where the flow of the program starts because we print the entry function as e.g., `[2] function void %26() #Entrypoint`.

I would say that if I care about how many threads I have, I would probably print the program after processing. Also, this is information you could fully derive from `Grid(th=3,sg=1,wg=3,qf=1,dv=1)`.

In short, I think for real GPU code, the Grid + the fact that the entry function are marked contains all info we need. I do not see the value in having the `Entry@%26 with` part.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/linux/LKMMStore.java
hernanponcedeleon: I think I can live with how things are now.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/printer/Printer.java
ThomasHaas: > I know the numerical name for the function/memory comes from the compiler, I just say that it looks ugly in the printer.

Well, nothing I can do about this. 

> 
> Before thread creation I would know where the flow of the program starts because we print the entry function as e.g., `[2] function void %26() #Entrypoint`.

Because I added this into this PR as well, otherwise you wouldn't know. I could only print the entry point on functions with all the details but it would be messy cause I cannot just use `entrypoint.toString` anymore without duplicating the function in the printing.

> 
> I would say that if I care about how many threads I have, I would probably print the program after processing. Also, this is information you could fully derive from `Grid(th=3,sg=1,wg=3,qf=1,dv=1)`.
> 
> In short, I think for real GPU code, the Grid + the fact that the entry function are marked contains all info we need. I do not see the value in having the `Entry@%26 with` part.

Ok, you just don't want the `Entry@%26 with` bit, but that comes from `Entrypoint.toString`. To be clear, the `Entrypoint.toString` prints `Entry@%26 with Grid(...)`, so I cannot simply print one half of it and skip the other without adding special code for that. And I really don't want to make the code more complex/less maintainable to just get rid of a minor redundancy in the printing which I think is actually a good redundancy (that's why I added the additional `#Entrypoint` marker on functions).

Also, it is very important that the printing is as complete as possible (this is the main motiviation of this PR!).
I wanted to dumb IR code into my thesis to show how Dartagnan works and explain the encoding on it. The first thing I had to say was "oh, btw. there is some entry point that you don't see, oh and memory objects, oh and constants...". And I had the same issue many times when analyzing the program for debugging/optimization purposes because information was missing all the time and I had to recover it from crawling through the debugger windows.
So I want to have a as-complete-as-possible printing as a baseline. That being said, we can later always add additional options to make it less verbose/complete (I already added some), but I want to mostly focus on the complete one for now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/arch/tso/TSOXchg.java
ThomasHaas: Actually, IIRC, we use this printing here because it aligns with how the original x86 instruction looks like (it does not mention the register twice).
Which is reasonable as this (uncompiled) event is supposed to represent the original one.
That being said, I don't have a strong opinion on this.

PR #925, opened 2025-09-09
Slightly better memory encoding

ThomasHaas:
This PR assigns a fixed memory range to dynamic allocations with known size and alignment irrespective of whether the allocation happens or not. As a result, in any program with only known-sized allocations, the memory encoding results in a statically fixed memory layout (the memory layout is identical for all program executions).

Some preliminary evaluations show that this encoding is roughly 10% faster on various benchmarks (e.g. LFDS).

NOTE: We could do the same for memory allocations that are of unknown size but have a known upper bound (e.g., computed by a value range analysis).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
@@ -6,8 +6,8 @@ import com.dat3m.dartagnan.expression.ExpressionFactory;
 import com.dat3m.dartagnan.expression.integers.IntLiteral;
 import com.dat3m.dartagnan.expression.type.IntegerType;
 import com.dat3m.dartagnan.expression.type.TypeFactory;
-import com.dat3m.dartagnan.program.Thread;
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -440,7 +440,10 @@ private BooleanFormula encodeMemoryLayout(Memory memory) {
                 final Expression zero = exprs.makeValue(BigInteger.ZERO, archType);
                 final Expression one = exprs.makeValue(BigInteger.ONE, archType);
 
-                size = exprs.makeITE(exec, cur.size(), zero);
+                // NOTE: If we know the size/alignment of the allocation, we can pre-reserve memory space.
```
hernanponcedeleon: Can you add "even if the allocation is not executed" as in the PR description?

ThomasHaas: sure

PR #927, opened 2025-09-19
Pointer expressions

MorMahd:
- Added dedicated pointer expressions.
- Pointers are still encoded like integers.


dartagnan/src/main/java/com/dat3m/dartagnan/expression/pointer/PointerLiteral.java
```diff
@@ -0,0 +1,39 @@
...
+import com.dat3m.dartagnan.expression.base.LiteralExpressionBase;
+import com.dat3m.dartagnan.expression.type.IntegerType;
+import com.dat3m.dartagnan.expression.type.PointerType;
+import com.google.common.base.Preconditions;
+
+import java.math.BigInteger;
+
+public final class PointerLiteral extends LiteralExpressionBase<PointerType> {
+
+    private final BigInteger value;
```
ThomasHaas: I am not sure if we want to have this class. Such a pointer literal can only really refer to a pointer with invalid provenance and would be equivalently represented by an `int2ptr` cast that does not attempt to recover provenance.
Such provenance-less casts can already be expressed by `nullptr + value`, i.e., `ptrlit(value) == int2ptrNoProv(value) == nullptr + value`, with the additional benefit that "value" does not have to be a constant.
Indeed, you could have a helper method `ExpressionFactory.castIntToPtrNoProvenance` that implements this idea and use it for all those cases where we temporarily need to embed an integer into a pointer due to the type system forcing us to (e.g., for implementing `pthread_t` or function pointers in `NaiveDevirtualisation`).
More generally, I could even imagine that we have a `int2PtrWithProv(Expr address, Expr prov)` expression that casts `address` to a pointer with provenance equivalent to that of `prov`, which would yield `int2PtrNoProv(value) == int2PtrWithProv(value, nullptr)`. But for now, this generelization is not needed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/VirtualAliasAnalysis.java
```diff
@@ -34,13 +35,25 @@ public List<Integer> mayMixedSizeAccesses(MemoryCoreEvent event) {
...
     private boolean samePhysicalAddress(MemoryCoreEvent e1, MemoryCoreEvent e2) {
         // TODO: Add support for pointers
-        if (!(e1.getAddress() instanceof VirtualMemoryObject addr1)
-                || !(e2.getAddress() instanceof VirtualMemoryObject addr2)) {
-            return false;
-        }
-        return addr1.getPhysicalAddress() == addr2.getPhysicalAddress();
+        // FIXME: Ugly code because somewhere a (virtual address + 0) is being generated.
+        VirtualMemoryObject v1;
+        VirtualMemoryObject v2;
```
ThomasHaas: See `EventFactory.newInit`. It had a special check to avoid generating `+0` offsets which you removed in your PR :).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -142,7 +154,14 @@ private static Event inlineBody(FunctionCall call, Snapshot callTarget, int scop
         final ArrayList<Event> parameterAssignments = new ArrayList<>();
         for (int j = 0; j < callTarget.parameters.size(); j++) {
             final Register register = registerMap.get(callTarget.parameters.get(j));
-            parameterAssignments.add(newLocal(register, arguments.get(j)));
+            if (register.getType() instanceof PointerType) {
```
MorMahd, line 157: should not be resolved here

PR #928, opened 2025-09-19
[SPIRV] Add support for OpNot

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBits.java
```diff
@@ -61,6 +62,24 @@ public Event visitOpBitwiseXor(SpirvParser.OpBitwiseXorContext ctx) {
...
+                Expression expression = HelperTypes.createResultExpression(id, type, op, IntUnaryOp.NOT);
+                Register register = builder.addRegister(id, typeId);
+                Local event = EventFactory.newLocal(register, expression);
+                return builder.addEvent(event);
+            }
+            throw new ParsingException("Illegal definition for '%s', vector expressions must have fixed size", id);
+        }
+        throw new ParsingException("Illegal definition for '%s', result type doesn't match operand types", id);
+    }
+
```
ThomasHaas: I would invert the conditions and avoid nesting:
```
if (!type.equals(op.getType())) {
   throw ...
}

if (type instance of ArrayType aType && !aType.hasKnownNumElement()) {
   throw ...
}

... // Actual logic of all checks pass.
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/helpers/HelperTypes.java
```diff
@@ -65,6 +66,21 @@ public static Expression createResultExpression(String id, Type type, Expression
...
     }
 
+    public static Expression createResultExpression(String id, Type type, Expression operand, IntUnaryOp op) {
+        if (isScalar(type)) {
+            return expressions.makeIntUnary(op, operand);
+        }
+        if (type instanceof ArrayType aType && aType.getElementType() instanceof IntegerType) {
+            List<Expression> elements = new ArrayList<>();
+            for (int i = 0; i < aType.getNumElements(); i++) {
+                Expression elementOp = operand instanceof ConstructExpr ? operand.getOperands().get(i) : expressions.makeExtract(operand, i);
```
ThomasHaas, line 76: Is this case distinction necessary? I guess it gives you better-to-read code after parsing, but otherwise the `ExprSimplifier` would do the same optimization I believe.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryOp.java
```diff
@@ -15,6 +15,7 @@ public String getSymbol() {
         return switch (this) {
             case CTLZ -> "ctlz ";
             case CTTZ -> "cttz ";
+            case NOT -> "not ";
```
ThomasHaas: Do you prefer "not" over "~"? I don't mind either one tbh.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/helpers/HelperTypes.java
hernanponcedeleon: This method follows the same implementation that the method above.

I remember we had a discussion about this simplification in the past. AFAIR the simplification is either not there (IIRC you promised to implement it :P) or there was some bug in some specific cases where several simplifications happen one after the other.

ThomasHaas: Isn't it already fixed by #876 ?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryOp.java
ThomasHaas: You added a space after "~". I think the space should go.

benchmarks/slang/complement.slang
```diff
@@ -0,0 +1,28 @@
...
+//; @Output: forall (%output2[0][0] == 4294967295 and %output2[0][1] == 4294967294 and %output2[0][2] == 4294967292)
+//; @Output: forall (%output3[0][0] == 4294967294 and %output3[0][1] == 4294967292 and %output3[0][2] == 4294967295)
+//; @Config: 3, 1, 1
+
+RWStructuredBuffer<uint8_t> input;
+RWStructuredBuffer<uint8_t> output1;
+RWStructuredBuffer<uint32_t> output2;
+RWStructuredBuffer<uint32_t> output3;
+
+[numthreads(2, 1, 1)]
```
natgavrilenko: Why numthreads is 2 here and not 3? Not sure if slang allows an odd numthreads, but maybe this test can be done with four threads.

dartagnan/src/test/java/com/dat3m/dartagnan/spirv/vulkan/basic/SpirvAssertionsTest.java
```diff
@@ -121,7 +121,8 @@ public static Iterable<Object[]> data() throws IOException {
                 {"push-constant-mixed.spvasm", 1, PASS},
                 {"bitwise-scalar.spvasm", 1, PASS},
                 {"bitwise-vector.spvasm", 1, PASS},
-                {"idx-overflow.spvasm", 1, PASS}
+                {"idx-overflow.spvasm", 1, PASS},
+                {"complement.spvasm", 1, PASS}
```
natgavrilenko: Could you rename it to something like "bits-complement" and put next to the bitwise tests, so that the tests for similar instructions are grouped together?

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBitsTest.java
```diff
@@ -102,8 +107,71 @@ public void testMismatchingTypes() {
         }
     }
 
+    @Test
+    public void testScalarNot() {
+        // given
+        MockProgramBuilder builder = new MockProgramBuilder();
+        builder.mockIntType("%int", 64);
+        builder.mockConstant("%value", "%int", 4);
+        String input = String.format("%%res = OpNot %%int %%value");
```
natgavrilenko: Unnecessary "String.format", also in the other tests

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsBitsTest.java
```diff
@@ -102,8 +107,71 @@ public void testMismatchingTypes() {
...
+            // when
+            visit(builder, input);
+            fail("Should throw exception");
+        } catch (ParsingException e) {
+            // then
+            assertEquals("Illegal definition for '%reg', " +
+                            "result type doesn't match operand types",
+                    e.getMessage());
+        }
+    }
```
natgavrilenko, line 190: I would add two more tests where result and operand types match but are illegal for OpNot - with a runtime array and with a totally different type, e.g. struct.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/helpers/HelperTypes.java
```diff
@@ -56,15 +57,30 @@ public static Expression createResultExpression(String id, Type type, Expression
         if (type instanceof ArrayType aType && aType.getElementType() instanceof IntegerType) {
             List<Expression> elements = new ArrayList<>();
             for (int i = 0; i < aType.getNumElements(); i++) {
-                Expression elementOp1 = op1 instanceof ConstructExpr ? op1.getOperands().get(i) : expressions.makeExtract(op1, i);
```
natgavrilenko, line 59: I don't think it is a good idea to rely on the "magic" of ExpressionSimplifier here. The job of a parser is to parse a program as it is, without making assumption of what happens to it on the later stages. Otherwise, we will quickly end up with a spaghetti code that is difficult to maintain and troubleshoot.

ThomasHaas: Oh, I see that we artificially generate the extract expressions here (they are not part of the original code).
Then I agree that we might want to consider keeping the case distinction there, but only for the sake of being closer to the parsed code.
That being said, I think the job of a parser is only to generate semantically equivalent code, which we have either way. In particular, the existence of `ExpressionSimplifier` is somewhat irrelevant and I think we would also encode the semantics of the unsimplified program correctly (i.e. if `ExpressionSimplifier` never ran at all).

PR #931, opened 2025-09-22
[SPIRV] Add support for OpBitCount

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/expression/integers/IntUnaryOp.java
```diff
@@ -3,7 +3,7 @@
 import com.dat3m.dartagnan.expression.ExpressionKind;
 
 public enum IntUnaryOp implements ExpressionKind {
-    CTLZ, CTTZ, NOT, MINUS;
+    COUNT, CTLZ, CTTZ, NOT, MINUS;
```
ThomasHaas: I think we should call this `CTPOP` to match with the LLVM name of the operation. This fits better with the existing `CTLZ/CTTZ`. If you don't mind deviating from the existing naming scheme, I would at least call it `(SET)BIT_COUNT` or similar, cause just `COUNT` is very unspecific.

natgavrilenko (approved):
Looks good to me. Regarding the operand name, any of COUNT, BIT_COUNT or CTPOP seems fine.

PR #932, opened 2025-09-23
[SPIRV] Add support for GLSL extensions FindILsb

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/extenstions/VisitorExtensionGlslStd.java
```diff
@@ -1,20 +1,67 @@
...
+                Expression elementValue = value instanceof ConstructExpr ? value.getOperands().get(i) : expressions.makeExtract(value, i);
+                Expression zero = expressions.makeValue(0, eType);
+                Expression mOne = expressions.makeValue(-1, eType);
+                Expression trailingZeros = expressions.makeCTTZ(elementValue);
+                Expression valueIsZero = expressions.makeEQ(elementValue, zero);
+                elements.add(expressions.makeITE(valueIsZero, mOne, trailingZeros));
+            }
+            return expressions.makeArray(aType, elements);
+        }
+        throw new ParsingException("Value type %s of FindILsb is not integer scalar or integer vector", type);
```
natgavrilenko: I would use valueIdRef in the error message. This way it will be easier for a user to understand where the problem is.

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/extensions/VisitorExtensionGlslStdTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.Test;
+
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.fail;
+
+public class VisitorExtensionGlslStdTest {
+
+    @Test
```
natgavrilenko, line 16: I would also add a positive test

hernanponcedeleon: I thought about this. The question is: what to test?

Differently to #928 or #931, we are creating a "complex" (ITE) expression rather than a "base" one (*). I do not think checking anything related to the expression makes much sense. Also, we cannot check anything related to the return register: this visitor class is similar to `VisitorExtensionOpenClStd` and for that one we do not have parser tests (not even negative one), I assume for the same reason as here.

(*) I initially added a new `IntUnaryOp`, but the fact that if the value is zero, we should return -1 seems too specific to me to have a general operator.

natgavrilenko: Fair enough. In principle, we could test the content of the ITE, but that wouldn't be a very valuable test.

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/extensions/VisitorExtensionGlslStdTest.java
```diff
@@ -0,0 +1,46 @@
+package com.dat3m.dartagnan.parsers.program.visitors.spirv.extensions;
+
+import com.dat3m.dartagnan.exception.ParsingException;
+import com.dat3m.dartagnan.program.event.core.Local;
```
natgavrilenko: Unused import

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/extensions/VisitorExtensionGlslStdTest.java
ThomasHaas: > (*) I initially added a new `IntUnaryOp`, but the fact that if the value is zero, we should return -1 seems too specific to me to have a general operator.

We should probably avoid too many specialed operators if we can express them already.
But what you can always do, at least in principle, is to add a new method to `ExpressionFactory` that constructs the desired semantics using existing operators. The factory does not really give syntactic guarantees anyways (the reason why it always returns most general `Expression`s rather than special subclasses).
The beneift of having it in `ExpressionFactory` would be that multiple frontends can make use of it.
For example, it would be straightforward to add vector operations (`ExpressionFactory.makeVectorOp(...)`) which could be used by both SPIRV and LLVM.
Of course, you pay the price that the parser has less control about the syntax of the generated expression (e.g., whether it inserts additional extract operations or avoids them). But, again, strictly speaking the parser had no real control to begin with since `ExpressionFactory` gives not syntactic gurantees. Consequently, all tests that check the syntactic structure of parsed expressions are fragile tests to begin with (which does not make the tests bad, but one has to expect to update them from time to time).

PR #933, opened 2025-09-25
[SPIRV] Fix support for OpShiftRightArithmetic

hernanponcedeleon:
There was a typo in the visitor and thus `OpShiftRightArithmetic` was not recognized.

I do not get why the parser tests were not complaining. Any idea @natgavrilenko?


benchmarks/slang/bitwise-shift.slang
```diff
@@ -0,0 +1,63 @@
+//; @Input: %input = {{0, 1, 3}}
```
natgavrilenko: Why not to have an example where arshift and rshift produce different values?

PR #934, opened 2025-09-29
 No static relation encoding by default + removed unnecessary code

ThomasHaas:
- `WmmEncoder` used to encode all unknown relations as static one's by default (i.e. `r(x,y) <=> exec(x) /\ exec(y)`).
This was very dangerous so I added a white list of known static relations; the default encoding throws an exception unless the relation is known to be static.
The code is still quite dangerous because it relies on the fact that RA's may-set computation for static relations is maximally precise, i.e., it is no over-approximation.
- I removed the `equals` implementations from `(Virtual)MemoryObject`: reference equality is the most sane implementation option anyway. One could even remove the `hashCode` implementation

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
@@ -245,8 +245,8 @@ public class WmmEncoder implements Encoder {
         final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
         final List<BooleanFormula> enc = new ArrayList<>();
 
-        // ASSUMPTION: The encode-set of a static relation only contains
-        //             possible edges. If not, this encodes wrong semantics.
+        // ASSUMPTION: The encode-set of the static relation is a subset of the most precise may-set.
+        //             This holds true as long as our RA computes the most precise may-set for static relations.
         private void visitStatic(Definition def) {
             final Relation rel = def.getDefinedRelation();
             final EncodingContext.EdgeEncoder edge = context.edge(rel);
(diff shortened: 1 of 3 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -227,17 +223,49 @@ private Map<Relation, EventGraph> initializeLazyEncodeSets(LazyRelationAnalysis
...
+                CartesianProduct.class
+        ));
+
         final Program program = context.getTask().getProgram();
         final RelationAnalysis ra = context.getAnalysisContext().requires(RelationAnalysis.class);
         final BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
         final List<BooleanFormula> enc = new ArrayList<>();
 
+        // ASSUMPTION: The encode-set of a static relation only contains
+        //             possible edges. If not, this encodes wrong semantics.
```
hernanponcedeleon: I find this comment unclear. Is this what you mean by "A's may-set computation for static relations is maximally precise, i.e., it is no over-approximation."?

ThomasHaas: Pretty much yes. But technically, only the encode set matters I guess. Even if the may-set is overapproximated, as long as the encode set is not, it would be fine.

hernanponcedeleon: Can you then use the text from the PR (just replace `may-set` by `encodeset`) for the comment?

ThomasHaas: I tried writing it with "encode set" but then it really does not make too much sense. "Maximally precise" and "over approximate" do not mean the same thing for encode sets, because the ground truth is totally different (and even unknown). I will just write it with the may sets.

PR #935, opened 2025-09-30
Add option to select entry point

hernanponcedeleon:


ThomasHaas (commented):
I like the change. What do you use it for?

I once wanted to add entry point annotations/attributes so you can write C benchmarks a bit like Litmus code. You would just put some attribute on each function that should run in its own thread and forget the main function entirely (as long as you don't need to join the threads).

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -207,6 +209,8 @@ public static void main(String[] args) throws Exception {
                         .withProgressModel(o.getProgressModel())
                         .withWitness(witness);
                 Program p = new ProgramParser().parse(f);
+                p.setEntrypoint(new Entrypoint.Simple(p.getFunctionByName(o.getEntryFunction()).orElseThrow(
+                    () -> new MalformedProgramException(String.format("Program has no function named %s. Select a different entry point.", o.getEntryFunction())))));
```
ThomasHaas: Isn't this piece of code always executed? Wouldn't it fail on litmus or SPIRV code that does not have a "main"?
I think you might want to put as default value for the entrypoint the empty string with the meaning that it defaults to whatever the parser decides. Only for non-empty values you overwrite the parser's decision.

PR #938, opened 2025-10-01
[SPIRV] Add support for OpUnreachable

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -823,6 +823,10 @@ public static GenericVisibleEvent newAvDevice() {
         public static GenericVisibleEvent newVisDevice() {
             return new GenericVisibleEvent("visdevice", Tag.Vulkan.VISDEVICE);
         }
+
+        public static GenericVisibleEvent newUnreachable() {
+            return new GenericVisibleEvent("OpUnreachable", Tag.UB);
+        }
```
ThomasHaas: If you add handling of this event into `stdlib.cat`, then I think the event should be part of the core section of the EventFactory (not SPIRV/Vulkan specific) and should be named "Unreachable" without the "Op"-prefix.
LLVM also has unreachable events btw., but I think we replace them by either `assert false` or `return` (not sure).

hernanponcedeleon: For LLVM, we threat those as `abort()`.

I already took a look to what the LLVM spec says about `unreachable`. It claims they do not have defined semantics, but I am not sure this is really UB, that is why I did not add it as such in this PR.

ThomasHaas: Reaching `unreachable` is surely UB. You don't need to change the way we do it in LLVM though. My point is that you shouldn't make a SPIRV-only variant of it and add it to stdlib.cat. Just generalize a bit more to a general-purpose unreachable event by just moving/renaming your code.

Btw. in LLVM we would always want to place the abort in addition to either an `assert(false)` or a unreachable event, because the control-flow needs to stop there. `assert(false)` also seems to be a reasonable solution to avoid special code.

PR #939, opened 2025-10-01
[RA] Fix propagation for negated/flagged axioms 

hernanponcedeleon:
For any program with at least two events in a thread, adding the constraint `~empty po` to the memory model should have no effect, however it is not currently the case:
```
> CFLAGS="-DACQ2RX" java -jar dartagnan/target/dartagnan.jar cat/vmm.cat benchmarks/locks/ttas.c
Test: benchmarks/locks/ttas.c
Result: FAIL
Reason: CAT specification violation found
Details:
        Flag w-data-race
        E181 / E181     @ttas.c#18 / @ttas.c#18
        E234 / E234     @ttas.c#18 / @ttas.c#18
Time: 0.980 secs

> echo "" >> cat/vmm.cat 
> echo "~empty po" >> cat/vmm.cat 

> CFLAGS="-DACQ2RX" java -jar dartagnan/target/dartagnan.jar cat/vmm.cat benchmarks/locks/ttas.c
Test: benchmarks/locks/ttas.c
Result: PASS
Time: 0.753 secs
```

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
@@ -351,7 +351,7 @@ public class NativeRelationAnalysis implements RelationAnalysis {
         @Override
         public Map<Relation, ExtendedDelta> visitEmptiness(Emptiness axiom) {
             Relation rel = axiom.getRelation();
-            return axiom.isNegated() ?
+            return axiom.isNegated() || axiom.isFlagged() ?
                 Map.of() :
                 Map.of(rel, new ExtendedDelta(knowledgeMap.get(rel).getMaySet(), new MapEventGraph()));
         }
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -389,7 +393,9 @@ public Map<Relation, ExtendedDelta> visitAcyclicity(Acyclicity axiom) {
             } while (!current.isEmpty());
             newDisabled.retainAll(knowledge.getMaySet());
             logger.debug("disabled {} edges in {}ms", newDisabled.size(), System.currentTimeMillis() - t0);
-            return Map.of(rel, new ExtendedDelta(newDisabled, new MapEventGraph()));
+            return axiom.isNegated() || axiom.isFlagged() ?
+                Map.of() :
+                Map.of(rel, new ExtendedDelta(newDisabled, new MapEventGraph()));
```
ThomasHaas: You should put the check at the beginning and do an early return (also for the other two methods).
This function here computes a whole lot of stuff and even logs output if the axiom is false/negated, all of which is unnecessary/wrong.

ThomasHaas (approved):
LGTM

PR #942, opened 2025-10-02
Add EncodingUtils class

ThomasHaas:
I added a new class `EncodingUtils` that can encode cardinality constraints and symmetry constraints (moved the existing `SymmetryEncoder.encodeLexLeader` into `EncodingUtils`). Simplified existing code to use the new class.

This PR is almost a pure refactor with one exception: the encoding of `rf` has slightly changed.
We used to encode the at-most-one constraint unconditionally for rf-edges, and only conditioned the at-least-one constraint on the execution of a read, i.e., `at-most-one(SetOfRfEdges) /\ (exec(r) => at-least-one(SetOfRfEdges)`.
Now both parts are conditioned: `exec(r) => (at-most-one(...) /\ at-least-one(...))`.
I don't know if this impacts performance in any measurable way (over our unit tests nothing really changes I think).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingUtils.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingUtils.java
@@ -1,0 +1,118 @@
+package com.dat3m.dartagnan.encoding;
+
+import com.dat3m.dartagnan.smt.FormulaManagerExt;
+import com.google.common.base.Preconditions;
+import org.sosy_lab.java_smt.api.BooleanFormula;
+import org.sosy_lab.java_smt.api.BooleanFormulaManager;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class EncodingUtils {
+
+    private final FormulaManagerExt fmgr;
+
+    public EncodingUtils(FormulaManagerExt fmgr) {
+        this.fmgr = fmgr;
+    }
+
+    public EncodingUtils(EncodingContext context) {
+        this(context.getFormulaManager());
+    }
+
+    // -----------------------------------------------------------------------------------------------
+    // Cardinality constraints
+
+    public BooleanFormula atLeastOne(Collection<BooleanFormula> formulas) {
+        return fmgr.getBooleanFormulaManager().or(formulas);
+    }
+
+    public BooleanFormula atMostOnePairwise(List<BooleanFormula> formulas) {
+        final BooleanFormulaManager bmgr = fmgr.getBooleanFormulaManager();
+        final List<BooleanFormula> enc = new ArrayList<>(formulas.size() * (formulas.size() - 1) / 2);
+        for (int i = 0; i < formulas.size(); i++) {
+            for (int j = i + 1; j < formulas.size(); j++) {
+                enc.add(bmgr.or(bmgr.not(formulas.get(i)), bmgr.not(formulas.get(j))));
+            }
+        }
+        return bmgr.and(enc);
+    }
+
+    // NOTE: This method generates helper variables using <uniqueIdent>, so the user must ensure
+    // that the identifier is unique.
+    public BooleanFormula exactlyOneSequence(List<BooleanFormula> formulas, String uniqueIdent) {
+        final BooleanFormulaManager bmgr = fmgr.getBooleanFormulaManager();
+        if (formulas.size() <= 1) {
+            return atLeastOne(formulas);
+        }
+
+        final String seqVarName = "seqVar_" + uniqueIdent + "#";
+        final List<BooleanFormula> enc = new ArrayList<>(2*formulas.size() - 1);
+        BooleanFormula lastSeqVar = formulas.get(0);
+        for (int i = 1; i < formulas.size(); i++) {
+            final BooleanFormula newSeqVar = bmgr.makeVariable(seqVarName + i);
+            enc.add(bmgr.equivalence(newSeqVar, bmgr.or(lastSeqVar, formulas.get(i))));
+            enc.add(bmgr.not(bmgr.and(formulas.get(i), lastSeqVar)));
+            lastSeqVar = newSeqVar;
+        }
+        enc.add(lastSeqVar); // NOTE: without this line, we get an at-most-one encoding
+
+        return bmgr.and(enc);
+    }
+
+    // -----------------------------------------------------------------------------------------------
+    // Symmetry constraints
+
+    /*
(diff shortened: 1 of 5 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingUtils.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private final FormulaManagerExt fmgr;
+
+    public EncodingUtils(FormulaManagerExt fmgr) {
+        this.fmgr = fmgr;
+    }
+
+    public EncodingUtils(EncodingContext context) {
+        this(context.getFormulaManager());
+    }
```
xeren: You could make the constructors package-private and define `FormulaManagerExt.getEncodingUtils()` as the public API.  Because of the 1-to-1 correspondence, `EncodingUtils` could even merge with `FormulaManagerExt`.

dartagnan/src/main/java/com/dat3m/dartagnan/smt/EncodingUtils.java
```diff
@@ -0,0 +1,118 @@
...
+        NOTE: This method generates helper variables using <uniqueIdent>, so the user must ensure
+              that the identifier is unique.
+    */
+    public BooleanFormula encodeLexLeader(List<BooleanFormula> r1, List<BooleanFormula> r2, String uniqueIdent) {
+        Preconditions.checkArgument(r1.size() == r2.size());
+        final BooleanFormulaManager bmgr = fmgr.getBooleanFormulaManager();
+        // Return TRUE if there is nothing to encode
+        if(r1.isEmpty()) {
+            return bmgr.makeTrue();
+        }
```
xeren, line 78: Space after `if`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingUtils.java
```diff
@@ -0,0 +1,118 @@
...
+            // NOTE: yi = TRUE means the prefixes (x1, x2, ..., xi) of the rows r1/r2 are equal
+            //       yi = FALSE means that no conditions are imposed on xi
+            // The first point, where y(i-1) is TRUE but yi is FALSE, is the breaking point
+            // where xi(r1) < xi(r2) holds (afterwards all yj (j >= i+1) are unconstrained and can be set to
+            // FALSE by the solver)
+            ylast = y;
+        }
+        // Final iteration for xn is handled differently as there is no variable yn anymore.
+        BooleanFormula a = r1.get(size-1);
+        BooleanFormula b = r2.get(size-1);
```
xeren: The variables declared here are `final`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/SymmetryEncoder.java
```diff
@@ -128,15 +128,16 @@ private BooleanFormula encodeSymmetryBreakingOnClass(EventGraph maySet, Encoding
...
             Function<Event, Event> p = symm.createEventTransposition(t1, t2);
             List<Tuple> t2Tuples = t1Tuples.stream().map(t -> new Tuple(p.apply(t.first()), p.apply(t.second()))).toList();
 
             List<BooleanFormula> r1 = t1Tuples.stream().map(t -> edge.encode(t.first(), t.second())).toList();
             List<BooleanFormula> r2 = t2Tuples.stream().map(t -> edge.encode(t.first(), t.second())).toList();
-            final String id = "_" + rep.getId() + "_" + i;
-            enc.add(encodeLexLeader(id, r2, r1, context)); // r1 >= r2
+            final String id = String.format("T%d_T%d", rep.getId(), i);
+            enc.add(utils.encodeLexLeader(r2, r1, id)); // r1 >= r2
             t1 = t2;
```
xeren: The numbers in `id` have different semantics: the first is the representant's thread ID, while the second is another thread's local index in the list of equivalent threads.  I would either omit the second `T` prefix, or use `t2.getId()` instead of `i`.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/symmetry/CoSymmetryBreaking.java
```diff
@@ -237,9 +238,8 @@ public BooleanFormula encode(EquivalenceClass<Thread> symmClass) {
                 r2.add(edge.encode(t.first(), t.second()));
             }
 
-            final String id = "_" + rep.getId() + "_" + i;
-            // NOTE: We want to have r1 >= r2 but lexLeader encodes r1 <= r2, so we swap r1 and r2.
-            enc.add(SymmetryEncoder.encodeLexLeader(id, r2, r1, context));
+            final String id = String.format("T%d_T%d", rep.getId(), i);
+            enc.add(utils.encodeLexLeader(r2, r1, id)); // r1 >= r2
```
xeren, line 243: Could this collide with `SymmetryEncoder`?  Also `i` should probably not get the `T` prefix.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingUtils.java
ThomasHaas: I deliberately did not merge `EncodingUtils` into `FormulaManagerExt`. I'm also not yet too sure if this class will not eventually get helper methods that are more closely related to Dartagnan's internals.
That's why I have the constructor that takes `EncodingContext`. 
My gut feeling is that we have some duplicated encodings here and there, but I didn't have time to check this thoroughly.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/symmetry/CoSymmetryBreaking.java
ThomasHaas: It would collide, yes. But you use either this encoding or the other, not both.
This class is never used and we might as well delete it. I wrote it back in the day because I thought we would try it once the XRA implementation was done, since the co symmetry breaking can give must co-edges across threads.
But XRA didn't turn out to be as flexible/general as I imagined.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingUtils.java
ThomasHaas: I made it package private as you suggested. It's easy to change it back if we find more specific utility that requires access to `EncodingContext`

xeren (approved):
LGTM

PR #944, opened 2025-10-07
Devirtualization with proper addresses

ThomasHaas:
- Devirtualization now creates new `MemoryObjects` to represent addresses of functions (it used to assign arbitary constants).
**Benefit:** It is much easier to understand the generated IR and see where function pointers are communicated/used.
**Possible future benefit:** AA can compute which function pointers are possible at a use-site, allowing us to remove impossible cases.

- `ExpressionSimplifier` optimizes comparisions on `MemoryObjects` now.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
@@ -97,10 +97,10 @@ public class NaiveDevirtualisation implements ProgramProcessor {
             return false;
         }
 
-        logger.debug("Assigned address to function \"{}\"", func);
         final MemoryObject funcAddr = func.getProgram().getMemory().allocate(1);
         funcAddr.setName(String.format("__funcAddr_%s", func.getName()));
         func2AddressMap.put(func, funcAddr);
+        logger.debug("Assigned address to function \"{}\"", func);
         return true;
     }
 


dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
```diff
@@ -165,6 +166,24 @@ public Expression visitIntCmpExpression(IntCmpExpr cmp) {
...
+        if (left instanceof MemoryObject lMem && right instanceof MemoryObject rMem) {
+            final boolean sameObj = lMem.equals(rMem);
+
+            final Boolean cmpResult = switch (op) {
+                case EQ -> sameObj;
+                case NEQ -> !sameObj;
+                case LT, ULT -> sameObj ? false : null;
+                case LTE, ULTE -> sameObj ? true : null;
+                default ->
+                    throw new VerifyException(String.format("Unexpected comparison operator '%s'. Missing normalization?", op));
```
hernanponcedeleon, line 179: The "greater" cases are missing. Once those are added you can get rid of the default case.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
```diff
@@ -97,16 +92,16 @@ private void findAndTransformAddressTakenFunctions(
...
-            func2AddressMap.put(func, expressions.makeValue(nextAvailableFuncAddress, ptrType));
-            nextAvailableFuncAddress += 8;
-            return true;
+    private boolean assignAddressToFunction(Function func, Map<Function, MemoryObject> func2AddressMap) {
+        if (func2AddressMap.containsKey(func)) {
+            return false;
         }
-        return false;
+
+        logger.debug("Assigned address to function \"{}\"", func);
```
hernanponcedeleon: nitpicking: this should probably come last, in case some of the following operations failed (which should never be the case, but still)

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
ThomasHaas: They are not missing. The code normalizes the operators to LT operations (it swaps the operands if necessary).
That's why the exception says "Missing normalization?". Btw. this is an internal assertion-like exception and should be impossible to trigger (that's why it's a `VerifyException`).

PR #953, opened 2025-12-16
Remove non-determinism in Llvm parser and Internal representation

xeren:
Removes two sources of non-determinism of the model checker:
- Identity-based iteration order of phi nodes in the LLVM parser
- Identity-based iteration order of relations in the memory model

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/VoidType.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/VoidType.java
@@ -10,4 +10,9 @@ public final class VoidType implements Type {
     public String toString() {
         return "void";
     }
+
+    @Override
+    public int hashCode() {
+        return 1;
+    }
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
@@ -1421,7 +1421,9 @@ public class VisitorLlvm extends LLVMIRBaseVisitor<Expression> {
 
     private interface MdNode extends Expression {
 
-        Type TYPE = new Type() { };
+        Type TYPE = new Type() {
+            @Override public int hashCode() { return -1; }
+        };
         ExpressionKind MdKind = new ExpressionKind() {
             @Override
             public String getSymbol() { return "Md"; }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/BooleanType.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/BooleanType.java
@@ -10,4 +10,9 @@ public final class BooleanType implements Type {
     public String toString() {
         return "bool";
     }
+
+    @Override
+    public int hashCode() {
+        return 2;
+    }
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/NaiveDevirtualisation.java
@@ -177,6 +177,7 @@ public class NaiveDevirtualisation implements ProgramProcessor {
         Preconditions.checkArgument(needsDevirtualization(call));
         return func2AddressMap.keySet().stream()
                 .filter(f -> f.getFunctionType() == call.getCallType())
+                .sorted(Comparator.comparingInt(Function::getId))
                 .collect(Collectors.toCollection(ArrayList::new));
     }
 


hernanponcedeleon (commented):
I understand that the `LinkedHashSet` guarantees determinism for the iteration order of relations, but I still do not see which part of the change guarantees iteration order of phi nodes.

I assume this is change that does it
```
for (final Map.Entry<BlockPair, List<Event>> phiNode : phiNodes.entrySet()) // before
for (final Map.Entry<Label, List<Event>> phiNode : phiNodes.getOrDefault(block.label, Map.of()).entrySet()) // after
```
but I do not really understand why.

It seems the filters in the Wmm class where not used at all, so it makes sense to get rid of them. However, does anybody remember what the role of those was? While I see the role of filters in programs, I have no clue what a filter might be in a memory model.

xeren:
>> It seems the filters in the Wmm class where not used at all, so it makes sense to get rid of them. However, does anybody remember what the role of those was? While I see the role of filters in programs, I have no clue what a filter might be in a memory model.
>
> I think the usage of filters was removed by Renè's PR that added unary relations. He probably forgot to remove the map in the WMM class.

Exactly.

Besides `Event` and `Relation`, there is another class in the IR that probably should become value-based: `Function`.  By extension, `Thread`, `Register` and `Expression` are currently identity-based.
`Function` does not have a persistent identifier.  There already exist two candidate fields.  Each of them has to be altered slightly.

- `Function.id` is set on creation and modified multiple times during program processing.  It also has multiple purposes:
  - Usually corresponds to the index in `program.functions` resp. `program.threads`.
  - Identifies the function for log messages and visualizations.
  - Partially identifies SMT variables for registers.
  - Is the numeric value of the thread handle used by the `pthread_*` intrinsics and the HSA and LOBE progress models.
  - Used in `"END_OF_T"` and `"RETURN_OF_T"` labels.

- `Function.name` is set on creation and modified once in `SymmetryReduction`.  Its purposes currently are:
  - Corresponds to the identifier in the source file, if parsed.
  - Sometimes used in log messages and visualizations, especially with `SyntacticContext`.
  - Used in `Program.getFunctionByName(String)`.
  - Metadata: Name of Init threads starts with `"Init_"`, Litmus threads start with "P", Threads spawned by `pthread_create` share the name with the function argument.

I would prefer `name` to be unique in the program and the sole solution for most of the above situations.  I would move `id` into `Thread` and reduce its usages to intrinsics and progress models.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -466,14 +463,14 @@ private Label getJumpLabel(LabelContext context) {
     }
 
     private List<Event> getPhiNode(Block from, Block to) {
-        final BlockPair pair = new BlockPair(from, to);
-        final List<Event> find = phiNodes.get(pair);
+        final Map<String, List<Event>> phiNodesFrom = phiNodes.computeIfAbsent(from.name, k -> new HashMap<>());
+        final List<Event> find = phiNodesFrom.get(to.name);
```
ThomasHaas, line 467: If you expect the name of blocks to be unique, you could overwrite `equals/hashCode` of `Block` to just compare/use the name. Then you can use `Map<Block, List<Event>>`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/VoidType.java
```diff
@@ -10,4 +10,9 @@ public final class VoidType implements Type {
     public String toString() {
         return "void";
     }
+
+    @Override
+    public int hashCode() {
+        return 1;
+    }
```
ThomasHaas, line 17: We don't have any case of using types in hashmaps, no?

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
xeren: Yes, but i deemed it not worth skipping the four times explicitly using `block.name`.  Anyone wondering what the `String` key is supposed to represent can find out by looking at the next statement.

This might not apply to the declaration of `this.phiNodes`, so I would consider renaming the field to `phiNodesByBlockNames`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/VoidType.java
xeren: Afaik the canonicalizer in `TypeFactory` might be the only instance.

hernanponcedeleon:
I am working on adding native compilation and in there, the non-determinism this PR tries to fix is more visible. On SVCOMP benchmarks I get a bunch of ERROR due to a `The jump does not belong to a loop` exception from loop unrolling when dumping the bounds to a file.

It would be great to finish this PR soon so I can make an SVCOMP run with a larger timeout and create a PR for the native compilation.

PR #950, opened 2025-11-18
Basic support for SVCOMP valid-memtrack

xeren:
The `TRACKABILITY` property checks if the execution exits normally and at least one heap-allocated object is not reachable from any static object.

### Examples

1. This program is unsafe, its property encoding is:
$(\mathtt{leak1}\Rightarrow\mathtt{exec0})\wedge\mathtt{leak1}\wedge\neg\mathtt{track1}$
```c
void main() {
    int* local_int = (int*) malloc(sizeof(int)); // object 1
    return 0;
}
```

2. This program is safe, as the encoding is unsatisfiable:
$(\mathtt{leak1}\Rightarrow\mathtt{exec0})\wedge(\mathtt{exec0}\Rightarrow\neg\mathtt{leak1})\wedge\mathtt{leak1}\wedge\neg\mathtt{track1}$
```c
void main() {
    int* local_int = (int*) malloc(sizeof(int)); // object 1
    free(local_int);
    return 0;
}
```

3. This program is safe as well:
$(\mathtt{leak1}\Rightarrow\mathtt{exec0})\wedge\mathtt{track0}\wedge(\mathtt{track0}\Rightarrow\mathtt{track1})\wedge\mathtt{leak1}\wedge\neg\mathtt{track1}$
```c
static int* global_int; // object 0
void main() {
    global_int = (int*) malloc(sizeof(int)); // object 1
    return 0;
}
```

PR #954, opened 2025-12-18
SVCOMP 2026

hernanponcedeleon:
This PR adds several changes related to SVCOMP 2026
- minor fixes (witness generation for data-races, UBSAN instrumentation options, removed unsound intrinsic),
- support for the mem-track property: while #950 was merged into the svcomp branch (because the results suggested the code was indeed correct), it was never properly reviewed so we need to do this as part of this PR, and
- the changes from #948 (@xeren: we can close that PR and make the review here)

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
@@ -9,6 +9,7 @@ import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.analysis.SyntacticContextAnalysis;
 import com.dat3m.dartagnan.program.event.MemoryEvent;
 import com.dat3m.dartagnan.program.event.RegWriter;
+import com.dat3m.dartagnan.program.event.core.Load;
 import com.dat3m.dartagnan.program.event.core.Local;
 import com.dat3m.dartagnan.program.event.core.MemoryCoreEvent;
 import com.dat3m.dartagnan.program.event.core.Store;
@@ -105,7 +106,7 @@ public class AndersenAliasAnalysis implements AliasAnalysis {
 
     @Override
     public Collection<MemoryObject> communicableObjects(MemoryCoreEvent a) {
-        return allObjects;
+        return a instanceof Load || a instanceof Store ? allObjects : Set.of();
     }
 
     @Override
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
@@ -4,7 +4,9 @@ import com.dat3m.dartagnan.program.Program;
 import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.event.Event;
 import com.dat3m.dartagnan.program.event.RegWriter;
+import com.dat3m.dartagnan.program.event.core.Load;
 import com.dat3m.dartagnan.program.event.core.MemoryCoreEvent;
+import com.dat3m.dartagnan.program.event.core.Store;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
 import com.dat3m.dartagnan.wmm.utils.graph.mutable.MapEventGraph;
 import com.dat3m.dartagnan.wmm.utils.graph.mutable.MutableEventGraph;
@@ -89,7 +91,7 @@ public class EqualityAliasAnalysis implements AliasAnalysis {
 
     @Override
     public Collection<MemoryObject> communicableObjects(MemoryCoreEvent a) {
-        return allObjects;
+        return a instanceof Load || a instanceof Store ? allObjects : Set.of();
     }
 
     @Override
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
@@ -204,7 +204,8 @@ public class InclusionBasedPointerAnalysis implements AliasAnalysis {
     @Override
     public Collection<MemoryObject> communicableObjects(MemoryCoreEvent e) {
         final DerivedVariable v = valueVariables.get(e);
-        return v == null ? objectVariables.keySet() : v.base.object != null ? Set.of(v.base.object)
+        return v == null ? e instanceof Load || e instanceof Store ? objectVariables.keySet() : Set.of()
+                : v.base.object != null ? Set.of(v.base.object)
                 : v.base.includes.stream().map(i -> i.source.object).collect(Collectors.toSet());
     }
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/FieldSensitiveAndersen.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/FieldSensitiveAndersen.java
@@ -108,7 +108,7 @@ public class FieldSensitiveAndersen implements AliasAnalysis {
 
     @Override
     public Collection<MemoryObject> communicableObjects(MemoryCoreEvent e) {
-        return allObjects;
+        return e instanceof Load || e instanceof Store ? allObjects : Set.of();
     }
 
     @Override


hernanponcedeleon:
> I guess the intention was to say that a load that reads from `Dealloc` can have any value (it doesn't matter) and so also the value of `Dealloc` itself doesn't matter.

If that was the intention, we could do the same making dealloc to write non-det values and avoid all those null checks, right?

xeren:
> which does not make any sense to me. @xeren why did you use this value?

I chose the value of the unit type to have it be potentially convertible from and to any other type.  It also avoids MSAs.

> This also explains this code in `EncodingContext.sameValue()`
> [...]
> which I am not a big fan of.

I think I experienced some cast issues here.

>> I guess the intention was to say that a load that reads from `Dealloc` can have any value (it doesn't matter) and so also the value of `Dealloc` itself doesn't matter.
>
> If that was the intention, we could do the same making dealloc to write non-det values and avoid all those null checks, right?

That would be the an alternative.  `Intrinsics` can use `EventFactory.Svcomp.newNondetChoice(...)` for that.  You just need to guess the right data type.

> Yes and no. First, since the value doesn't really matter anymore (reading from Dealloc is a bug), we could also just default to 0 instead of using a non-det value. However, I think the major problem is that you have to come up with some type for that value (int, pointer, float, aggregate, and what size?) so that any load can read from it and ideally without introducing MSA. Renè's code avoids this problem by just letting the load decide its value based on its own type independent of the value/type of the Dealloc.
> Still, I don't like having null as a value here and we should find a better way to describe this situation.

Alternatively avoid ever calling `EncodingContext.value(MemoryCoreEvent)` on `Dealloc`.  This would be more explicit, e.g.
```java
return first instanceof Dealloc || second instanceof Dealloc
        ? booleanFormulaManager.makeTrue()
        : exprEncoder.equal(value(first), value(second), cmode);
```

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
```diff
@@ -14,8 +14,9 @@ public enum Property implements OptionInterface {
     PROGRAM_SPEC,        // Litmus queries OR assertion safety in C-code
     TERMINATION,         // All executions terminate (absence of deadlocks/livelocks)
     CAT_SPEC,            // CAT-spec defined via flagged axioms in .cat file (~bug specification)
-    DATARACEFREEDOM;     // Special option for data-race detection in SVCOMP only
-
+    DATARACEFREEDOM,     // Special option for data-race detection in SVCOMP only
+    TRACKABILITY,  // All heap-allocated objects are either freed or remain reachable at the end of the execution.
```
hernanponcedeleon: Align at `//`

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
```diff
@@ -31,6 +32,7 @@ public String toString() {
             case DATARACEFREEDOM -> "Data-race freedom (SVCOMP only)";
             case TERMINATION -> "Termination";
             case CAT_SPEC -> "CAT specification";
+            case TRACKABILITY -> "Memory Trackability";
```
hernanponcedeleon, line 35: Unless we see a use case of this outside SVCOMP, I would add "(SVCOMP only)" for consistency with data race freedom

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -190,7 +190,11 @@ public BooleanFormula sameResult(RegWriter first, RegWriter second) {
     }
 
     public BooleanFormula sameValue(MemoryCoreEvent first, MemoryCoreEvent second, ExpressionEncoder.ConversionMode cmode) {
-        return exprEncoder.equal(value(first), value(second), cmode);
+        final TypedFormula<?, ?> firstValue = value(first);
+        final TypedFormula<?, ?> secondValue = value(second);
+        return firstValue == null || secondValue == null
+                ? booleanFormulaManager.makeTrue()
+                : exprEncoder.equal(firstValue, secondValue, cmode);
```
hernanponcedeleon: As discussed, we need a cleaner solution than this

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/NonTerminationEncoder.java
```diff
@@ -386,7 +386,7 @@ private BooleanFormula areEquivalent(Event x, Event y) {
         }
         if (x instanceof MemoryCoreEvent e && y instanceof MemoryCoreEvent f) {
             equality = bmgr.and(equality, context.sameAddress(e, f));
-            if (x instanceof Store || x instanceof Load) {
+            if (x instanceof Store && !(x instanceof Dealloc) || x instanceof Load) {
```
hernanponcedeleon: Add parenthesis to make it clear how this should bind

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -398,6 +402,93 @@ public TrackableFormula encodeDataRaces() {
...
+                .collect(Collectors.toMap(o -> o, o -> new Var(context.leakVariable(o), context.trackVariable(o))));
+        // Requires all threads to exit normally.
+        for (Event event : program.getThreadEvents()) {
+            if (Stream.of(Tag.BOUND, Tag.EXCEPTIONAL_TERMINATION, Tag.NONTERMINATION).anyMatch(event::hasTag)) {
+                enc.add(bmgr.not(context.execution(event)));
+            }
+        }
+        // Do not leak unallocated objects.
+        for (Map.Entry<MemoryObject, Var> entry : variables.entrySet()) {
+            enc.add(bmgr.or(context.execution(entry.getKey().getAllocationSite()), bmgr.not(entry.getValue().leak)));
```
hernanponcedeleon: It took me a while to figure out this was `leak => allocated`. Since `p => q = ~p \/ q` can you invert the disjunction?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -398,6 +402,93 @@ public TrackableFormula encodeDataRaces() {
...
+        return true;
+    }
+
+    private BooleanFormula referencesObject(Store store, boolean isValue, MemoryObject object) {
+        //TODO enhance with provenance
+        final TypedFormula<?, ?> address = isValue ? context.value(store) : context.address(store);
+        final Formula base = context.address(object).formula();
+        final Formula size = context.size(object).formula();
+        final int addressSize = types.getMemorySizeInBytes(address.type());
+        if (address.formula() instanceof BitvectorFormula a && base instanceof BitvectorFormula b && size instanceof BitvectorFormula s) {
```
hernanponcedeleon: What about the integer encoding?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
```diff
@@ -398,6 +402,93 @@ public TrackableFormula encodeDataRaces() {
...
+            final BitvectorFormulaManager bvmgr = context.getFormulaManager().getBitvectorFormulaManager();
+            final BitvectorFormula end = bvmgr.add(b, s);
+            final BitvectorFormula c = isValue ? end : bvmgr.subtract(end, bvmgr.makeBitvector(bvmgr.getLength(end), addressSize));
+            final BooleanFormula lowerBound = bvmgr.lessOrEquals(b, a, false);
+            final BooleanFormula upperBound = bvmgr.lessOrEquals(a, c, false);
+            return bmgr.and(lowerBound, upperBound);
+        }
+        throw new UnsupportedOperationException("inArrayBounds(%s, %s, %s)".formatted(address.formula(), base, size));
```
hernanponcedeleon: I do not think this error tells much

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
```diff
@@ -78,6 +83,17 @@ public boolean mayAlias(MemoryCoreEvent a, MemoryCoreEvent b) {
...
     }
 
+    @Override
+    public Collection<MemoryObject> addressableObjects(MemoryCoreEvent a) {
+        return allObjects;
+    }
+
+    @Override
+    public Collection<MemoryObject> communicableObjects(MemoryCoreEvent a) {
+        final int size = a.getAccessType() instanceof IntegerType t ? t.getBitWidth() : 0;
```
hernanponcedeleon: Why is the "fallback" 0 here but 1 in Andersen?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -124,6 +122,10 @@ public static Alloc newAlignedAlloc(Register register, Type allocType, Expressio
         return new Alloc(register, allocType, arraySize, alignment, isHeapAlloc, doesZeroOutMemory);
     }
 
+    public static Dealloc newDealloc(Expression address) {
+        return new Dealloc(address, expressions.makeConstruct(types.getAggregateType(List.of()), List.of()));
```
hernanponcedeleon: As I already mentioned, I am not sure using the fake aggregate type is the cleanest solution

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -71,7 +71,16 @@ private boolean canInline(FunctionCall call) {
...
-        int scopeCounter = 0;
+        int scope = 0;
+        // Advance scope counter.
+        for (Register r : function.getRegisters()) {
+            final int i = r.getName().indexOf(":");
+            if (i == -1) { continue; }
+            try {
+                scope = Integer.max(scope, Integer.parseInt(r.getName().substring(0, i)));
+            } catch (NumberFormatException ignore) {
+            }
```
hernanponcedeleon: No need for the extra line

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
```diff
@@ -71,7 +71,16 @@ private boolean canInline(FunctionCall call) {
...
 
     private void inlineAllCalls(Function function, Map<Function, Snapshot> snapshots) {
-        int scopeCounter = 0;
+        int scope = 0;
+        // Advance scope counter.
+        for (Register r : function.getRegisters()) {
+            final int i = r.getName().indexOf(":");
+            if (i == -1) { continue; }
+            try {
+                scope = Integer.max(scope, Integer.parseInt(r.getName().substring(0, i)));
```
hernanponcedeleon: Why do we need all this?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
```diff
@@ -61,7 +61,7 @@ private Matcher analyze(Function function) {
         // Initially, all locally-allocated addresses are potentially promotable.
         for (final Alloc allocation : function.getEvents(Alloc.class)) {
             // Allocations will usually not have users.  Otherwise, their object is not promotable.
-            if (allocation.getUsers().isEmpty()) {
+            if (allocation.getUsers().isEmpty() && !allocation.isHeapAllocation()) {
```
hernanponcedeleon, line 66: We do we need this now and not before?

hernanponcedeleon (commented):
First pass up to "RemoveDeadFunctions" pass (minus the pthread classes). Will try to review the rest in the next few days.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
ThomasHaas: Just use `bmgr.implication` :)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/MemToReg.java
ThomasHaas: My guess is that promoting heap allocations hides memory leaks if said heap allocation is never freed.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
ThomasHaas: I would also refactor this code into a helper function `getNextFreeScopeId`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadFunctions.java
```diff
@@ -59,22 +67,71 @@ private Set<Function> findReachableFunctions(Program program) {
...
 
+    private void liveFunctions(Function function, FunctionCollector liveFunctions) {
+        final var liveRegisters = new HashSet<Register>();
+        final var jumps = new HashMap<Event, Set<Register>>();
+        final List<Event> events = function.getEvents();
+        for (int i = events.size() - 1; i >= 0; i--) {
+            Event event = events.get(i);
+            if (IRHelper.isAlwaysBranching(event)) { liveRegisters.clear(); }
+            final Set<Register> incoming = jumps.get(event);
+            if (incoming != null) { liveRegisters.addAll(incoming); }
```
hernanponcedeleon: I do not like this kind of null checking. Can we rather initialize `jumps` with the empty set for each jump in the function? This also allows us to get rid of the `computeIfAbssent` below

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadFunctions.java
```diff
@@ -59,22 +67,71 @@ private Set<Function> findReachableFunctions(Program program) {
...
+            if (event instanceof Label label) {
+                for (CondJump jump : label.getJumpSet()) {
+                    if (!jump.isDead() && jumps.computeIfAbsent(jump, k -> new HashSet<>()).addAll(liveRegisters)) {
+                        // Redo, if there is a backjump.
+                        i = Integer.max(i, events.indexOf(jump) + 1);
```
hernanponcedeleon: can't we end up looping forever because of this?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadFunctions.java
```diff
@@ -59,22 +67,71 @@ private Set<Function> findReachableFunctions(Program program) {
...
+            public Expression visitExpression(Expression x) {
+                if (!x.equals(exception)) {
+                    expressions.add(x);
+                }
+                return x;
+            }
+        };
+        reader.transformExpressions(collector);
+        return expressions;
+    }
```
hernanponcedeleon, line 127: I do not understand this method

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/ResolveAborts.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class ResolveAborts implements FunctionProcessor {
+
+    private ResolveAborts() {}
+
+    public static ResolveAborts newInstance() { return new ResolveAborts(); }
+
+    @Override
+    public void run(Function function) {
+        checkArgument(function instanceof Thread, "Non-thread %s", function.getName());
```
hernanponcedeleon: The error msg is not very informative, please add a reference to the pass

dartagnan/src/main/java/com/dat3m/dartagnan/utils/ExitCode.java
```diff
@@ -5,6 +5,7 @@ public enum ExitCode {
     NORMAL_TERMINATION, BOUNDED_RESULT,
     // Verification errors
     PROGRAM_SPEC_VIOLATION, CAT_SPEC_VIOLATION, TERMINATION_VIOLATION, DATA_RACE_FREEDOM_VIOLATION,
+    MEMORY_SAFETY_VIOLATION,
```
hernanponcedeleon: Memory safety can be more general than trackability. This should be more "mem-track" specific

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadLocalCreate.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override public void setResultRegister(Register r) { key = checkNotNull(r); }
+
+    @Override
+    public Set<Register.Read> getRegisterReads() {
+        return Register.collectRegisterReads(destructor, Register.UsageType.OTHER, new HashSet<>());
+    }
+
+    @Override
+    public void transformExpressions(ExpressionVisitor<? extends Expression> transformer) {
+        destructor = destructor.accept(transformer);
```
hernanponcedeleon, line 49: why is the key not transformed? Is it because it is in the LHS?

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
```diff
@@ -48,16 +48,23 @@ public class MemoryObject extends LeafExpressionBase<Type> {
...
+
+    public record PosixThreadLocal(MemoryObject destructor) implements ThreadLocalMode {}
+
     public boolean hasName() { return name != null; }
     public String getName() { return name; }
     public void setName(String name) { this.name = name; }
 
     public boolean isStaticallyAllocated() { return allocationSite == null; }
     public boolean isDynamicallyAllocated() { return !isStaticallyAllocated(); }
+    public boolean isHeapAllocated() { return allocationSite != null && allocationSite.isHeapAllocation(); }
```
hernanponcedeleon, line 63: Can we have a heap allocation without an allocation site? May any of our passes remove the allocation site?

ThomasHaas: We consider static allocations (without a allocation site) as heap allocated. For dynamic allocations, the memory object is typically created at the very end of the processing pass, so it is unlikely that an allocation site gets deleted then. That being said, it could happen but we will likely get a runtime exception during encoding then, because the memory object references an invalid event.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/dat3m/DynamicThreadLocalCreate.java
ThomasHaas: Yes.

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Property.java
ThomasHaas: The comment `(SVCOMP only)` does not mean that we currently only use it for `SVCOMP` (why would we care?) but because the input memory model needs to be `svcomp.cat` to get the desired semantics.
This is different for trackability: it should work and make sense for all memory models.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/PropertyEncoder.java
ThomasHaas: I agree.  `context.address` and `context.size` return `TypedFormulas` which are expressions, meaning you can use our standard `ExpressionFactory` to construct the encoding. You just need to call `exprEnc.encodeBooleanFinal(...)` (not sure about the name) to convert the final expression into a `BooleanFormula`. 
This way, the formula you construct is agonistic about the concrete encoding we use and hence automatically also works for integers.

Indeed, I think most if not all encodings related to our IR types/expressions should be done via `ExpressionFactory`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -192,6 +195,22 @@ public boolean mustAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
+                : v.base.includes.stream().map(i -> i.source.object).collect(Collectors.toSet());
+    }
+
+    @Override
+    public Collection<MemoryObject> communicableObjects(MemoryCoreEvent e) {
+        final int size = e.getAccessType() instanceof IntegerType t ? t.getBitWidth() : 0;
+        if (size < 64) { return Set.of(); }
+        final DerivedVariable v = valueVariables.get(e);
+        return v == null ? objectVariables.keySet() : v.base.object != null ? Set.of(v.base.object)
+                : v.base.includes.stream().map(i -> i.source.object).collect(Collectors.toSet());
```
ThomasHaas, line 209: Does this reasoning even hold up in the presence of MSA? A pointer store can get teared so that none of the subevents "communicate" objects anymore.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
ThomasHaas: The "fake aggregate type" is actually a "unit type". I'm not sure if using the unit type here is a good idea or not, but if you want to use it, you should definitely make this more explicit and add methods `TypeFactory.getUnitType` and `ExpressionFactory.makeUnit()` and use those.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadFunctions.java
ThomasHaas: It can't happen because of the condition before: the entries in `jumps` are monotonically increasing and bounded, so eventually the loop will stop.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadFunctions.java
ThomasHaas: Btw. `SingleAccessMemoryEven` is an implementation detail and should not be checked against anywhere.
I'm not sure why you even have the code like that? It looks like a really marginal optimization to me.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/EqualityAliasAnalysis.java
xeren: You're right.  It should return 1 here like in the other two implementations.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Inlining.java
xeren: > Why do we need all this?

TSS destructor calls appear after thread creation.  Thus we need a second `Inlining` pass (see `ProcessingManager`).  During the second pass we would get name collisions for new registers.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/RemoveDeadFunctions.java
xeren: This method collects those expressions in `event`, that any contained function pointer may end up in a call event.  It contains a black list:  a function is only considered live, if its pointer is used in a situation other than
- unused assignments
- addresses
- guards

Though I am not so sure anymore if we really need to exclude anything other than guards: function pointers in addresses should be rare and unused assignments could be removed in `DeadAssignmentElimination`.

The reason for this method is that for thread-specific storage destructors to work, `NaiveDevirtualisation`, as early in the pipeline as it is, must not replace occurrences of `Function` by new instances of `MemoryObject`, as it did before.  When `RemoveDeadFunctions` happens afterwards, there are still many reachable 'usages' of `Function`, some may even be stored in memory.  When less functions got removed, then there are a lot more to run later passes on.  I experienced considerable performance issues.  So I eased the impact a bit with a weaker definition for dead functions.

dartagnan/src/main/java/com/dat3m/dartagnan/program/memory/MemoryObject.java
xeren: Yes, `EncodingContext.execution(Event)` would return `null` for the removed allocation site, instead of a formula.  Usually this causes a `NullPointerException` later.
(Otherwise, the encoding could confuse the removed allocation site with the event inheriting its global ID.  The object being allocated in the execution and e.g. violating trackability might depend on some originally-unrelated event.)
We could add a verification pass for this into `ProcessingManager`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
hernanponcedeleon: Other than this open comment, I am fine with the current status

hernanponcedeleon: @xeren ping!

xeren: You are right: It does not hold up.

I modified the encoding so that it reconstructs the final value from the subevents.  There are now also a few small benchmarks for this feature.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AndersenAliasAnalysis.java
```diff
@@ -101,13 +100,12 @@ public boolean mustAlias(MemoryCoreEvent x, MemoryCoreEvent y) {
...
     public Collection<MemoryObject> addressableObjects(MemoryCoreEvent a) {
-        return allObjects;
+        return getMaxAddressSet(a).stream().map(l -> l.base).collect(Collectors.toSet());
     }
 
     @Override
     public Collection<MemoryObject> communicableObjects(MemoryCoreEvent a) {
-        final boolean fitsAddress = a.getAccessType() instanceof IntegerType t && t.getBitWidth() >= 64;
-        return fitsAddress ? allObjects : Set.of();
+        return allObjects;
```
ThomasHaas: I would probably either check `a instanceof Store` or return `List.of()` for non-stores.
Or does this also reason about the value returned by the load? In that case, it is fine I guess.

xeren: Yes, load could potentially have communicable objects, although we are only concerned about stores for now.  At least `GenericMemoryEvent` and `Dealloc` should probably always have an empty value set.

PR #957, opened 2025-12-28
New memory type & explicit encoding of assignments

ThomasHaas:
This PR adds a new family of types `MemoryType<N>` (`N` is the number of bits) with the following operations;
- Casts between `MemoryType<N>` and any type `T` that is storable in memory with`memorySizeInBits(T) == N` (currently all our types are storable in memory, but this need not be the case in principle). A round-trip `T -> MemoryType<N> -> T` is lossless, meaning `MemoryType<N>` must be able to precisely capture any type T of appropriate size (that is storable in memory).
- Concat/Extract. Round-trips over extract/concat `MemoryType<N> -extract-> X*MemoryType<N/X> -concat-> MemoryType<N>` must be lossless, i.e., if we split the type into chunks and then concat them again correctly, we get the original object.

### The purposes of `MemoryType<N>`:

(1) Any data-flow over memory, i.e., between stores and loads, can be modelled in a more principled manner.
Conceptually, every `store(addr, val)` stores `val to MemoryType<N>` and every load `T r := load(addr)` loads a value `MemoryType<N> loadVal` and then converts it `r = loadVal to T`. By the lossless round-trip property, if the type of the load and the store match, then we get the expected semantics.
However, if the types mismatch, a sort of reinterpret happens over the intermediary memory type: `T1 -> MemoryType<N> -> T2`. This gives a unified way to deal with type mismatches/conversions over memory.

(2) Since `MemoryType<N>` supports concat/extract with lossless round-trips, we can uniformly model `Tearing` for all types:
- A store of type T gets teared like this:`T -cast-> MemoryType<N> -extract-> X*MemoryType<N/X> -store-> ...`
- A load of Type T gets teared like this: `... -load-> X*MemoryType<N/X> -concat-> MemoryType<N> -cast-> T`

This should enable automatic support for tearing of `FloatType` ( and `PointerType` #927). That being said, the current implementation does not yet encode all possible type conversions and treats `MemoryType<N>` identical to `IntegerType<N>` in the encoding.
To support lossless round-trips, the encoding of `MemoryType<N>` will likely need to become a tuple `<metadata, bits>` to keep track of additional information such as provenance, address space, or even original type.

### Distuingishing between assignments and equality

To better control where and when type conversions should occur, I added `assignEqual` methods to `ExpressionEncoder`. 
The usual `equals` methods now always encode a strict equality (no conversion) whereas `assignEqual` takes an additional parameter to define what conversion should happen (if any).
The encoding should use `assignEqual` whenever it encodes assignments, i.e., any sort of data-flow like read-from and  register assignments.

### DIFF
--- a/dartagnan/src/test/resources/miscellaneous/floats_1.ll
+++ b/dartagnan/src/test/resources/miscellaneous/floats_1.ll
@@ -13,56 +13,60 @@ define dso_local i32 @main() #0 !dbg !10 {
   %1 = alloca i32, align 4
   %2 = alloca float, align 4
   %3 = alloca double, align 8
+  %4 = alloca double, align 8
   store i32 0, i32* %1, align 4
   call void @llvm.dbg.declare(metadata float* %2, metadata !16, metadata !DIExpression()), !dbg !18
-  %4 = call float @__VERIFIER_nondet_float(), !dbg !19
-  store float %4, float* %2, align 4, !dbg !18
+  %5 = call float @__VERIFIER_nondet_float(), !dbg !19
+  store float %5, float* %2, align 4, !dbg !18
   call void @llvm.dbg.declare(metadata double* %3, metadata !20, metadata !DIExpression()), !dbg !22
-  %5 = call double @__VERIFIER_nondet_double(), !dbg !23
-  store double %5, double* %3, align 8, !dbg !22
-  %6 = load float, float* %2, align 4, !dbg !24
-  %7 = fcmp uno float %6, %6, !dbg !24
-  br i1 %7, label %8, label %15, !dbg !26
+  %6 = call double @__VERIFIER_nondet_double(), !dbg !23
+  store double %6, double* %3, align 8, !dbg !22
+  call void @llvm.dbg.declare(metadata double* %4, metadata !24, metadata !DIExpression()), !dbg !25
+  %7 = call double @__VERIFIER_nondet_double(), !dbg !26
+  store double %7, double* %4, align 8, !dbg !25
+  %8 = load float, float* %2, align 4, !dbg !27
+  %9 = fcmp uno float %8, %8, !dbg !27
+  br i1 %9, label %10, label %17, !dbg !29
 
-8:                                                ; preds = %0
-  %9 = load float, float* %2, align 4, !dbg !27
-  %10 = load float, float* %2, align 4, !dbg !27
-  %11 = fcmp une float %9, %10, !dbg !27
-  br i1 %11, label %12, label %13, !dbg !30
+10:                                               ; preds = %0
+  %11 = load float, float* %2, align 4, !dbg !30
+  %12 = load float, float* %2, align 4, !dbg !30
+  %13 = fcmp une float %11, %12, !dbg !30
+  br i1 %13, label %14, label %15, !dbg !33
 
-12:                                               ; preds = %8
-  br label %14, !dbg !30
+14:                                               ; preds = %10
+  br label %16, !dbg !33
 
-13:                                               ; preds = %8
-  call void @__assert_fail(i8* noundef getelementptr inbounds ([7 x i8], [7 x i8]* @.str, i64 0, i64 0), i8* noundef getelementptr inbounds ([54 x i8], [54 x i8]* @.str.1, i64 0, i64 0), i32 noundef 20, i8* noundef getelementptr inbounds ([15 x i8], [15 x i8]* @__PRETTY_FUNCTION__.main, i64 0, i64 0)) #4, !dbg !27
-  unreachable, !dbg !27
+15:                                               ; preds = %10
+  call void @__assert_fail(i8* noundef getelementptr inbounds ([7 x i8], [7 x i8]* @.str, i64 0, i64 0), i8* noundef getelementptr inbounds ([54 x i8], [54 x i8]* @.str.1, i64 0, i64 0), i32 noundef 18, i8* noundef getelementptr inbounds ([15 x i8], [15 x i8]* @__PRETTY_FUNCTION__.main, i64 0, i64 0)) #4, !dbg !30
+  unreachable, !dbg !30
 
-14:                                               ; preds = %12
-  br label %15, !dbg !31
+16:                                               ; preds = %14
+  br label %17, !dbg !34
 
-15:                                               ; preds = %14, %0
-  %16 = load double, double* %3, align 8, !dbg !32
-  %17 = fcmp uno double %16, %16, !dbg !32
-  br i1 %17, label %18, label %25, !dbg !34
+17:                                               ; preds = %16, %0
+  %18 = load double, double* %3, align 8, !dbg !35
+  %19 = fcmp uno double %18, %18, !dbg !35
+  br i1 %19, label %20, label %27, !dbg !37
 
-18:                                               ; preds = %15
-  %19 = load double, double* %3, align 8, !dbg !35
-  %20 = load double, double* %3, align 8, !dbg !35
-  %21 = fcmp une double %19, %20, !dbg !35
-  br i1 %21, label %22, label %23, !dbg !38
+20:                                               ; preds = %17
+  %21 = load double, double* %3, align 8, !dbg !38
+  %22 = load double, double* %3, align 8, !dbg !38
+  %23 = fcmp une double %21, %22, !dbg !38
+  br i1 %23, label %24, label %25, !dbg !41
 
-22:                                               ; preds = %18
-  br label %24, !dbg !38
+24:                                               ; preds = %20
+  br label %26, !dbg !41
 
-23:                                               ; preds = %18
-  call void @__assert_fail(i8* noundef getelementptr inbounds ([7 x i8], [7 x i8]* @.str.2, i64 0, i64 0), i8* noundef getelementptr inbounds ([54 x i8], [54 x i8]* @.str.1, i64 0, i64 0), i32 noundef 21, i8* noundef getelementptr inbounds ([15 x i8], [15 x i8]* @__PRETTY_FUNCTION__.main, i64 0, i64 0)) #4, !dbg !35
-  unreachable, !dbg !35
+25:                                               ; preds = %20
+  call void @__assert_fail(i8* noundef getelementptr inbounds ([7 x i8], [7 x i8]* @.str.2, i64 0, i64 0), i8* noundef getelementptr inbounds ([54 x i8], [54 x i8]* @.str.1, i64 0, i64 0), i32 noundef 19, i8* noundef getelementptr inbounds ([15 x i8], [15 x i8]* @__PRETTY_FUNCTION__.main, i64 0, i64 0)) #4, !dbg !38
+  unreachable, !dbg !38
 
-24:                                               ; preds = %22
-  br label %25, !dbg !39
+26:                                               ; preds = %24
+  br label %27, !dbg !42
 
-25:                                               ; preds = %24, %15
-  ret i32 0, !dbg !40
+27:                                               ; preds = %26, %17
+  ret i32 0, !dbg !43
 }
 
 ; Function Attrs: nofree nosync nounwind readnone speculatable willreturn
@@ -86,7 +90,7 @@ attributes #4 = { noreturn nounwind }
 !llvm.ident = !{!9}
 
 !0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, producer: "Ubuntu clang version 14.0.0-1ubuntu1.1", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, splitDebugInlining: false, nameTableKind: None)
-!1 = !DIFile(filename: "/home/drc/git/Dat3M/benchmarks/miscellaneous/floats.c", directory: "/home/drc/git/Dat3M", checksumkind: CSK_MD5, checksum: "016a68962542b9f94ab88bda45b97e2a")
+!1 = !DIFile(filename: "/home/drc/git/Dat3M/benchmarks/miscellaneous/floats.c", directory: "/home/drc/git/Dat3M", checksumkind: CSK_MD5, checksum: "655be55c9424586b3c1ea9b9681db180")
 !2 = !{i32 7, !"Dwarf Version", i32 5}
 !3 = !{i32 2, !"Debug Info Version", i32 3}
 !4 = !{i32 1, !"wchar_size", i32 4}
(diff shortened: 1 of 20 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -171,16 +190,25 @@ public void setEvent(Event e) {
...
             assert typedFormula.formula() instanceof IntegerFormula || typedFormula.formula() instanceof BitvectorFormula;
             return (TypedFormula<IntegerType, ?>) typedFormula;
         }
 
+        @SuppressWarnings("unchecked")
+        public TypedFormula<MemoryType, ?> encodeMemoryExpr(Expression expression) {
+            checkArgument(expression.getType() instanceof MemoryType);
+            final TypedFormula<?, ?> typedFormula = encode(expression);
+            assert typedFormula.getType() == expression.getType();
+            assert typedFormula.formula() instanceof IntegerFormula || typedFormula.formula() instanceof BitvectorFormula;
```
hernanponcedeleon: There was some assert in other part of the code suggesting that `MemoryType` cannot be encoded as integer (which made sense), but then it seems to me the left-hand side here is useless

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -552,6 +580,72 @@ public TypedFormula<?, TupleFormula> visitInsertExpression(InsertExpr insert) {
...
+            final Type targetType = expr.getTargetType();
+
+            // TODO: Do actual conversions
+            return new TypedFormula<Type, Formula>(targetType, inner.formula());
+        }
+
+        @Override
+        public TypedFormula<?, ?> visitMemoryConcatExpression(MemoryConcat expr) {
+            checkArgument(!expr.getOperands().isEmpty());
+            Preconditions.checkState(!context.useIntegers);
```
hernanponcedeleon: For consistency with the other changes in the PR, remove the `Preconditions`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -552,6 +580,72 @@ public TypedFormula<?, TupleFormula> visitInsertExpression(InsertExpr insert) {
...
+                final IntegerFormula lowBitValue = imgr.makeNumber(BigInteger.TWO.pow(expr.getLowBit()));
+                final IntegerFormula op = (IntegerFormula) operand;
+                final IntegerFormula extracted = expr.isExtractingHighBits() ? op : imgr.modulo(op, highBitValue);
+                enc = expr.isExtractingLowBits() ? extracted : imgr.divide(extracted, lowBitValue);
+            } else {
+                final BitvectorFormulaManager bvmgr = bitvectorFormulaManager();
+                enc = bvmgr.extract((BitvectorFormula) operand, expr.getHighBit(), expr.getLowBit());
+            }
+            return new TypedFormula<>(expr.getType(), enc);
+        }
```
hernanponcedeleon, line 831: The "partial" support for `context.useIntegers` is annoying. Do the unit test really require integer encodings? Overall I would be fine with moving more and more to precisely encode the program semantics by trying to always use BVs

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -338,7 +350,66 @@ public ScopedPointerVariable makeScopedPointerVariable(String id, ScopedPointerT
...
+
+        Expression exprMem = makeToMemoryCast(expr);
+        if (targetSize < sourceSize) {
+            exprMem = makeMemoryExtract(exprMem, 0, targetSize - 1);
+            exprMem = makeFromMemoryCast(exprMem, types.getCompatibleTypeOfMemorySize(targetType, targetSize));
+            exprMem = makeCast(exprMem, targetType, signed);
+        } else if (targetSize == sourceSize) {
+            exprMem = makeFromMemoryCast(exprMem, targetType);
+        } else {
+            assert (targetSize > sourceSize);
```
hernanponcedeleon: what's the point of the assert ... it is clear it holds from the control flow

dartagnan/src/main/java/com/dat3m/dartagnan/expression/memory/MemoryEqualExpr.java
```diff
@@ -0,0 +1,22 @@
...
+import com.dat3m.dartagnan.expression.ExpressionVisitor;
+import com.dat3m.dartagnan.expression.base.BinaryExpressionBase;
+import com.dat3m.dartagnan.expression.type.BooleanType;
+import com.dat3m.dartagnan.expression.type.MemoryType;
+import com.dat3m.dartagnan.expression.utils.ExpressionHelper;
+
+public class MemoryEqualExpr extends BinaryExpressionBase<BooleanType, ExpressionKind> {
+
+    public MemoryEqualExpr(BooleanType type, Expression left, Expression right) {
+        super(type, () -> "EQ", left, right);
```
hernanponcedeleon, line 14: This and other hardcode strings are ugly. We need a proper `MemoryXOp` class

hernanponcedeleon (commented):
@ThomasHaas when I was working on adding better support for floats, I remember we discussed about some problem of floats + tearing, but I do not remember the details. Do these changes "solve" all problems, and we should easily get tearing working with floats?

Do we also need these changes to get correct semantics for #927?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: I initially put such assertions everywhere where `MemoryType` was encoded. However, some of those assertions failed and I removed them again. However, I think I still kept those that did not fail.

The reason is that the `MemoryType` basically always occurs somewhere if you have rf-edges between mismatching types, either due to MSA (mismatching type sizes) or implicit conversions over memory (e.g., storing a float but loading it as an int).
We currently have a few such cases even when doing integer encodings.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: Oh, I didn't even notice this change. I guess I accidentally let my IDE add a static import. I would rather keep the full `Precondtions.` prefix though.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: Depends on what you mean by "require". Currently, some unit tests use integer encoding and if we keep the tests as such, we need the support.
I think in this particular case it was that some init value was `i64` but we have a load using `i32` or smaller (maybe due to `memcpy` in the test?). We could surely change the unit tests to use BV encoding though.
There might also be some such cases among the Litmus tests, but I'm not sure.

All that being said, I agree that the integer encoding support is a bit annoying. I don't really mind the little bit of extra code, but the fact that the semantics of many things like MSA, type conversions, signedness and so on become unclear when dealing with integers.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
ThomasHaas: That is kinda the point. I mean, assertions are supposed to hold :P.
I just like having the condition being explicitly stated rather than inferred from the context. But that is subjective opinion.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/memory/MemoryEqualExpr.java
ThomasHaas: Honestly, I would rather get rid of the `ExpressionsKind.Other` enum and use these "hardcoded" strings whenever they are irrelevant as they have a one-to-one correspondence with the expression class they are in.
Maybe it is even worth getting rid of `ExpressionKind` altogether and only use it for those expressions where it makes sense.

In the particular case of equality, I also think we should introduce a generic `EqualityExpr` that works over all types. It is unnecessary to have separate equality expressions for each type if they all mean the same thing (only for `FloatExpr` there are even different notions of equality).

ThomasHaas:
> @ThomasHaas when I was working on adding better support for floats, I remember we discussed about some problem of floats + tearing, but I do not remember the details. Do these changes "solve" all problems, and we should easily get tearing working with floats?

The idea is to streamline the handling of all data going through memory. The conceptual idea is that when logical data goes into memory it gets "serialized" into some byte representation and when it gets out of memory it gets "deserialized" into logical data again.
This serialization/deserialization process should deal (in principle) with stuff like padding, memory layout, endianness, and tearing.
All that needs to be defined per type is how that type gets serialized/deserialized. So for example, tearing of floats should work out-of-the-box as long as long as you define how conversions between `floatX` and `memX` work.
In principle it also solves type conversions between `floatX` to `iX` in the form of reinterpret casts.
That being said, the current implementation in this PR does not even perform an actual conversion, it just reinterprets a formula of type `T` as a formula of type `memX`, which works fine as long as we convert the formula eventually back to type `T` and not to some other type `T2`, i.e., we don't do type casting through memory (which does not currently happen as we basically only ever store `IntegerType` in memory).

> Do we also need these changes to get correct semantics for #927?

They are not "needed", but would make #927 a lot easier to implement because it streamlines tearing of pointers.

ThomasHaas:
> Btw. I'm not yet sure how to best handle types whose logical size is smaller than their memory size. For example, `bv1` should probably get converted to `mem8` rather than `mem1`. I implemented it this way, but to do the conversion, I first do a zero extension `bv1 -> bv8` and then the cast to `mem8`. Maybe it is better to first convert `bv1->mem1` and then do a zero extension `mem1->mem8`. Or maybe it is even better to just stay with `mem1` and not extend it at all? I don't know... I think both approaches are better than the current one, because not all types will inherently support the concept of zero extension.

I changed the approach now: Instead of having a loss-free cast between `bvX <-> memX`, we always cast to/from the next larger memory type whose size is a multiple of 8. So, e.g., `bv1 <-> mem8` and `bv13 <-> mem16`, reflecting the fact that each type's byte representation is a multiple of 8.

ThomasHaas:
The branch works better than I expected.
However, there is still one problem with `Mem2Reg` which currently inserts regular type-casts instead of bitcasts.
Apart from that, I think the branch should work.

ThomasHaas:
It seems the error in `float9` is due to Z3 being wrong.
So I think the PR is almost ready to be reviewed (I will do minor cleanup).
What we need to check is if the new memory cast operations make AA more imprecise or inhibit some other optimizations.

Another thing that could be done in principle is to replace the mem type by our int type for now as the memory type does not yet carry any metadata to differentiate it from ints. However, even if we did this, I would still keep the API of memory types/operations to make the intention of the operations clear.

hernanponcedeleon:
> So I think the PR is almost ready to be reviewed (I will do minor cleanup).

When it is really just remove the DRAFT dtatus

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: I added several simplifications to `ExprSimplifier` regarding casts between memory and integer types.
The simplifications seem to be enough that without MSA and without mixed type accesses (MTA?) all memory operations get eliminated. So I added checks that `!context.useIntegers` holds whenever memory expressions are encoded.
Btw. we probably also want to check `!context.useIntegers` when dealing with floats: many operations like `isinf` or `signbit` result in float->int conversions.

hernanponcedeleon:
@ThomasHaas can you please fix rebase / conflicts?

benchmarks/miscellaneous/floats.c
```diff
@@ -148,6 +144,35 @@ int main(void) {
         assert(isnan(x));
     }
 
+ #elif TEST_ID == 15
+     /* --- fmin vs. fmax and signed zero --- */
+     double d2 = __VERIFIER_nondet_double();
```
hernanponcedeleon: I would rather define `d2` at the top rather than for 15 and 16. This is consistent with e.g., `d` not being used by Test_3

benchmarks/miscellaneous/floats.c
```diff
@@ -148,6 +144,35 @@ int main(void) {
...
+         #ifdef FAIL
+             assert (maxNegImpliesMinNeg);
+         #endif
+     }
+
+#elif TEST_ID == 16
+    /* --- fmax and signed zero corner case --- */
+    double d2 = __VERIFIER_nondet_double();
+    if (!isnan(d) && !isnan(d2)) {
+        assert (fmax(d, d2) == fmax(d, d2));
```
hernanponcedeleon, line 165: is this supposed to check that the SMT solver is deterministic for max? or that it can return once +0 and once -0, but then equality holds?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -702,6 +745,109 @@ public TypedFormula<?, TupleFormula> visitInsertExpression(InsertExpr insert) {
...
+        @Override
+        public TypedFormula<MemoryType, ?> visitToMemoryCastExpression(ToMemoryCast expr) {
+            requireBVEncoding(expr);
+            checkMemoryCastSupport(expr.getSourceType());
+
+            final TypedFormula<?, ?> inner = encode(expr.getOperand());
+            final MemoryType targetType = types.getMemoryTypeFor(expr.getSourceType());
+
+            final Formula enc;
+            if (inner.getType() instanceof IntegerType iType) {
```
hernanponcedeleon: `inner.getType()` is used twice ... for consistency with `visitFromMemoryCastExpression` store it into a variable

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -702,6 +745,109 @@ public TypedFormula<?, TupleFormula> visitInsertExpression(InsertExpr insert) {
...
+            if (targetType instanceof IntegerType bvType) {
+                final BitvectorFormulaManager bvmgr = bitvectorFormulaManager();
+                final int targetSize = bvType.getBitWidth();
+                if (targetSize < expr.getSourceType().getBitWidth()) {
+                    enc = bvmgr.extract((BitvectorFormula) inner.formula(), targetSize - 1, 0);
+                } else {
+                    enc = inner.formula();
+                }
+            } else if (targetType instanceof FloatType fType) {
+                enc = floatingPointFormulaManager().fromIeeeBitvector((BitvectorFormula) inner.formula(), getFloatFormulaType(fType));
```
hernanponcedeleon, line 799: should we have some assert about bit width?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -702,6 +745,109 @@ public TypedFormula<?, TupleFormula> visitInsertExpression(InsertExpr insert) {
...
+            requireBVEncoding(expr);
+
+            final List<? extends TypedFormula<MemoryType, ?>> operands = expr.getOperands().stream()
+                    .map(this::encodeMemoryExpr)
+                    .toList();
+            Formula enc = operands.get(0).formula();
+            final BitvectorFormulaManager bvmgr = bitvectorFormulaManager();
+            for (TypedFormula<MemoryType, ?> op : operands.subList(1, operands.size())) {
+                enc = bvmgr.concat((BitvectorFormula) op.formula(), (BitvectorFormula) enc);
+            }
```
hernanponcedeleon: Not sure it makes much difference, but maybe better to cast `enc` at initialization rather than at each iteration

dartagnan/src/main/java/com/dat3m/dartagnan/expression/floats/FloatBinaryOp.java
```diff
@@ -18,8 +18,8 @@ public String getSymbol() {
             case FMUL -> "*";
             case FDIV -> "/";
             case FREM -> "%";
-            case FMAX -> "fmax ";
-            case FMIN -> "fmin ";
+            case FMAX -> "fmax";
+            case FMIN -> "fmin";
```
hernanponcedeleon, line 22: IIRC the spaces were on purpose ... don't we get things like `2 max1` now?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
```diff
@@ -378,6 +398,82 @@ public Expression visitAggregateCmpExpression(AggregateCmpExpr expr) {
...
+
+        return super.visitMemoryEqualExpression(expr);
+    }
+
+    @Override
+    public Expression visitMemoryConcatExpression(MemoryConcat expr) {
+        final List<Expression> inner = expr.getOperands().stream().map(e -> e.accept(this)).toList();
+
+        // (int x to mem) :: (int y to mem) == (int x::y) to mem
+        // TODO: only true if the int types sizes are byte-multiples
```
hernanponcedeleon: Does this mean this simplification is unsound otherwise? Why dont we add a check that the `bitWidth % 8 == 0` ?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/MemoryType.java
```diff
@@ -0,0 +1,59 @@
...
+          Generally, metadata like provenance will not be retained, however, for metadata-less types like
+          integers, the concatenation may be reasonable.
+ */
+public class MemoryType implements Type {
+
+    private final int bitWidth;
+
+    MemoryType(int bitWidth) {
+        Preconditions.checkArgument(bitWidth > 0, "Size for memory type must be positive: %s", bitWidth);
+        Preconditions.checkArgument((bitWidth & 7) == 0, "Size must be a multiple of 8: %s", bitWidth);
```
hernanponcedeleon: Why %7 and not %8?

Related to my comment about of checking this modulo ... doesn't this guarantee that memory types are always at the byte granular and those TODO can be removed?

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -209,24 +208,24 @@ public Expression makeIntegerCast(Expression operand, IntegerType targetType, bo
     // -----------------------------------------------------------------------------------------------------------------
     // Floats
 
-    public FloatLiteral makeZero(FloatType type) {
-        return makeValue(BigDecimal.ZERO, type);
+    public FloatLiteral makePlusZero(FloatType type) {
+        return makeValue(BigDecimal.ZERO, false, type);
```
hernanponcedeleon, line 212: This confused me a lot ... see comment about in `FloatLiteral` class

dartagnan/src/main/java/com/dat3m/dartagnan/expression/floats/FloatLiteral.java
```diff
@@ -74,9 +77,9 @@ public String toString() {
             return "+INF";
         } else if (isMinusInf()) {
             return "-INF";
-        } else if (isNaN) {
+        } else if (isNaN()) {
             return "NaN";
         }
-        return String.format("%s(%s)", getType(), value);
+        return String.format("%s(%s%s)", getType(), sign ? "-" : "", absValue);
```
hernanponcedeleon: I found this rather confusing. My first impression was that "true = +" and I was confused when I saw `makePlusX` in the factory get `false` as parameter.

Rather than sign it might be more clear to call it `isNegative`

dartagnan/src/main/java/com/dat3m/dartagnan/expression/floats/FloatBinaryOp.java
ThomasHaas: No, we always added spaces before/after each binary operator. Before we got `2 max  1`.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/MemoryType.java
ThomasHaas: It's `& 7` not `% 7` (but I can change it to `% 8`).
The memory type is currently byte-granular (maybe we will change this at some point?).
However, int types are not byte-granular, so e.g. a `bv1` will do a round-trip like this `bv1 -> mem8 -> bv1`, which basically zero-extends the `bv1` and then truncates it again. Because of this implicit extension/truncation, some of the optimizations are not sound.
For example: `(bv1(1) :: bv1(1)) to mem != (bv1(1) to mem) :: (bv1(1) to mem)` since the former yields a `mem8` with the lowest two bits set, and the latter yields a `mem16` with the first and 9-th bit set.
It would be sound if `mem1` was a valid type though. But for now I'm inclined to assume that our memory is byte-granular.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/floats/FloatLiteral.java
ThomasHaas: Technically, the cleanest variant is to use an `enum Sign { POSITIVE, NEGATIVE; }` I think. But I can also change the boolean to a `isNegative`

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
ThomasHaas: That's why it was a todo (maybe I should have put FIXME). It boils down to the question whether we allow the mem-type to have sizes that are not multiples of 8. If so, the optimization is sound, if not, we need to restrict it (see answer to your question below).

benchmarks/miscellaneous/floats.c
ThomasHaas: The assertion by itself is quite pointless (although it ensures that the result is not NaN) but I put it as contrast to the coming assertions that swap the order of `d` and `d2`.
Also, the very next assertion checks sign-preservation, so those two together ensure that multiple calls to e.g. `fmax(0, -0)` consistenly return either 0 or -0.

benchmarks/miscellaneous/floats.c
ThomasHaas: I changed it, but I did not bother with regenerating all float tests, meaning there is now a slight mismatch between the llvm and the C code.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
ThomasHaas: The optimization is now sound and written so that it works with bit-granular and byte-granular memory types

dartagnan/src/main/java/com/dat3m/dartagnan/expression/floats/FloatLiteral.java
ThomasHaas: Changed to `isNegative`

dartagnan/src/main/java/com/dat3m/dartagnan/expression/type/MemoryType.java
ThomasHaas: Changed to `% 8`. 
Also fixed the optimizations to be as general as possible and always sound.

hernanponcedeleon:
After the changes of this PR, can we fix [this](https://github.com/hernanponcedeleon/Dat3M/blob/30640123b88d4fc5ca01403f936d45c54d83a9c1/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java#L1803)?

ThomasHaas:
> After the changes of this PR, can we fix [this](https://github.com/hernanponcedeleon/Dat3M/blob/30640123b88d4fc5ca01403f936d45c54d83a9c1/dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java#L1803)?

I think the issue is unrelated to this PR but is instead related to verification with `--mixedSize=false`. With MSA detection, the check can just be removed. Without MSA, the memset is only correctly modelled for `0` and `-1` and there is no easy way around it.

hernanponcedeleon:
The intrinsics class knows if MSA is enabled, so we can at least remove the exception in that case

ThomasHaas:
I would do it in a separate PR though as there are plenty things that can be improved, for example, generating proper loops for memset/memcpy so that we support dynamic size arguments.

PR #963, opened 2026-01-25
Large refactor of ModelChecker class and its users

ThomasHaas:
- ModelCheckers now create their own SolverContext/ProverEnvironment
- Creation of ModelCheckers is more streamlined.
- Dartagnan interacts with ModelCheckers only via an IREvaluator but no direct inspection of the SMT model.

### DIFF
--- a/dartagnan/pom.xml
+++ b/dartagnan/pom.xml
@@ -276,6 +276,7 @@
                             <className>com.dat3m.dartagnan.program.processing.ThreadCreation</className>
                             <className>com.dat3m.dartagnan.program.processing.compilation.Compilation</className>
                             <className>com.dat3m.dartagnan.utils.options.BaseOptions</className>
+                            <className>com.dat3m.dartagnan.verification.solving.ModelChecker$SMTConfig</className>
                             <className>com.dat3m.dartagnan.wmm.Wmm$Config</className>
                             <className>com.dat3m.dartagnan.wmm.analysis.RelationAnalysis$Config</className>
                             <className>com.dat3m.dartagnan.wmm.analysis.WmmAnalysis</className>
--- a/dartagnan/src/main/resources/META-INF/native-image/com.dat3m.dartagnan/dartagnan/reflect-config.json
+++ b/dartagnan/src/main/resources/META-INF/native-image/com.dat3m.dartagnan/dartagnan/reflect-config.json
@@ -115,6 +115,11 @@
         "queryAllDeclaredMethods":true
       },
       {
+      "name":"com.dat3m.dartagnan.verification.solving.ModelChecker$SMTConfig",
+      "allDeclaredFields":true,
+      "queryAllDeclaredMethods":true
+      },
+      {
         "name":"com.dat3m.dartagnan.witness.graphml.WitnessBuilder",
         "allDeclaredFields":true,
         "queryAllDeclaredMethods":true


dartagnan/src/main/java/com/dat3m/dartagnan/smt/SMTHelper.java
```diff
@@ -0,0 +1,39 @@
...
+    private SMTHelper() { }
+
+    private static void loadLibrary(String libName) {
+        // Try using NativeLibraries::loadLibrary. Fallback to System::loadLibrary
+        // if NativeLibraries failed, for example, because the operating system is
+        // not supported,
+        try {
+            NativeLibraries.loadLibrary(libName);
+        } catch (Exception e) {
+            // TODO: Is this really needed?
```
hernanponcedeleon: This does not work on weird OSs ... see #835

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -1,52 +1,127 @@
...
+    protected ModelChecker(VerificationTask task) throws InvalidConfigurationException {
+        this.task = Preconditions.checkNotNull(task);
+        this.smtConfig = new SMTConfig();
+        task.getConfig().inject(smtConfig);
+    }
+
+    public static ModelChecker create(VerificationTask task, Method method) throws InvalidConfigurationException {
+        if (task.getProperty().contains(DATARACEFREEDOM)) {
+            // For DATARACEFREEDOM, we always use EAGER
+            // Shall we just throw an exception here?
```
hernanponcedeleon: I would rather just log a warning

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -57,6 +132,83 @@ public boolean hasModel() {
...
     }
 
+    public IREvaluator getModel() throws SolverException {
+        Preconditions.checkState(hasModel(), "No model available");
+        return context.newEvaluator(prover);
+    }
+
+    public ExecutionModelNext getExecutionGraph() throws SolverException {
+        Preconditions.checkState(hasModel(), "No model available");
+        try (IREvaluator evaluator = getModel()) {
```
hernanponcedeleon, line 132: Why do we need the `try` if there is a model? Is it so it auto-closes?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -151,116 +119,65 @@ public static void main(String[] args) throws Exception {
...
-                        .withConfig(config)
-                        .withProgressModel(o.getProgressModel())
-                        .withWitness(witness);
-                Program p = new ProgramParser().parse(f);
+                final Program p = new ProgramParser().parse(f);
                 if (o.overrideEntryFunction()) {
                     p.setEntrypoint(new Entrypoint.Simple(p.getFunctionByName(o.getEntryFunction()).orElseThrow(
                         () -> new MalformedProgramException(String.format("Program has no function named %s. Select a different entry point.", o.getEntryFunction())))));
                 }
+                // FIXME? Should we reparse the memory model each iteration?
```
hernanponcedeleon: The modifications are not related to the program (i.e., it is some kind of memory model normalization) so this should not be a problem, right?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -151,116 +119,65 @@ public static void main(String[] args) throws Exception {
...
+                }
+                // We only generate SVCOMP witnesses if we are not validating one.
+                if (o.getWitnessType().equals(GRAPHML) && !o.runValidator()) {
+                    generateWitnessIfAble(task, modelChecker, summary.reason() + "\n" + summary.details());
                 }
             } catch (InterruptedException e) {
-                final long time = 1000L * o.getTimeout();
-                summary = new ResultSummary(f.toString(), "", TIMEDOUT, "", "", "", time, TIMEOUT_ELAPSED);
+                final String details = "\t" + Optional.ofNullable(e.getMessage()).orElse("Unknown error occurred");
+                summary = new ResultSummary(f.toString(), "", TIMEDOUT, "", "", details, 0, TIMEOUT_ELAPSED);
```
hernanponcedeleon: Why are you passing the whole exception msg rather than the time here?

dartagnan/src/test/java/com/dat3m/dartagnan/llvm/LibvsyncTest.java
```diff
@@ -43,8 +43,9 @@ protected Provider<Integer> getBoundProvider() {
     }
 
     @Override
-    protected Configuration getConfiguration(){
-        return Configuration.defaultConfiguration();
+    protected ConfigurationBuilder additionalConfig(ConfigurationBuilder builder) {
+        return builder
+                .setOption(OptionNames.USE_INTEGERS, "false");
```
hernanponcedeleon: I would make it a one-liner

dartagnan/src/main/java/com/dat3m/dartagnan/smt/SMTHelper.java
ThomasHaas: I understand, but I found it somewhat surprising, cause the source code of `NativeLibraries.loadLibrary` contains the same call that we do, but just in an `else`-case rather than on exception. Maybe we should ask JavaSMT to make the else case a proper fallback :)

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
ThomasHaas: Yes, it's for auto-closing. We could avoid this, if we assumed that the ModelChecker generates a single cached model that it always returns. For now, the calls to `getModel` create a new SMT model.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: Well, it works right now. But it is a bit dangerous and reparsing the memory model is probably not much overhead.
For example, in the unit tests we do reparse the whole model over and over again, even for the thousands of litmus tests.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: Because it was a bit problematic to track the timeout here since the timeout option was moved into the ModelChecker class. This means that Dartagnan has to create the ModelChecker (within the try-block), query the timeout and store it in a variable that is declared outside of the try-block so that it can be accessed in the catch-block (quite ugly!).
However, the error message produced by the ModelChecker already contains the exact timeout, so it was nicer to just reuse it. Notice that the message is produced by us (see ModelChecker.run): it is not some default error message.

Generally speaking, I think it is better to keep the error message rather than overwriting it arbitrarily anyways.
In fact, it is technically possible to interrupt the ModelChecker due to other reasons besides timeouts, which would be reflected now.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: I added reparsing now.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -310,7 +320,6 @@ private void runInternal(SolverContext ctx, ProverWithTracker prover, Verificati
             }
         } else {
             res = FAIL;
-            saveFlaggedPairsOutput(baselineModel, prover, context, task.getProgram());
         }
```
xeren, line 323: Don't we need this anymore?

ThomasHaas: It just got moved to `Dartagnan.java`, where all the code for output generation is at the moment

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/ArrayValidTest.java
```diff
@@ -54,18 +51,16 @@ public ArrayValidTest(String path, Wmm wmm) {
...
-            AssumeSolver s = AssumeSolver.run(ctx, prover, task);
-            assertEquals(PASS, s.getResult());
-        } catch (Exception e) {
-            fail("Missing resource file");
+    public void test() throws Exception {
+        Program program = new ProgramParser().parse(new File(path));
+        VerificationTask task = VerificationTask.builder()
+                .withSolverTimeout(60)
+                .withTarget(Arch.LKMM)
+                .build(program, wmm, EnumSet.of(Property.PROGRAM_SPEC));
```
xeren, line 59: This has been `Property.getDefault()`.  I assume there is neither a `CAT_SPEC`, nor a `TERMINATION` issue to check in these benchmarks?

ThomasHaas: I think the property doesn't matter for these tests.
That being said, it's probably a good idea to have tests specify exactly what they are checking rather than relying on `getDefault`. That's why I changed it

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
hernanponcedeleon: I just checked the current error msg and I am fine with it.

The problem with other types of interruption is that then the exit code (`TIMEDOUT`) would be wrong. Even if JavaSMT allows other kinds of interrupts, do we really use any of them or from dat3m side, this is the only

dartagnan/src/test/java/com/dat3m/dartagnan/litmus/comparison/AbstractComparisonTest.java
```diff
@@ -75,7 +75,6 @@ protected Provider<EnumSet<Property>> getPropertyProvider() {
     protected final Configuration getConfiguration() throws InvalidConfigurationException {
         var configBase = Configuration.builder()
                 .setOption(SOLVER, Z3.name())
-                .setOption(TARGET, targetProvider.get().name())
```
hernanponcedeleon: Why did you remove this one?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: I agree with the exit code, but that can be fixed. Indeed, if you run Dartagnan via command line and interrupt it with ctrl+c, you get the TIMEDOUT result with details "canceled" in this PR.
On development, you just get ` TIMEDOUT` and if you also passed `--timeout=X` it will report `Time: X` which is clearly wrong.

Btw. our unit tests also have a custom timeout rule that stops the solver, but that one does not run the `Dartagnan.main` code so it doesn't really matter.

dartagnan/src/test/java/com/dat3m/dartagnan/litmus/comparison/AbstractComparisonTest.java
ThomasHaas: Because the comparisons tests set two different targets so it cannot be done over the config (unless we use two configs) but has to be done on a per-task basis. The setting per task overwrites whatever is in the config, so setting it there doesn't matter.

PR #964, opened 2026-01-28
Add compilation to native

hernanponcedeleon:
This PR adds compilation to native. Overall, the whole verification time is generally speed-up. I have seen between 10x and 100x improvements for litmus tests, and more modest (but still good) improvements for large benchmarks (30% reduction in a full svcomp run).


hernanponcedeleon:
> If I execute `maven` without `-Pnative`, does it install only the original JVM version?

Yes

ThomasHaas:
Why did you change the logging library to SLF4j and Logback? Is this a compatibility thing with GraalVM?

Issue #955, opened 2025-12-23
Different verification result depending on encoding.integers


tcherrou:
For the following C program: 

```
#include <assert.h>

unsigned char a = 128;

int main() {
    assert(a==128);
    return 0;
}

```
Running `dartagnan` with different encoding options gives a different verification results.
More concretely: 

- Bit vectors 

```
$ ./scripts/dartagnan -cat cat/sc.cat program.c --encoding.integers=false
Test: program.c
Result: PASS
Time: 0.516 secs
```
And with using the option `--enable-integers=true`: 

- Integers

```
$ ./scripts/dartagnan -cat cat/sc.cat program.c --encoding.integers=true 
Test: program.c
Result: FAIL
Reason: Program specification violation found
Details:
	E12:	@program.c#6: user assertion
Time: 0.514 secs
```

This is counter-intuitive to me as I expect the answer to be the same.
I am not sure if this is intended or it is an issue on my side (or i missed some option).
I am using the latest commit from the master branch (eb981729200b7239373d1be4287e9d8db16d9047) 
Clang version: `18.1.8`
My guess it has something to do with how integers are represented in LLVMIR.
Thought this was worth sharing, although I am not sure if this is common enough to worry about.

ThomasHaas:
#956 should fix at least this example.

Issue #941, opened 2025-10-01
[Benchmark] Local data-flow benchmark


ThomasHaas:
I just want to share a simple benchmark that benchmarks our handling of local data flow.

```C
#include <assert.h>
#include <dat3m.h>

#define N 100

int main()
{
    int r1 = 0;
    for (int i = 0; i < N+1; i++) {
        if (__VERIFIER_nondet_bool()) { break; }
        r1 += i;
    }

    // Intermediate variable "phi = r1" should act like a phi-node. 
    // We need to make the assignment more complex/convoluted so that Dartagnan does not propagate it away.
    // int phi = (r1 + r1) / 2;

    int r2 = 0;
    for (int i = 0; i < N; i++) {
        if (__VERIFIER_nondet_bool()) { break; }
        r2 += r1;
        // r2 += phi;
    }

    assert(r2 <= N*(N*(N+1))/2);
}
```

Verifying this benchmark takes 25-40 seconds on my machine. With the intermediary phi assignment, it takes < 10 seconds.
I think this benchmark (or variants thereof) can also be used to test different data-flow encodings.
Maybe someone is interested in making Dartagnan faster using this benchmark :)
For example, we could automatically insert the phi node as mentioned here #576, or somehow make the data-flow encoding smarter to better reuse values.

hernanponcedeleon:
> If we reduce the size to `N=10`, the eager encoding solves it almost instantly (< 0.5s) but the lazy solver needs around 20 seconds.

I cannot reproduce this in the current development branch
```
> cat test.c

#include <assert.h>
#include <dat3m.h>

#define N 10
volatile int r1 = 0;

int main()
{
    
    for (int i = 0; i < N+1; i++) {
        if (__VERIFIER_nondet_bool()) { break; }
        r1 += i;
    }

    // Intermediate variable "phi = r1" should act like a phi-node. 
    // We need to make the assignment more complex/convoluted so that Dartagnan does not propagate it away.
    // int phi = (r1 + r1) / 2;

    int r2 = 0;
    for (int i = 0; i < N; i++) {
        if (__VERIFIER_nondet_bool()) { break; }
        r2 += r1;
        // r2 += phi;
    }

    assert(r2 <= N*(N*(N+1))/2);
}

> java -jar dartagnan/target/dartagnan.jar cat/vmm.cat test.c --method=eager

Test: test.c
Result: PASS
Time: 0.925 secs

[...]

PR #965, opened 2026-01-30
Interval Analysis

tcherrou:
# Interval Analysis
## Overview
This pull request adds interval analysis (also known as range analysis or value analysis) to `dartagnan`'s analysis pipeline after preprocessing.
The analysis determines the intervals of integer register in the intermediate representation.
These intervals can be used for program simplification. 
Here, it is used to add bounds to variables in the smt encoding to speed up verification.
This work was done for my master's thesis and concluded that using interval analysis.
[master_thesis.pdf](https://github.com/user-attachments/files/24967903/master_thesis.pdf)
The experimental results are a bit outdated as the most recent version is more conservative during the encoding phase (mostly with the bit vector encoding).

### Features 
- An interval analysis for integer registers that derives bounds.
- Use of interval analysis to encode bounds on SMT variables to limit the search space of the solver.
- Tests for interval analysis and interval objects.
- Encoding support for bit vector and integer encoding.

#### Representation
An interval for a register consists of a lower bound and upper bound such that r is within that range of values.
Since registers have a certain bitwidth, the length of the interval is bounded by the number of values representable with that bitwidth.
For example, for an 8-bit registers, the widest possible interval is `[-128,255]`.

#### Signedness
To my knowledge, registers are neither signed nor unsigned and signedness by determined by operations.
This ambiguity is dealt with by not making a choice on signedness.
A drawback with this approach is that some interval are sensitive the sign while some are not.
For example, the interval `[-1,1]` is sound when interpreted as signed but unsound when interpreted as unsigned.
These type of intervals can occur when

In contrast, the interval `[-5,-1]` has the same set of values unsigned as signed in terms of their bit vector representation.

#### No wrapping
The "top" value of the interval lattice is `[MIN_SIGNED,MAX_UNSIGNED]`.
Additionally, interval are not wrapped. 
This means that overflows and underflows are conservatively dealt with by widening to the top value for that specific bitwidth.

#### Note
As in #955 llvm represents values above the signed upper bound (e.g. 128 for an 8-bit integer) as their signed versions. 
This can lead to unsound intervals as values are interpreted as they are represented in the IR. 
Suggestions to remedy this are always welcome.

### Analysis
The pull request introduces an interface for querying the interval of an integer register at a specific program event.
There are three analyses, which differ in their precision.
1. Naive Analysis just returns top for each register.
2. Local Analysis analyses each thread in isolation by looking at `Local` events and propagating information using a work list algorithm.
3. Global Analysis extends local analysis by leveraging `dartagnan`'s relation analysis and employing `read-from` pairs to analyse `Load` instructions for additional precision.

#### Note
Since Global analysis depends on Relation analysis, it must be performed after Relation analysis.

### Encoding
The interval of a register is used in the encoding phase by using the bounds to constrain the SMT variable associated. 

For example, a register `r0` with interval `[0,5]` will have the SMT constraints (assuming the integer encoding) `(assert (0 <= r0))` and `(assert (r0 <= 5))`.
#### Notes 
To the extent of my knowledge, knowledge of the sign bit is necessary when encoding bounds on bit vectors.
As registers have inherently no sign, only sign insensitive interval are encoded.
Maybe it is possible to add some sort of sign analysis here but I am unsure.

### Usage
By default, naive analysis is enabled. One can choose a specific by using the option: `program.analysis.interval`.
The following choices are available:  
- `naive` 
- `local`
- `global` 
### Tests
- Analysis correctness tests
- Unit tests for interval operations and initialisation.

Existing tests pass on my machine besides the ones that ran out of memory.
The `RelationAnalysisTest` file has been updated to include the updated analysis pipeline.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
@@ -115,9 +115,9 @@ public interface IntervalAnalysis {
     // Only executed if logger.isInfoEnabled = true
     // Rounds half up for decimal values.
     private void computeAnalysisMetrics(Program program) {
-        double totalRegWrites = 0;
-        double totalIntervalsReduced = 0;
-        double totalIntervalsTop = 0;
+        int totalRegWrites = 0;
+        int totalIntervalsReduced = 0;
+        int totalIntervalsTop = 0;
         BigDecimal totalReducedIntervalSize = BigDecimal.ZERO;
 
         for (Event e : program.getThreadEvents()) {
@@ -157,9 +157,9 @@ public interface IntervalAnalysis {
 
         logger.info("""
             \n======== IntervalAnalysis Summary ========
-            \t#Register writes: {}
-            \t#Register bounded {}
-            \t#Registers top {}
+            \t#Register Writes: {}
+            \t#Register-Reads bounded: {}
+            \t#Register-Reads top: {}
             \t%Bounded: {}
             \t%Top: {}
             \tAverage reduced interval size: {}


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -740,5 +743,44 @@ private BooleanFormula encodeProgressForwarding(ThreadHierarchy group, ProgressM
             return bmgr.implication(hasForwardProgress(group), bmgr.and(enc));
         }
     }
+
+    // ============= Bounds ============= 
+
+    public void encodeBounds(ProverWithTracker prover) throws InterruptedException {
+        FormulaManagerExt formulaManager = context.getFormulaManager();
+        for(var entry : context.getVarToInterval().entrySet()) {
```
ThomasHaas: Rather than passing a prover and asserting each bound separately, it is better to generate a `BooleanFormula` that contains the conjunction of all bounds and return that formula as result (just like how the other functions do it).
The caller can assert it to the prover.
This can heavily improve the performance in some edge-cases because `prover.assert()` is an expensive operation. For example,  the benchmark from #941 is fully solved by the interval analysis but the verification still takes around 6 seconds on my machine, 4 of which are just spent asserting the bounds to the prover.
By generating the `BooleanFormula` and asserting only once, the 4 second overhead vanishes fully.

ThomasHaas:
>     1. Could you please elaborate on your definition of "small"?

Small is anything that does not get too close to the bit vector sizes, but I think your analysis reaches TOP in most of those cases anyways. From what I have seen on some local tests is that all your intervals are basically small or TOP.

>     2. Only considering strictly positive or negative intervals is an interesting idea and something I have not tried yet.
>        I will try it out actually and see what the effects are. Which type of intervals yield the best performance is still an open question to me unfortunately. The software verification benchmark suite that seemed to benefit the most of  interval analysis was  `pthread-wmm` and all found an average interval size of around less than 3.

After testing a little bit locally, I noticed that your approach already avoids encoding intervals crossing a sign-boundary, at least if doing BV encodings. So in that sense, you get that result already. However, what could be interesting is to only encode the signedness (>0. <0, =0) as an extreme lightweight encoding. IIRC, this information was useful on the fib_bench benchmarks (the unsat versions), especially when doing the lazy/caat method (and integer encoding?).

>     3. I briefly encoded TOP intervals at the start but removed it later. The reason was that it did not seem to make sense in the bit vector encoding I think. As a result for ease of code I also do not add them for the integer encoding.

For BVs, they don't make sense, I agree. I think they should also be avoided on integers, cause IIRC the bounds just degrade the performance of the solver heavily. But maybe you can try once to see for yourself :)

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
tcherrou: Thanks for the suggestion, I found an improvement on my machine as well. Changed in baf210af3e9854d654e790c4042e39dc038b30c8

tcherrou:
There were duplicate commits in the commit history after rebasing. Removed them with a force push

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
```diff
@@ -0,0 +1,10 @@
+package com.dat3m.dartagnan.configuration;
+
+public enum IntervalAnalysisOptions implements OptionInterface{
+    NAIVE,LOCAL,GLOBAL;
```
hernanponcedeleon: Use space between enums

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -82,11 +84,13 @@ public final class EncodingContext {
     private final Map<Event, TypedFormula<?, ?>> addresses = new HashMap<>();
     private final Map<MemoryObject, TypedFormula<?, ?>> objAddress = new HashMap<>();
     private final Map<MemoryObject, TypedFormula<IntegerType, ?>> objSize = new HashMap<>();
+    private final Map<TypedFormula<IntegerType,?>, Interval> varToInterval = new HashMap<>();
```
hernanponcedeleon: space after `,`.

There are many places where you miss the spaces. Please add them everywhere

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -156,6 +158,19 @@ public BooleanFormula equalAt(Expression left, Event leftAt, Expression right, E
     // ====================================================================================
     // Private implementation
 
+    @SuppressWarnings("unchecked")
+    private void saveIntervalOfVariable(Event event, Register register, TypedFormula<?,?> variable, Type type) {
+        IntervalAnalysis intervalAnalysis = context.getAnalysisContext().get(IntervalAnalysis.class);
+        if (event != null && intervalAnalysis != null && type instanceof IntegerType) {
+            Interval interval = intervalAnalysis.getIntervalAt(event,register);
+            if(interval != null && !interval.isTop()) {
```
hernanponcedeleon: Are all these null really necessary?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -156,6 +158,19 @@ public BooleanFormula equalAt(Expression left, Event leftAt, Expression right, E
     // ====================================================================================
     // Private implementation
 
+    @SuppressWarnings("unchecked")
+    private void saveIntervalOfVariable(Event event, Register register, TypedFormula<?,?> variable, Type type) {
+        IntervalAnalysis intervalAnalysis = context.getAnalysisContext().get(IntervalAnalysis.class);
+        if (event != null && intervalAnalysis != null && type instanceof IntegerType) {
+            Interval interval = intervalAnalysis.getIntervalAt(event,register);
```
hernanponcedeleon: Based on the name of the method, it would make more sense to have the register as the first argument and the event as the second

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -156,6 +158,19 @@ public BooleanFormula equalAt(Expression left, Event leftAt, Expression right, E
...
+            if(interval != null && !interval.isTop()) {
+                assert variable.getType() instanceof IntegerType;
+                assert variable.formula() instanceof BitvectorFormula || variable.formula() instanceof IntegerFormula;
```
hernanponcedeleon: Doesn't this assert trivially hold from the fact that type is `IntegerType`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -156,6 +158,19 @@ public BooleanFormula equalAt(Expression left, Event leftAt, Expression right, E
...
+                context.putVarToInterval((TypedFormula<IntegerType, ?>) variable,interval);
```
hernanponcedeleon: Why not making the input parameter of `saveIntervalOfVariable` to be directly `TypedFormula<IntegerType, ?>`? Do we have any case where you would pass another kind of variable?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -577,7 +592,11 @@ public TypedFormula<?, TupleFormula> visitInsertExpression(InsertExpr insert) {
             final String name = event == null ?
                     reg.getName() + "_" + reg.getFunction().getId() + "_final" :
                     reg.getName() + "(" + event.getGlobalId() + ")";
-            return makeVariable(name, reg.getType());
+            Type type = reg.getType();
+            TypedFormula<?,?> variable = makeVariable(name, type);
+            // Side effect!
+            saveIntervalOfVariable(event,reg,variable,type);
```
hernanponcedeleon: Do we really need to do this here?
Can't we retrieve the SMT variable from the Register via calling the visitor in `ProgramEncoder.encodeBounds()`?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -743,5 +746,47 @@ private BooleanFormula encodeProgressForwarding(ThreadHierarchy group, ProgressM
...
+            } else if (formula instanceof IntegerFormula variable) {
+                IntegerFormulaManager integerFormulaManager = formulaManager.getIntegerFormulaManager();
+                BigInteger lowerbound = interval.getLowerbound();
+                BigInteger upperbound = interval.getUpperbound();
+                BooleanFormula constraintGTE = integerFormulaManager.greaterOrEquals(variable, integerFormulaManager.makeNumber(lowerbound));
+                BooleanFormula constraintLTE = integerFormulaManager.lessOrEquals(variable, integerFormulaManager.makeNumber(upperbound));
+                encoding.add(constraintLTE);
+                encoding.add(constraintGTE);
+            }
+        }
```
hernanponcedeleon: Can we merge a big part of this by using `TypedFormula`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/EventNotFoundException.java
```diff
@@ -0,0 +1,9 @@
+package com.dat3m.dartagnan.program.analysis.interval;
+
+import com.dat3m.dartagnan.program.event.Event;
+
+public class EventNotFoundException extends RuntimeException {
```
hernanponcedeleon: Under which situations we might call `getIntervalAt` and do not find the event?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+
+final public class Interval {
+
+    private final BigInteger lowerbound;
+    private final BigInteger upperbound;
+    private final IntegerType type;
+
+    static Set<Object> unsupportedOperators = new HashSet<>();
+    @SuppressWarnings("unused")
+    Logger logger = LoggerFactory.getLogger(Interval.class);
```
hernanponcedeleon: If the logger is not used, simply remove

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+// E.g. a register bv8 r0 has a lower bound of -128 (signed) and 255 (unsigned)
+// We do not assume anything about the signedness of a register associated with an interval.
+
+final public class Interval {
+
+    private final BigInteger lowerbound;
+    private final BigInteger upperbound;
+    private final IntegerType type;
+
+    static Set<Object> unsupportedOperators = new HashSet<>();
```
hernanponcedeleon: Do we currently have unsupported operators? Are there operators for which we would definitely not be able to add support? If not, I would rather make java complain if there is any operator that it is not implemented and thus force us to do it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+    public IntegerType getType() {
+        return type;
+    }
+
+    public static Set<Object> getUnsupportedOperatorsFound() {
+        return unsupportedOperators;
+    }
+
+    public static Interval getTop(IntegerType type) {
+        return new Interval(type.getMinimumValue(true),type.getMaximumValue(false),type);
```
hernanponcedeleon: does it make sense to use unsigned max for top?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    public BigInteger size() {
+        return (this.upperbound.subtract(this.lowerbound)).add(BigInteger.ONE);
+    }
+
+
+    public Interval join(Interval interval2) {
```
hernanponcedeleon:
```suggestion
    public Interval join(Interval other) {
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+        else if(op instanceof IntBinaryOp binop && intervals.length == 1 && !intervals[0].isTop()) {
+            opFunc = selectCurriedBinaryOperatorMethod(binop,intervals[0]);
+        } else if (intervals.length > 2) {
+            throw new IllegalArgumentException("More than 2 intervals specified");
+        }
+
+        if(opFunc != null && !this.isTop()) {
+            return opFunc.get();
+        } else return Interval.getTop(type);
+    }
```
hernanponcedeleon: Rather than having a single generic method and a bunch of checks to be sure there is only 1 or 2 intervals, why not having two specific methods and avoid all the `if ... then` checking?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+        return switch (op) {
+            case MINUS -> this::negate;
+            default -> {
+                unsupportedOperators.add(op);
+                yield null;
+            }
+        };
+    }
+
+    private Interval subtract(Interval i2) {
```
hernanponcedeleon:
```suggestion
    private Interval subtract(Interval other) {
```

The same applies for all other similar methods

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,387 @@
...
+        BigDecimal mul1 =  lb1.multiply(lb2);
+        BigDecimal mul2 =  lb1.multiply(ub2);
+        BigDecimal mul3 =  ub1.multiply(lb2);
+        BigDecimal mul4 =  ub1.multiply(ub2);
+        BigDecimal resultBoundLowerBounds = mul1.min(mul2).min(mul3).min(mul4);
+        BigDecimal resultBoundUpperBounds = mul1.max(mul2).max(mul3).max(mul4);
+        BigInteger resultBoundLowerBoundsInt = resultBoundLowerBounds.setScale(0,RoundingMode.FLOOR).toBigInteger();
+        BigInteger resultBoundUpperBoundsInt = resultBoundUpperBounds.setScale(0,RoundingMode.CEILING).toBigInteger();
+        return new Interval(resultBoundLowerBoundsInt,resultBoundUpperBoundsInt,this.getType());
+    }
```
hernanponcedeleon: I don't think having two methods really make it more readable. You can easily merge these

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: I'm not sure about using `TypedFormula` since there is a distinction between which encoding is actually used, i.e., it does care about the inner formula (though some merging is still possible). That being said, the function can be reduced to 1/3 of its code even with the distinction in place:
- inline the `upperbound/lowerbound` variables (they are used only once after definition)
- `lowerboundConstraint/upperboundConstraint` are dead declarations. Similarly `constraintGTE/constraintLTE` can be inlined into their single usage.
- Rename the long formular manager names to `bvmgr` and `imgr` just like we usually do.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
ThomasHaas: What is even the use of `BigDecimal`? Is it for a hypothetical support of floating point numbers? Cause right now, you cannot possibly get fraction values anywhere.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
tcherrou: Judging from the other code in the visitor, I think event can be null. 
But the other null check are indeed not necessary in theory.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
tcherrou: Am I understanding that you mean we iterate over the events of the program in `ProgramEncoder.encodeBounds()` and using the expression visitor to get the register name?
   If that is the case would the expression encoder not just make a new variable (I am not sure if the smt interface re uses variables)?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: For binary operators, I have not thought about how to do the remainder operations yet.
   `XOR` can be done with the morgan laws I think.
   I think shifting is trivial but I could not find many sources on it.
   For the unary operators, not quite sure how to add `CTTZ`, `CTLZ` and `CTPOP`.
So yes, there are still unsupported operators.
I am not opposed to the idea of just letting Java complain, this was mostly to help me find operator.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: I do not know the sign of the register and this covers for possible integer values for unsigned numbers.
   It also helps with cast expressions.
   Please let me know your main concerns for where this potentially can go wrong.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: It is not for readability, the second multiply is also used in division.
   BigDecimals are used because in division, I need to calculate the reciprocal. If I use BigIntegers there it would (almost) always be zero. 
   Then again, I think maybe a fix is to entirely skip the reciprocal calculation step and just calculate the endpoints directly. It will lead to slightly smaller intervals because currently I take the ceiling for the upper bound but I think (correct me if I am wrong) integer division is always truncated.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: SMT Variables are uniqued, i.e., there are no two distinct variables with the same name.
So `ExpressionEncoder.encodeAt(<register>, <event>).formula()` should give you the correct formula that you want to bound (assuming you restrict the values read from registers, not the values written into them).

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
ThomasHaas: Integer division is always truncating.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
tcherrou: Modifications in 64e0cf88b7c31525d22756f0a320371e688acf50. Logic related to encoding bounds for registers is now contained in `ProgramEncoder`. Thanks for the helpful information.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: See 48b4811918fdc858f82f58b80057b0ec4875cfbc.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/EventNotFoundException.java
tcherrou: It can happen from improper use of the method but perhaps I am just being overly cautious.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
ThomasHaas: `CTTZ` and co. are easy to support: they can only return values between 0 and the number of bits in the integer type. So `cttz bv32 x` will always return a value in '[0, 32]'.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -743,5 +747,43 @@ private BooleanFormula encodeProgressForwarding(ThreadHierarchy group, ProgressM
...
+        return encoding;
+    }
+
+    public BooleanFormula encodeBounds() {
+        List<BooleanFormula> encoding = new ArrayList<>();
+        for (RegReader e : context.getTask().getProgram().getThreadEvents(RegReader.class)) {
+            for (Register.Read read : e.getRegisterReads()) {
+                Register r = read.register();
+                if (r.getType() instanceof IntegerType) {
+                    Interval interval = context.getAnalysisContext().requires(IntervalAnalysis.class).getIntervalAt(e, r);
```
hernanponcedeleon: get the IntervalAnalysis only once outside the loop

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -743,5 +747,43 @@ private BooleanFormula encodeProgressForwarding(ThreadHierarchy group, ProgressM
...
+                    encoding.addAll(encodeRegisterBounds(context.getExpressionEncoder().encodeAt(r, e).formula(), interval));
```
hernanponcedeleon: why `addAll`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,393 @@
...
+
+    public static Set<Object> getUnsupportedOperatorsFound() {
+        return unsupportedOperators;
+    }
+
+    public static Interval getTop(IntegerType type) {
+        return new Interval(type.getMinimumValue(true), type.getMaximumValue(false), type);
+    }
+
+    public BigInteger size() {
```
hernanponcedeleon, line 72: Do you really need this to be a `BigInteger` rather than e.g., long? Do we need to handle e.g., 128 bits vars?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,393 @@
...
+        return (this.upperbound.subtract(this.lowerbound)).add(BigInteger.ONE);
+    }
+
+
+    public Interval join(Interval other) {
+        return new Interval(this.lowerbound.min(other.lowerbound), this.upperbound.max(other.upperbound), this.getType());    }
```
hernanponcedeleon: Put `}` in a new line

hernanponcedeleon (commented):
Minor comments after a quick check.

@tcherrou please fix the conflicts

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
tcherrou: `encodeRegisterBounds` returns a list of  the lower/upper bound constraints. Add all just adds them as I don't think I want a nested list.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -103,6 +104,7 @@ private EncodingContext(VerificationTask t, Context a, FormulaManager m, Collect
         verificationTask = checkNotNull(t);
         analysisContext = checkNotNull(a);
         a.requires(BranchEquivalence.class);
+        a.requires(IntervalAnalysis.class);
```
ThomasHaas: This `requires` here is unnecessary

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: This was done for generality, but if it is not necessary I can change it. I cannot immediately think of architectures  that support such registers.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: See c260524d683526c6e3e65dad9fac3f02b14c7683

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
ThomasHaas: `BigInteger` is correct; our type system easily supports integers larger than 64 bits.
Also, as long as the bounds are stored as `BigInteger`, the size must also be `BigInteger`.
In fact, the size of your `TOP` interval exceeds `long` even for 64 bit integers.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: See c3a0dc5c68c618b92f7eb9cf7c7d08fa3873c93f.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
hernanponcedeleon: For consistency with the rest of the encoding methods, I would make `encodeRegisterBounds` to return directly `BooleanFormula`

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,384 @@
...
+import java.math.BigInteger;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.UnaryOperator;
+import java.util.function.Supplier;
+
+
+
+//Class to represent closed integer intervals for registers.
```
hernanponcedeleon: Add space

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
```diff
@@ -0,0 +1,384 @@
...
+// E.g. a register bv8 r0 has a lower bound of -128 (signed) and 255 (unsigned)
+// We do not assume anything about the signedness of a register associated with an interval.
+
+final public class Interval {
+
+    private final BigInteger lowerbound;
+    private final BigInteger upperbound;
+    private final IntegerType type;
+
+    static Set<Object> unsupportedOperators = new HashSet<>();
```
hernanponcedeleon: Is there a type more concrete than Object we can use?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
```diff
@@ -0,0 +1,185 @@
...
+            analysis.computeAnalysisMetrics(program);
+        }
+        return analysis;
+    }
+
+    @Options
+    class Config {
+        @Option(
+            name = INTERVAL_ANALYSIS_METHOD,
+            description = "Indicates the desired precision of the interval analysis"
```
hernanponcedeleon: I would something similar to the description of `AliasAnalysis`, e.g., "General type of analysis that bounds integer registers."

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
```diff
@@ -0,0 +1,185 @@
...
+
+            ============== Interval Analysis Summary ====================
+            Total register writes: {}
+            Total register reads: {}
+            Total register bounded {}
+            Total registers top {}
+            Percentage bounded: {}%
+            Percentage top: {}%
+            Average reduced interval size: {}
+            ============== Interval Analysis Summary End ================
```
hernanponcedeleon: Make the layout match that of `======== Program statistics ========` and `======== RelationAnalysis summary ========`

We are bounding the write of a register, right? i.e. if we have
```
r0 = r1 + r2
```
Here we are bounding r0. If so, why are we reporting # of registers read?
Also, isn't # of wirtes (or reads) the sum of bounded and top? If so, we do not gain much reporting all

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisGlobal.java
```diff
@@ -0,0 +1,86 @@
...
+    // Calculate the interval of a memory address.
+    // Takes into account all stores from a load can read from.
+    // Join the intervals of all possible stores.
+    private Interval calculatePossibleInterval(Set<Store> stores, Register r) {
+        if (!stores.isEmpty()) {
+            Interval interval = null;
+            for (Store s : stores) {
+                Map<Register, Interval> eventState = eventStates.getOrDefault(s, new HashMap<>());
+                Expression value = s.getMemValue();
+                Interval resultInterval = new AbstractExpressionEvaluator((IntegerType) r.getType(), value, eventState).getResultInterval();
```
hernanponcedeleon: Is this cast always safe? Why not add a precondition?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisGlobal.java
```diff
@@ -0,0 +1,86 @@
...
+        Set<Event> potentialStoreEvents =
+        relationAnalysis
+        .getKnowledge(
+            task
+            .getMemoryModel()
+            .getRelation(RF)
+        )
+        .getMaySet()
+        .getInMap()
+        .get(event);
```
hernanponcedeleon: You can use `getOrDefault(_, empty)` to avoid the null check before

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
```diff
@@ -0,0 +1,330 @@
...
+            "reg=" + reg +
+            ", interval=" + interval +
+            '}';
+        }
+    }
+
+
+
+    // Analyse an event to calculate an interval for a register (i.e. the transfer function).
+    // TODO: Maybe use the event visitor
```
hernanponcedeleon: This would certainly be cleaner

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
```diff
@@ -0,0 +1,330 @@
...
+                } else {
+                    return analyseOther(rw);
+                }
+            }
+        }
+        return null;
+    }
+
+    // Transfer functions to be implemented by subclasses
+    protected abstract RegisterState analyseLoad(Load l, Map<Register, Interval> eventState);
```
hernanponcedeleon: I think you can move the one computing top here

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }
+
+        // Add remaining registers
+        lessIntervals.putAll(moreIntervals);
+        return lessIntervals;
+    }
+
+    // For debugging
+    // TODO: Probably can be removed in final version
```
hernanponcedeleon: Remove if not really ncessary

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
```diff
@@ -0,0 +1,10 @@
+package com.dat3m.dartagnan.configuration;
+
+public enum IntervalAnalysisOptions implements OptionInterface {
+    NAIVE, LOCAL, GLOBAL;
```
hernanponcedeleon: Would it make sense to have a NONE method rather than NAIVE and avoid doing any computation?

Or do we want it to do the computation so we can assume we will always have an interval (even if it is top)?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
```diff
@@ -71,6 +72,10 @@ protected void runInternal() throws InterruptedException, SolverException, Inval
         prover.addConstraint(symmetryEncoder.encodeFullSymmetryBreaking());
 
         BooleanFormulaManager bmgr = context.getBooleanFormulaManager();
+        // Adding bounds
+        prover.writeComment("Bounds over variables");
+        BooleanFormula bounds = programEncoder.encodeBounds();
+        prover.addConstraint(bounds);
```
hernanponcedeleon:
```suggestion
        prover.addConstraint(programEncoder.encodeBounds());
```

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -232,6 +233,10 @@ protected void runInternal()
         prover.addConstraint(baselineEncoder.encodeFullMemoryModel());
         prover.writeComment("Symmetry breaking encoding");
         prover.addConstraint(symmetryEncoder.encodeFullSymmetryBreaking());
+        // Bounds
+        prover.writeComment("Bounds over variables");
+        BooleanFormula bounds = programEncoder.encodeBounds();
+        prover.addConstraint(bounds);
```
hernanponcedeleon:
```suggestion
        prover.addConstraint(programEncoder.encodeBounds());
```

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/IntervalAnalysisTest.java
```diff
@@ -0,0 +1,390 @@
...
+
+
+public class IntervalAnalysisTest {
+    private static final TypeFactory types = TypeFactory.getInstance();
+    private static final ExpressionFactory expressions = ExpressionFactory.getInstance();
+    private IntervalAnalysis runIntervalAnalysis(Program p, IntervalAnalysisOptions method) throws InvalidConfigurationException, IOException {
+        final String modelPath = "cat/sc.cat";
+        Configuration config = Configuration.builder()
+                .setOption(INTERVAL_ANALYSIS_METHOD,method.asStringOption())
+                .setOption(ENABLE_EXTENDED_RELATION_ANALYSIS, "false")
```
hernanponcedeleon: Why do you need to disable this one?

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/IntervalAnalysisTest.java
```diff
@@ -0,0 +1,390 @@
...
+        ProcessingManager.fromConfig(config).run(p);
+        final Wmm wmm = new ParserCat().parse(new File(getRootPath(modelPath)));
+        final VerificationTask task = VerificationTask.builder()
+                .withConfig(config)
+                .withBound(1)
+                .withTarget(Arch.C11)
+                .build(p, wmm, EnumSet.of(PROGRAM_SPEC));
+        wmm.configureAll(task.getConfig());
+        final Context context = Context.create();
+        performStaticProgramAnalyses(task,context,config);
```
hernanponcedeleon: Add spaces between `,` (same applies for all classes)

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
ThomasHaas: I think `NONE` is a fair option. The interval analysis is the first analysis we have that is fully optional, so there is no need for dummy values. The `encodeBounds` method could then also instantly return rather than traversing the whole program and all its register accesses.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: And move the `context.getExpressionEncoder()` to the outside of the loop

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
ThomasHaas: I think the analysis actually bounds the register reads `r1` and `r2` rather than the result register `r0`.
I'm wondering if it would be better to bound the writes instead, because there are certainly fewer of those.
Maybe bounding both could also be an option.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/Interval.java
tcherrou: `ExpressionKind` most likely.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
tcherrou: See 32c835aa993ba7ef1abcbe3750e07a4c446cdf1b

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
tcherrou: The analysis indeed bounds reads.  

Reporting separate metrics was useful when analysing the analysis.
As only `RegisterWrite` events where the ones that were analysed.

> Also, isn't # of wirtes (or reads) the sum of bounded and top? If so, we do not gain much reporting all

Yes this can be simplified.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisGlobal.java
tcherrou: In context of the analysis code it should be. On its own probably not.

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
tcherrou: NONE sounds interesting. It might depend on the use of TOP intervals, if there is any.

tcherrou: Since this analysis would be the first fully optional with `NONE`  there are several options I think 

1. Do not register the analysis in the solver if the option is NONE 
2. Register it, but NONE will probably be `null` 
3. Register it and return some dummy class that errors when trying to get an interval

If going with the first two options there will probably be null checks that are not very descriptive 
To avoid null checks, the `program.analysis.interval` option can be used but I am not sure how to do it cleanly, again there are several options.

The third option can circumvent the need to get the option in different classes.

ThomasHaas: I think not registering it in the first place is the best option. 
All passes/analyses that make use of it will have to be aware that the analysis can be missing and do the appropriate checks.

To do so, I think returning `Optional<IntervalAnalysis>` from `IntervalAnalysis.fromConfig` is the cleanest option. The caller then has control over whether to register the analysis or not.
I would probably provide a new `Context.registerOptional(Class<T>, Optional<T>)` function that registers the value only if it is present.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
```diff
@@ -0,0 +1,180 @@
...
+     * @param task
+     *  Verification task required to get the memory model when using the RelationAnalysis.
+     * @param config
+     * - Contains configuration options for choosing the desired analysis.
+     * @return
+     *  A completed interval analysis.
+     * @throws InvalidConfigurationException
+     */
+
+    static IntervalAnalysis fromConfig(Program program, Context analysisContext, VerificationTask task, Configuration config) throws InvalidConfigurationException {
```
ThomasHaas: I don't like this signature. Just take the task as input, it already contains both the program and the memory model. In fact, it even contains the Config.
If you want the memory model to be optional (only needed for Global analysis?), then I would specialize the `VerificationTask` to `Wmm`. Or do you need more than the `Wmm` here?

tcherrou: The signature can indeed be simplified to: 

`static IntervalAnalysis fromConfig(Program program, Context analysisContext, Wmm memoryModel, Configuration config) throws InvalidConfigurationException`

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
tcherrou: > I think not registering it in the first place is the best option.
All passes/analyses that make use of it will have to be aware that the analysis can be missing and do the appropriate checks.

If I understand correctly, the potential absence is reflected when registering the analysis but not while accessing the analysis.
If I do `analysisContext.get(IntervalAnalysis.class)`, then a `null` check is necessary (or a try catch if using `analysisContext.requires(IntervalAnalysis.class)`.

ThomasHaas: Yes, but that is the nature of optional analyses. The user has to check whether the analysis is present or not, 
either by `.get(...) + null-check` or by `.has(...)`.
You don't want to do `try-catch + .requires(...)` unless you have two optional analyses A and B, and B necessarily depends on A. Then if the user enables B but not A, it is fair to throw an exception in B via `.requires(...)`.

tcherrou: Please take a look at dfa4e767c038ea87beaef1a48434f2f53f326b40.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/IntervalAnalysisTest.java
tcherrou: No exact reason to be honest.
I removed it see 063a66e75c93b89ab2f90e4196b2b35ccd1fe6c6

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
ThomasHaas: Looks good. But code-wise, you should prefer early returns to avoid indentation and make the code cleaner overall. In fact, the code will look like your original version which assumed the presence of an interval analysis, modulo three lines of early-return code:
```Java
public BooleanFormula encodeBounds() {
   if (!analysisContext.has(IntervalAnalysis.class)) {
      return context.getBooleanFormulaManager().makeTrue();
   }

   List<BooleanFormula> enc = new ArrayList<>();
   IntervalAnalysis intervalAnalysis = analysisContext.get(IntervalAnalysis.class);
   ...
}
```

tcherrou: Thanks for the tip.
It can probably be used in more places in my code.
I suspect it is to separate preconditions from the main body for readability and indeed less indentation.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
tcherrou: See 351aa8e9d5c9e84bae258f0adccb9bf63d3d2f2b.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisGlobal.java
tcherrou: I think I will add `assert` statements to the code

hernanponcedeleon: I would rather use a `Precondition` to the method

tcherrou: See e7d5e5824723f83bc72d50241a8983512b8544ae

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/EventNotFoundException.java
tcherrou: @hernanponcedeleon Can I assume the event is always present?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
tcherrou: If you mean giving it a default implementation,  with f87fb3687afa293f16a7a963756b1b3a7d92876f every event (granted it is a  `RegWriter` to an integer register) has a default behaviour of returning top if there is no visit method implemented.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/EventNotFoundException.java
hernanponcedeleon: I think you can add a `Precondition` to the `getIntervalAt` method, and then get rid of this class.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
```diff
@@ -0,0 +1,190 @@
...
+                        logger.warn("No interval found at event {} for register {}. Defaulting to top interval", event.toString(), r.getName());
+                    }
+                    return Interval.getTop(itype);
+                }
+            } else {
+                throw new EventNotFoundException(event);
+            }
+        } else {
+            throw new InvalidRegisterTypeException(r.getType());
+        }
```
hernanponcedeleon: You should rather throw early ...
```
if (!(r.getType() instanceof IntegerType itype)) {
   throw new EventNotFoundException(event);
}
if (eventStates.containsKey(event)) {
...
}
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisGlobal.java
```diff
@@ -0,0 +1,92 @@
...
+            return new RegisterState(l.getResultRegister(), interval);
+        }
+    }
+
+    // Calculate the interval of a memory address.
+    // Takes into account all stores from a load can read from.
+    // Join the intervals of all possible stores.
+    private Interval calculatePossibleInterval(Set<Store> stores, Register r) {
+        Preconditions.checkArgument(r.getType() instanceof IntegerType);
+        IntegerType registerType = (IntegerType) r.getType();
```
hernanponcedeleon, line 64: not sure if naming after an `instanceof` check works in preconditions, but if it does, it would be better to use
```
Preconditions.checkArgument(r.getType() instanceof IntegerType registerType);
```

ThomasHaas: It doesn't work

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
```diff
@@ -0,0 +1,10 @@
+package com.dat3m.dartagnan.configuration;
+
+public enum IntervalAnalysisOptions implements OptionInterface {
```
hernanponcedeleon: For consistency with the other option classes, I would simply call this Interval (and probably rename `com.dat3m.dartagnan.program.analysis.interval.Interval as IntegerInterval`)

ThomasHaas: We don't have consistency: `Alias` and `RelationAnalysisMethod` are the options for other analyses.
I would rather rename `Alias` to `AliasAnalysisMethod` and the interval option to `IntervalAnalysisMethod`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+    // Iterate over the program.
+    // For each register check if their bound is reduced.
+    // Only executed if logger.isInfoEnabled = true
+    // Rounds half up for decimal values.
+    private void computeAnalysisMetrics(Program program) {
+        double totalRegWrites = 0;
+        double totalIntervalsReduced = 0;
+        double totalIntervalsTop = 0;
```
hernanponcedeleon: These are always int values, the decimal in the output is annoying
```
	#Register writes: 478.0
	#Register bounded 167.0
	#Registers top 290.0
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysisWorklist.java
tcherrou: What about replacing the exceptions fully with preconditions?

```java
public Interval getIntervalAt(Event event, Register r) {
        Type type = r.getType();
        Preconditions.checkArgument(type instanceof IntegerType, "Intervals not supported for type " + type);
        Preconditions.checkArgument(eventStates.containsKey(event), String.format("Event %s is not found", event));
        Map<Register, Interval> registerStates = eventStates.get(event);

        if (!registerStates.containsKey(r)) {
            if (logger.isWarnEnabled()) {
                logger.warn("No interval found at event {} for register {}. Defaulting to top interval", event.toString(), r.getName());
            }
            return Interval.getTop((IntegerType) type);
        }
        return registerStates.get(r);
    }
```

hernanponcedeleon: This code definitely looks much better

tcherrou: See 8957a2c131652fd3be9c74698a976efadba3b7e1

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/EventNotFoundException.java
tcherrou: See 8957a2c131652fd3be9c74698a976efadba3b7e1

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/IntervalAnalysisOptions.java
tcherrou: See 8a69b7384e7e7e929030e94add69ad73ebf2ba97

hernanponcedeleon:
@tcherrou there is only the open discussion in `IntervalAnalysis.java` about the statistics report in the log. Other than that, unless you have further planned changes, you can already remove the draft status. I think this is almost ready to be merged.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/interval/IntervalAnalysis.java
tcherrou: See a51105549b485fe2d2ef09ebcd7c565e7bcbfb82

tcherrou:
There is one more operator I want to try and implement and I want to run some benchmarks with some of the suggestions in this conversation.
However, this can be done in a separate pull request.

hernanponcedeleon:
> There is one more operator I want to try and implement and I want to run some benchmarks with some of the suggestions in this conversation. However, this can be done in a separate pull request.

Yes, I would suggest getting this one merged (I am fine with the current status, so unless @ThomasHaas has further suggestions I will merge).

If you come up with improvements, PRs will always be welcomed :D

ThomasHaas:
I am fine with merging.

>There is one more operator I want to try and implement and I want to run some benchmarks with some of the suggestions in this conversation.

What I would find interesting is if you bounded the result of register writes rather than register reads, because the bounds on register writes imply bounds on register reads, but there are less writers than readers in general.
Another interesting thing is exploiting control-flow branching conditions to reduce the intervals.

If you ever feel bored, you can try out these things :P.

tcherrou:
> What I would find interesting is if you bounded the result of register writes rather than register reads, because the bounds on register writes imply bounds on register reads, but there are less writers than readers in general.

This sounded interesting to me too.
Only encoding strictly positive or negative bounds for integers is something I want to try too.


> Another interesting thing is exploiting control-flow branching conditions to reduce the intervals.

This was actually present in an early version of the analysis but removed for simplicity. 
It will probably be easier to add now with the cleaner code.

hernanponcedeleon:
Thanks @tcherrou for bearing with us over the long review. 

Happy to see your thesis work finally merged (@apaolillo might also be happy about this).

apaolillo:
Excellent! Congratulations @tcherrou ! :tada:

tcherrou:
I am happy to see it finally merged.
Thank you for the lengthy review it improved the code a lot.

PR #966, opened 2026-02-02
Minor fix in vulkan model

natgavrilenko:


natgavrilenko:
> Are you sure this is correct?
> 
> The allow model explicitly has this
> 
> ```
> fr = (~rf . ((W -> W) & locord)) + (~rf . asmo) + ((stor[RFINIT]) . sloc . (stor[W])) - iden
> ```
> 
> If init events are part of `mutordatom` and thus `asmo`, there would not be a need for the `((stor[RFINIT]) . sloc . (stor[W]))`.
> 
> I know the alloy model has other problems, so the above does not indeed mean much, but the spec also states _The initialization is not considered to be a memory operation_.

In the old definition, initial writes wouldn't get into asmo anyway because of the intersection with mutordatom. In the fr definition, it still creates edges to non-atomic writes.

hernanponcedeleon:
> In the old definition, initial writes wouldn't get into asmo anyway because of the intersection with mutordatom. In the fr definition, it still creates edges to non-atomic writes.

I know the old definition was not adding inits to `asmo`, but unless I am wrong, the new definition does add them and this is what I am asking if you are sure is correct.

The spec says _"The initialization is not considered to be a memory operation."_ and _"For a given atomic write A, all atomic writes that are mutually-ordered with A occur in an order known as A’s scoped modification order. A’s scoped modification order relates no other operations."_ which I understand as _"inits are not part of asmo"_.

PR #968, opened 2026-02-03
Fix mathsat5 in native mode

hernanponcedeleon:
I also checked the other solvers in native mode: bitwuzla currently complains about JNI stuff, but even if I fix that, I eventually get an `Attempted to invoke pure virtual method` error. I do not think this is something we can fix on our side, so for that solver we would need to use JVM mode. For the other solvers there are no more problems related to reflection/JNI (at least for the few benchmarks I tried).


ThomasHaas (approved):
LGTM

PR #972, opened 2026-02-04
Remove RefinementModel

xeren:
Unifies the two `Wmm` instances maintained by `RefinementSolver` into one model equipped with the cut, a subset of axioms and definitions.  The `WmmEncoder` encodes the cut and all its dependents.  The `WmmSolver` builds its constraint system on all remaining axioms and definitions of the model.

- There is no second `VerificationTask` anymore.
- The second pass of static WMM analyses is no longer needed.
- `ForceEncodeAxiom` is no longer used.  Instead, the encode graphs get culled a bit more precisely.  This removes some dead parts from the encoding.


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -106,48 +105,23 @@ public static WmmEncoder withContext(EncodingContext context) throws InvalidConf
...
-                        Iterables.transform(Wmm.ANARCHIC_CORE_RELATIONS, memoryModel::getRelation), // base relations
-                        Iterables.transform(memoryModel.getAxioms(), Axiom::getRelation) // axiom relations
-                )
-        );
-        RelationEncoder v = new RelationEncoder();
-        for (Relation rel : depGraph.getNodeContents()) {
-            logger.trace("Encoding relation '{}'", rel);
-            rel.getDefinition().accept(v);
+        final Set<Constraint> constraints = context.getConstraintsToEncode();
+        final Collection<? extends Constraint> toEncode = DependencyGraph.from(constraints).getNodeContents();
```
ThomasHaas: The previous solution added `Wmm.ANARCHIC_CORE_RELATIONS` to the set of encoded relations which, IIRC, was related to encoding degenerate models like the empty memory model. You should double check that this still works and does not result in exceptions.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
```diff
@@ -41,10 +36,10 @@ public class ExecutionGraph {
     // Some are not declared final for purely technical reasons but all of them get
     // assigned during construction.
 
-    private final RefinementModel refinementModel;
-    private final BiMap<Relation, CAATPredicate> predicateToRelationMap;
-    private final BiMap<Axiom, Constraint> constraintMap;
-    private final Set<Relation> cutRelations;
+    private final Wmm memoryModel;
+    private final Set<? extends com.dat3m.dartagnan.wmm.Constraint> encodedConstraints;
```
ThomasHaas: the full qualification of `Constraint` is unnecessary, no?

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
```diff
@@ -227,16 +204,14 @@ private RelationGraph createGraphFromRelation(Relation rel) {
         final Class<?> relClass = rel.getDefinition().getClass();
         final List<Relation> dependencies = rel.getDependencies();
 
-        if (cutRelations.contains(rel)) {
-            graph = new DynamicDefaultWMMGraph(refinementModel.translateToBase(rel));
-        } else if (relClass == ReadFrom.class) {
-            graph = new ReadFromGraph();
+        if (encodedConstraints.contains(rel.getDefinition())) {
+            graph = new DynamicDefaultWMMGraph(rel);
```
ThomasHaas: You removed the special casing from rf and co, assuming they are always part of the cut.
This assumption can be made, but defaulting to `DynamicDefaultWMMGraph` might incur unnecessary overhead. 

@hernanponcedeleon, could you maybe try this branch on your mutex example that shows high population time and see how it is affected?

If the effect is really not big, we could delete `ReadFromGraph` and `CoherenceGraph` entirely.
Alternatively, you just put rf/co as special cases inside the `encodedConstraints` case to use the more efficient representation.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Constraint.java
```diff
@@ -1,18 +1,20 @@
...
 import com.dat3m.dartagnan.wmm.utils.graph.EventGraph;
+import com.google.common.collect.Collections2;
 import org.sosy_lab.java_smt.api.BooleanFormula;
 
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 
-public interface Constraint {
+public interface Constraint extends Dependent<Definition> {
```
ThomasHaas: I'm not sure about adding the interface here. I think it might be best to avoid it, and just use an appropriate overload of `DependencyGraph.from` that takes the dependency mapping as input.
I already thought a few times about getting rid of this interface as it serves too special of a purpose.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -51,6 +54,7 @@ public final class EncodingContext {
     private final RelationAnalysis relationAnalysis;
     private final FormulaManagerExt formulaManager;
     private final BooleanFormulaManager booleanFormulaManager;
+    private final Set<Constraint> constraintsToEncode;
```
ThomasHaas: I'm wondering if this set should be downward closed like the original cuts were.
Cause right now, the `WmmEncoder` and the `ExecutionGraph` classes separately compute the downward closure and I think there is no use for the non-closed version. 
There was at least a reason why I had defined cuts to be downward closed :)

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
hernanponcedeleon: > @hernanponcedeleon, could you maybe try this branch on your mutex example that shows high population time and see how it is affected?
> 
> If the effect is really not big, we could delete `ReadFromGraph` and `CoherenceGraph` entirely. Alternatively, you just put rf/co as special cases inside the `encodedConstraints` case to use the more efficient representation.

These are some numbers comparing both branches (I use vsyncer with the automatic loop unrolling, thus there are many entries for a single verification run)
```
development
   -- Population time: 2.867 secs
   -- Population time: 3.447 secs
   -- Population time: 7.509 secs
   -- Population time: 7.291 secs
   -- Population time: 7.480 secs
   -- Population time: 5.070 secs
   -- Population time: 6.791 secs
   -- Population time: 5.530 secs
   -- Population time: 6.689 secs
   -- Population time: 12.178 secs

PR
   -- Population time: 1.711 secs
   -- Population time: 1.739 secs
   -- Population time: 6.076 secs
   -- Population time: 6.592 secs
   -- Population time: 5.576 secs
   -- Population time: 5.347 secs
   -- Population time: 8.753 secs
   -- Population time: 8.599 secs
   -- Population time: 5.738 secs
   -- Population time: 8.620 secs
   -- Population time: 9.335 secs
   -- Population time: 13.193 secs
   -- Population time: 15.005 secs
``` 
I don't think this is much different from what I have seen between different runs on the same codebase (there is usually variance from run to run), thus probably we should not worry about any regression here.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
xeren: It conflicts with `com.dat3m.dartagnan.solver.caat.constraints.Constraint`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
xeren: I agree.  The anarchic relations containing a part of the proper program encoding is a quirk of the `encoding` package and should be resolved there.  The PR had `RefinementSolver` ensure this.  I will move this into `EncodingContext`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
xeren: I can unify these into `EncodingContext`.  `ExecutionGraph` is just concerned with membership.  `WmmEncoder` on the other hand iterates this collection.  The iteration order should probably be deterministic.

ThomasHaas: Sure. I would avoid explicity storing the `DependencyGraph` though. I think constructing a `Set` (`LinkedHashSet`) along the stratification is fine. That being said, it might still not be 100% deterministic.

dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/ExecutionGraph.java
xeren: I would rather not remove these classes, as there are other parts making similar assumptions, e.g. `CoreReasoner`.  Also, `ExecutionModel` already extracts this data and is a better object for exposing it.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
@xeren can you fix the conflict so we can merge?

PR #973, opened 2026-02-05
Add support for floats

hernanponcedeleon:
I tested the current version of the code on the 1400 SVCOMP float benchmarks.

Unfortunately, I cannot upload the results, but here is a sumamry
```
Correct: 101
Incorrect: 108
```
From the 108 incorrect ones, 104 are actually variants of [this](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/float-newlib/double_req_bl_0460.c?ref_type=heads) and [this](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/float-newlib/float_req_bl_0460.c?ref_type=heads) benchmarks, both of which use MSA. The wrong results  are probably due to that.

I still need to take a closer look to these 4 one and thus the DRAFT status.
```
float-benchs/interpolation2.c.p%2Bcfa-reducer.yml
float-benchs/interpolation2.c.v%2Bnlh-reducer.yml
float-benchs/interpolation2.yml
float-benchs/inv_sqrt_Quake.c.v%2Bcfa-reducer.yml
```


ThomasHaas:
You needed only few, clean changes to support floats. I like :)
I thought the code base was in a worse shape.

hernanponcedeleon:
> I still need to take a closer look to these 4 one and thus the DRAFT status.
> 
> ```
> float-benchs/interpolation2.c.p%2Bcfa-reducer.yml
> float-benchs/interpolation2.c.v%2Bnlh-reducer.yml
> float-benchs/interpolation2.yml
> float-benchs/inv_sqrt_Quake.c.v%2Bcfa-reducer.yml
> ```

Some update on this ... for the first 3 benchmarks, I get wrong results using z3 and cvc5, but correct results using mathsat5.

hernanponcedeleon:
I just noticed `float-benchs/inv_sqrt_Quake.c.v%2Bcfa-reducer.yml` also [has MSA](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/float-benchs/inv_sqrt_Quake.c.v+cfa-reducer.c?ref_type=heads#L11-14) ... so the only problematic benchmarks are the 3 above for which we get different results depending on the solver

hernanponcedeleon:
The problem has nothing to do with floats / encodings but rather with `memcpy`.

I rewrote this initialization
```
  float min[] = { 5,  10, 12, 30,  150 };
  float max[] = { 10, 12, 30, 150, 300 };
  float org[] = { 1, 5, -10, 30, 20 };
  float slope[] = { 4, -15, 40, -10, 0 };
```
as
```
  float min[5];
  float max[5];
  float org[5];
  float slope[5];

  min[0] = 5; min[1] = 10; min[2] = 12; min[3] = 30; min[4] = 150;
  max[0] = 10; max[1] = 12; max[2] = 30; max[3] = 150; max[4] = 300;
  org[0] = 1; org[1] = 5; org[2] = -10; org[3] = 30; org[4] = 20;
  slope[0] = 4; slope[1] = -15; slope[2] = 40; slope[3] = -10;slope[4] = 0;
```
and now I get correct results.

That problem is orthogonal to the changes of this PR, so now I am relatively confident that the encoding is correct.

benchmarks/miscellaneous/floats.c
```diff
@@ -0,0 +1,166 @@
+#include <math.h>
+#include <float.h>
+#include <assert.h>
+
+/* Nondeterministic generators */
+extern float  __VERIFIER_nondet_float(void);
+extern double __VERIFIER_nondet_double(void);
```
ThomasHaas: We should also add these in `dat3m.h`

benchmarks/miscellaneous/floats.c
```diff
@@ -0,0 +1,166 @@
...
+        double r2 = a + (b + c);
+        assert(r1 != r2);
+    }
+
+#elif TEST_ID == 12
+    /* --- Subnormal boundaries --- */
+    if (fabs(d) == DBL_MIN) {
+        double x = d / 2.0;
+        assert(fabs(x) < DBL_MIN);
+    }
```
ThomasHaas: This test looks similar to `Test 5`. If you changed `Test 5` to check `fabs(d) <= DBL_MIN` it would cover both tests, no?

benchmarks/miscellaneous/floats.c
```diff
@@ -0,0 +1,166 @@
...
+    {
+        float ff = (float)1e-300;
+        assert(ff == 0.0f);
+    }
+
+#elif TEST_ID == 15
+    /* --- Division corner cases --- */
+    if (!isnan(d) && !isinf(d) && d != 0.0) {
+        double x = d / d;
+        assert(x == 1.0);
```
ThomasHaas, line 144: Have you checked if the compiler actually generates this division? I could imagine he simplifies it immediately

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -71,6 +73,11 @@ public final class EncodingContext {
             secure = true)
     boolean useIntegers = false;
 
+    @Option(name = ROUNDING_MODE_FLOATS,
+            description = "Rounding mode for floating point operations.",
+            secure = true)
+    FloatingPointRoundingMode roundingModeFloats = NEAREST_TIES_TO_EVEN;
+
```
ThomasHaas, line 80: I know that we deliberately do not simplify float arithmetic right now and so it is fine to have the rounding mode only for encoding, but in principle it would make sense to have it accessible at a more general place. I'm not sure where to place it though.

Maybe we should add something (not in this PR) like a `ProgramSemantics` class that contains all options that affect the program semantics itself: the progress model, the rounding modes for floats, pointer provenance model, ...

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -512,6 +531,131 @@ public TypedFormula<BooleanType, BooleanFormula> visitIntCmpExpression(IntCmpExp
             return new TypedFormula<>(expr.getType(), enc);
         }
 
+        public TypedFormula<FloatType, ?> visitIntToFloatCastExpression(IntToFloatCast expr) {
+            final Formula operand = encodeIntegerExpr(expr.getOperand()).formula();
+            final FloatType fType = (FloatType) expr.getTargetType();
+            final int exponentBits = fType.getExponentBits();
+            final int mantissaBits = fType.getMantissaBits();
+            final FloatingPointType targetType = FormulaType.getFloatingPointType(exponentBits, mantissaBits);
```
ThomasHaas: These three steps occur many times. You could use a helper function `getSMTType(FloatType)` instead

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
```diff
@@ -213,10 +213,58 @@ public FloatLiteral makeZero(FloatType type) {
         return makeValue(BigDecimal.ZERO, type);
     }
 
+    public FloatLiteral makeInf(FloatType type) {
+        return new FloatLiteral(type, null, false, true);
+    }
```
ThomasHaas: If you want to support signs, we could use a BigDecimal of `+1.0` or `-1.0` to remember the sign in `FloatLiteral`. We should add preconditions that stop the user from accessing the underlying `BigDecimal` if `isInf/isNan` is true.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
```diff
@@ -598,27 +599,81 @@ public Expression visitICmpInst(ICmpInstContext ctx) {
...
     }
 
+    @Override
+    public Expression visitFCmpInst(FCmpInstContext ctx) {
+        final Expression left = visitTypeValue(ctx.typeValue());
+        final Expression right = checkExpression(left.getType(), ctx.value());
+        final String operator = ctx.fPred().getText();
+        final Expression compared = switch (operator) {
+            // TODO oeq vs ueq
+            case "oeq" -> expressions.makeFEQ(left, right, true);
```
ThomasHaas, line 606: What does the todo mean?

benchmarks/miscellaneous/floats.c
hernanponcedeleon: This remains in our IR
```
25:      if((float64 __r_nondet_double / float64 __r_nondet_double) o== float64(1.000000)); then goto 22.28
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLlvm.java
hernanponcedeleon: Outdated comment ... in the first commit I still had
```
            // TODO oeq vs ueq
            case "oeq" -> expressions.makeEQ(left, right);
            ...
            case "ueq" -> expressions.makeEQ(left, right);
```

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
hernanponcedeleon: You mean, rather than passing a null value, passing +1.0 or -0.0 and checking the sign bit of the value when the object is INF?

ThomasHaas: Yes. `FloatLiteral` would just need to have some checks to avoid the user calling `getValue` and getting the `+1.0/-1.0` dummy value although the value is actually infinity.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
ThomasHaas: Why don't you just pass in the `FloatType` rather than extracting the exponent/mantissa explicitly?
I think that would be even nicer.

dartagnan/src/main/java/com/dat3m/dartagnan/expression/ExpressionFactory.java
hernanponcedeleon: Now I just need to figure out why we still have wrong results for TEST_3

hernanponcedeleon: The problem was due to the fact our support for bitcast is wrong. This should be fixed by #957 thus I just disable the test for now.

ThomasHaas (approved):
LGTM

PR #976, opened 2026-02-09
[IR] LLVM cmpxchg Rework

xeren:
Restructures the `LlvmCmpXchg` event from #332 into a single-result event using the `AggregateType` from #493.  This was the only event in our IR with two result registers.

Also adds proper compilation of *weak* `LlvmCmpXchg`.  The intrinsics of `pthread_mutex_trylock`, `pthread_rwlock_tryrdlock` and `pthread_rwlock_trywrlock` now use strong events instead of weak ones.  The API seems not to specify these to fail spuriously.


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/llvm/LlvmCmpXchg.java
```diff
@@ -1,64 +1,32 @@
...
+    public LlvmCmpXchg(Register oldValueSuccessRegister, Expression address, Expression expectedValue,
+            Expression value, String mo, boolean isStrong) {
+        super(oldValueSuccessRegister, address, expectedValue, value, isStrong, mo);
         Preconditions.checkArgument(!mo.isEmpty(), "LLVM events cannot have empty memory order");
-        Preconditions.checkArgument(cmpRegister.getType() instanceof BooleanType,
-                "Non-boolean result register of LlvmCmpXchg.");
-        this.cmpRegister = cmpRegister;
+        final var type = oldValueSuccessRegister.getType() instanceof AggregateType t ? t : null;
+        Preconditions.checkArgument(type != null && type.getFields().size() == 2,
+                "Non-aggregate result register of LlvmCmpXchg.");
```
ThomasHaas: why not just `Preconditions.checkArgument(oldValueSuccessRegister.getType() instanceof AggregateType t && t.getFields().size() == 2`)?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -480,12 +480,9 @@ public static LlvmXchg newExchange(Register register, Expression address, Expres
             return new LlvmXchg(register, address, value, mo);
         }
 
-        public static LlvmCmpXchg newCompareExchange(Register oldValueRegister, Register cmpRegister, Expression address, Expression expectedAddr, Expression desiredValue, String mo, boolean isStrong) {
-            return new LlvmCmpXchg(oldValueRegister, cmpRegister, address, expectedAddr, desiredValue, mo, isStrong);
-        }
-
-        public static LlvmCmpXchg newCompareExchange(Register oldValueRegister, Register cmpRegister, Expression address, Expression expectedAddr, Expression desiredValue, String mo) {
-            return newCompareExchange(oldValueRegister, cmpRegister, address, expectedAddr, desiredValue, mo, false);
+        public static LlvmCmpXchg newCompareExchange(Register actualSuccess, Expression address, Expression expected,
```
ThomasHaas: What is the meaning of `actualSuccess`? What would be a "not-actual success"?

xeren: A better name could be `oldValueAndSuccess`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/llvm/LlvmCmpXchg.java
```diff
@@ -1,64 +1,32 @@
...
+    public LlvmCmpXchg(Register oldValueAndSuccessRegister, Expression address, Expression expectedValue,
+            Expression value, String mo, boolean isStrong) {
+        super(oldValueAndSuccessRegister, address, expectedValue, value, isStrong, mo);
         Preconditions.checkArgument(!mo.isEmpty(), "LLVM events cannot have empty memory order");
-        Preconditions.checkArgument(cmpRegister.getType() instanceof BooleanType,
-                "Non-boolean result register of LlvmCmpXchg.");
-        this.cmpRegister = cmpRegister;
+        Preconditions.checkArgument(oldValueAndSuccessRegister.getType() instanceof AggregateType t
+                        && t.getFields().size() == 2,
+                "Non-aggregate result register of LlvmCmpXchg.");
```
hernanponcedeleon, line 18: Why not add a check to be sure the part of the aggregate representing the comparison is bool?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorBase.java
```diff
@@ -83,6 +90,27 @@ public List<Event> visitLlvmStore(LlvmStore e) {
         throw error(e);
     }
 
+    @Override
+    public final List<Event> visitLlvmCmpXchg(LlvmCmpXchg e) {
+        final var type = e.getResultRegister().getType() instanceof AggregateType t ? t : null;
+        verify(type != null && type.getFields().size() == 2, "Invalid result type of '%s'", e);
```
hernanponcedeleon: Why not check directly in the `verify` that the type is an aggregate?

xeren: Because this method needs `type` later.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/lang/llvm/LlvmCmpXchg.java
xeren: Because it is `bv1` when parsed from LLVM IR.  During compilation, `VisitorBase` creates a `bool` register.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -861,20 +862,20 @@ private List<Event> newPthreadUnlock(Register oldValueRegister, Expression addre
...
         final Label spinLoopHead = EventFactory.newLabel("__spinloop_head");
         final Label spinLoopEnd = EventFactory.newLabel("__spinloop_end");
         return List.of(
                 spinLoopHead,
-                newPthreadTryLock(oldValueRegister, successRegister, address, true),
-                EventFactory.newJump(successRegister, spinLoopEnd),
+                newPthreadTryLock(oldValueSuccessRegister, address),
+                EventFactory.newJump(expressions.makeExtract(oldValueSuccessRegister, 1), spinLoopEnd),
                 EventFactory.newGoto(spinLoopHead),
                 spinLoopEnd
```
ThomasHaas, line 869: I just noticed that our loop unrolling will unroll the spinloop but the dynamic side-effect detection will actually bound the loop to one iteration. I think it would be fair to generate a `__VERIFIER_loopbound(1)` event here to restrict the loop.
Well, at least if dynamic spin loop detection is enabled... but I think we might want to assume it is enabled?

This optimization could be applied to every CAS-loop that we generate ourselves (if there are more such cases).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorBase.java
hernanponcedeleon: I don't get it. When you define `type` you can generate a null (if the instance check does not hold) but then in the verify you do not allow type to be null. Why not simlply
```
verify(e.getResultRegister().getType() instanceof AggregateType);
verify(e.getResultRegister().getType().getFields().size() == 2, "Invalid result type of '%s'", e);

```

ThomasHaas: He basically says that he wants to avoid doing the the cast to `AggregateType` part twice:
```
verify (type instance of AggregateType t && ...)
AggregateType t = (AggregateType) type; // "second" cast
// do something else with "t"
```
But I would totally prefer the version with the second cast instead of using nulls (`null` is always ugly).

hernanponcedeleon:
LGTM ... unless @ThomasHaas has further comments we can merge

ThomasHaas (approved):
LGTM

PR #980, opened 2026-02-10
Refactor WmmEncoder and add ActiveSetAnalysis class

ThomasHaas:
- Add `ActiveSetAnalysis` to compute active sets (encode sets). It contains the previous `EncodeSet` and `LazyEncodeSet` classes
- Move all Constraint encoding into `WmmEncoder`: Axioms don't encode themselves anymore
- Move all active set computation into `ActiveSetAnalysis`: Axioms don't compute their own active sets anymore

~EDIT: I think there is still more refactoring that can be done, but for now it should be in a good shape (a lot better than before!!!).~

EDIT 2: The `ActiveSetAnalysis` is in the encoding namespace, but it could also be put into the wmm.analysis namespace. I think both places are justifiable.


dartagnan/pom.xml
```diff
@@ -259,6 +259,7 @@
                             <className>com.dat3m.dartagnan.wmm.axiom.Emptiness</className>
                             <className>com.dat3m.dartagnan.wmm.axiom.Irreflexivity</className>
                             <className>com.dat3m.dartagnan.Dartagnan</className>
+                            <className>com.dat3m.dartagnan.encoding.ActiveSetAnalysis</className>
```
hernanponcedeleon, line 262: This is not enough to make it work with native. You also need to register the file [here](https://github.com/hernanponcedeleon/Dat3M/blob/master/dartagnan/src/main/resources/META-INF/native-image/com.dat3m.dartagnan/dartagnan/reflect-config.json). Maybe also update the comment `<!-- Every class should also be registered in OptionsInfo -->` to point to that file

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionNames.java
```diff
@@ -34,7 +34,7 @@ public class OptionNames {
     // Encoding Options
     public static final String USE_INTEGERS = "encoding.integers";
     public static final String ENABLE_ACTIVE_SETS = "encoding.activeSets";
-    public static final String REDUCE_ACYCLICITY_ENCODE_SETS = "encoding.wmm.reduceAcyclicityEncodeSets";
+    public static final String REDUCE_ACYCLICITY_RELEVANT_SETS = "encoding.wmm.reduceAcyclicityRelevantSets";
```
hernanponcedeleon, line 37: Changing options name is never a good idea, so unless there is a really good reason to do so, we should avoid it.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -0,0 +1,866 @@
...
+    public static ActiveSetAnalysis newInstance(Wmm memoryModel, Context analysisContext, Configuration config) throws InvalidConfigurationException {
+        return new ActiveSetAnalysis(memoryModel, analysisContext, config);
+    }
+
+    public static ActiveSetAnalysis newInstance(VerificationTask task, Context analysisContext) throws InvalidConfigurationException {
+        return newInstance(task.getMemoryModel(), analysisContext, task.getConfig());
+    }
+
+    private void logConfig() {
+        logger.info("{}: {}", ENABLE_ACTIVE_SETS, enableActiveSetComputation);
```
hernanponcedeleon, line 108: We should also log `REDUCE_ACYCLICITY_RELEVANT_SETS`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -0,0 +1,866 @@
...
+        final AxiomRelevantSets axiomRelevantSetsVisitor = new AxiomRelevantSets();
+        axiom2RelevantSets = new HashMap<>();
+        for (Axiom axiom : memoryModel.getAxioms()) {
+            axiom2RelevantSets.put(axiom, axiom.accept(axiomRelevantSetsVisitor));
+        }
+
+        if (!enableActiveSetComputation) {
+            // FIXME: This is wrong if XRA was performed!
+            this.definition2ActiveSets = memoryModel.getRelations().stream()
+                    .collect(Collectors.toMap(Relation::getDefinition, r -> ra.getKnowledge(r).getMaySet()));
```
hernanponcedeleon: Why is this wrong? Was it always the case it was wrong? It is the reason why e.g., `AnalysisTest` disables XRA? What needs to be done to fix this?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -0,0 +1,866 @@
...
+                    for (DependencyGraph<Event>.Node node2 : scc) {
+                        Event e1 = node1.getContent();
+                        Event e2 = node2.getContent();
+                        if (maySet.contains(e1, e2)) {
+                            result.add(e1, e2);
+                        }
+                    }
+                }
+            }
+
```
hernanponcedeleon, line 255: Useless new line

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -107,16 +96,19 @@ private EncodingContext(VerificationTask t, Context a, FormulaManager m, Collect
         executionAnalysis = a.requires(ExecutionAnalysis.class);
         aliasAnalysis = a.requires(AliasAnalysis.class);
         relationAnalysis = a.requires(RelationAnalysis.class);
-        formulaManager = new FormulaManagerExt(m);
-        booleanFormulaManager = formulaManager.getBooleanFormulaManager();
+        fmgr = new FormulaManagerExt(m);
+        bmgr = fmgr.getBooleanFormulaManager();
+        exprEncoder = new ExpressionEncoder(this);
+
+
```
hernanponcedeleon: Useless new line

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -298,23 +271,30 @@ public EdgeEncoder edge(Relation relation) {
...
-            }
-            if (must.contains(e1, e2)) {
-                return execution(e1, e2);
-            }
-            return variable.encode(e1, e2);
+            return !may.contains(e1, e2)
+                    ? bmgr.makeFalse()
+                    : must.contains(e1, e2)
+                    ? execution(e1, e2)
+                    : variable.encode(e1, e2);
```
hernanponcedeleon, line 278: Add indentation

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Assumption.java
```diff
@@ -23,8 +23,8 @@ public Assumption(Relation relation, EventGraph maySet, EventGraph mustSet) {
     public EventGraph getMustSet() { return must; }
 
     @Override
-    public Set<Relation> getConstrainedRelations() {
-        return Set.of(rel);
+    public List<Relation> getConstrainedRelations() {
+        return List.of(rel);
```
hernanponcedeleon, line 27: Why should we be able to get the same Relation more than once?

dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionNames.java
ThomasHaas: I think that option is never used externally and there is no reason to. Honestly, we could just remove the option. But if we keep it, I want it to match with the internal naming of things instead of having a comment that says "btw. encode sets == relevant sets" and forever living with this ugly discrepancy.

That being said, I could revert the renaming and use `encodeSets` again. The biggest issue I had with the naming was that `encodeSets` was misused to represent both "relevant sets" and "active sets", but I fixed that issue.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
ThomasHaas: It was always wrong: it is not true that `active-set(def(r)) \subseteq may-set(r)` if XRA was run. A big reason why I wanted to do this PR and to correctly define active sets on definitions rather than relations.

> It is the reason why e.g., AnalysisTest disables XRA

It is only one test in `AnalysisTest` that disables XRA and I think it does so for no reason. You can just enable it. We don't have any tests that run into this "wrong" case, because it only happens if the disable the active set computation altogether! This is not done anywhere.

If you mean `RelationAnalysisTest`, then XRA is disabled because it doesn't work for lazy RA.

> What needs to be done to fix this?

The easiest fix is to just not allow disabling active set computations :). The other one is doing something similar to the `runNative` method that computes the missing active entries by running a part of RA again.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Assumption.java
ThomasHaas: We shouldn't, but it is consistent with the axioms which also return lists. I think it also simplified some call site (I don't remember where though). Notice that uniqueness is not the only relevant part here, but also ordering: a reason why `Constraint.getConstrainedRelation` by default returns a List (also in general duplicate relations are permitted and necessary!).
In the case of singleton constraints, neither uniqueness nor ordering are relevant, and you can only capture one of them in the signature (unless we used `SequencedSet` from Java 21). Using the default signature is better IMO.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
ThomasHaas: I guess this depends on how you read the code:
```Java
// Chaining
if (!may.contains(e1. e2)) 
    bmgr.makeFalse();
else if (must.contains(e1, e2)) 
    execution(e1, e2);
else
    variable.encode(e1, e2)

// Nesting
if (!may.contains(e1. e2)) 
    bmgr.makeFalse();
else {
     if (must.contains(e1, e2)) 
          execution(e1, e2);
     else
          variable.encode(e1, e2)
}
```

My IDE IntelliJ defaults to the former when auto-formatting and it cannot be configure to produce the latter anymore. It seems to have been removed in 2017 or so. I think it is due to people using nested ternaries mostly to express chaining rather than proper tree-like nesting. I think this is very reasonable.

ThomasHaas:
Btw. the `ActiveSetAnalysis` contains two variants based on the underlying type of RA that was used. 
This distinction never made sense to me, `ActiveSetAnalysis` shouldn't really care about that. We can delete one of the two variants. I don't know if `LazyActiveSets` is supposed to be faster or something, but I'm inclined to delete that one, as it uses a lot more code.
Any objections?

ThomasHaas:
I just went ahead and deleted `LazyActiveSets` and simplified the whole code. Also, "disabled active sets + XRA" works correctly now.
With `LazyActiveSets` being gone, I think part of `RelationAnalysisTest` also became obsolete.

EDIT: I deleted the part of `RelationAnalysisTest` related to active sets and moved the file to the directory with the other tests.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
ThomasHaas: I fixed this.

ThomasHaas:
> Error:  Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.1.2:test (default-test) on project dartagnan: No tests matching pattern "com/dat3m/dartagnan/encoding/**/*" were executed!

Where is this test pattern specified? I don't see it in any `pom.xml` files. Or is this just a bug with outdated metadata being used?

hernanponcedeleon:
https://github.com/hernanponcedeleon/Dat3M/blob/30640123b88d4fc5ca01403f936d45c54d83a9c1/.github/workflows/maven.yml#L130

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
```diff
@@ -68,124 +72,228 @@ public class WmmEncoder implements Encoder {
             secure = true)
     private boolean allowMultiReads = false;
 
-    // =====================================================================
+    @Option(
+            name=IDL_TO_SAT,
```
hernanponcedeleon: Put them on a single line as the other options

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/LazyRelationAnalysis.java
```diff
@@ -86,13 +85,8 @@ public void runExtended() {
...
     @Override
-    public long countMaySet() {
-        return lazyKnowledgeMap.values().stream().mapToLong(k -> k.getMaySet().size()).sum();
-    }
-
-    @Override
-    public long countMustSet() {
-        return lazyKnowledgeMap.values().stream().mapToLong(k -> k.getMustSet().size()).sum();
+    public void collectDiscrepancies(Set<Relation> relations, Map<Relation, List<EventGraph>> discrepancies) {
+        // Without XRA, there are no discrepancies
```
hernanponcedeleon:
```suggestion
        // Without XRA (runExtended() is currently no implemented), there are no discrepancies
```

hernanponcedeleon (commented):
Code-wise LGTM (modulo the two tiny comments). I will do some further testing and if all is ok, I will merge

PR #937, opened 2025-09-30
[SPIRV] Add support for SRem, SMax, UMax, SMin and UMin

hernanponcedeleon:

PR #987, opened 2026-02-13
[IR] Maximum & Minimum

xeren:
Adds signed and unsigned integer maximum and minimum as binary operations.  Adds support for respective atomic operations in LLVM IR.


hernanponcedeleon:
We should also be able to simplify much of the code from #937, right?

ThomasHaas:
I don't think there is much code on #937 to be simplified, is there? It looks like you would only replace some `ITE(LTE(..), .., ..)` construct with a single operator but that's it.

Btw. I think a better long-term solution would actually be to have "Expressions with holes" and then instead of generating an RMWOp with some `(op, operand)`-pair you would just provide the expression. Then we don't even need ops for min/max, because you could put `min(_, val) := ITE( _ < val, _, val)` as RMW operation where `_` gets replaced by the register you load into. Basically, an RMWOp could compute an arbitrary expression over the loaded value instead of the handful of predefined one's.

EDIT: I just prototyped a version for `LLVM_rmw_op` with holes like so (taken from an IMM test)
```
8:       bv32 r3 = llvm_rmw_op(&x, bv32 {_} + bv32(1), REL) // amounts to fetch_add(&x, 1, REL)
```
I think it is certainly feasible and would could unify a lot. It would also cover Xchg (don't add a hole), possibly any form of CAS by adding an optional cmp-expr with hole, and generally also any fancy operation we can think of in our expression system, e.g. atomic_fetch_cttz, and any atomics on floats.

EDIT 2: I experimented some more with a very general `RMWReturn` that can model all returning RMWs it seems (technically it could even model non-returning RMWs to a certain extend). It uses three expressions with holes: one for the store value, one for an optional condition, and one for the return value. Maybe we should add these kinds of generalized events before we start adding more support for specific operations.
That being said, we can still merge this PR before doing so.

ThomasHaas:
The PR looks good to me. The only thing that could be added is simplifications of min/max in `ExprSimplifier`, but I think those are not important for now.
So from my side, it can be merged.

hernanponcedeleon:
@xeren can you use the new operators in the spirv parser from #937?

After those changes I will merge

ThomasHaas:
@hernanponcedeleon I made the changes you requested.

hernanponcedeleon (approved):

PR #988, opened 2026-02-13
[DRAFT] Expression templates and generic RMW events

ThomasHaas:
This PR is just a proof-of-concept implementation for generic RMW events that can model (hopefully) all standard RMWs by utilizing expression templates (expressions with holes).

~Right now, many unit tests will fail because I didn't add compilation schemes for the new events to all architectures but the LLVM parser will generate a new event already for CAS operations. However, one of the (two) new events can be compiled to IMM/LLVM and so tests on the language level should work.~


ThomasHaas:
I have replaced all LLVM RMW atomics by a single generic event and it seems to work. The compilation scheme for that event is also almost identical for all architectures (I just copy&pasted the code and changed the memory orders).

However, while everything seems to work now, the generic event is not suitable to capture spurious failure (weak CAS) right now. And I think it might be too much to also try to capture that in the generic event (at some point it tries to do too much).
I think it might be better to retain a dedicated CAS event (which could still be generalized in some aspects).

PR #989, opened 2026-02-15
Fix help option in native mode

### LINKED ISSUE #983: Option --help does not work in native mode
Guava’s ClassPath (used in `OptionInfo`) does not work reliably in native images so we might need a different solution than the current one.

hernanponcedeleon:
Fixes #983.

As a side-effect: now the outcome is immediate, while in the past it took a few seconds to collect all classes.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionInfo.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionInfo.java
@@ -1,5 +1,34 @@
 package com.dat3m.dartagnan.configuration;
 
+import com.dat3m.dartagnan.wmm.RelationNameRepository;
+import com.dat3m.dartagnan.configuration.OptionNames;
+import com.dat3m.dartagnan.wmm.axiom.Acyclicity;
+import com.dat3m.dartagnan.wmm.axiom.Emptiness;
+import com.dat3m.dartagnan.wmm.axiom.Irreflexivity;
+import com.dat3m.dartagnan.Dartagnan;
+import com.dat3m.dartagnan.encoding.EncodingContext;
+import com.dat3m.dartagnan.encoding.ProgramEncoder;
+import com.dat3m.dartagnan.encoding.SymmetryEncoder;
+import com.dat3m.dartagnan.encoding.WmmEncoder;
+import com.dat3m.dartagnan.program.analysis.ReachingDefinitionsAnalysis;
+import com.dat3m.dartagnan.program.analysis.alias.AliasAnalysis;
+import com.dat3m.dartagnan.program.processing.BranchReordering;
+import com.dat3m.dartagnan.program.processing.Inlining;
+import com.dat3m.dartagnan.program.processing.Intrinsics;
+import com.dat3m.dartagnan.program.processing.LoopUnrolling;
+import com.dat3m.dartagnan.program.processing.MemoryAllocation;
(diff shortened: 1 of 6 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/configuration/OptionInfo.java
```diff
@@ -14,65 +14,62 @@
...
+                "com.dat3m.dartagnan.wmm.Wmm$Config",
+                "com.dat3m.dartagnan.wmm.analysis.RelationAnalysis$Config",
+                "com.dat3m.dartagnan.wmm.analysis.WmmAnalysis",
+                "com.dat3m.dartagnan.wmm.processing.WmmProcessingManager"
+        ).map(name -> {
+            try {
+                return Class.forName(name);
+            } catch (ClassNotFoundException e) {
+                throw new RuntimeException("Cannot load class: " + name, e);
+            }
```
ThomasHaas: Can't you just list the classes explicitly e.g.,`WmmAnalysis.class` or are some private inner classes not accessible this way? Because with an explicit listing, the code becomes robust against refactoring/renaming

hernanponcedeleon: No, `Class.forName` requires the fully qualified name of the desired class.

ThomasHaas: My point is that you don't need to call `Class.forName` at all, you can just write a big list like this:
```Java
return Stream.of(
   WmmAnalysis.class,
   WmmProcessingManager.class,
   OptionNames.class,
   ...
)
```
The only limitation I can think of is that you cannot access private inner classes this way. I'm not sure if we have such classes. And even if, we could make them public as long as their constructor remains private.

hernanponcedeleon: How would I call `Options o = c.getAnnotation(Options.class);` if I do not have access to the class?

> I'm not sure if we have such classes.

We do have them ... see the strings having `$`

ThomasHaas: I don't understand. The stream still returns all classes just like before. The only difference is that you don't do a lookup by name but instead just put the class reference directly. 
I think that, e.g., `Class.fromName("com.dat3m.dartagnan.Dartagnan") == Dartagnan.class` will hold. Unless `Class.fromName` does more than just returning the class instance.

hernanponcedeleon: Ok, now I got what the suggestion was

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -77,7 +77,7 @@ static AliasAnalysis fromConfig(Program program, Context analysisContext, Config
     }
 
     @Options
-    final class Config {
+    public final class Config {
```
ThomasHaas, line 80: FYI, I think inner classes of interfaces are always public by default. In fact, you cannot even make them private.
But you can keep the explicit public modifier.

ThomasHaas (approved):
LGTM

PR #994, opened 2026-02-17
Distinguish release version from unstable ones

hernanponcedeleon:
This PR changes a bit how we handle version development. Rather than always reporting the commit id, we will distinguish between stable versions (those that get tagged) and development versions.

As soon as we make a release, the first commit after that will increase the version number (to mark the start of a new development cycle) and mark it as apha. Then, until we make a new release, we will report "next-stable-version-id-alpha (commit id)".

The last commit before a release will remove the alpha from the version string in the pom file, and we will tag, effectively reporting  this version as "latest-stable-version-id"

### DIFF
--- a/dartagnan/pom.xml
+++ b/dartagnan/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>com.dat3m</groupId>
         <artifactId>dat3m</artifactId>
-        <version>4.3.1</version>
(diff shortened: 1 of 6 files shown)


ThomasHaas:
Since I have barely any opinion on version numbering, I'm fine with your proposal :)

ThomasHaas (approved):
LGTM

PR #998, opened 2026-02-23
Implement missing visitors in SVCOMP events

hernanponcedeleon:


ThomasHaas:
Why do you need those visitor methods? Do you plan to overwrite them somewhere explicitly? Cause otherwise nothing changes as they already inherited those methods with the same implementation as the default in `EventVisitor`.

~EDIT: nvm, you already closed the branch :)~

hernanponcedeleon:
> Why do you need those visitor methods? Do you plan to overwrite them somewhere explicitly? Cause otherwise nothing changes as they already inherited those methods with the same implementation as the default in `EventVisitor`.

I am playing around with some code where I needed the visitors and I noticed they were missing. Even if they are not explicitly used, the same is true for the other SVCOMP events (on probably many more) and we still have visitors for all of them for completeness.

ThomasHaas:
If you have a use-case, it's fine. But generally, I would prefer to reduce the `EventVisitor` rather than making it larger, especially when it comes to marker events: we also have `StringAnnotation`, `FunCallMarker`, and `FunReturnMarker`, all of which have no special visitor methods.
Also, we should definitely move `LoopBound` and `NonDetChoice` somewhere else. They are hardly related to Svcomp.
Maybe such a small PR is suitable to also make this refactor :)

hernanponcedeleon:
> If you have a use-case, it's fine. But generally, I would prefer to reduce the `EventVisitor` rather than making it larger, especially when it comes to marker events: we also have `StringAnnotation`, `FunCallMarker`, and `FunReturnMarker`, all of which have no special visitor methods. 

Annotations do not have semantics, and you are right probably we do not really need a visitor for them, thus I removed the visitor for bound annotation.

> Also, we should definitely move `LoopBound` and `NonDetChoice` somewhere else. They are hardly related to Svcomp. Maybe such a small PR is suitable to also make this refactor :)

I move the bound + the events related to spinloop annotation to the annotations folder. I also move NonDetChoice to core

ThomasHaas:
Taking a quick look at the package structure, I think `NonDetChoice` should be in `event/lang/dat3m` because the event does get removed by a pass eventually.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -269,6 +267,22 @@ public static Assert newAssert(Expression expr, String errorMessage) {
...
+        return new LoopBegin();
+    }
+
+    public static SpinStart newSpinStart() {
+        return new SpinStart();
+    }
+
+    public static SpinEnd newSpinEnd() {
+        return new SpinEnd();
+    }
```
ThomasHaas: We never make use of those events, right? Shall we just delete them?

hernanponcedeleon: I was about to do that yesterday, but actually, even if we removed them from our headers, they might still appear in the llvm code and I thought we should keep them for backward compatibility.

The question I am now asking myself is, what happens if we remove the event and the corresponding code from the intrinsics? Would we report this as a "unknown function called reached"?

ThomasHaas: We could inline them as NOOP in our intrinsics. And in the headers, we could implement them as an empty function/macro. Then old code that still calls the functions will get handled internally as NOOP, and newly-compiled code will not even have the function calls as intrinsics (regular inlining handles them).

This way, we would still completely remove the events internally and only have the three intrinsics for backward-compatibility.

hernanponcedeleon: I would not add them to our headers: this kind of suggests the use can use it and I want to prevent this. If libvsync want to generate code using them, then it is its responsibility to have the headers. For us, we just need to be sure we parse them and the NOOP solution works.

ThomasHaas: You could use `inlineAsZero` which has the same semantics as the new noop. Though I don't mind having a distinguished `inlineNoOp` method. It's up to you.

hernanponcedeleon: Ohhh I did not noticed that acts as a NOOP if the function is non-value-returning

ThomasHaas (approved):
LGTM

Issue #977, opened 2026-02-09
[Bug] `memcpy_s` unit tests fails with `--mixedSize=true`


ThomasHaas:

hernanponcedeleon:
@xeren can you take a look?

xeren:
1. The test failed due to a register redefinition in `Intrinsics`.  All translations of `memcpy` share a common namespace for the intermediate registers.  These are used to load `archType`-sized chunks from the source array.  I did not account for name conflicts in #840, when I introduced the translation with one big access covering the entire area, to avoid spurious tearing points.
2. Fixing that, I discovered that the test timed out.  This was because the `memcpy_s` translation still had no big-access variant.  This lead to byte-sized tearing points everywhere.  I added the variants in the PR.
3. There were null-pointer accesses in the test from the first call `ret = memcpy_s(NULL, 3*sizeof(int), a, 3*sizeof(int));`.  The accesses are dead code that do not get eliminated, because the jump-over condition was `(...) || True`.  After I added shortcircuiting to the translation in `Intrinsics`, the accesses get properly eliminated.

ThomasHaas:
This was fixed by #1002 right?

hernanponcedeleon:
Yes

PR #1002, opened 2026-03-02
Fix mixed-size `memcpy_s`

xeren:
Modifies the intrinsics for the `string.h` standard library functions to be compatible with mixed-size accesses.  With the current default settings, dynamic memory accesses are torn into units of `archType`.  E.g. `memcpy(dest, src, 16)` translates to
```
bv64 __memcpy_1 = load(bv64 dest)
store(bv64 src, bv64 __memcpy_1)
bv64 __memcpy_2 = load(bv64 dest + (bv64) 8)
store(bv64 src + (bv64) 8, bv64 __memcpy_2)
```
With the option `--mixedSize=true`, more informed tearing is applied later.  So that call translates to
```
bv128 __memcpy#1 = load(bv64 dest)
store(bv64 src, bv128 __memcpy#1)
```
WIth this PR, that scheme also applies to `memcpy_s`, `memcmp` and `memset`.  This should fix at least #977.


dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1606,38 +1606,16 @@ private List<Event> inlineNonDet(FunctionCall call) {
     // For proper support, we need at least alias information and most likely also proper support for mixed-sized accesses
```
hernanponcedeleon: The mixed-size comment can go

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1756,37 +1717,60 @@ private List<Event> inlineMemCpyS(FunctionCall call) {
...
+        final boolean bigEndian = caller.getProgram().getMemory().isBigEndian();
+        assert bigEndian != caller.getProgram().getMemory().isLittleEndian();
 
-        if (!(numExpr instanceof IntLiteral numValue)) {
-            final String error = "Cannot handle memcmp with dynamic num argument: " + call;
-            throw new UnsupportedOperationException(error);
-        }
-        final int count = numValue.getValueAsInt();
+        final IntegerType resultType = returnReg.getType() instanceof IntegerType t ? t : null;
+        final Expression zero = expressions.makeZero(resultType);
```
hernanponcedeleon: I do not like this kind of code. If we expect the result register to be of type integer (otherwise we will get some exception somewhere else when trying to create the zero value with null type), we should have a proper precondition to the method

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1837,6 +1816,51 @@ private List<Event> inlineMemSet(FunctionCall call) {
...
+        }
+        throw new UnsupportedOperationException("Cannot handle dynamic count argument: %s".formatted(call));
+    }
+
+    private void forEachMemSpan(List<Event> replacement, Expression countExpr, FunctionCall call, MemAction action) {
+        final Slice count = toSlice(countExpr, call);
+        checkArgument(0 <= count.min && count.min <= count.max && 0 < count.step, "Invalid count %s: %s", count,  call);
+        final IntegerType countType = countExpr.getType() instanceof IntegerType t ? t : noIntegerType();
+        final Label end = EventFactory.newLabel("__end");
+        // FIXME Unsound optimization.  This should always be one.
```
hernanponcedeleon: Do we currently have the same unsound optimization, and this was just moved, or does the new code introduce the unsoundness?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1837,6 +1816,51 @@ private List<Event> inlineMemSet(FunctionCall call) {
...
+        final IntegerType firstType = types.getIntegerType(8 * firstBytes);
+        final IntegerType restType = types.getIntegerType(8 * restBytes);
+        for (int offset = 0; offset < count.min; offset += firstBytes) {
+            action.run(offset, firstType);
+        }
+        for (int offset = count.min; offset < count.max; offset += restBytes) {
+            if (offset - count.min % count.step == 0) {
+                final Expression offsetValue = expressions.makeValue(offset, countType);
+                replacement.add(EventFactory.newJump(expressions.makeLTE(countExpr, offsetValue, false), end));
+            }
```
ThomasHaas: What does this part mean? The condition looks strange  and currently always resolves to `offset - 0 == 0` which is never satisfied. Are you sure that it shouldn't be `(offset - count.min) % count.step == 0`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1807,28 +1791,23 @@ private List<Event> inlineMemSet(FunctionCall call) {
...
-        if (!(countExpr instanceof IntLiteral countValue)) {
-            final String error = "Cannot handle memset with dynamic count argument: " + call;
-            throw new UnsupportedOperationException(error);
-        }
         if (!(fillExpr instanceof IntLiteral fillValue && fillValue.isZero())) {
             //FIXME: We can soundly handle only 0 (and possibly -1) because the concatenation of
             // byte-sized 0's results in 0's of larger types. This makes the value robust against mixed-sized accesses.
             final String error = "Cannot handle memset with non-zero fill argument: " + call;
             throw new UnsupportedOperationException(error);
         }
```
ThomasHaas: I think this code is outdated. If MSA is enabled, we should be able to handle other `fillValue`s.
But careful: you would need to concatenate the `fillValue` byte with itself until you get a value that aligns with your step size.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1756,37 +1717,60 @@ private List<Event> inlineMemCpyS(FunctionCall call) {
...
+            final int byteSize = bigEndian ? type.getBitWidth() : 8;
+            for (int cmpByte = 0; cmpByte < type.getBitWidth(); cmpByte += byteSize) {
+                final Expression byte1 = expressions.makeIntExtract(regSrc1, cmpByte, cmpByte + byteSize - 1);
+                final Expression byte2 = expressions.makeIntExtract(regSrc2, cmpByte, cmpByte + byteSize - 1);
+                final Expression nonNegative = expressions.makeGTE(byte1, byte2, false);
+                final Expression nonPositive = expressions.makeLTE(byte1, byte2, false);
+                final Event constrainNegative = EventFactory.newAssume(expressions.makeOr(nonNegative, returnNegative));
+                final Event constrainPositive = EventFactory.newAssume(expressions.makeOr(nonPositive, returnPositive));
+                final Event breakIfNonzero = EventFactory.newJump(expressions.makeNEQ(byte1, byte2), endCmp);
+                replacement.addAll(List.of(constrainNegative, constrainPositive, breakIfNonzero));
```
ThomasHaas: I wouldn't go all the way to return arbitrary values in case of difference really(*). While the spec technically allows for arbitrary values, it is common to just return `byte2-byte1` or `sign(byte2-byte1)`.
More importantly though, a purely "operational" implementation is more friendly for other approaches like SMC or maybe generally for static analysis, e.g., an value range analysis could bound the output size.

And if you want to defend the "spec" implementation, you definitely need to put a comment in the code of why you have to guess return values.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1756,37 +1717,60 @@ private List<Event> inlineMemCpyS(FunctionCall call) {
...
-                    EventFactory.newJump(expressions.makeNEQ(src1, src2), endCmp)
-            ));
-        }
+            final Register regSrc1 = caller.newUniqueRegister("__memcmp_src1", type);
+            final Register regSrc2 = caller.newUniqueRegister("__memcmp_src2", type);
+            final Event load1 = EventFactory.newLoad(regSrc1, src1Addr);
+            final Event load2 = EventFactory.newLoad(regSrc2, src2Addr);
+            replacement.addAll(List.of(load1, load2));
+            final int byteSize = bigEndian ? type.getBitWidth() : 8;
+            for (int cmpByte = 0; cmpByte < type.getBitWidth(); cmpByte += byteSize) {
```
ThomasHaas: `byteSize` is an odd name for anything that is not always `8` :).
Maybe you want to call it `chunkSize` or similar.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1690,64 +1656,59 @@ private List<Event> inlineMemCpyS(FunctionCall call) {
...
         ));
 
         // Condition 2: dest != NULL && destsz <= RSIZE_MAX && (src == NULL || count > destsz || overlap(src, dest)) 
         // ----> return error > 0 and zero out [dest, dest+destsz)
         // The first two are guaranteed by not matching cond1
-        final Expression cond2 = expressions.makeOr(expressions.makeOr(srcIsNull, invalidCount), overlap);
-        CondJump skipE2 = EventFactory.newJump(expressions.makeNot(cond2), success);
+        CondJump check2part1 = EventFactory.newJump(srcIsNull, check2fail);
+        CondJump check2part2 = EventFactory.newJump(invalidCount, check2fail);
+        CondJump check2part3 = EventFactory.newJumpUnless(overlap, success);
```
ThomasHaas: I'm not sure about splitting the checks, but since you renamed `cond` to `check`, you have should fix the comments here. You can change `cond1` to `Condition 1`, then the comment is independent of the variable names.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: Support for even non-constant values should be implemented now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1796,23 +1793,21 @@ private List<Event> inlineMemSet(FunctionCall call) {
...
-            replacement.add(newStore(destAddr, zero));
+        // Generate stores
+        final Expression fillByte = expressions.makeIntegerCast(fillExpr, types.getByteType(), false);
+        final Map<Integer, Expression> fillByBitWidth = new HashMap<>();
+        fillByBitWidth.put(8, fillByte);
+        forEachMemSpan(replacement, countExpr, call, (offset, type) -> {
+            final Expression offsetExpr = expressions.makeValue(offset, types.getArchType());
+            final Expression destAddr = expressions.makeAdd(dest, offsetExpr);
+            final Expression fill = fillByBitWidth.computeIfAbsent(type.getBitWidth(),
+                    n -> expressions.makeIntConcat(IntStream.range(0, n).mapToObj(x -> fillByte).toList()));
```
ThomasHaas: This should use the number of bytes, no? I think this is off by factor 8.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: You are right, the parentheses were missing.  The condition is supposed to reduce the number of `countExpr` checks, if `count.step` is above one.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: It was not there before.  I removed it now.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1821,10 +1816,13 @@ private List<Event> inlineMemSet(FunctionCall call) {
         return replacement;
     }
 
-    private record Slice(int min, int max, int step) {}
+    // Represents a value space for size expressions, as usually occurring in memory operations with dynamic size.
+    private record Slice(int start, int end, int step) {}
```
ThomasHaas: I feel the comment says nothing really. But the renaming of the record members suggests now that we have it (roughly) refers to a concrete range `{int i = start; i <= end; i+=step}`? Is that correct?

After reading the the remaining code, it appears as if a `Slice` currently represents more than the range above, but it also implicitly represents the range `(0, start)` which I suggest to represent explicitly as a another slice `(0, start, start)`. See some reasoning further below.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1821,10 +1816,13 @@ private List<Event> inlineMemSet(FunctionCall call) {
...
 
+    // To be performed
     private interface MemAction { void run(int offset, IntegerType accessType); }
```
ThomasHaas: Just remove the comment and put the interface declaration right before `forEachMemorySpan` so the connection becomes clear.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1821,10 +1816,13 @@ private List<Event> inlineMemSet(FunctionCall call) {
...
 
+    // Extracts relevant information from a size expression.
     private Slice toSlice(Expression expression, FunctionCall call) {
```
ThomasHaas: Again, this comment means nothing. What is "relevant"? Just remove the comment.
Also, instead of saying "size expression" in the comment, just rename the input parameter to `sizeExpr`.
Lastly, it is confusing how a `sizeExpr` becomes the slice `Slice(sizeExpr, sizeExpr, sizeExpr)`.
So a slice represents a single point if the size expression is constant? Does the step-size mean anything if `start==end`?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1833,28 +1831,39 @@ private Slice toSlice(Expression expression, FunctionCall call) {
...
-        final int bytes = 4;
-        final int firstBytes = detectMixedSizeAccesses ? count.min : bytes;
-        final int restBytes = detectMixedSizeAccesses ? count.step : bytes;
-        final IntegerType firstType = types.getIntegerType(8 * firstBytes);
-        final IntegerType restType = types.getIntegerType(8 * restBytes);
-        for (int offset = 0; offset < count.min; offset += firstBytes) {
-            action.run(offset, firstType);
+        // The span from 0 to count.start does not need any checks for `countExpr`.
+        if (count.start > 0) {
+            final int firstBytes = detectMixedSizeAccesses ? count.start : 1;
```
ThomasHaas: This code suggest to me that you actually want `sizeExpr -> Slice(0, sizeExpr, 1)` in case of no MSA, no? 
Or maybe something like `Slice(0, sizeExpr, msa ? sizeExpr : 1)`.

It certainly feels strange that you generate a Slice and then basically ignore all of its entries in some case as if the slice is just wrong.
Also, it is strange to generate a slice that seemingly represents a range (x, y), but then special-treat the range (0, x). It sounds like you want two slices `initialSlice=(0, min, min)` and `slice=(min, max, step)`. If `(max - min) % step != 0 =: x`, you might even want to refine `slice=(min, max-x, step)` and add a third slice `restSlice(max - x, max, x)` for the final step.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1833,28 +1831,39 @@ private Slice toSlice(Expression expression, FunctionCall call) {
...
+            }
         }
-        for (int offset = count.min; offset < count.max; offset += restBytes) {
-            if (offset - count.min % count.step == 0) {
+        // Each span from count.start to count.end needs one check for `countExpr`.
+        final int restBytes = detectMixedSizeAccesses ? count.step : 1;
+        final IntegerType restType = types.getIntegerType(8 * restBytes);
+        for (int offset = count.start; offset < count.end; offset += restBytes) {
+            // Continue only if `countExpr` allows it.  Check this only when needed.
+            if ((offset - count.start) % count.step == 0) {
```
ThomasHaas: With the idea of multiple slices, I think this whole code can be merged into the above. Basically generate a bunch of slices (a slice decomposition if you will) and then loop through them.

Currently, this loop is completely dead because of `count.start==count.end` being guarateed to hold, right?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: No, the `Slice` should over-approximate the possible values for `countExpr`.  It should only depend on the program.  It determines the exact offsets where a check for `countExpr` is inserted.  On all other bytes, I want to omit the check.

What you call `initialSlice` describes what I call the 'first' span $[0,\mathtt{min}-1]$.  I describe the 'rest' spans being $[\mathtt{offset},\mathtt{offset}+\mathtt{step}-1]$ for $\mathtt{offset}\in\lbrace \mathtt{min},\ldots,\mathtt{max}\rbrace$.  I would not want to model a 'final' span.

Take for example `memcpy(dest, src, (either 4 or 6))`.  I want this to translate to either
```
  bv32 r0 = load(src + 0)
  store(dest + 0, r0)
  if 4 >= (either 4 or 6), goto __end
  bv16 r1 = load(src + 4)
  store(dest + 4,r1)
__end:
```
or
```
  bv8 r0 = load(src + 0)
  store(dest + 0, r0)
  bv8 r1 = load(src + 1)
  store(dest + 1, r1)
  bv8 r2 = load(src + 2)
  store(dest + 2, r2)
  bv8 r3 = load(src + 3)
  store(dest + 3, r3)
  if 4 >= (either 4 or 6), goto __end
  bv8 r4 = load(src + 4)
  store(dest + 4, r4)
  bv8 r5 = load(src + 5)
  store(dest + 5, r5)
__end:
```
So even with `--mixedSize=false`, I want this translation to skip checks where possible.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: I could have outlined these parts into another method right away, but I was not sure if they would stay this way.

Yes, the loop body becomes reachable, once the `Slice` `count` describes more than one possible value for `countExpr`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
ThomasHaas: > No, the Slice should over-approximate the possible values for countExpr.

Then I think you should write a comment that says exactly that.

> It should only depend on the program. It determines the exact offsets where a check for countExpr is inserted. On all other bytes, I want to omit the check.

I understood this optimization, but without a clear definition of what `Slice` was supposed to mean, it looked like it could mean both "a way to traverse the memory chunks" and "an approximation of countExpr".
In my suggestion above I took it more like the first way, in particular, (0, start, start) would represent a single step in the traversal and hence imply only a single bound check (if at all).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: I know no canonical way of describing a slice that represents one value.  With the `min` and `max` scheme, any `Slice(v,v,s)` works for a positive `s`.  I swapped to [Python's scheme](https://docs.python.org/3/reference/expressions.html#slicings) with `start` and `end`, where any `Slice(i,i+s,s)` is a valid constant.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: Almost, `Slice(start,end,step)` stands for `{int i = start; i < end; i += step}`.

`Slice` does not describe spans directly, but instead describes the integer values that `countExpr` *may* have in an execution.
It indirectly yields one *span* of size `start` and `(end-start)/step` spans of size `step`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: Actually, I replaced `cond1` and `cond2` by `destIsNull`, `invalidDestsz`, `srcIsNull` and `invalidCount`.  The jumps `checkXpartY` replace `checkX`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
xeren: I returned to the `byte2-byte1` scheme.

ThomasHaas:
I think the code looks a lot better now, especially with the clarification what a `Slice` is.
I have not checked all details, but I think it can be merged.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Intrinsics.java
```diff
@@ -1691,65 +1655,60 @@ private List<Event> inlineMemCpyS(FunctionCall call) {
...
-        Local retError1 = EventFactory.newLocal(resultRegister, errorCodeFail);
+        final Label check1 = EventFactory.newLabel("__memcpy_s_check_1");
+        final Label check1fail = EventFactory.newLabel("__memcpy_s_fail_1");
+        final Label check2 = EventFactory.newLabel("__memcpy_s_check_2");
+        final Label check2fail = EventFactory.newLabel("__memcpy_s_fail_2");
+        final Label success = EventFactory.newLabel("__memcpy_s_success");
+        final Label end = EventFactory.newLabel("__memcpy_s_end");
+
+        final Expression errorCodeFail = expressions.makeOne((IntegerType)resultRegister.getType());
+        final Expression errorCodeSuccess = expressions.makeZero((IntegerType)resultRegister.getType());
```
hernanponcedeleon: `errorCodeSuccess` -> having error/success together is weird. I would call it `returnCode`

PR #1003, opened 2026-03-07
Refactor Event Semantics

ThomasHaas:
- Removed `Event.encodeExec` and instead moved the encoding into `ProgramEncoder.encodeEventSemantics`.
- `ControlBarrier.execScope` is now used in the encoding: the previous code always defaulted to `Vulkan.WG/PTX.CTA` irrespective of the specified execution scope.
- `CoreCodeVerification` now checks that all `ControlBarriers` are placed correctly: the control barriers execution scope matches with the thread it is in.

EDIT
- `ThreadCreate` cannot spuriously fail anymore. Instead, the inlining of `pthread_create` explicitly models the failure case via guessing.


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -14,18 +14,15 @@
...
-import com.dat3m.dartagnan.program.event.core.CondJump;
-import com.dat3m.dartagnan.program.event.core.ControlBarrier;
-import com.dat3m.dartagnan.program.event.core.Label;
-import com.dat3m.dartagnan.program.event.core.NamedBarrier;
+import com.dat3m.dartagnan.program.event.core.*;
+import com.dat3m.dartagnan.program.event.core.threading.ThreadArgument;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadJoin;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadReturn;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.dat3m.dartagnan.program.memory.Memory;
```
hernanponcedeleon, line 19:
```suggestion
import com.dat3m.dartagnan.program.event.core.threading.*;
```

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
```diff
@@ -293,14 +281,20 @@ private BooleanFormula encodeThreadJoining() {
         return bmgr.and(enc);
     }
 
+    private String getScopeId(ControlBarrier barrier) {
+        final ScopeHierarchy hierarchy = barrier.getThread().getScopeHierarchy();
+        final int id = hierarchy.getScopeId(barrier.getExecScope());
+        assert id >= 0;
+        return barrier.getExecScope() + id;
```
hernanponcedeleon, line 285: I do not understand this addition

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CoreCodeVerification.java
```diff
@@ -54,5 +55,30 @@ public void run(Function function) {
...
             throw new MalformedProgramException("Found non-core events.\n" + msg);
         }
+
+        validateCoreEvents(function);
+    }
+
+    private void validateCoreEvents(Function function) {
+        StringBuilder msg = new StringBuilder();
+
+        for (Event e : function.getEvents()) {
```
hernanponcedeleon: Why not `function.getEvents(ControlBarrier.class)` and avoid the if?

ThomasHaas: Here I did it in anticipation that we will add more checks for other events in the future.
Those could all be done with a separate `function.getEvents(...)` loop, but then we iterate through the whole function many times rather than once which - to be fair - shouldn't be performance relevant.

I can change it if you prefer multiple specialized loops over a single general loop.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProgramEncoder.java
ThomasHaas: The old code used to just return `id` assuming that it means the workgroup id.
Since the new code is generalized to any scope, I need to distinguish the ids by scope (`@wg,1 != @cta,1`), so I prepend the scope name to the `id`.

hernanponcedeleon: Ohhhh the return type is a String, I thought it was an int and I could not figure out why this sound would result in a unique identifier

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/CoreCodeVerification.java
hernanponcedeleon: Seeing that the CoreCodeVerification pass has not changed drastically in quite some time, I have doubts that we might extend it to other events in the near future.

I would thus go with the specific event type for now, and if we ever end up adding more types of events, we can go back to the `for (any) { if (specif) {...} }` pattern.

hernanponcedeleon (approved):

PR #1005, opened 2026-03-09
Delete unused code / minor refactor to reduce IDE warnings

ThomasHaas:
There are no functional changes in this PR.

EDIT: Okay, there is one bug fix that I accidentally committed as well. It's a single line of code added in cutting:
```Java
   ... || def instanceof Free;
```


dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -12,20 +12,32 @@
 import com.dat3m.dartagnan.program.Register;
 import com.dat3m.dartagnan.program.Thread;
 import com.dat3m.dartagnan.program.event.arch.*;
-import com.dat3m.dartagnan.program.event.arch.ptx.*;
+import com.dat3m.dartagnan.program.event.arch.ptx.PTXAtomCAS;
+import com.dat3m.dartagnan.program.event.arch.ptx.PTXAtomExch;
+import com.dat3m.dartagnan.program.event.arch.ptx.PTXAtomOp;
+import com.dat3m.dartagnan.program.event.arch.ptx.PTXRedOp;
```
hernanponcedeleon: Revert this (I do not even understand why your IDE goes from explicit to * in other parts and from * to explicit here).

There are other files where this gets expanded, revert those too

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -440,10 +440,25 @@ private boolean checkProgress(List<RefinementIteration> trace) {
...
+    private static boolean isUnknownDefinitionForCAAT(Definition def) {
+        // TODO: We should probably automatically cut all "unknown relation",
+        //  i.e., use a white-list of known relations instead of a blacklist of unknown one's.
+        return def instanceof LinuxCriticalSections // LKMM
+                || def instanceof AMOPairs || def instanceof SameInstruction
+                || def instanceof CASDependency // IMM
+                // GPUs
+                || def instanceof SameScope || def instanceof SyncWith
+                || def instanceof SyncFence || def instanceof SyncBar || def instanceof SameVirtualLocation
+                || def instanceof Free;
```
hernanponcedeleon: now it kind of looks like this is related to GPUs. I would move it together with `AMOPairs` and `SameInstruction`

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
ThomasHaas: Do we have a rule-of-thumb of when we want to use explicit imports? Would you already want to use `.*` for just two imports?

EDIT: It seems my IDE collapses import to `.*` when 5 classes get imported from the same package.

hernanponcedeleon: If the whole package only contains two imports (my comment _"There are other files ..."_ refers to svcomp package which only has two classes), then I would say yes.

As a rule-of-thumbs I would say: _"do not import e.g., the whole java.utils unnecessarily, but if the package just contains a few classes (this would cover e.g., all our event and expressions packages) go for *"_

hernanponcedeleon (approved):

PR #1006, opened 2026-03-11
Add debug information to litmus format

hernanponcedeleon:
We currently provide no useful debug information to the user for litmus programs . For example
```
> java -jar dartagnan/target/dartagnan.jar cat/vmm.cat --property=cat_spec litmus/C11/auto/cyc_na.litmus
Test: litmus/C11/auto/cyc_na.litmus
Result: FAIL
Reason: CAT specification violation found
Details:
        Flag r-data-race
        E3 / E3 @unknown / @unknown
Time: 0.582 secs
```
The same is also true for witnesses.

This PR adds line number debug information to litmus code. So far it only does so for a few instructions (thus the draft status), but once we agree on code and the format we want for the output/witness, I will move the code in `addChildWithMetadata` to `addChild` and pass the LOC in all litmus visitors.

**Output format:** once we have proper lines of code for every format, which we already have for spirv/llvm (unless they fail to provide debug info, but I have rarely seen this happen), I think the information about the events id becomes completely useless for the user (it is probably already useless unless they know how to use the printer). However, it might still be useful for developers, so I will keep it, but at the end of the output.

The output for the example above looks like this now
```
> cat -n litmus/C11/auto/cyc_na.litmus 
     1  C cyc_na
     2  { [x] = 0; [y] = 0; }
     3
     4  P0 (volatile int* x, volatile int* y) {
     5    int r0 = *x;
     6    if (r0) {
     7      *y = 1;
     8    }
     9  }
    10
    11  P1 (volatile int* x, volatile int* y) {
    12    int r1 = *y;
    13    if (r1) {
    14      *x = 1;
    15    }
    16  }
    17
    18  exists (0:r0=1 /\ 1:r1=1)

> java -jar dartagnan/target/dartagnan.jar cat/vmm.cat --property=cat_spec litmus/C11/auto/cyc_na.litmus
Test: litmus/C11/auto/cyc_na.litmus
Result: FAIL
Reason: CAT specification violation found
Details:
        Flag r-data-race
        P1#12 / P1#12   (E14 / E14)
Time: 0.489 secs
```
This is even more useful for litmus syntax where each thread is a column
```
> cat -n /home/drc/git/Dat3M/litmus/VULKAN/Data-Race/mpnotinscope2-filter.litmus
     1  Vulkan mpnotinscope2-filter
     2  {
     3  x=0;
     4  y=0;
     5  }
     6  P0@sg 0, wg 0, qf 0            | P1@sg 0, wg 1, qf 0             ;
     7  st.atom.rel.wg.sc0.semsc0 x, 1 | ld.atom.acq.dv.sc0.semsc0 r0, y ;
     8  st.atom.rel.dv.sc0.semsc0 y, 1 | ld.atom.acq.wg.sc0.semsc0 r1, x ;
     9
    10  filter
    11  (P1:r0 == 1)

> java -jar dartagnan/target/dartagnan.jar cat/vulkan.cat --target=vulkan --property=cat_spec /home/drc/git/Dat3M/litmus/VULKAN/Data-Race/mpnotinscope2-filter.litmus
Test: /home/drc/git/Dat3M/litmus/VULKAN/Data-Race/mpnotinscope2-filter.litmus
Filter: bv64 r0 == bv64(1)
Result: FAIL
Reason: CAT specification violation found
Details:
        Flag racy
        P0#7 / P1#8     (E1 / E7)
Time: 0.661 secs
```

For the witness, since the nodes already contain the thread, I only use the LOC.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
@@ -17,6 +17,7 @@ import com.dat3m.dartagnan.program.event.RegWriter;
 import com.dat3m.dartagnan.program.event.core.Label;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.dat3m.dartagnan.program.event.metadata.OriginalId;
+import com.dat3m.dartagnan.program.event.metadata.LineNumber;
 import com.dat3m.dartagnan.program.memory.Memory;
 import com.dat3m.dartagnan.program.memory.MemoryObject;
 import com.dat3m.dartagnan.program.memory.VirtualMemoryObject;
@@ -193,6 +194,13 @@ public class ProgramBuilder {
         return child;
     }
 
+    // TODO: make this the default addChild implementation
+    public Event addChildWithMetadata(int fid, Event child, int ln) {
+        Event e = addChild(fid, child);
+        e.setMetadata(new LineNumber(ln));
+        return e;
+    }
+
     // ----------------------------------------------------------------------------------------------------------------
     // Memory & Constants
 
@@ -322,7 +330,7 @@ public class ProgramBuilder {
     }
 
     public void newScopedThread(Arch arch, int id, int ...ids) {
-        newScopedThread(arch, String.valueOf(id), id, ids);
+        newScopedThread(arch, "P" + String.valueOf(id), id, ids);
     }
 
     // ----------------------------------------------------------------------------------------------------------------
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
@@ -194,10 +194,9 @@ public class ProgramBuilder {
         return child;
     }
 
-    // TODO: make this the default addChild implementation
-    public Event addChildWithMetadata(int fid, Event child, int ln) {
+    public Event addChildWithSourceLocation(int fid, Event child, int lineOfCode) {
         Event e = addChild(fid, child);
-        e.setMetadata(new SourceLocation.Litmus(ln));
+        e.setMetadata(new SourceLocation.Litmus(lineOfCode));
         return e;
     }
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
@@ -254,20 +254,20 @@ public class ProgramBuilder {
         getOrNewMemoryObject(locName).setInitialValue(0, iValue);
     }
 
-    public void initRegEqLocPtr(int regThread, String regName, String locName, Type type) {
+    public void initRegEqLocPtr(int regThread, String regName, String locName, Type type, int lineOfCode) {
         MemoryObject object = getOrNewMemoryObject(locName);
         Register reg = getOrNewRegister(regThread, regName, type);
-        addChild(regThread, EventFactory.newLocal(reg, object));
+        addChildWithSourceLocation(regThread, EventFactory.newLocal(reg, object), lineOfCode);
     }
 
-    public void initRegEqLocVal(int regThread, String regName, String locName, Type type) {
+    public void initRegEqLocVal(int regThread, String regName, String locName, Type type, int lineOfCode) {
         Register reg = getOrNewRegister(regThread, regName, type);
-        addChild(regThread, EventFactory.newLocal(reg,getInitialValue(locName)));
+        addChildWithSourceLocation(regThread, EventFactory.newLocal(reg,getInitialValue(locName)), lineOfCode);
     }
 
-    public void initRegEqConst(int regThread, String regName, Expression value){
+    public void initRegEqConst(int regThread, String regName, Expression value, int lineOfCode){
         Preconditions.checkArgument(value.getRegs().isEmpty());
-        addChild(regThread, EventFactory.newLocal(getOrNewRegister(regThread, regName, value.getType()), value));
+        addChildWithSourceLocation(regThread, EventFactory.newLocal(getOrNewRegister(regThread, regName, value.getType()), value), lineOfCode);
     }
 
     private Expression getInitialValue(String name) {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
@@ -185,20 +185,19 @@ public class ProgramBuilder {
         throw new MalformedProgramException("Function or Thread with id " + fid + " does not exist");
     }
 
-    public Event addChild(int fid, Event child) {
+    public Event addChild(int fid, Event child, int lineOfCode) {
         // Every event in litmus tests is non-optimisable
         if (program.getFormat().equals(Program.SourceLanguage.LITMUS)) {
             child.addTags(NOOPT);
         }
         getFunctionOrError(fid).append(child);
+        final String threadName = child.getThread().getName();
+        child.setMetadata(new SourceLocation.Litmus(threadName, lineOfCode));
         return child;
     }
 
-    public Event addChildWithSourceLocation(int fid, Event child, int lineOfCode) {
-        Event e = addChild(fid, child);
-        final String threadName = child.getThread().getName();
-        e.setMetadata(new SourceLocation.Litmus(threadName, lineOfCode));
-        return e;
+    public Event addChild(int fid, Event child) {
+        return addChild(fid, child, -1);
     }
 
     // ----------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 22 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
```diff
@@ -193,6 +194,13 @@ public Event addChild(int fid, Event child) {
         return child;
     }
 
+    // TODO: make this the default addChild implementation
+    public Event addChildWithMetadata(int fid, Event child, int ln) {
+        Event e = addChild(fid, child);
+        e.setMetadata(new LineNumber(ln));
+        return e;
+    }
```
ThomasHaas: It should probably be named `addChildWithSourceInfo/DebugInfo` or even `addChildWithLineNumber`.
Of course, if you eventually decide that this is the default `addChild` implementation, then the name can be simplified again.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/SyntacticContextAnalysis.java
```diff
@@ -222,6 +221,13 @@ public static String getSourceLocationString(Event ev) {
         return loc != null ? String.format("@%s#%s", loc.getSourceCodeFileName(), loc.lineNumber()) : "@unknown";
     }
 
+    public static String getLitmusLocationString(Event ev, boolean showThread) {
+        final String prefix = showThread ? ev.getThread().getName() : "";
+        return ev.hasMetadata(LineNumber.class) ?
+            String.format("%s#%s", prefix, ev.getMetadata(LineNumber.class).lineNumber()) :
+            "@unknown";
+    }
+
```
ThomasHaas: I'm not conviced that the use-site should know what shape the source location comes in.
Maybe you should actually have `SourceLocation` as the root type, and then subtypes like `LineNumber`.
Maybe it is even better to have language-specific subtypes like `LitmusSourceLoc` and `LLVMSourceLoc` (or `GenericSourceLoc` for high-level languages?). You can define those even as nested types `SourceLocation.Litmus` and `SourceLocation.LLVM/SourceLocation.Generic` within `SourceLocation`.

The `getSourceLocationString` method can do the case distinction based on the type of the source location.
The caller might still need to do a case-distinction based on whether he wants the thread to be printed or not though (unless we simply ignore that parameter for LLVM source locations).
However, you should fix `MetadataMap.contains` to look for subclasses instead of strict class equality (oddly enough, `MetadataMap.get` is based on subclassing?!). Then just use `MetadataMap.get(SourceLocation.class)`.
In fact, if you overwrite `toString` for both types of source location appropriately, you don't even need a case distinction at all: just use `String.format("@%s%s", prefix, srcLoc)` (I would still put "@" for litmus code!)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java
hernanponcedeleon: Yes, as the PR description mentions, I will eventually merge both methods into a default `addChild`

ThomasHaas:
Btw. we should stay with the name `SourceLocation`. I think `DebugInfo` is more suitable to store information like source-level types and variable names.

ThomasHaas:
I pushed a variant of what I have suggested. Feel free to revert or modify the changes if you don't like them.
I have also not tested if the output has changed or not.
After the refactor, it looks like the two variants of `SourceLocation` can be merged again by making the `filePath` optional. Then we could add static factory methods like `SourceLocation.withFilePath()` or `SourceLocation.onlyLineNumber()`.

hernanponcedeleon:
The output is still the same and the code LGTM (thanks!). Maybe I would just change `public static String getSourceLocationString(Event ev, boolean showThread)` to `public static String getSourceLocationString(Event ev, boolean showThreadForLitmus)` to make it clear in the method signature what the implementation actually do.

I'll change all the litmus visitors to make use of the new metadata.

ThomasHaas:
> The output is still the same and the code LGTM (thanks!). Maybe I would just change `public static String getSourceLocationString(Event ev, boolean showThread)` to `public static String getSourceLocationString(Event ev, boolean showThreadForLitmus)` to make it clear in the method signature what the implementation actually do.

We could also have `SourceLocation.Litmus` contain a `String threadName`? I think this would make sense. For example, it would allow us to rename our internal threads (for whatever reason) without breaking source location information. After all, the source location should be independent of whatever we do internally.
But then I would unconditionally print it. Is the only case you want to avoid having the thread printed twice in the execution witness?

hernanponcedeleon:
I thought about it, but then we could not simply use `toString()` ... we still need a way to enable/disable showing the thread.

There are only two places where we use `SourceLocation`: command output and witness (we might want to use this in the UI, but AFAIR we do not do it now). We want to show it in the first one (this is particularly useful for the litmus variants where threads are in columns), but not for the later (to avoid repetition)

ThomasHaas:
> I thought about it, but then we could not simply use `toString()` ... we still need a way to enable/disable showing the thread.

That's why I'm asking if we should unconditionally print the thread. Our events always have an IR-level location and optionally a source-level location, and currently we just print both of them (which is IMO fine).
I'm not sure if it is worth working around cases where both types of locations happen to have something in common.
And we could always have an option that generates execution witnesses only with source location information and omits our internal structure. Then the duplication would automatically be gone.

> There are only two places where we use `SourceLocation`: command output and witness (we might want to use this in the UI, but AFAIR we do not do it now). We want to show it in the first one (this is particularly useful for the litmus variants where threads are in columns), but not for the later (to avoid repetition)

Is it really worth it to avoid the repetition though?

hernanponcedeleon:
After giving it another try, maybe it does not look as bad as I thought
image

ThomasHaas:
Yeah, I think Litmus code is very forgiving with the redundancy because the thread names are so short.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/SyntacticContextAnalysis.java
```diff
@@ -218,8 +216,8 @@ private Map<Event, LoopAnalysis.LoopIterationInfo> getLoopMarkerTypesMap(Thread
     // ============================================================================
 
     public static String getSourceLocationString(Event ev) {
-        SourceLocation loc = ev.getMetadata(SourceLocation.class);
-        return loc != null ? String.format("@%s#%s", loc.getSourceCodeFileName(), loc.lineNumber()) : "@unknown";
+        final SourceLocation loc = ev.getMetadata(SourceLocation.class);
+        return String.format("@%s", loc != null ? loc : "#unknown");
```
ThomasHaas: Why the `#` for unknown? Won't we now generate`@#unknown`?

hernanponcedeleon: Before  40f3b2f29ae1efff7acbe2f0f1f2ab8e3214f674 we would generate things like `P0unknown`, but after that commit, the change is useless so I will revert it

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
```diff
@@ -76,28 +76,29 @@ public void reachingDefinitionMustOverride() throws InvalidConfigurationExceptio
...
     private void reachingDefinitionMustOverride(ReachingDefinitionsAnalysis.Method method) throws InvalidConfigurationException {
         ProgramBuilder b = ProgramBuilder.forLanguage(SourceLanguage.LITMUS);
+        final int lineOfCode = -1;
         b.newThread(0);
         Register r0 = b.getOrNewRegister(0, "r0");
         Register r1 = b.getOrNewRegister(0, "r1");
         Register r2 = b.getOrNewRegister(0, "r2");
         Label alt = b.getOrCreateLabel(0, "alt");
-        b.addChild(0, newJump(b.newConstant(types.getBooleanType()), alt));
+        b.addChild(0, newJump(b.newConstant(types.getBooleanType()), alt), lineOfCode);
```
ThomasHaas: This makes me question if you really want to force line numbers in the program builder.
I can see that it makes sense to enforce it so that you find all usages in the Litmus parsers, but now that you have covered them, I think it makes sense to add an overload that does not add source information specifically programs that are generated programatically and not from parsing (like in all these tests here).

hernanponcedeleon: This was exactly what I did [before the last commit](https://github.com/hernanponcedeleon/Dat3M/blob/3314321ce237ca519e924cd56a60d0fc1a0fa670/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/ProgramBuilder.java#L199).  The problem is not the current state of the visitors, but the fact that if we do not enforce the method to use the `lineOfCode`, if we even extend those grammars or come up with new litmus formats (both situations are very realistic), we might silently add events without debug information. 

This is why I went with the solution of painfully add the lines to the current tests but make the code "bullet-proved" for the future.

ThomasHaas: I guess the fundamental question is whether the `ProgramBuilder` is to be interpreted as `LitmusProgramBuilder` and should only be used for parsed code but not for general-purpose program construction.
Either way, in a parser you could easily add a post-parsing pass that checks if all instructions have source location information and either give a warning or throw an error. This is a pass that you only need to write once for all litmus code and the parser should invoke it after parsing.
Another alternative is to have an explicit `addChildWithoutSourceLoc` method because then the caller cannot accidentally forget to add source information.

I just think that we should avoid using dummy source locations like `-1` if possible. For analysis/unit tests, it is not that big of a deal, but generally speaking, there is a clear difference between having no source information and having wrong source information :)

EDIT: Actually, I think there is also a general design problem if you need to explicitly add source information every single time you create an event.
For example, if the visitor method for events actually returned a `List<Event>` instead of adding the events by itself, you could do this:
```Java
for (Instruction instr : InstructionList) {
    List<Event> events = instr.visit(this);
    for (Event e : events) {
        e.addMetadata(new SourceLocation.Litmus(curThread.getName, instr.getStart().getLine())
        curThread.addEvent(e);
    }
}
```

A variant that is easier to implement in our current setting is to have variable `List<Event> generatedEvents` and `visitInstructionXYZ` just adds the generated events to that list, while the outer loop around the instructions resets the list between visiting of different instructions.
This is vaguely similar to what the LLVM parser does.

ThomasHaas (approved):
LGTM

PR #1008, opened 2026-03-13
ModifierTrait

xeren:
Adds more control over the used lattice in `InclusionBasedPointerAnalysis`.  More implementations can be added later.

- `--program.analysis.alias=EXPERIMENTAL_FIELD_INSENSITIVE` uses the singleton lattice $\lbrace\mathbb Z\rbrace$ implemented in `ModifierTrait.VoidTrait`.  It lets the analysis be truely field-insensitive.  It is not generally, but mainly, less precise than `FIELD_INSENSITIVE` implemented in `AndersenAliasAnalysis`.
- `--program.analysis.alias=EXPERIMENTAL_FIELD_SENSITIVE` uses the lattice $\mathbb F(\mathbb Z)+\lbrace\mathbb Z\rbrace$ of finite sets.  It is implemented in `ModifierTrait.Offsets` and comparable with `FIELD_INSENSITIVE` and `FIELD_SENSITIVE`.
- `--program.analysis.alias=EXPERIMENTAL_LINEAR_1D` uses $\lbrace n+\mathbb Zs\mid n,s\in\mathbb Z\rbrace$ in `ModifierTrait.SdLinear`.
- `--program.analysis.alias=FULL` remains using $\lbrace n+\sum_{i=0}^k\mathbb Zs_k\mid k\in\mathbb N,n,s_1,\ldots,s_k\in\mathbb Z\rbrace$ which has moved to `ModifierTrait.MdLinear`.


ThomasHaas:
Interesting changes, but why the changes to the IntervalAnalysis? It was a deliberate choice to make it `Optional`.
Apart from the fact that those changes shouldn't be part of this PR in the first place :)

xeren:
You are right, this had nothing to do there.

ThomasHaas:
Do you intend to do more changes in this PR? It is marked as DRAFT after all :)

xeren:
I needed to add some unit tests before removing the `[DRAFT]` from this PR.  Now that this is done, you can properly take a look at it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/ModifierTrait.java
```diff
@@ -0,0 +1,302 @@
...
+    /// Checks if `compose(m, modifier) == m` for all `m`.
+    boolean isIdentity(Modifier modifier);
+
+    /// Checks if there may be integers `x` and `y` with `x[left]y` and `x[right]y`.
+    /// This method must not have false negatives, but is allowed to have false positives.
+    boolean overlaps(Modifier left, Modifier right);
+
+    /// Checks if for all integers `x` and `y`, `x[smaller]y` must imply `x[larger]y`.
+    /// This method must not have false positives, but is allowed to have false negatives.
+    boolean includes(Modifier larger, Modifier smaller);
```
ThomasHaas, line 28: Maybe you should include the approximation direction in the name of the methods: 
`mayOverlap(...)` and `mustInclude(...)`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/ModifierTrait.java
```diff
@@ -0,0 +1,302 @@
...
+
+    /// Describes the transitive closure of `modifier`,
+    /// i.e. `{ (x,y_n) | exists n,y_1...y_n. x[modifier]y_1 ... [modifier]y_n }`.
+    Modifier accelerate(Modifier modifier);
+
+    /// Intersects `[modifier]` with `{ (x,y) | x <= y < x + objectSize }`.
+    /// If this intersection is empty, the associated access must be out-of-bounds.
+    /// <p>
+    /// This is an unsound strengthening that assumes no out-of-bounds accesses, given a memory object of known size.
+    Modifier postProcess(Modifier modifier, int objectSize);
```
ThomasHaas, line 58: The description forces `postProcess` to have a very concrete semantics despite its very generic name. I would call it `bound` or `restrict` or something like that.

ThomasHaas:
What is the correlation between the old alias analyses and the inclusion-based one with weaker modifier traits?
Do they coincide in terms of precision? What about computation time?

Also, do you think there are other reasonable instantiations of the `ModifierTrait` that could yield higher precisions? E.g., is the interface suitable to implement tracking of tagged pointers or is it too specialized to this kind of linear reasoning?

xeren:
Here are some summaries from our unit test suite.

**Table 1**: Alias methods
| ID | Name
| - | -
| A0 | FIELD_SENSITIVE
| A1 | FIELD_INSENSITIVE
| A2 | FULL
| A3 | EXPERIMENTAL_FIELD_INSENSITIVE
| A4 | EXPERIMENTAL_FIELD_SENSITIVE
| A5 | EXPERIMENTAL_LINEAR_1D

~**Table 2**~: Number of tests in `AbstractCTest` with at least one pair in `A.mayAlias` without `B.mayAlias`.
| A\B | A0 | A1 | A2 | A3 | A4 | A5
| - | - | - | - | - | - | - |
| A0 | | 18 | 44 | 0 | 40 | 44
| A1 | 150 | | 150 | 150 | 150 | 150
| A2 | 21 | 18 | | 0 | 0 | 0
| A3 | 284 | 256 | 284 | | 284 | 284
| A4 | 23 | 18 | 6 | 0 | | 6
| A5 | 21 | 18 | 0 | 0 | 0

~**Table 3**~: Number of tests in `AbstractCTest` with at least one pair in `A.mustAlias` without `B.mustAlias`.
| A\B | A0 | A1 | A2 | A3 | A4 | A6
| - | - | - | - | - | - | - |
| A0 | | 65 | 21 | 583 | 21 | 20
| A1 | 18 | | 18 | 583 | 18 | 18
| A2 | 80 | 124 | | 583 | 0 | 0
| A3 | 0 | 0 | 0 | | 0 | 0
| A4 | 80 | 124 | 0 | 583 | | 0
| A5 | 80 | 125 | 2 | 583 | 2

**Table 4**: Total analysis runtime for tests in `AbstractCTest` relative to `A2`.  *Note that I did only one test yet, so 'min' and 'max' deviate a lot.*
| A | min | avg | max
| - | - | - | -
| A0 | 4% | 52% | 600%
| A1 | 1% | 152% | 40170%
| A2 | 100% | 100% | 100%
| A3 | 11% | 85% | 1095%
| A4 | 8% | 73% | 195%
| A5 | 8% | 74% | 476%

ThomasHaas:
Oh well, I didn't expect such tables. It's interesting that A0 and A1 are sometimes more precise than all the other analyses (unsoundly so?).
How come that A5 is more precise than A2 in terms of must-aliases? I would expect `A3 < A4 < A5 < A2` in terms of precision for both may and must.

Btw. I think you should transpose one of the two tables, because right now an entry at (A/B) in the first table means how many times A was worse than B, but in the second table it means how many times A was better than B.
I would also order the rows based on expected precision within their groups: `A1, A0` and then `A3, A4, A5, A2`.
That being said, I'm not sure whether the tables are all that useful.

xeren:
I got the sides mixed-up exactly the wrong way.  The following tables are corrected and have your suggested ordering.

**~Table 2~** (Version 2): Number of tests in `AbstractCTest` with at least one pair in `A.mayAlias` without `B.mayAlias`.
| A\B | A1 | A0 | A3 | A4 | A5 | A2
| - | - | - | - | - | - | - |
| A1 | | 18 | 256 | 18 | 18 | 18
| A0 | 150 | | 284 | 23 | 21 | 21
| A3 | 150 | 0 | | 0 | 0 | 0
| A4 | 150 | 40 | 284 | | 0 | 0
| A5 | 150 | 44 | 284 | 6 | | 0
| A2 | 150 | 44 | 284 | 6 | 0 |

**~Table 3~** (Version 2): Number of tests in `AbstractCTest` with at least one pair in `B.mustAlias` without `A.mustAlias`.
| A\B | A1 | A0 | A3 | A4 | A5 | A2
| - | - | - | - | - | - | - |
| A1 | | 18 | 583 | 18 | 18 | 18
| A0 | 65 | | 583 | 21 | 20 | 21
| A3 | 0 | 0 | | 0 | 0 | 0
| A4 | 124 | 80 | 583 | | 0 | 0
| A5 | 125 | 80 | 583 | 2 | | 2!
| A2 | 124 | 80 | 583 | 0 | 0 |

The value 2! is unexpected and occurs in `cachedq-opt.ll` and `paper-E3.5.ll`.

xeren:
### Latest results:

**Table 2** (Version 3): Number of tests in `AbstractCTest` with at least one pair in `A.mayAlias` without `B.mayAlias`.  A value of 0 indicates that `B.mayAlias` may imply `A.mayAlias`.  The differences to the former version come from the change to `shrinkToBounds` / `postProcess`.
| A\B | 1 | 0 | 3 | 4 | 5 | 2
| - | - | - | - | - | - | -
| 1 |  | 18 | 256 | 18 | 18 | 18
| 0 | 150 |  | 284 | 23 | 21 | 21
| 3 | 150 | 0 |  | 0 | 0 | 0
| 4 | 150 | 40 | 284 |  | 0 | 0
| 5 | 150 | 44 | 284 | 6 |  | 0
| 2 | 150 | 63 | 284 | 25 | 19 |

**Table 3** (Version 3): Number of tests in `AbstractCTest` with at least one pair in `B.mustAlias` without `A.mustAlias`.  A value of 0 indicates that `A.mustAlias` may imply `B.mustAlias`.
| A\B | 1 | 0 | 3 | 4 | 5 | 2
| - | - | - | - | - | - | -
| 1 |  | 18 | 583 | 18 | 18 | 18
| 0 | 65 |  | 583 | 21 | 20 | 20
| 3 | 0 | 0 |  | 0 | 0 | 0
| 4 | 124 | 80 | 583 |  | 0 | 0
| 5 | 125 | 80 | 583 | 2 |  | 0
| 2 | 144 | 99 | 583 | 20 | 19 |

**Table 4** (Version 2): Alias analysis runtime in `AbstractCTest` relative to A2.  This was done with 10 iterations, so these results deviate less than before.  The max-outlier for A1 is the program `cachedq-opt.ll`.  `mcslock-opt.ll` and `rec_mcslock-opt.ll` are also slow for A1.  The max-outliers for A3 are the programs `ck_ebr.ll` and `progressUnfair.ll`.  The min-outliers are `recursion.ll`, `nondet_loop.ll`, `floats_2.ll`, `floats_15.ll`, but also `treiber.ll` for A1.
| A | total | min/test | avg/test | median/test | max/test
| - | - | - | - | - | -
| 1 | 668% | 2% | 146% | 58% | 33269%
| 0 | 83% | 3% | 56% | 53% | 487%
| 3 | 186% | 12% | 88% | 78% | 1296%
| 4 | 84% | 10% | 73% | 76% | 186%
| 5 | 83% | 9% | 72% | 76% | 369%
| 2 | 100% | 100% | 100% | 100% | 100%

**Table 5** (Version 1): Verification runtime in `AbstractCTest` relative to A2.  One iteration again, thus tests with `Fail` verdicts like `safe_stack.ll` play a dominant part for the total runtime.  Maybe I can filter them out.
| A | Eager | min/test | avg/test | median/test | max/test | Lazy | min/test | avg/test | median/test | max/test
| - | - | - | - | - | - | - | - | - | - | -
| 1 | 112% | 36% | 101% | 100% | 317% | 86% | 31% | 100% | 99% | 279%
| 0 | 122% | 24% | 102% | 100% | 453% | 87% | 27% | 101% | 100% | 278%
| 3 | 108% | 33% | 101% | 100% | 317% | 95% | 27% | 100% | 99% | 216%
| 4 | 107% | 49% | 101% | 100% | 344% | 92% | 34% | 101% | 100% | 298%
| 5 | 101% | 25% | 101% | 100% | 327% | 104% | 27% | 100% | 99% | 335%
| 2 | 100% | 100% | 100% | 100% | 100% | 100% | 100% | 100% | 100% | 100%

ThomasHaas:
Yeah, you have to avoid `safe_stack`. Anyway, I believe your current FULL alias analysis is very good and so I don't think the other one's really matter too much (it's good to have them without additional maintenance overhead though).

What I think is far more interesting is whether you think your new `ModifierTrait` interface is able to represent even more accurate analyses. For example, is it imaginable to capture pointer tagging with this (i.e., writing to the lowest/highest bits of pointers and undoing that before using the pointers)?

hernanponcedeleon:
I haven't checked the code yet ... but if I understand the tables correctly, the new analysis are better than `FIELD_SENSITIVE` and `FIELD_INSENSITIVE` and at the same time they fit in the "framework" that allows to just change the lattice.

If this is the case, I would go all to way and get rid of the old analysis to reduce code maintenance.

xeren:
> What I think is far more interesting is whether you think your new ModifierTrait interface is able to represent even more accurate analyses. For example, is it imaginable to capture pointer tagging with this (i.e., writing to the lowest/highest bits of pointers and undoing that before using the pointers)?

It should get easier to implement than before.  If you want to assume correct usage of *Pointer Tagging* for some specific number of lower/higher bits, you need to extend `InclusionBasedPointerAnalysis.Collector`, at least.  To model incorrect use (dereference with uncleared tags), you would also need to add another `ModifierTrait<TaggedModifier>` for some `TaggedModifier` and extend the interface.

There might be hidden issues with the composition order, as it is currently commutative for all traits.

> if I understand the tables correctly, the new analysis are better than `FIELD_SENSITIVE` and `FIELD_INSENSITIVE` and at the same time they fit in the "framework" that allows to just change the lattice.

The new analyses try to model the internal representation of the old analyses inside the framework of `FULL`.  `FIELD_INSENSITIVE` (A1) primarily uses the same as (A3), with a bit of (A4) post-analysis.  `FIELD_SENSITIVE` (A0) has (A4) for the address sets and (A5) for constraints.

> If this is the case, I would go all to way and get rid of the old analysis to reduce code maintenance.

That is possible.  I take naming suggestions:
- `EXPERIMENTAL_FIELD_INSENSITIVE` $\to$ (new) `FIELD_INSENSITIVE`
- `EXPERIMENTAL_FIELD_SENSITIVE` $\to$ `FINITE_FIELDS`
- `EXPERIMENTAL_LINEAR_1D` $\to$ `LINEAR_FIELDS` aliasing `FIELD_SENSITIVE`
- `FULL` $\to$ `LINEAR_MD_FIELDS` aliasing `FULL` for compatibility.

ThomasHaas:
All you said sounds good to me. We can get rid of the old alias analyses and rename the current one's.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
```diff
@@ -77,9 +73,22 @@ static AliasAnalysis fromConfig(Program program, Context analysisContext, Config
...
+            if (methodString == null) {
+                return;
+            }
+            methodString = methodString.toUpperCase();
+            method = switch (methodString) {
+                case "FULL" -> Alias.LINEAR_MD_FIELDS;
+                case "FIELD_SENSITIVE" -> Alias.LINEAR_FIELDS;
+                default -> Alias.valueOf(methodString);
+            };
+        }
```
ThomasHaas: I'm not sure if we need this. 

@hernanponcedeleon do you think any user relies on the current names of the option values of AA?
This PR changes the names of the options but still allows the usage of the old names in addition to the new one's. In the UI, however, we will only see the new names. Should we drop the old names entirely?
We could keep "FULL" as an alias for the most precise AA we have, i.e., if we add more precise alias analyses in the future, "FULL" will automatically refer to the new one.
I would still drop "FIELD_SENSITIVE" though.

ThomasHaas (commented):
The PR looks good to me.
From my side, the only discussion point left is the naming of the alias analysis options.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
hernanponcedeleon: I am not aware of any use case where an AA other than default is used. I think we can drop the old names

hernanponcedeleon:
The native build is currently broken
```
Exception in thread "main" org.graalvm.nativeimage.MissingReflectionRegistrationError: The program tried to reflectively invoke method private void com.dat3m.dartagnan.program.analysis.alias.AliasAnalysis$Config.setMethod(java.lang.String) without it being registered for runtime reflection. Add private void com.dat3m.dartagnan.program.analysis.alias.AliasAnalysis$Config.setMethod(java.lang.String) to the reflection metadata to solve this problem. See https://www.graalvm.org/latest/reference-manual/native-image/metadata/#reflection for help.
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.reflect.MissingReflectionRegistrationUtils.forQueriedOnlyExecutable(MissingReflectionRegistrationUtils.java:72)
        at java.base@21.0.2/java.lang.reflect.Method.acquireMethodAccessor(Method.java:77)
        at java.base@21.0.2/java.lang.reflect.Method.invoke(Method.java:577)
        at org.sosy_lab.common.configuration.Configuration.setOptionValueForMethod(Configuration.java:652)
        at org.sosy_lab.common.configuration.Configuration.inject(Configuration.java:488)
        at org.sosy_lab.common.configuration.Configuration.inject(Configuration.java:435)
        at org.sosy_lab.common.configuration.Configuration.inject(Configuration.java:410)
        at com.dat3m.dartagnan.program.analysis.alias.AliasAnalysis$Config.<init>(AliasAnalysis.java:122)
        at com.dat3m.dartagnan.program.analysis.alias.AliasAnalysis.fromConfig(AliasAnalysis.java:57)
        at com.dat3m.dartagnan.verification.solving.ModelChecker.performStaticProgramAnalyses(ModelChecker.java:187)
        at com.dat3m.dartagnan.verification.solving.RefinementSolver.runInternal(RefinementSolver.java:202)
        at com.dat3m.dartagnan.verification.solving.ModelChecker.run(ModelChecker.java:117)
        at com.dat3m.dartagnan.verification.TaskSolver.run(TaskSolver.java:112)
        at com.dat3m.dartagnan.Dartagnan.main(Dartagnan.java:97)
        at java.base@21.0.2/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -109,28 +113,27 @@ public class InclusionBasedPointerAnalysis implements AliasAnalysis {
...
-    public static InclusionBasedPointerAnalysis fromConfig(Program program, Context analysisContext, AliasAnalysis.Config config) {
+    public static InclusionBasedPointerAnalysis<?> fromConfig(Program program, Context analysisContext, AliasAnalysis.Config config) {
         final ReachingDefinitionsAnalysis def = analysisContext.requires(ReachingDefinitionsAnalysis.class);
-        final var analysis = new InclusionBasedPointerAnalysis(program, def);
+        final var analysis = switch (config.method) {
+            case FIELD_INSENSITIVE -> new InclusionBasedPointerAnalysis<>(new ModifierTrait.VoidTrait(), program, def);
+            case FINITE_FIELDS -> new InclusionBasedPointerAnalysis<>(new ModifierTrait.Offsets(), program, def);
+            case LINEAR_FIELDS -> new InclusionBasedPointerAnalysis<>(new ModifierTrait.SdLinear(), program, def);
+            case LINEAR_MD_FIELDS -> new InclusionBasedPointerAnalysis<>(new ModifierTrait.MdLinear(), program, def);
+        };
```
hernanponcedeleon: I would rather make the switch on the trait and then unconditionally set `analysis = new InclusionBasedPointerAnalysis<>(trait, program, def);`

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/InclusionBasedPointerAnalysis.java
```diff
@@ -725,45 +695,48 @@ private boolean addInto(List<IncludeEdge> list, IncludeEdge element, boolean isG
...
         if (isGraphModification) {
-            addIntoGraphSucceesses++;
+            addIntoGraphSuccesses++;
         } else {
             addIntoCyclesSuccesses++;
         }
-        list.removeIf(o -> element.source.equals(o.source) && includes(element.modifier, o.modifier));
+        list.removeIf(o -> {
+            if (!element.source.equals(o.source)) return false;
+            return trait.mustInclude(element.modifier, o.modifier);
```
hernanponcedeleon:
```suggestion
            return !element.source.equals(o.source) ?  false : trait.mustInclude(element.modifier, o.modifier);
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/alias/AliasAnalysis.java
hernanponcedeleon: @xeren this discussion still remains open

xeren: I am fine with removing them.  I had kept the old option names just in case that I missed something outside this project that relied on them.

hernanponcedeleon:
@xeren can you please fix the conflicts and rebase?

xeren:
The CI got unlucky on test `com.dat3m.dartagnan.llvm.LFDSTest.testRefinement[32: safe_stack, target=TSO]` (verdict: `FAIL` / satisfiable), with a timeout after 16:40 min.  I repeated it multiple times on my machine with runtimes spanning from 1:51 min to 10:58 min.  The difference to current `development` is slim, as the alias analysis takes only 60ms and produces equal results.

I could increase the timeout for the test, but IMO this has nothing to do with this PR.

To check equality, I compared the results of the following snippet in both `development` and `alias-modifiertrait`:
```java
    private static String detailString(AliasAnalysis alias, Program program) {
        final List<MemoryCoreEvent> events = program.getThreadEvents(MemoryCoreEvent.class);
        final List<MemoryObject> objects = List.copyOf(program.getMemory().getObjects());
        final List<StringBuilder> message = new ArrayList<>();
        final TypeFactory types = TypeFactory.getInstance();
        final int maxAccessSize = events.stream()
                .mapToInt(e -> types.getMemorySizeInBytes(e.getAccessType()))
                .max().orElse(0);
        for (MemoryCoreEvent event : events) {
            final var b = new StringBuilder();
            for (MemoryCoreEvent other : events) {
                final boolean may = alias.mayAlias(event, other);
                final boolean must = alias.mustAlias(event, other);
                b.append(may ? must ? '1' : '0' : must ? '2' : ' ');
            }
            b.append('\t');
            final Set<MemoryObject> addressable = Set.copyOf(alias.addressableObjects(event));
            final Set<MemoryObject> communicable = Set.copyOf(alias.communicableObjects(event));
            for (MemoryObject object : objects) {
                final boolean a = addressable.contains(object);
                final boolean c = communicable.contains(object);
                b.append(a ? c ? 'B' : 'A' : c ? 'C' : ' ');
            }
            b.append('\t');
            final Set<Integer> tearingPoints = Set.copyOf(alias.mayMixedSizeAccesses(event));
            for (int i = 0; i < maxAccessSize; i++) {
                b.append(tearingPoints.contains(i) ? '|' : ' ');
            }
            b.append('\t').append(event);
            message.add(b);
        }
        return String.join(System.lineSeparator(), message);
    }
```

hernanponcedeleon:
It is not too uncommon to get `safe_stack` timeout in some PR. 

As you said, this has never anything to do with the changes in the PR, but simply with the fact that the encoding is SAT and there are inherently differences in solving times from run to run.

The must/may sets of `rf`/`co` (which can get affected by the alias analysis) match between this branch and `development`
```
        #must|may rf edges: 3|3182
        #must|may co edges: 246|3138
```
so, I think there is nothing to worry about.

PR #1009, opened 2026-03-13
Refactor EventFactory, Compilers and Parsers

xeren:
Mostly cleans up some imports, but also
- outlines some memory barrier generations in compilers.
- outlines uses of `ProgramBuilder.addChild` in litmus parsers.
- integrates `FenceNameRepository` into `EventFactory`

#### Added
- `AArch64.newRMWLoadExclusive`, `AArch64.newRMWStoreExclusive`
- `X86.newLoadFence`, `newStoreFence` (still not implemented, but merges two places)
- `RISCV.newRMWLoadExclusive`
- `Power.newRMWLoadExclusive`

#### Modified
- `newRMWStoreExclusive`, `newRMWStoreExclusiveWithMo`
- `AArch64.newDmbBarrier`, etc.
- `AArch64.newBarrier(String,String)` (moved from `newFenceOpt`)
- `RISCV.newRMWStoreConditional`
- `newFakeCtrlDep(Register)` (returns `List<Event>` and does no longer accept a `Label`)


hernanponcedeleon:
@xeren can you please rebase and fix the conflicts?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -215,12 +212,11 @@ protected List<Event> newLlvmCmpXchg(Register oldValue, Register success, Expres
...
     @Override
     public List<Event> visitLlvmFence(LlvmFence e) {
-        String mo = e.getMo();
-        Event fence = mo.equals(C11.MO_RELEASE) || mo.equals(C11.MO_ACQUIRE_RELEASE) || mo.equals(C11.MO_SC) ?
-                AArch64.DMB.newISHBarrier() :
-                mo.equals(C11.MO_ACQUIRE) ?
-                        AArch64.DSB.newISHLDBarrier() :
-                        null;
+        Event fence = switch (e.getMo()) {
+            case Tag.C11.MO_RELEASE, Tag.C11.MO_SC -> AArch64.newDmbIshBarrier();
```
hernanponcedeleon: You are missing the `mo.equals(C11.MO_ACQUIRE_RELEASE)` case

hernanponcedeleon (commented):
Why are you adding `newRMWLoadExclusive` for each arch if they all end up calling the general case? While I could see an argument in the factory for having the method for each arch, we do have other "common" events that use a single method, e.g. `newRMWStoreExclusiveWithMo`

ThomasHaas:
> Why are you adding `newRMWLoadExclusive` for each arch if they all end up calling the general case? While I could see an argument in the factory for having the method for each arch, we do have other "common" events that use a single method, e.g. `newRMWStoreExclusiveWithMo`

I agree with you but maybe from a different point of view. I think it becomes more and more unclear of whether a `XYZ.newEvent` factory method generates a core event or not. For parsers, you want to NOT generate core events to support compilation(*), but for the `Compilation` passes you need to ensure to generate only core events.
I think this is particularly true for the `C11.newThreadFence` events which should NOT be called by the parser (for the most part) and only during Compilation. And those can reasonably well be moved in the dedicated compilation pass.
For example, the `LKMM` compiler has some helper methods for the lowering that are not otherwise exposed to parsers/other callers.

(*) It is fine to generate core events from the parser, but not if you intend to support compilation to other targets on those events (at least in our current architecture).

xeren:
With the different `newRMWLoadExclusive` methods, I wanted to group all the usages in the program target-wise.  It should have made them easier to be altered per-architecture.  Ideally, `Common`, `newRMWLoadExclusive`, `newLoadWithMo`, `newStoreWithMo` and `newFence` are never used directly by parsers.  As you noticed, my refactoring was incomplete.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -266,311 +264,307 @@ public Object visitWhileExpression(LitmusCParser.WhileExpressionContext ctx) {
...
-        programBuilder.addChild(currentThread, event, ctx.getStart().getLine());
+        return addCmpXchg(register, address, expectedAdd, value, C11.DEFAULT_MO, false, null, ctx);
+    }
+
+    private Expression addCmpXchg(Register register, Expression address, Expression expectedAddress, Expression value,
+            String mo, boolean strong, OpenCLScopeContext scope, ParserRuleContext ctx) {
+        final Event event = EventFactory.Atomic.newCompareExchange(register, address, expectedAddress, value, mo, strong);
+        addScopeTag(event, scope);
+        append(event, ctx);
         return register;
```
ThomasHaas: I suggest to not return the `register` as it is totally unclear why this function should do that.
I would rather have the call-sites explicitly returning the register.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
```diff
@@ -722,4 +716,9 @@ private void addScopeTag(Event event, LitmusCParser.OpenCLScopeContext ctx) {
             }
         }
     }
+
+    private Object append(Event event, ParserRuleContext ctx) {
+        final int line = ctx.getStart().getLine();
+        return programBuilder.addChild(currentThread, event, line);
+    }
```
ThomasHaas: The return type should be as specific as possible, i.e., `Event`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusPPC.java
```diff
@@ -106,118 +104,123 @@ public Object visitInstructionRow(LitmusPPCParser.InstructionRowContext ctx) {
...
 
     @Override
-    public Register visitRegister(LitmusPPCParser.RegisterContext ctx) {
+    public Register visitRegister(RegisterContext ctx) {
         return programBuilder.getOrNewRegister(mainThread, ctx.getText(), archType);
     }
 
+    private Object append(Event event, ParserRuleContext ctx) {
+        return programBuilder.addChild(mainThread, event, ctx.getStart().getLine());
+    }
```
ThomasHaas: Return type `Event`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusPTX.java
```diff
@@ -314,27 +309,30 @@ public Object visitBarrier(LitmusPTXParser.BarrierContext ctx) {
...
-    public Object visitJump(LitmusPTXParser.JumpContext ctx) {
+    public Object visitJump(JumpContext ctx) {
         Label label = programBuilder.getOrCreateLabel(mainThread, ctx.Label().getText());
-        return programBuilder.addChild(mainThread, EventFactory.newGoto(label), ctx.getStart().getLine());
+        return append(EventFactory.newGoto(label), ctx);
     }
 
+    private Object append(Event event, ParserRuleContext ctx) {
+        return programBuilder.addChild(mainThread, event, ctx.getStart().getLine());
+    }
```
ThomasHaas: Return type `Event`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusRISCV.java
```diff
@@ -252,30 +238,34 @@ public Object visitFence(LitmusRISCVParser.FenceContext ctx) {
...
-	private String getMo(LitmusRISCVParser.MoRISCVContext mo1, LitmusRISCVParser.MoRISCVContext mo2) {
+	private String getMo(MoRISCVContext mo1, MoRISCVContext mo2) {
 		String moR = mo1 != null ? mo1.mo : "";
 		String moW = mo2 != null ? mo2.mo : "";
 		return !moR.isEmpty() ? (!moW.isEmpty() ? Tag.RISCV.MO_ACQ_REL : moR) : moW;
 	}
+
+    private Object append(Event event, ParserRuleContext ctx) {
+        return programBuilder.addChild(mainThread, event, ctx.getStart().getLine());
+    }
```
ThomasHaas: Return type `Event`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusVulkan.java
```diff
@@ -325,19 +324,23 @@ public Object visitDeviceOperation(LitmusVulkanParser.DeviceOperationContext ctx
...
-        if (ctx.getRuleContext(LitmusVulkanParser.MoAcqRelContext.class, 0) != null) {
+        if (ctx.getRuleContext(MoAcqRelContext.class, 0) != null) {
             return Tag.Vulkan.ACQ_REL;
         }
         return defaultMo;
     }
+
+    private Object append(Event event, ParserRuleContext ctx) {
+        return programBuilder.addChild(mainThread, event, ctx.getStart().getLine());
+    }
```
ThomasHaas: Return type `Event`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusVulkan.java
```diff
@@ -233,29 +235,26 @@ public Object visitAtomicRmwInstruction(LitmusVulkanParser.AtomicRmwInstructionC
...
+    public Object visitMemoryBarrierInstruction(MemoryBarrierInstructionContext ctx) {
+        final String scope = ctx.scope().content;
+        final List<String> semantics = ctx.semSc().stream().map(c -> c.content).toList();
+        final boolean av = ctx.semAv() != null;
+        final boolean vis = ctx.semVis() != null;
+        final Event fence = switch (getMemoryOrderOrDefault(ctx, null)) {
+            case Tag.Vulkan.ACQUIRE -> EventFactory.Vulkan.newAcqBarrier(scope, semantics, vis);
+            case Tag.Vulkan.RELEASE -> EventFactory.Vulkan.newRelBarrier(scope, semantics, av);
+            case Tag.Vulkan.ACQ_REL -> EventFactory.Vulkan.newAcqRelBarrier(scope, semantics, av, vis);
+            default -> throw new ParsingException("Unknown barrier type");
```
ThomasHaas: On first sight, it looks like the semantics has slightly changed because `mo=null` is now rejected, but was permitted before. However, the grammar seems to enforce `mo != null` so the semantics is the same.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusX86.java
```diff
@@ -101,82 +98,82 @@ public Object visitInstructionRow(LitmusX86Parser.InstructionRowContext ctx) {
...
+            case "lfence" -> EventFactory.X86.newLoadFence();
+            case "sfence" -> EventFactory.X86.newStoreFence();
+            default -> throw new ParsingException("Unrecognised fence '%s'", ctx.getText());
+        };
+        return append(fence, ctx);
+    }
+
+    private Object append(Event event, ParserRuleContext ctx) {
+        return programBuilder.addChild(mainThread, event, ctx.getStart().getLine());
     }
```
ThomasHaas: Return type `Event`

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -642,6 +669,9 @@ public static GenericMemoryEvent newSrcuSync(Expression address) {
             return srcuSync;
         }
 
+        public static GenericVisibleEvent newFence(String fenceType) {
+            return EventFactory.newFence(fenceType);
+        }
```
ThomasHaas: I'm not sure about this one. I can see the point somewhat, but I would at least rename the function to `newBarrier` to align with the nomenclature of the kernel.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -807,6 +830,27 @@ public static VulkanCmpXchg newVulkanCmpXchg(Expression address, Register regist
             return new VulkanCmpXchg(register, address, expected, value, mo, scope);
         }
 
+        public static GenericVisibleEvent newAcqBarrier(String scope, List<String> semantics, boolean visible) {
+            final GenericVisibleEvent barrier = newFence(Tag.FENCE);
```
ThomasHaas: I understand that the previous code was already using `newFence(Tag.FENCE)`, but maybe we can fix this here? I would use a generic visible event with name `Membar`. If one wants to be more precise, we could even add the mo/scope/semantics/visibility tags.

For reference, in litmus vulkan, a memory barrier looks like this:
```
membar.rel.dv.semsc0.semav
// membar.<mo>.<scope>.<semantics>.<vis/avail>
```
There is one theoretical disadvantage, namely, if the tagging of the event changes for whatever reason, the detailed name becomes wrong. On the other hand, there is no reason to change the tagging later on really.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -467,6 +463,34 @@ public static AtomicXchg newExchange(Register register, Expression address, Expr
...
+            return newThreadFence(Tag.C11.MO_RELEASE);
+        }
+
+        public static GenericVisibleEvent newThreadFenceAcquireRelease() {
+            return newThreadFence(Tag.C11.MO_ACQUIRE_RELEASE);
+        }
+
+        public static GenericVisibleEvent newThreadFenceSequentiallyConsistent() {
+            return newThreadFence(Tag.C11.MO_SC);
+        }
```
ThomasHaas: I don't see much benefit in these 4 functions. They are not really much cleaner just than calling `newThreadFence(Tag.C11.MO_XYZ)`

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
```diff
@@ -252,15 +249,11 @@ public List<Event> visitAtomicCmpXchg(AtomicCmpXchg e) {
...
-        if (e.isWeak()) {
-            Register statusReg = e.getFunction().newRegister("status(" + e.getLocalId() + ")", types.getBooleanType());
-            optionalExecStatus = newExecutionStatus(statusReg, storeValue);
-            optionalUpdateCasCmpResult = newLocal(booleanResultRegister, expressions.makeNot(statusReg));
-        }
+        Load loadValue = newRMWLoadExclusive(regValue, address, extractLoadMoFromCMo(mo));
+        Store storeValue = newRMWStoreExclusive(address, value, strong, extractStoreMoFromCMo(mo));
+        Register statusReg = strong ? null : e.getFunction().newRegister("status(" + e.getLocalId() + ")", types.getBooleanType());
+        ExecutionStatus optionalExecStatus = strong ? null : newExecutionStatus(statusReg, storeValue);
+        Local optionalUpdateCasCmpResult = strong ? null : newLocal(booleanResultRegister, expressions.makeNot(statusReg));
```
ThomasHaas: Just FYI, `eventSequence` can actually take nested lists as input and it will flatten them.
So one could write:
```Java
List<Event> optional = strong ? List.of() : List.of(
    newExectutionStatus(...),
    newLocal(...)
);
```
and then use the optional list as input to `eventSequence`. I don't know whether it's cleaner though.

Actually, I originally added this feature because I wanted to use it for fake control dependencies, where `newFakeCtrlDep(Register reg)` would return both the jump and the label... Well, we never used this feature and I believe that, except for me, probably no one knew that `eventSequence` could do this :P.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
```diff
@@ -100,24 +85,13 @@ public List<Event> visitLlvmLoad(LlvmLoad e) {
 
     @Override
     public List<Event> visitLlvmStore(LlvmStore e) {
-        Event optionalBarrierBefore = null;
+        Event optionalBarrierBefore = switch (e.getMo()) {
+            case C11.MO_SC -> leadingSync() ? newSync() : newLwSync();
+            case C11.MO_RELEASE -> newLwSync();
+            default -> null;
+        };
```
ThomasHaas: You added `optionalBarrierBefore/After` methods that you could use here.
I think you decided halfway during the changes to add the helper function but forgot to use it for the already changed code :P

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -574,21 +571,23 @@ public List<Event> visitLKMMOpAndTest(LKMMOpAndTest e) {
...
         Expression testResult = expressions.makeNot(expressions.makeBooleanCast(dummy));
 
-        Load load = newRMWLoadExclusive(dummy, address); // TODO: No mo on the load?
+        // TODO: No mo on the load?
+        String storeMo = mo.equals(Tag.Linux.MO_MB) ? MO_REL : "";
+
+        Load load = newRMWLoadExclusive(dummy, address, "");
         Local localOp = newLocal(dummy, expressions.makeIntBinary(dummy, e.getOperator(), e.getOperand()));
-        Store store = newRMWStoreExclusiveWithMo(address, dummy, true, mo.equals(Tag.Linux.MO_MB) ? Tag.RISCV.MO_REL : "");
+        Store store = newRMWStoreExclusiveWithMo(address, dummy, true, false, storeMo);
```
ThomasHaas: This should be a `newRMWStoreConditional`, no?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -603,18 +602,67 @@ public List<Event> visitLKMMLock(LKMMLock e) {
         // https://github.com/westerndigitalcorporation/RISC-V-Linux/blob/master/linux/arch/riscv/include/asm/spinlock.h
         // We replace AMO instructions with LL/SC
         return eventSequence(
-                newRMWLoadExclusive(dummy, e.getLock()),
+                newRMWLoadExclusive(dummy, e.getLock(), ""),
                 newAssume(expressions.makeEQ(dummy, zero)),
-                newRMWStoreExclusive(e.getLock(), one, true),
-                RISCV.newRRWFence()
+                newRMWStoreExclusive(e.getLock(), one, true, false),
```
ThomasHaas: Same here: conditional store

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
xeren: This and `C11.newThreadFence(String)` were intended to mimic the barrier constructors of other architectures.  At some point I wanted to outline all direct calls to `EventFactory.newFence(String)`.
In retrospect, that was not a good idea.  This method should probably be in `VisitorLKMM` as it is the only place in this project that produces core fences for `linux-kernel.cat`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
xeren: I remove them again.
They and `newThreadFence(String)` are also core event constructors like `Linux.newFence(String)` and should remain in `VisitorC11` and `VisitorIMM`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
xeren: As a private method, `addCmpXchg` can be tailored to the four call sites.  I would have preferred to merge the contexts in `LitmusC.g4` instead, but did not want to modify grammars in this PR.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
xeren: The barriers for loads and stores do not match with those for RMWs.  I could not do an outlining here.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
xeren: I think so.  This and the other line had `requiresMatchingAddresses=false`.  It became visible when I removed the overloaded method in `EventFactory`.  The PR fixes this bug now.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusC.java
ThomasHaas: It's not a matter of the method being private or not. Most visitor methods had the pattern `append(event); return register;` or `return append(event);`. But `return addCmpXchg(...)` looked like a variant of the latter `return append(...)` while behaving like the former `append(...); return register`.
Also, from a developer's perspective, private and public methods are not so different, both need to be easily readable :)

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorPower.java
ThomasHaas: I don't understand. The barriers you use in `visitLlvmStore` and `visitLlvmLoad` are exactly those that `optionalBarrierBefore/After` produce.
For comparison:
```Java
       // Llvm store...
        Event optionalBarrierBefore = switch (e.getMo()) {
            case C11.MO_SC -> leadingSync() ? newSync() : newLwSync();
            case C11.MO_RELEASE -> newLwSync();
            default -> null;
        };

      // ... vs. helper method
      private Event optionalBarrierBefore(String mo) {
        return switch (mo) {
            case C11.MO_SC -> leadingSync() ? newSync() : newLwSync();
            case C11.MO_RELEASE, C11.MO_ACQUIRE_RELEASE -> newLwSync();
            default -> null;
        };
    }
```
The only difference is that the helper method also matches ACQ_REL, but that should be irrelevant for stores/loads.

xeren: Semantically, you are right wrt. half of the expressions.

| where | seq_cst + leading sync | seq_cst | acquire | release |
| - | - | - | - | - |
| `optionalBarrierBefore` | `sync` | `lwsync` | - | `lwsync` |
| `LlvmLoad` | `sync` | - | - | - |
| `LlvmStore` | `sync` | `lwsync` | - | `lwsync` |
| `AtomicLoad` | `sync` | - | - | - |
| `AtomicStore` | `sync` | `lwsync` | - | `lwsync` |
| `optionalBarrierAfter` | `isync` | `sync` | `isync` | - |
| `LlvmLoad` | `isync` | `sync` | `isync` | - |
| `LlvmStore` | - | `sync` | - | - |
| `AtomicLoad` | `isync` | `sync` | `isync` | - |
| `AtomicStore` | - | `sync` | - | - |

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorArm8.java
xeren: That is a useful transformation.  I have applied it to `EventFactory.newFakeCtrlDep()` as well, so it creates the label itself.

xeren:
In the latest commit I removed the individual barrier constructors in `EventFactory`, like `AArch64.newDmbIshBarrier()`.  There is no use in having them around.  The only users were the respective compilers, where I added individual methods in earlier commits.

From my side, this PR should be ready to merge.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -656,8 +757,13 @@ public static TSOXchg newExchange(MemoryObject address, Register register) {
             return new TSOXchg(address, register);
         }
 
-        public static GenericVisibleEvent newMemoryFence() {
-            return newFence(MFENCE);
+        public static Event newMemoryFence(String type) {
+            final String name = switch (type.toLowerCase()) {
+                case "mfence" -> "mfence";
+                case "lfence", "sfence" -> throw new UnsupportedOperationException("X86 fence '%s'.".formatted(type));
```
ThomasHaas, line 764: I'm not convinced about this. We support those fences by simply not throwing an exception :).
It was fair when it was a `ParsingException` in the parser, but having this exception here is odd IMO.

In general, moving checks from the parsers into the EventFactory causes previous `ParsingExceptions` to become `IllegalArgumentExceptions` and similar which I think is not so clean.

xeren: There are a lot of `ParsingException` with a `"No support"` message in the project.  How about making those `UnsupportedOperation` too and wrapping them in `ProgramParser` when they occur?  We would event catch some more exceptions like `IllegalState: "Register 0:x is not initialised"`.
```java
try {
    program = parser.parse(charStream);
} catch (ParsingException | MalformedProgramException exception) {
    throw exception;
} catch (RuntimeException exception) {
    throw new ParsingException(exception, exception.getMessage());
}
```

ThomasHaas: I guess that is a possibility.

xeren: Then again, those are supposed to be `IllegalStateException`, as stated in `ExceptionsTest`.

ThomasHaas: I think you can change the `ExceptionsTest` to expect `ParsingException` instead.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -527,53 +527,159 @@ public static class AArch64 {
         private AArch64() {
         }
 
-        public static class DMB {
-            private DMB() {
-            }
+        public enum MemoryOrder {
+            RELAXED, ACQUIRE, RELEASE, ACQ_REL
+        }
```
ThomasHaas, line 532: I would rename `RELAXED` to `PLAIN` (maybe even `NONE`?) as it better conveys that there is no tagging generated. Same for the `RISCV` case.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusRISCV.java
```diff
@@ -103,179 +102,158 @@ public Object visitInstructionRow(LitmusRISCVParser.InstructionRowContext ctx) {
...
-		return !moR.isEmpty() ? (!moW.isEmpty() ? Tag.RISCV.MO_ACQ_REL : moR) : moW;
+	private EventFactory.RISCV.MemoryOrder getMo(List<MoRISCVContext> mo) {
+        boolean acq = false;
+        boolean rel = false;
+        for (MoRISCVContext ctx : mo) {
+            acq |= ctx.Acq() != null;
+            rel |= ctx.Rel() != null;
+        }
+		return acq ? rel ? EventFactory.RISCV.MemoryOrder.ACQ_REL : EventFactory.RISCV.MemoryOrder.ACQUIRE
+                : rel ? EventFactory.RISCV.MemoryOrder.RELEASE : EventFactory.RISCV.MemoryOrder.RELAXED;
```
ThomasHaas: Maybe use a static import.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
```diff
@@ -157,8 +157,8 @@ public List<Event> visitLlvmXchg(LlvmXchg e) {
         Expression address = e.getAddress();
         String mo = e.getMo();
 
-        Load load = newRMWLoadExclusiveWithMo(resultRegister, address, Tag.C11.loadMO(mo));
-        Store store = newRMWStoreExclusiveWithMo(address, e.getValue(), true, Tag.C11.storeMO(mo));
+        Load load = newRMWLoadExclusiveWithMo(resultRegister, address, loadMO(mo));
+        Store store = newRMWStoreExclusiveWithMo(address, e.getValue(), true, false, storeMO(mo));
```
ThomasHaas: Maybe we can sneak a fix into this PR :). I think the C11 compilation should try to stick to either LX/SX-based RMWs or AMO-based RMWs in the compilation.
Personally, I prefer the AMO variants though we need to use LX/SX for the weak CAS.
So either we make everything LX/SX or we make everything AMO with an exception for spuriously failing events, i.e., (weak) CAS (I would also compile strong CAS to LX/SX).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
```diff
@@ -603,18 +584,70 @@ public List<Event> visitLKMMLock(LKMMLock e) {
...
+
+    private Load newRMWLoadExclusive(Register value, Expression address, String mo) {
+        return newRMWLoadExclusiveWithMo(value, address, mo);
+    }
+
+    private Store newRMWStoreConditional(Expression address, Expression value, boolean strong, String mo) {
+        Store store = newRMWStoreExclusiveWithMo(address, value, strong, true, mo);
+        store.addTags(STCOND);
+        return store;
+    }
```
ThomasHaas: We should name both methods according to either
(i) our core events (LX/SX)
or (ii) the RISCV instructions (load-reserved/store-conditional).
But mixing both is a bit ugly :)

I guess the same argument can be made in the case of Power.

xeren: I'd prefer (i) in the compilers and (ii) in the parsers.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
xeren: This can happen in another PR.  I know that X86 and RISC-V have this issue as well.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
ThomasHaas: For the parsers I 100% agree. For the compilers, I'm not sure. I think you could also go with `newRMWStoreConditionalAsCore` to capture both the fact that you want to express a`RMWStoreConditional` but also the fact that it is implemented via core events.
I mean, you cannot call it `newRMWStoreExclusive` (following the core) because it name-clashes with the core factory methods but it also conveys less information (the method name would be the same in all compilers, but each of them might work slightly differently).

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
ThomasHaas: I don't see the issue with those two. `X86` consistenly uses AMOs since it has no LX/SX, and RISC-V consistently uses LX/SX, though it technically also supports AMOs.
I'm pretty sure that all hardware compilers are consistent.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorRISCV.java
xeren: Let's call them `newCore{LoadReserved|StoreConditional}` for RISC-V and Power and `newCore{Load|Store}Exclusive` for ARM.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/compilation/VisitorC11.java
xeren: I looked at the wrong file, I guess.  It was IMM, not X86.  For RISC-V, we have todos indicating that AMOs are intended once they are supported, which they already are.

ThomasHaas: For RISC-V both implementations are possible. Technically it depends on the assembly instructions used. I'm not sure what the standard compiler mappings do, but I also think it really doesn't matter either way.

ThomasHaas:
Looks good to me. @hernanponcedeleon might want to take look once he has time.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/VisitorLitmusX86.java
```diff
@@ -101,82 +97,76 @@ public Object visitInstructionRow(LitmusX86Parser.InstructionRowContext ctx) {
...
-        Event event = EventFactory.X86.newExchange(object, register);
-        return programBuilder.addChild(mainThread, event, ctx.getStart().getLine());
+        return append(EventFactory.X86.newExchange(object, register), ctx);
     }
 
     @Override
-    public Object visitIncrementLocation(LitmusX86Parser.IncrementLocationContext ctx) {
+    public Object visitIncrementLocation(IncrementLocationContext ctx) {
         // TODO: Implementation
         throw new ParsingException("INC is not implemented");
```
hernanponcedeleon: For consistency with other visitors we should use `UnsupportedOperationException` (same for other methods in this class)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
```diff
@@ -71,20 +72,31 @@ public Program parse(String raw, String path, String format, String cflags) thro
...
+            default -> throw new ParsingException("Unknown input file type");
+        }
+        return parseAndWrap(parser, CharStreams.fromString(raw));
+    }
+
+    private Program parseAndWrap(ParserInterface parser, CharStream charStream) {
+        try {
+            return parser.parse(charStream);
+        } catch (ParsingException exception) {
+            throw exception;
```
hernanponcedeleon: Why to catch it if you are going to eventually throw the same exception?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -527,53 +527,159 @@ public static class AArch64 {
...
-        public static class DSB {
-            private DSB() {
-            }
+        public static Event newStoreOp(Expression address, IntBinaryOp operator, Expression operand, MemoryOrder mo) {
+            final Event st = EventFactory.Common.newRmwOp(address, operator, operand);
+            // ignore acquire tag
+            st.addTags(toRelTag(mo));
+            st.setMetadata(STOP_PRINTING);
+            return st;
+        }
```
hernanponcedeleon, line 565: I am confused about this method and `newLoadOp`. The names suggest they are simply load or store operations, but then we call RMW methods. Also, the load one adds both release and acquire tags, but the store one ignores the acquire one. Why?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
```diff
@@ -807,13 +923,25 @@ public static VulkanCmpXchg newVulkanCmpXchg(Expression address, Register regist
             return new VulkanCmpXchg(register, address, expected, value, mo, scope);
         }
 
+        public static Event newBarrier(String mo, String scope, List<String> semantics, boolean av, boolean vis) {
```
hernanponcedeleon: Vulkan has both memory and control barriers. I would be more explicit here and call the method `newMemoryBarrier`

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
ThomasHaas: It's cause he wants to catch everything BUT the ParsingException. I don't think Java has a way to express this directly. The alternative would be
```Java
catch (Exception e) {
   if (e instance ParsingException)
       throw e; // Exception is already ParsingException
   throw new ParsingException(e); // Wrap into ParsingException
}
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
ThomasHaas: Those are `LDADD/STADD` (and its variants, i.e, `LD<OP>` and `ST<OP>`) in `arm8`. That's also why the `StoreOp` ignores acquire.
He just moved the code from the `Arm8Visitor` into the `EventFactory`.

hernanponcedeleon: But then why does not the load one ignore the release?

ThomasHaas: Maybe you forgot what these operations mean :)
`LD<OP> == fetch_op`, i.e., it has a proper load followed by a proper store both of which can have memory orderings.
`ST<OP> == store_op` (atomic reduction), i.e., it technically has no load (non-returning) and only its store can have a memory ordering. To be fair, the code should probably throw an exception when trying to use ACQ/ACQ_REL memory orderings for `ST<OP>`.

xeren: Yes, like with `newBarrier`/`newFence`, `newLoadReserved`/`newLoadExclusive`, etc..  I chose these names to correspond with the instructions of the associated ISAs/Languages (and to abstract from our IR's).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
xeren: Your code looks easier to read.

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/EventFactory.java
hernanponcedeleon: > To be fair, the code should probably throw an exception when trying to use ACQ/ACQ_REL memory orderings for `ST<OP>`.

This is still open

PR #1011, opened 2026-03-14
Fix jitpack build

natgavrilenko:


hernanponcedeleon:
I have no clue what this is about

ThomasHaas:
The Jitpack docu says
```
Maven projects that specify a target version in their pom will be built with that target version.
```
Is our `pom.xml` not set up correctly? If so, can we fix that instead of adding the `jitpack.yml`?

hernanponcedeleon:
> Is our `pom.xml` not set up correctly? 

Our pom is correct. Actually, I do not think the problem is coming from our code, [jitpack was working until at least until 7e188c4266](https://www.[jitpack.io/com/github/hernanponcedeleon/Dat3M/development-7e188c4266-1/build.log](https://www.jitpack.io/com/github/hernanponcedeleon/Dat3M/development-7e188c4266-1/build.log)). I suspect the problem is some transitive dependency, but as far as I can see, we only updated the mockito version after that commit, and reverting to the previous version [does not fix the build in jitpack](https://www.[jitpack.io/com/github/hernanponcedeleon/Dat3M/test-13d5530049-1/build.log](https://www.jitpack.io/com/github/hernanponcedeleon/Dat3M/test-13d5530049-1/build.log)).

> If so, can we fix that instead of adding the `jitpack.yml`?

According to AI the documentation of jitpack is misleading. It does not mean “_JitPack will use the JDK version specified in your pom.xml”_, but rather _“maven will compile your code targeting that version”_. By default, jitpack uses java 8. According to AI adding the `jitpack.yml` is the proper way of solving the problem.

ThomasHaas:
I see. Then the PR looks good to me

PR #1013, opened 2026-03-21
Support upper-case inline assembly

ThomasHaas:


ThomasHaas:
Hmm, my easy solution doesn't work. I hoped making everything lower case works, but the lexers have exactly one upper-case rule for `Q`. That's unfortunate.

ThomasHaas:
Upper case letters are now recognized by the lexer but they will never match a rule.
However, inline assembly seems to be case insensitive and so, e.g., `mfence` and `MFENCE` should both be valid, but we only recognize the former.

ThomasHaas:
see `cna.h` in #1016, where I had to comment out `asm("PAUSE")` because it was tripping up the parser.
With this PR, the code is parsed and `asm("PAUSE")` is treated as unknown inline assembly (replaced by NOOP + logger warning).

hernanponcedeleon (approved):

hernanponcedeleon:
Let get this one merged first so you can keep the original code in #1016

Issue #642, opened 2024-03-22
Discrepancy between eager encoding and CAAT with overflowing accesses


ThomasHaas:
In the presence of buffer-overflows our alias analysis is unsound and there can be different same-address accesses that are not considered aliasing and thus will not appear in the `loc` relation of an eager encoding.
However, the CAAT solver established the `loc` relation in an execution by comparing the address values directly thus ignoring the (unsound) alias analysis.
This results in cases where the CAAT solver and the eager encoding return different verdicts.

What do we do about this?
The alias analysis is intentionally unsound in overflow cases, so we cannot blame this issue on the analysis.
I guess the most reasonable fix is to let CAAT inherit this unsoundness by querying the alias analysis between same-address accesses?
Alternatively we live with this discrepancy as it only appears on benchmarks where we are unsound anyways.

PR #1014, opened 2026-03-21
Fix CAAT adding spurious co-edges to executions with OOB accesses (+ code simplification)

ThomasHaas:
If two writes were accessing the same address due to OOB in such a way that it violated AA, we had two same-address writes without a co-edge in the encoding. `ExecutionModel` "completed" such missing coherence edges arbitrarily which could result in a very, very rare inconclusiveness bug.

This PR ensures that we just use the coherence edges appearing in the encoding. This also aligns the behaviors of lazy and eager solving a bit more in the presence of OOB.

Furthermore, I simplified `ExecutionModel`, `EventData`, `ReadFromGraph` and `CoherenceGraph` a lot.
Due to the introduction of `ExecutionModelNext`, `ExecutionModel` is likely never going to be used outside of `RefinementSolver` and so it can be specialized more to its only use-case.

I also fixed a little bug in `ExecutionGraphVisualizer` when invoking it with a `null`-config (we had such code!?).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/basePredicates/CoherenceGraph.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/solver/caat4wmm/basePredicates/CoherenceGraph.java
@@ -1,138 +1,33 @@
 package com.dat3m.dartagnan.solver.caat4wmm.basePredicates;
 
 
-import com.dat3m.dartagnan.solver.caat.misc.EdgeDirection;
+import com.dat3m.dartagnan.encoding.EncodingContext;
+import com.dat3m.dartagnan.encoding.IREvaluator;
 import com.dat3m.dartagnan.solver.caat.predicates.relationGraphs.Edge;
 import com.dat3m.dartagnan.verification.model.EventData;
-import com.google.common.collect.Iterators;
 
-import java.util.*;
-import java.util.function.Function;
-import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


ThomasHaas:
This PR is partially related to #642 which talks about CAAT adding `loc`-edges between same-address accesses that are OOB.
This has not changed in this PR; it only affects the `co` relation.

hernanponcedeleon:
Why not adding the cna code that exposed this bug as a benchmark? 

Besides being a good regression, you already mentioned it has several other "weird" behaviors so it would be good to have the code around (and it would be another "large" benchmark which you always complain we do not have enough).

ThomasHaas:
> Besides being a good regression, you already mentioned it has several other "weird" behaviors so it would be good to have the code around (and it would be another "large" benchmark which you always complain we do not have enough).

I wanted to add the C-code of the CNA version somewhere, but I need to clean it up first. Especially because the code is not mine and comes from an external repo. Also, there are two variants of the code, a buggy one and a correct one.

> Why not adding the cna code that exposed this bug as a benchmark?

I'm not sure if it is suitable for a unit test since IIRC it takes quite a bit of time + it is non-deterministic.
But I can see how bad it is once I cleaned up the code.

hernanponcedeleon (approved):
As discussed privately, lets merged this one without the new benchmark and add that one as a separated PR which will move some other benchmarks to a new directory.

PR #1019, opened 2026-03-28
Moved floating point rounding mode into Program

ThomasHaas:
This is basically the less controversial half of #978 :)
I kept the `Program.SemanticConfig` class even though it contains only a single option right now.

The biggest question is where to best inject the `Configuration` into the `Program`. For simplicity, I do this when constructing the `VerificationTask`, but it is a bit sketchy.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -294,7 +294,7 @@ public class Dartagnan extends BaseOptions {
         StringBuilder details = new StringBuilder();
         // We only show the condition if this is the reason of the failure
         String condition = "";
-        final boolean ignoreFilter = task.getConfig().hasProperty(IGNORE_FILTER_SPECIFICATION) && task.getConfig().getProperty(IGNORE_FILTER_SPECIFICATION).equals("true");
+        final boolean ignoreFilter = Objects.equals(task.getConfig().getProperty(IGNORE_FILTER_SPECIFICATION), "true");
         final boolean nonEmptyFilter = !(p.getFilterSpecification() instanceof BoolLiteral bLit) || !bLit.getValue();
         final String filter = !ignoreFilter && nonEmptyFilter ? p.getFilterSpecification().toString() : "";
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
@@ -39,6 +39,9 @@ public class VerificationTask {
         this.property = checkNotNull(property);
         this.witness = checkNotNull(witness);
         this.config = checkNotNull(config);
+
+        // TODO: Is it a good idea to inject configs into the program here?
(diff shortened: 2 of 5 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -294,7 +294,7 @@ private static ResultSummary summaryFromResult(VerificationTask task, ModelCheck
         StringBuilder details = new StringBuilder();
         // We only show the condition if this is the reason of the failure
         String condition = "";
-        final boolean ignoreFilter = task.getConfig().hasProperty(IGNORE_FILTER_SPECIFICATION) && task.getConfig().getProperty(IGNORE_FILTER_SPECIFICATION).equals("true");
+        final boolean ignoreFilter = Objects.equals(task.getConfig().getProperty(IGNORE_FILTER_SPECIFICATION), "true");
```
hernanponcedeleon, line 297: Does this now handle the case where we do not have the property?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
```diff
@@ -39,6 +39,9 @@ protected VerificationTask(Program program, Wmm memoryModel, ProgressModel.Hiera
         this.property = checkNotNull(property);
         this.witness = checkNotNull(witness);
         this.config = checkNotNull(config);
+
+        // TODO: Is it a good idea to inject configs into the program here?
```
hernanponcedeleon, line 43: What would be the alternative?

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: Yeah, it seems `getProperty` just returns `NULL` in that case.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTask.java
ThomasHaas: Inject it right after parsing the program.

hernanponcedeleon: Code-wise that would be bad, because we have different "places" where program is parsed (CLI, UI, unit tests) and we would have to inject it in all of them.

One could claim that we do not care the rounding mode unless we are doing the verification (and thus having the inject the VerificationTask is ok), but then this raises the question if the rounding mode should then indeed be part of the Program.

ThomasHaas: Well, we have a central `ProgramParser` that is always used to parse the program. If that class was configurable via a `Configuration`, we could inject the settings there. Generally, making the parsers configurable would be a good idea and allow us to give options like `ignore inline assembly` or `throw on unknown inline assembly` etc.

If the rounding mode was part of the verification task, we would need to lift all program passes to verification task passes (at least in principle) for otherwise the rounding mode was inaccessible.

But for now, I think having the injection in `VerificationTask` is okayish.

PR #1020, opened 2026-03-28
Remove all thrown exceptions in Dartagnan.main()

hernanponcedeleon:
This solves some SVCOMP problems in which the runner was looping forever e.g., if we would pass the wrong file path. 

However, we still have the looping problem if dartagnan runs OOM. It seems the jvm returns 1 as exit code, which is the same value we use for `ExitCode.BOUNDED_RESULT` (which is also what we used for the looping condition in the runner).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -98,7 +98,7 @@ public class Dartagnan extends BaseOptions {
     }
 
 
-    public static void main(String[] args) throws Exception {
+    public static void main(String[] args) {
 
         initGitInfo();
 
@@ -109,17 +109,30 @@ public class Dartagnan extends BaseOptions {
 
         if (Arrays.asList(args).contains("--version")) {
             final MavenXpp3Reader mvnReader = new MavenXpp3Reader();
-            final FileReader fileReader = new FileReader(System.getenv("DAT3M_HOME") + "/pom.xml");
-            final String base = mvnReader.read(fileReader).getVersion();
-            final String version = base.equals(getGitTags()) ? base : String.format("%s (commit %s)", base, getGitId());
-            System.out.println(version);
+            final String pomPath = System.getenv("DAT3M_HOME") + "/pom.xml";
+            try (FileReader fileReader = new FileReader(pomPath)) {
+                final String base = mvnReader.read(fileReader).getVersion();
+                final String version = base.equals(getGitTags()) ? base : String.format("%s (commit %s)", base, getGitId());
+                System.out.println(version);
+            } catch (Exception e) {
+                logger.warn("Failed to read version from {}", pomPath, e);
+                System.exit(UNKNOWN_ERROR.asInt());
+            }
             return;
         }
 
         logGitInfo();
 
-        final Configuration config = loadConfiguration(args);
-        final Dartagnan o = new Dartagnan(config);
+        final Configuration config;
+        final Dartagnan o;
+        try {
+            config = loadConfiguration(args);
+            o = new Dartagnan(config);
+        } catch (IOException | InvalidConfigurationException e) {
+            logger.error(e.getMessage());
+            System.exit(UNKNOWN_ERROR.asInt());
+            return;
+        }
 
         final File fileModel = new File(Arrays.stream(args).filter(a -> a.endsWith(".cat")).findFirst()
                 .orElseThrow(() -> new IllegalArgumentException("CAT model not given or format not recognized")));
@@ -129,7 +142,13 @@ public class Dartagnan extends BaseOptions {
         final WitnessGraph witness;
         if (o.runValidator()) {
             logger.info("Witness path: {}", o.getWitnessPath());
-            witness = new ParserWitness().parse(new File(o.getWitnessPath()));
+            try {
+                witness = new ParserWitness().parse(new File(o.getWitnessPath()));
+            } catch (IOException e) {
+                logger.error("Failed to read witness file {}", o.getWitnessPath());
+                System.exit(WRONG_WITNESS_FILE.asInt());
+                return;
+            }
         } else {
             witness = new WitnessGraph();
         }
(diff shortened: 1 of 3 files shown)


ThomasHaas:
I think we should simply use different exit codes and avoid the common one's like `1` (see [here](https://tldp.org/LDP/abs/html/exitcodes.html) for some examples). I also asked ChatGPT and it mentions that several of the lower numbers `0-4` are used. I would play it safe and use exit codes starting at `30` or so.

ThomasHaas:
I think we should just put `UNKNOWN_ERROR` on exit code 1, and then even if we have uncaught exceptions or the JVM runs out of memory, the error code is fine.
But I guess that any changes in the exit codes would require you to update Vsyncer again?

hernanponcedeleon:
> I think we should just put `UNKNOWN_ERROR` on exit code 1, and then even if we have uncaught exceptions or the JVM runs out of memory, the error code is fine. 

As you said, that will break vsyncer. However, it might indeed be the best solution. Since there is no hurry for these changes, I will put the PR on hold until we also rework all options (which will also have some breaking changes); then we can make a new release with major version update

ThomasHaas:
The changes in this PR conflict with #1022. I think #1022 will get merged before this one and then you will need to "reapply" the changes, so maybe we should just close this PR for now.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -63,11 +65,21 @@ public static void main(String[] args) throws Exception {
...
+        try {
+            config = loadConfigurationFromArgs(args);
+            o = new Dartagnan(config);
+            catFile = getCatFileFromArgs(args);
+            progFiles = getProgramFilesFromArgs(args);
+        } catch (IOException | InvalidConfigurationException e) {
+            logger.error(e.getMessage());
+            System.exit(UNKNOWN_ERROR.asInt());
+            return;
+        }
```
ThomasHaas, line 83: I think you should just catch all Exceptions here, no? What if the thrown exception is neither of the two cases (not sure if this is possible)?
Generally, I assume you want to ensure that no exception is left uncatched, right?

Also, should we just log the error message here? If an exception appears during solving, we print it to stdout instead of logging it. This is inconsistent.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -120,14 +132,16 @@ public static void printOptions() {
...
+    private static void printVersion() {
         final MavenXpp3Reader mvnReader = new MavenXpp3Reader();
         final Path pomPath = getHomeDirectory().resolve("pom.xml");
 
         try (BufferedReader reader = Files.newBufferedReader(pomPath)) {
             final String base = mvnReader.read(reader).getVersion();
             final String version = base.equals(getGitTags()) ? base : String.format("%s (commit %s)", base, getGitId());
             System.out.println(version);
+        } catch (IOException | XmlPullParserException e) {
+            logger.warn("Failed to load {}", pomPath);
```
ThomasHaas: Similarly to above, I would just catch all exceptions.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
hernanponcedeleon: > Also, should we just log the error message here? If an exception appears during solving, we print it to stdout instead of logging it. This is inconsistent.

I don't get what you mean.

ThomasHaas: During solving we catch exceptions and generate a `ResultSummary` from them. The `ResultSummary` is later printed to stdout. This is different to the exceptions during parsing which are only logged but not printed to stdout. In particular, if logging were disabled, there would be no output at all.

EDIT: Btw. there is an overload `logger.error(String msg, Throwable e)` that seems better suited for exceptions.

EDIT 2: I guess the question is if logging at that point in the code is even needed. Conceptually, I think the error message itself is the tool's output to the user at that point, and I wouldn't produce user-facing output via logging.

ThomasHaas: I didn't mean that you should use `getSummaryFromException`, just that printing to stdout should be preferred.

hernanponcedeleon: Just using printing makes it harder to keep consistency with the rest. This way is easier to guarantee that we get the same kind of input we get for other exceptions (I even tried to use `OutputLogger` to make the code work even if there are changes, but since `catFile` and `config` might be `NULL` it look ugly)

ThomasHaas: Having `Test: Unknown` in the output is also ugly, no? Well, I guess the ugliness will go away with #1068 anyway, so I don't mind too much for now.

ThomasHaas (approved):
LGTM

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
hernanponcedeleon: The problem is that at this point you might not yet have a test: the case I was using to test this is when you give a wrong path, thus we cannot have a test because there is no proper code to parse. In theory you could also be in batch mode in which case reporting a single test (which one?) would also be weird

ThomasHaas: The point is that you are forced to print to a test because you use `ResultSummary`. `ResultSummary` is the wrong class to capture the output: 90% of its fields are unused/wrong (e.g. the test name which doesn't exist in this case). Batch mode is also irrelevant because the error here is before batch mode is even decided (the user input is fundamentally broken, not some test in a batch).

That's why I said #1068 will fix this ugliness, because it replaces `ResultSummary` and generalizes `TaskResultAnalyzer` (which was meant to analyze task results... we don't even have a task yet at all) to `OutputGenerator`.

PR #1022, opened 2026-04-04
Add TaskSolver as abstraction layer over ModelChecker

ThomasHaas:
This PR adds a new abstraction layer before the ModelChecker. For now, this layer does not do much apart from choosing the appropriate ModelChecker and managing timeouts.
However, the intend is to make this layer more powerful in the future and enable more interesting solving approaches such as:
- Automatic bound unrolling without external script
- Running SMC for some time and if SMC solves the task in, say, < 1000 explored executions, then we are done. Otherwise, fall back to BMC but exploit information gathered from SMC (loop bounds, indirect call targets, aliasing information, summary of sequential prefix, etc.).
- Parallel solving via decomposition?
- "Proof mode" that first unrolls loops until completion and only then checks the properties of interest. This can be a lot more efficient if the program is likely to be correct.
- State-space exploration mode for compilation correctness.

~**TODO:** Update reflection metadata for native build.~

**EDIT:** I also added a `TaskResultAnalyzer` class that is now responsible for generating the `ResultSummary` and witness files. Basically, I moved all output generation from `Dartagnan.java` to `TaskResultAnalyzer`

**EDIT 2:** I took the freedom to clean up `Dartagnan.java`. I think it is a lot easier to follow the logic of Dartagnan now.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/TaskSolver.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/TaskSolver.java
@@ -42,7 +42,7 @@ public class TaskSolver implements AutoCloseable {
             name = TIMEOUT,
             description = "Timeout before interrupting the task solver. Can specify time units ns, ms, s (default), min, and h.")
     @TimeSpanOption(min = 0, codeUnit = TimeUnit.MILLISECONDS, defaultUserUnit = TimeUnit.SECONDS)
-    private int timeout = 1200;
+    private int timeout = 0;
 
     private boolean hasTimeout() {
         return timeout > 0;
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/verification/TaskSolver.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/verification/TaskSolver.java
@@ -1,0 +1,113 @@
+package com.dat3m.dartagnan.verification;
+
+import com.dat3m.dartagnan.configuration.Method;
+import com.dat3m.dartagnan.encoding.IREvaluator;
+import com.dat3m.dartagnan.utils.Result;
+import com.dat3m.dartagnan.verification.model.ExecutionModelManager;
+import com.dat3m.dartagnan.verification.model.ExecutionModelNext;
+import com.dat3m.dartagnan.verification.solving.ModelChecker;
+import com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.sosy_lab.common.ShutdownManager;
+import org.sosy_lab.common.configuration.InvalidConfigurationException;
+import org.sosy_lab.common.configuration.Option;
+import org.sosy_lab.common.configuration.Options;
+import org.sosy_lab.java_smt.api.SolverException;
+
+import static com.dat3m.dartagnan.configuration.OptionNames.METHOD;
+
+@Options
+public class TaskSolver implements AutoCloseable {
+
+    private static final Logger logger = LoggerFactory.getLogger(TaskSolver.class);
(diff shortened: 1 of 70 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/verification/TaskSolver.java
```diff
@@ -0,0 +1,184 @@
...
+            try {
+                final long timeoutInMillis = timeout;
+                Thread.sleep(timeoutInMillis);
+                final String error = String.format("Timeout of %s exceeded.", Utils.toTimeString(timeoutInMillis));
+                modelChecker.requestShutdown(error);
+            } catch (InterruptedException e) {
+                // Verification ended, nothing to be done.
+            }
+        });
+    }
```
xeren, line 183: You should capture `timeout` and `modelChecker` in the lambda's closure, or join `timeoutThread` in `run()`.  The accesses here are potentially racy, e.g. with `modelChecker = null` in `close()`.

ThomasHaas: I guess we could add a `join()` after the `interrupt()`. The join should then return more or less instantly.

The case you have in mind seems to be an extreme edge-case where the model checker returns right before the timeout, the interrupt gets send but doesn't arrive quickly enough (though the target thread is sleeping and therefore listens to interrupts) before the task solver gets closed and the timeout thread reaches its timeout.

hernanponcedeleon:
Witness generation is not working with the native build
```
CFLAGS="-DREL2RX" ./dartagnan/target/dartagnan cat/vmm.cat benchmarks/locks/ttas.c --bound.save=output/bound.csv --bound.load=output/bound.csv --witness=png
Fatal error reported via JNI: Could not allocate library name

...  
```

dartagnan/src/test/java/com/dat3m/dartagnan/litmus/comparison/AbstractComparisonTest.java
```diff
@@ -77,7 +77,8 @@ protected final Configuration getConfiguration() throws InvalidConfigurationExce
                 .setOption(SOLVER, Z3.name())
                 .setOption(PHANTOM_REFERENCES, "true")
                 .setOption(INITIALIZE_REGISTERS, "true")
-                .setOption(USE_INTEGERS, "true");
+                .setOption(USE_INTEGERS, "true")
+                .setOption(METHOD, Method.EAGER.asStringOption());
```
hernanponcedeleon, line 81: why are you changing this?

ThomasHaas: We always used the `EAGER` method here by explicitly using the `AssumeSolver`. Now we can configure this via the `Configuration` (this was not possible before because the method option was injected into the`Dartagnan` class).
To be fair, I should rename the method `testAssume` to just `test` then.

I guess the question is: do we want to have one test method that can be configured to use eager/lazy or use dedicated test methods for each? We use the latter approach for several test cases where we run both solvers.
For compilation/comparison tests we always used only one solver so I thought I make it at least configurable.

hernanponcedeleon:
> Witness generation is not working with the native build
> 
> ```
> CFLAGS="-DREL2RX" ./dartagnan/target/dartagnan cat/vmm.cat benchmarks/locks/ttas.c --bound.save=output/bound.csv --bound.load=output/bound.csv --witness=png
> Fatal error reported via JNI: Could not allocate library name
> 
> ...  
> ```

The problem has nothing to do with this PR (there is already a proposed fix in #1023) and the rest LGTM after the latest commit.

PR #1023, opened 2026-04-15
Remove awt dependency in ColorMap

hernanponcedeleon:
Fixes the problem with witness generation and native build identified in #1022

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
@@ -32,6 +32,8 @@ class ColorMap {
         "\"#F4C800\"", // yellow
         "\"#593315\"", // brown
         "\"#F13A13\""  // reddish orange
+        "\"#232C16\"", // olive green
+        "\"#232C16\""  // purplish red
     };
 
     ColorMap() {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
@@ -1,65 +1,38 @@
 package com.dat3m.dartagnan.witness.graphviz;
 
-import java.awt.Color;
-import java.util.Set;
-import java.util.HashSet;
 import java.util.Map;
 import java.util.HashMap;
 
 import static com.dat3m.dartagnan.wmm.RelationNameRepository.*;
 
-
 class ColorMap {
-    private final Map<String, String> fixedColorMap = new HashMap<>();
-    private final Set<String> usedColors = new HashSet<>();
-    private int currentHue = 0;
-    private int step = 72;
-    private float saturation = 1.0f;
-    private float brightness = 1.0f;
+
+    private final Map<String, String> relationColors = new HashMap<>();
+    private int nextColorIndex = 0;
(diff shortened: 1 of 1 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
```diff
@@ -1,65 +1,38 @@
...
-        }
-        return generateColor();
-    }
+    // Kelly palette
+    private static final String[] KELLY_COLORS = {
+        "\"#FFB300\"", "\"#803E75\"", "\"#FF6800\"", "\"#A6BDD7\"", "\"#C10020\"",
+        "\"#CEA262\"", "\"#817066\"", "\"#007D34\"", "\"#F6768E\"", "\"#00538A\"",
+        "\"#FF7A5C\"", "\"#53377A\"", "\"#FF8E00\"", "\"#B32851\"", "\"#F4C800\"",
+        "\"#7F180D\"", "\"#93AA00\"", "\"#593315\"", "\"#F13A13\"", "\"#232C16\""
+    };
```
ThomasHaas, line 33: Maybe you want to skip colors that are too close to the 4 predefined one's (green and red are certainly in the list somewhere).

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
```diff
@@ -1,65 +1,38 @@
...
-            currentHue -= 360;
-            step /= 2;
-            if (currentHue == 0) {
-                updateHue();
-            }
-        }
+    private String nextPaletteColor() {
+        String color = KELLY_COLORS[nextColorIndex];
+        nextColorIndex = (nextColorIndex + 1) % KELLY_COLORS.length;
+        return color;
```
ThomasHaas, line 53: I would probably do `Preconditions.checkState(nextColorIndex < KELLY_COLORS.length - 1, "Not enough colors")` rather than doing the silent wraparound. In practice it won't matter (no one shows so many relations), but I think it is better to be notified if we ever run into this problem.

hernanponcedeleon: That is certainly not the behavior I want dartagnan to have. Even if the user requested many relations, we should handle this case and thus I made the wrap around on purpose.

That being said, we could log a warning the let the user know it might not be a good idea to request that many relations.

ThomasHaas: Logging a warning is fine.

dartagnan/src/main/java/com/dat3m/dartagnan/witness/graphviz/ColorMap.java
ThomasHaas: I don't think you need to remove all green and red tones, just the one's that are too close to rf/co colors.
Colors like olive green and purplish red are very distinct and we can keep them.

ThomasHaas: You made a copy&paste error for the two added colors: they have the same color string.

ThomasHaas (approved):
I haven't checked what the colors look like and if we should change their order (most distinct colors first), but I don't care too much really :P
So LGTM.

PR #1025, opened 2026-04-15
Fix logging of program path

hernanponcedeleon:
Fixes a problem introduced in #1022

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -155,8 +155,13 @@ public class Dartagnan extends BaseOptions {
         final List<File> files = new ArrayList<>();
         Stream.of(args).map(Paths::get).filter(Files::exists)
                 .forEach(path -> {
-                    logger.info("Program(s) path: {}", path.normalize());
-                    files.addAll(getProgramFiles(path));
+                    List<File> supported = getProgramFiles(path).stream()
+                        .filter(ProgramParser::isSupported)
+                        .toList();
+                    if (!supported.isEmpty()) {
+                        logger.info("Program(s) path: {}", path.normalize());
+                        files.addAll(supported);
+                    }
                 });
         if (files.isEmpty()) {
             throw new IllegalArgumentException("Path to input program(s) not given or format not recognized");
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -155,8 +155,13 @@ private static List<File> getProgramFilesFromArgs(String[] args) {
...
                 .forEach(path -> {
-                    logger.info("Program(s) path: {}", path.normalize());
-                    files.addAll(getProgramFiles(path));
+                    List<File> supported = getProgramFiles(path).stream()
+                        .filter(ProgramParser::isSupported)
+                        .toList();
+                    if (!supported.isEmpty()) {
+                        logger.info("Program(s) path: {}", path.normalize());
+                        files.addAll(supported);
+                    }
```
ThomasHaas: `getProgramFiles` only returns supported files, so no need to run the filter again.
Do we want to only log the directory or also all files inside? The old solution did both.
If we want to log both, we could do (inside the lambda):
```Java
final List<File> progFiles = getProgramFiles(path);
if (progFiles.isEmpty()) {
    return;
}

if (Files.isDirectory(path)) {
    logger.info("Programs path: {}", path.normalize());
}
for (File progFile : getProgramFiles(path)) {
    logger.info("Program path: {}", progFile.toPath().normalize());
    files.add(progFile);
}
```

Btw. the `Path` class is the more modern version of the `File` class. We should eventually switch over to using `Path` everywhere.

hernanponcedeleon: I would not add a log entry for every file: I use batch mode with litmus tests and I do not want to get ~5K entries

ThomasHaas: Sure. I was just saying this because it used to be like that.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
```diff
@@ -41,6 +41,10 @@ public static boolean isSupported(Path filePath) {
         return SUPPORTED_EXTENSIONS.contains(extension);
     }
 
+    public static boolean isSupported(File file) {
+        return isSupported(file.toPath());
+    }
+
```
ThomasHaas: this method is not needed anymore.

ThomasHaas (approved):
LGTM

PR #1030, opened 2026-05-10
Add more program simplifications

ThomasHaas:
(1) Added SROA (scalar replacement of aggregates), which is mainly applied to aggregate registers generated by spin loop instrumentation.

(2) Added a new `NO_CARRY_DEP` tag for events which signifies that registers read by this event do not carry dependencies, i.e., we are allowed to optimize aggressively.  For now, only the events created by the spin loop instrumentation use that tag.
NOTE: For now, this is only used in optimizations. We still generate dependency edges in the encoding!

(3) Made `AssignmentInlining` more aggressive when inlining ITE expressions. Added more expression simplifications.

Pretty much all changes are related to eliminating the spinloop instrumentation code if we can statically infer that the loop must be side-effectful.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/expression/processing/ExprSimplifier.java
@@ -1,14 +1,10 @@
 package com.dat3m.dartagnan.expression.processing;
 
 
-import com.dat3m.dartagnan.expression.BinaryExpression;
-import com.dat3m.dartagnan.expression.Expression;
-import com.dat3m.dartagnan.expression.ExpressionKind;
-import com.dat3m.dartagnan.expression.aggregates.AggregateCmpExpr;
-import com.dat3m.dartagnan.expression.aggregates.AggregateCmpOp;
-import com.dat3m.dartagnan.expression.aggregates.ConstructExpr;
-import com.dat3m.dartagnan.expression.aggregates.ExtractExpr;
+import com.dat3m.dartagnan.expression.*;
+import com.dat3m.dartagnan.expression.aggregates.*;
 import com.dat3m.dartagnan.expression.booleans.*;
+import com.dat3m.dartagnan.expression.floats.FloatUnaryOp;
 import com.dat3m.dartagnan.expression.integers.*;
 import com.dat3m.dartagnan.expression.memory.*;
 import com.dat3m.dartagnan.expression.misc.ITEExpr;
@@ -29,7 +25,7 @@ public class ExprSimplifier extends ExprTransformer {
     // If set to "false", the simplifier will not destroy register dependencies, i.e.,
     // it will maintain "expr.getRegs() == simplified(expr).getRegs()".
     // For example, "0*r" will not get simplified to "0";
-    private final boolean aggressive;
+    protected boolean aggressive;
 
     public ExprSimplifier(boolean aggressive) {
         this.aggressive = aggressive;
@@ -59,6 +55,34 @@ public class ExprSimplifier extends ExprTransformer {
         ).accept(this);
 
         return hoistedIte;
+    }
+
+    // Tries to push unary operations into ITE expressions.
+    private Expression tryGeneralRewrite(UnaryExpression expression) {
+        final Expression operand = expression.getOperand().accept(this);
+        if (!(operand instanceof ITEExpr ite)) {
+            return null;
+        }
+
+        final ExpressionKind op = expression.getKind();
+        if (op instanceof BoolUnaryOp || op instanceof IntUnaryOp || op instanceof FloatUnaryOp) {
+            // ------- Standard unary expressions -------
+            return expressions.makeITE(
+                    ite.getCondition(),
+                    expressions.makeUnary(expression.getKind(), ite.getTrueCase()),
+                    expressions.makeUnary(expression.getKind(), ite.getFalseCase())
+            ).accept(this);
+        } else if (expression instanceof IntSizeCast cast) {
+            // ------- Integer casts -------
+            return expressions.makeITE(
+                    ite.getCondition(),
+                    expressions.makeIntegerCast(ite.getTrueCase(), cast.getTargetType(), cast.preservesSign()),
+                    expressions.makeIntegerCast(ite.getFalseCase(), cast.getTargetType(), cast.preservesSign())
+            ).accept(this);
+        } else {
+            return null;
+        }
+
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


ThomasHaas:
The CI failure is not due to wrong optimizations, but to the integer-based encoding being messed up.
We compare `zext bv8(128) == zext bv8(-128)` which is true because `bv8(128)==bv8(-128)`.
The integer encoding for the former will eliminate the sign and generate `128 == 128` but it won't eliminate the sign for the latter and generate `128 == -128` which is clearly false.

hernanponcedeleon:
Are you saying the integer encoding is treating `zext bv8(-128)` as `128`? Why is that? for the integer encoding a `zext` should be a noop

ThomasHaas:
See #956: `zext` makes values positive because they will always be positive as BV (the sign bit is guaranteed to be 0).
So it correct that `zext bv8(-128)` is treated as `128`. However, this PR adds an optimization that removes the `zext` which ends up giving different results for BV and Int encoding:
```
zext bv8(128) == zext bv8(-128) // True for BV encoding and int encoding
// ... but
bv8(128) == bv8(-128) // True for BV encoding, but not for int encoding
```
I guess we could normalize the constants to unsigned form when doing the optimization, i.e., replace `-128` by the equivalent `128`.

ThomasHaas:
Normalization doesn't solve anything... my bad. The value that is compared against is loaded from memory, so I cannot normalize it (it's not a constant). 

Either we live the fact that integer encodings are possibly affected (they are imprecise anyhow) or we remove this particular optimization.

hernanponcedeleon:
Can we come up with a program that would show that the integer encoding is already imprecise in `development`?

ThomasHaas:
> Can we come up with a program that would show that the integer encoding is already imprecise in `development`?

Depends on what imprecision you want to highlight. For example, `assert (UINT_MAX + 1 == 0)` will fail for integer encodings (but you need to load UINT_MAX from memory so that SCCP does not simplify the program).
Maybe it is also possible to write a variant of the failing test that naturally does not have the `zext`.

hernanponcedeleon:
I would be fine with merging all optimizations if indeed the integer encoding is already imprecise. So I am asking for an example that shows that, because the imprecision does not show up in the examples from the Ci

ThomasHaas:
```C
#include <assert.h>
#include <limits.h>

unsigned long a = -128;
unsigned long b = ULONG_MAX / 2;

int main() {
    assert(a == (2*b - 126));
    return 0;
}
```

This one is a variant of the failing test that fails without any optimizations, i.e., it already fails in development.

hernanponcedeleon:
Great. Then simply change `MiscellaneousTest` to force the failing test to use BVs instead of ints so the CI passes. I will review the code over the weekend.

ThomasHaas:
Btw. the failing test is the one we added in #956 because of #955.

ThomasHaas:
> Great. Then simply change `MiscellaneousTest` to force the failing test to use BVs instead of ints so the CI passes. I will review the code over the weekend.

The test was specifically made to check zext for integer encodings, so using BVs makes it meaningless. 
So the only reasonable options are deleting the test or obscuring it so that simplifications do not apply.

hernanponcedeleon:
Which is the optimization that converts
```
3:       bv8 r2 = load(&a)
4:       bv32 r3 = zext bv8 r2 to bv32
5:       bv1 r4 = (bv32 r3 == bv32(128)) ? bv1(1) : bv1(0)
6:       if(bv1 r4 != bv1(0)); then goto 0.5
```
into
```
1:       bv8 r2 = load(&a)
2:       if(bv8 r2 == bv8(128)); then goto 0.5
```
?
This optimization is only sound if we assume BVs. 

The questions is: if we are gonna have optimizations that exploit BVs structure, then maybe we should not support encoding things as ints in the first place.

ThomasHaas:
> Which is the optimization that converts
> 
> ```
> 3:       bv8 r2 = load(&a)
> 4:       bv32 r3 = zext bv8 r2 to bv32
> 5:       bv1 r4 = (bv32 r3 == bv32(128)) ? bv1(1) : bv1(0)
> 6:       if(bv1 r4 != bv1(0)); then goto 0.5
> ```
> 
> into
> 
> ```
> 1:       bv8 r2 = load(&a)
> 2:       if(bv8 r2 == bv8(128)); then goto 0.5
> ```

It's some case in the `ExpressionSimplifier` for `IntSizeCast` involving constants.

> This optimization is only sound if we assume BVs.

Well, we don't assume BVs, the program's original semantics is defined in terms of BVs not in terms of mathematical integers.

> The questions is: if we are gonna have optimizations that exploit BVs structure, then maybe we should not support encoding things as ints in the first place.

Integer encoding was always imprecise/unsound with and without optimizations.
There are effectively two different reasons:
- Integers have no overflow semantics.
- Both LLVM IR and our IR do not distinguish between signed and unsigned integer/BV types. This is the most fundamental issue that makes integer encodings unsound no matter what we do (it's impossible to resolve I think).
Maybe you rememeber that also SMACK had problems with this.

The only role optimizations play here is that they sometimes make the encoding "less sound" but sometimes also "more sound", i.e., they go both ways. To see the latter point, consider the example I showed above but make 'a' and 'b' local:
```C
#include <assert.h>
#include <limits.h>
int main() {
    unsigned long a = -128;
    unsigned long b = ULONG_MAX / 2;
    assert(a == (2*b - 126));
    return 0;
}
```
This program constant-folds to `assert (true)`. If we did not apply SCCP/constant-folding, an integer encoding would end up violating the assertion, i.e., be unsound. In that sense, even simple constant-folding "exploits BV structure". Notice that the integer encoding here is unsound even if do not apply any optimizations at all (the optimzations only make it more sound).

hernanponcedeleon:
When I said, "we assume BVs", I was not referring to the encoding, but as you said that the program is defined in terms of BV semantics (maybe the term "we assume" was confusing).

Yesterday I was remembering that we had the same issue with SMACK also.

But then again, do we want to keep the integer encoding if
(1) we know it is unsound, and
(2) there is not much we can do about it?

One could say that since the encoding is not on by default, most users will never get unsound results anyway, and we could still benefit of having the extra encoding (which should give better performance if bitwise operations are not used). However, this is not even true as witnessed in #955.

Additionally, getting rid of the integer encoding would remove quite a bit of code, specially those extra "equal" utils we had to deal with BVs vs ints.

ThomasHaas:
I don't mind getting rid of the integer encoding, but I don't see how #955 shows performance differences (I think you linked the wrong PR?). I think many of our LFDS benchmarks are like 2x or more slower when doing BV instead of integers (our unit tests use integers).  But apart from our unit tests, I have never really used the integer encoding to really get a feeling for its performance characteristics.

> Additionally, getting rid of the integer encoding would remove quite a bit of code, specially those extra "equal" utils we had to deal with BVs vs ints.

I think nowadays all the relevant encoding is fully encapsulated in the `ExpressionEncoder`, so it is not that much code at least.

PR #1034, opened 2026-05-22
Remove integer encoding

hernanponcedeleon:
As discussed in #1030, the integer encoding is unsound and there is not much we can do about it (since programs are defined in terms of BVs semantics).

This PR removes the option to use integer encoding. While we use the BVs encoding by default, the CI still used integers for several benchmarks. Hopefully there is not much regression.


ThomasHaas:
You should also check for usages of the integer formula type. For example, the interval bound encoding has a case for integer formulas.

ThomasHaas:
I think you should revert the changes to `FormulaManagerExt`. Everything in the `smt` package is general-purpose code that should be able to handle integer formulas just fine.

ThomasHaas (approved):
LGTM

PR #1036, opened 2026-05-27
Add back the ability to pass user-specified witness file names

ThomasHaas:
Note: passing witness names in batch mode is broken because all witnesses will have the same name.

@hernanponcedeleon Feel free to do changes to this PR as you see fit

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -99,7 +99,7 @@ public class Dartagnan extends BaseOptions {
 
                 // ----------- Generate output-----------
                 summary = resultAnalyzer.getSummaryFromSolver(taskSolver, progFile.getPath());
-                resultAnalyzer.generateWitnessIfAble(taskSolver, o.getWitnessType(), Utils.getNameWithoutExtension(progFile), o.generateWitnessForUnknown());
+                resultAnalyzer.generateWitnessIfAble(taskSolver, o.getWitnessType(), getWitnessFilename(progFile, o), o.generateWitnessForUnknown());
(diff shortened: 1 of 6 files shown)


hernanponcedeleon:
> Note: passing witness names in batch mode is broken because all witnesses will have the same name.

I fixed this by adding a postfix to the filename in batch mode. To be able to know from which program the witness comes from, I replaced the previous "Iteration 1" (which never made sense) by the program name, except when the graph is generated for debugging the refinement solver.

ThomasHaas:
I think the debugging code for Refinement can be removed. No one uses it.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -99,7 +100,8 @@ public static void main(String[] args) throws Exception {
 
                 // ----------- Generate output-----------
                 summary = resultAnalyzer.getSummaryFromSolver(taskSolver, progFile.getPath());
-                resultAnalyzer.generateWitnessIfAble(taskSolver, o.getWitnessType(), Utils.getNameWithoutExtension(progFile), o.generateWitnessForUnknown());
+                resultAnalyzer.generateWitnessIfAble(taskSolver, o.getWitnessType(), getWitnessFilename(progFile, o, isBatchMode ? "_batch_#" + String.valueOf(it) : ""), o.generateWitnessForUnknown());
```
ThomasHaas: I think you should put the witness file name in a separate variable. The line of code gets too long otherwise.

ThomasHaas:
LGTM, but I cannot approve because it is my PR.

hernanponcedeleon (dismissed):

hernanponcedeleon (approved):

PR #1037, opened 2026-05-28
Add unit tests to check all witness options work as expected

hernanponcedeleon:
The new unit tests should prevent bugs like those from #1036 and #1031.

The CI is expected to fail until we merge #1036 .

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -199,6 +199,11 @@ public class Dartagnan extends BaseOptions {
 
         ProcessBuilder pb = new ProcessBuilder(command);
         Process process = pb.start();
-        process.waitFor();
+        int exitCode = process.waitFor();
+
+        if (exitCode != 0) {
+            String error = new String(process.getErrorStream().readAllBytes());
+            fail("Dartagnan finished with exit code " + exitCode + ". Error: " + error);
+        }
     }
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -194,6 +194,8 @@ public class Dartagnan extends BaseOptions {
         command.add(programPath.toAbsolutePath().toString());
         command.addAll(Arrays.asList(options));
 
+        System.out.println(command);
+
         ProcessBuilder pb = new ProcessBuilder(command);
         Process process = pb.start();
         process.waitFor();
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -73,6 +73,7 @@ public class Dartagnan extends BaseOptions {
         final OutputLogger output = new OutputLogger(catFile, config);
         final TaskResultAnalyzer resultAnalyzer = TaskResultAnalyzer.create();
         ResultSummary summary = null;
+        int it = 0;
         for (File progFile : progFiles) {
             try {
                 // ----------- Generate verification task -----------
@@ -99,7 +100,9 @@ public class Dartagnan extends BaseOptions {
 
                 // ----------- Generate output-----------
                 summary = resultAnalyzer.getSummaryFromSolver(taskSolver, progFile.getPath());
-                resultAnalyzer.generateWitnessIfAble(taskSolver, o.getWitnessType(), Utils.getNameWithoutExtension(progFile), o.generateWitnessForUnknown());
+                final String witnessFileName = getWitnessFilename(progFile, o, isBatchMode ? "_batch_#" + String.valueOf(it) : "");
+                resultAnalyzer.generateWitnessIfAble(taskSolver, o.getWitnessType(), witnessFileName, o.generateWitnessForUnknown());
+                it++;
             } catch (Exception e) {
                 summary = resultAnalyzer.getSummaryFromException(e, progFile.getPath());
             }
@@ -202,5 +205,11 @@ public class Dartagnan extends BaseOptions {
             String error = new String(process.getErrorStream().readAllBytes());
             logger.warn("Dartagnan finished with exit code {}. Error:", exitCode, error);
         }
+    }
+
+    private static String getWitnessFilename(File progFile, BaseOptions options, String postfix) {
+        return options.hasWitnessFilename()
+                ? options.getWitnessFilename() + postfix
+                : Utils.getNameWithoutExtension(progFile);
     }
 }
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
@@ -74,6 +74,7 @@ public class Dartagnan extends BaseOptions {
         final TaskResultAnalyzer resultAnalyzer = TaskResultAnalyzer.create();
         ResultSummary summary = null;
         for (File progFile : progFiles) {
+            System.out.println(progFile);
             try {
                 // ----------- Generate verification task -----------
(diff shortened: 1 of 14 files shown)


dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/WitnessGenerationTest.java
```diff
@@ -0,0 +1,84 @@
...
+        File expectedFile = Paths.get(dat3mHome, "output", programName + ".png").toFile();
+
+        ProcessBuilder pb = new ProcessBuilder(
+            "java",
+            "-jar",
+            dat3mHome + "/dartagnan.jar",
+            "cat/vmm.cat",
+            dat3mHome + "/dartagnan/src/test/resources/locks/" + programName + ".ll",
+            "--witness=png"
+        );
```
ThomasHaas: I think this code could be put into a utlity method including the retrieval of `DAT3M_HOME`.
Something like
```
ProcessBuilder Dartagnan.buildProcess(String catPath, String program Path, String... options)
```
Or a variant that also starts and waits for the process.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/WitnessGenerationTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class WitnessGenerationTest {
+
+    @Test
+    public void testGeneratesDefaultWitness() throws Exception {
+        final String dat3mHome = System.getenv("DAT3M_HOME");
+        assertNotNull("DAT3M_HOME environment variable is not set!", dat3mHome);
+        
+        final String programName = "ttas-acq2rx";
+        File expectedFile = Paths.get(dat3mHome, "output", programName + ".png").toFile();
```
ThomasHaas: Use `Path.of` instead of `Paths.get` (the former is more modern). 

Also, you don't need to convert `Path` to a `File`. `Path` is the modern version of `File` and should be preferred. We should probably refactor our code eventually and replace all usage of `File` by `Path`.

hernanponcedeleon:
I did not anticipate that some changes to the CI were also needed, but now indeed there is only a single failure related to #1036.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -185,6 +185,30 @@ private static List<File> getProgramFiles(Path path) {
         }
     }
 
+    public static void runAsApplication(Path programPath, Path catPath, String... options) throws Exception {
+        final Path dat3mHome = Path.of(System.getenv("DAT3M_HOME"));
```
ThomasHaas: We shouldn't have the hardcoded environment variable here. 
`GlobalSettings` contains helper functions to get the correct variables, though I think none of them fit for this particular use case yet.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/WitnessGenerationTest.java
```diff
@@ -0,0 +1,58 @@
...
+public class WitnessGenerationTest {
+
+    private final Path dat3mHome = Path.of(System.getenv("DAT3M_HOME"));
+    private final Path testDir = dat3mHome.resolve("dartagnan")
+                                     .resolve("src")
+                                     .resolve("test")
+                                     .resolve("resources")
+                                     .resolve("locks");
+    private final Path catPath = dat3mHome.resolve("cat").resolve("vmm.cat");
+    private final Path outputDir = dat3mHome.resolve("output");
```
ThomasHaas: Same issue with hardcoded paths. 
At least the home directory and the cat directory should come from `GlobalSettings`, possibly also the output directory. Maybe we should even have a method to retrieve the `resources` directory.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
hernanponcedeleon: `GlobalSettings` returned different paths in application vs test mode, so I have to make some further changes. Please check it did not end up over-complicating the solution.

ThomasHaas: I know, that's why I said that there is no fitting function yet :)

svcomp/src/main/java/com/dat3m/svcomp/SVCOMPRunner.java
```diff
@@ -95,16 +96,16 @@ public static void main(String[] args) throws Exception {
...
                 cmd.add("-DLOGNAME=" + Files.getNameWithoutExtension(programPath));
-                cmd.add("-Djava.library.path=" + System.getenv().get("DAT3M_HOME") + "/dartagnan/target/libs/");
+                cmd.add("-Djava.library.path=" + getHomeDirectory() + "/dartagnan/target/libs/");
             } else {
                 cmd.add("java");
                 cmd.add("-DlogLevel=info");
                 cmd.add("-DLOGNAME=" + Files.getNameWithoutExtension(programPath));
                 cmd.add("-jar");
-                cmd.add(System.getenv().get("DAT3M_HOME") + "/dartagnan/target/dartagnan.jar");
+                cmd.add(getHomeDirectory() + "/dartagnan/target/dartagnan.jar");
```
ThomasHaas: We should also use the `Path` class here to construct the paths. Maybe helper methods to get the Dartagnan jar/native paths are also reasonable.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/WitnessGenerationTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    public void testGeneratesDefaultWitness() throws Exception {
+        final String programName = "ttas-acq2rx";
+        final Path programPath = testDir.resolve(programName + ".ll");
+        final Path witnessPath = outputDir.resolve(programName + ".png");
+
+        runDartagnanApplication(programPath, catPath, String.format("--%s=png", WITNESS));
+        assertTrue("Witness not found at: " + witnessPath.toAbsolutePath().toString(), Files.exists(witnessPath));
+    }
```
ThomasHaas, line 32: Do these tests also run locally? I guess it is not too big of a deal, but they will generate a permanent file in the output directory.

svcomp/src/main/java/com/dat3m/svcomp/SVCOMPRunner.java
```diff
@@ -153,4 +153,9 @@ private static List<String> filterOptions(Configuration config) {
             collect(Collectors.toList());
     }
     
+    public static String getNameWithoutExtension(Path path) {
+       String fileName = path.getFileName().toString();
+        int dotIndex = fileName.lastIndexOf('.');
+        return (dotIndex == -1) ? fileName : fileName.substring(0, dotIndex);
+    }
```
ThomasHaas, line 158: `Files.getNameWithoutExtension` already exists :)

hernanponcedeleon: yes, but that one is in `com.google.common.io.Files` which clashes with `java.nio.file.Files`. I can use the fully qualified name if you think it is cleaner.

ThomasHaas: It's better than duplicating the method here at least. But I get the point.
We could have a wrapper in `Utils`, i.e., `Utils.getNameWithoutExtension(...)` which simply calls the google version. Then it can be used anywhere without causing name clashes.

ThomasHaas: Btw. even in this particular case, you could have implemented your own `getNameWithoutExtension` by just calling the google version with fully-qualified name. Then you only write the fully qualified name once, and the one time you write it out it is still a lot shorter than the whole reimplementation :P

ThomasHaas (approved):
LGTM

Issue #1026, opened 2026-04-15
Non-descriptive error msg when trying to parse unknown relations


hernanponcedeleon:
```
> cat lala.cat 
let a = b & rf
acyclic a

> java -jar dartagnan/target/dartagnan.jar lala.cat benchmarks/locks/ttas.c
Test: benchmarks/locks/ttas.c
Result: ERROR
Reason: IllegalArgumentException
Details:
        Non-unary relation in [[b] := b, [rf] := rf]
```

xeren:
Explanation:  The parser treats all undefined names as implicitly-predefined tag sets.  In this example, `b` is interpreted as a tag set, while `rf` is explicitly predefined as a binary relation.  Trying to built the intersection of the two fails.

The message is different with `let rec a = b & rf`:
```
Test: benchmarks/locks/ttas.c
Result: ERROR
Reason: ParsingException
Details:
        Incompatible kinds UNARY and BINARY
```
The example should rather throw a `ParsingException` as well.  Its details should at least tell that this is related to `lala.cat`.

hernanponcedeleon:
> The message is different with `let rec a = b & rf`:

I do not get a different msg with this definition
```
> cat lala2.cat 
let rec a = b & rf
acyclic a

> java -jar dartagnan/target/dartagnan.jar lala2.cat benchmarks/locks/ttas.c
Test: benchmarks/locks/ttas.c
Result: ERROR
Reason: IllegalArgumentException
Details:
        Non-binary relation in [[b] := b, [rf] := rf]
```

hernanponcedeleon:
Ping @xeren

xeren:
Seems like my example was too small and `rf` had not made it into the parser's `namespace` as I expected from `stdlib.cat`.  These should now show what I meant.
```
let rf1 = rf
let a = b & rf // Non-binary relation in [[b] := b, [rf, rf1] := rf]
```
```
let rf1 = rf
let a = b & rf1 // Non-binary relation in [[b] := b, [rf, rf1] := rf]
```
```
let rf1 = rf
let rec a = b & rf // Non-binary relation in [[b] := b, [rf, rf1] := rf]
```
```
let rf1 = rf
let rec a = b & rf1 // Incompatible kinds UNARY and BINARY
```

hernanponcedeleon:
> Explanation: The parser treats all undefined names as implicitly-predefined tag sets. 

[...]

PR #1038, opened 2026-05-29
Better messages for errors during Cat parsing.

xeren:
Gives better error messages during parsing of `.cat` files.  Adds the file name and the line where the error was encountered.  Solves #1026.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
@@ -48,6 +48,9 @@ class VisitorCat extends CatBaseVisitor<Object> {
     private final Map<String, Integer> nameOccurrenceCounter = new HashMap<>();
     // Used to handle recursive definitions properly
     private Relation relationToBeDefined;
+    // Used for error messages.
+    private String currentFileName;
+    private ParserRuleContext currentInstructionContext;
 
     private record FuncDefinition(String name, List<String> params, String expression,
                                   Map<String, Object> capturedNamespace) {
@@ -61,9 +64,10 @@ class VisitorCat extends CatBaseVisitor<Object> {
         }
     }
 
-    VisitorCat(Path includePath) {
+    VisitorCat(Path includePath, String currentFileName) {
         this.includePath = includePath;
         this.wmm = new Wmm();
+        this.currentFileName = currentFileName;
         includeStdlib();
     }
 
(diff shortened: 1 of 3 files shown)


hernanponcedeleon:
I think it would make sense to report `Incompatible kinds UNARY and BINARY` (+ add the relation's names) rather than the current error. 

On the one side, it makes it clearer that the arity of relations does not match, rather than explicitly assuming one must be unary/binary.

On the other side, the error msg is different depending on if there is recursion or not involved (which should not have any impact)
```
> cat test.cat 
let rf1 = rf
let rec a = b & rf1

> dartagnan test.cat test.c
Test: test.c
Result: ERROR
Reason: ParsingException
Details:
        Non-binary relation in [[b] := b, [rf, rf1] := rf] (test.cat at line 2)

> cat test.cat 
let rf1 = rf
let a = b & rf1

> dartagnan test.cat test.c
Test: test.c
Result: ERROR
Reason: ParsingException
Details:
        Non-unary relation in [[b] := b, [rf, rf1] := rf] (test.cat at line 2)
```

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -471,15 +487,32 @@ public Relation.Arity visitExprMinus(ExprMinusContext c) {
...
+        final Relation relationFromGlobalNamespace = wmm.getRelation(name);
+        if (relationFromGlobalNamespace != null) {
+            return relationFromGlobalNamespace;
+        }
+        return RelationNameRepository.contains(name) ? wmm.getOrCreatePredefinedRelation(name) : null;
+    }
+
+    private ParsingException parsingException(ParserRuleContext ctx, Throwable cause, String message, Object... arguments) {
+        final String format = message + " (%s at line %d)".formatted(currentFileName, ctx.getStart().getLine());
+        return new ParsingException(cause, format, arguments);
```
hernanponcedeleon: This looks a bit abusive to me: we end up with calls like `parsingException(e, e.getMessage(), ...)` where the string is not necessary, or calls like `parsingException(null, reason, ...)` with the ugly `null` passing.

The only downside I see of having two methods with the correct signature is that we need to keep them consistent on the format of how we report the error.

xeren: You could call it abusive.  I prefer nullable parameters over overrides, because it forces programmers to give some argument.  It shows if they forgot something.

The best I can think of is to have their method bodies be short and positioned close to each other.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -97,23 +107,29 @@ public Object visitLetFuncDefinition(LetFuncDefinitionContext ctx) {
...
         }
 
+        final String currentFileParent = currentFileName;
         try {
             final CatParser parser = getParser(CharStreams.fromPath(filePath));
+            currentFileName = fileName;
             return parser.mcm().accept(this);
         } catch (IOException e) {
-            throw new ParsingException(e, String.format("Error parsing file '%s'", filePath));
+            throw parsingException(ctx, e, "Error parsing file '%s'", filePath);
```
hernanponcedeleon: Why is `"Error parsing file '%s'"` a better msg than `e.getMessage()` in this case?

If we use `e.getMessage()`, AFAICT all calls to `parsingException` passing the `Throwable ` would always use `e.getMessage()` which means we can simplify the signature of the method.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
hernanponcedeleon: Reading my own msg out loud made me realized it sounded quite harsh. This was not the intention. Sorry about that

PR #1041, opened 2026-06-11
Fix weekly job

hernanponcedeleon:
The weekly job runs `WitnessGenerationTest` which requires `dartagnan.jar` to exists and the container to have graphviz installed.


ThomasHaas (approved):
LGTM

PR #1044, opened 2026-06-18
Compact sets in `RelationAnalysis`

xeren:
- Adds `IndexedDomain<T>`, a list of distinct elements with a hashtable-powered `indexOf` method.
- Adds `IndexedSet<T>`, an implementation of `Set<T>` that uses `IndexedDomain<T>` and a bit array to describe its contents.
- Memoizes the results of `ExecutionAnalysis` (EA) with `IndexedSet`.
- Replaces usages `HashSet<T>` with `IndexedSet<T>` in `RelationAnalysis` (RA) and `ActiveSetAnalysis` (ASA).
- Adds ~`VisibilityAnalysis`~ `RelationEventDomains` to allow using a smaller *domain* on most relations in `RelationAnalysis`.
- Enhances the precision of the Extended Relation Analysis (XRA):
  Given a transitive closure $r\mathrel{:=}s^+$.  The event pair set $\delta$ has previously been disabled in $s$, such that $\mathsf{may}(s)\cap\delta$ is empty.
  The bottom-up propagation should compute $\mathsf{may}(r)\setminus\mathsf{may}(s)^+$.
  In development, the result was smaller $(\delta\cup(\delta\mathrel;\mathsf{may}(r)))\setminus\mathsf{may}(s)^+$.
  Without this, XRA and ASA produce the same results as `development` in all tests.
- *Edit:* Adds `EventDomainRepository` to create domains to be used by analyses.

**Table 1** - Performance for suite `benchmarks/challenging` in seconds (only the mentioned analyses).  For each pair of rows, the upper describes `development` and the lower describes this PR.
| Benchmark | EA | RA | XRA | ASA | Total
| - | - | - | - | - | -
| `cna.c rc11.cat` | 0.0 | 2.2 | 2.0 | 6.6 | 10.8
| | 0.1 | 0.2 | 0.5 | 0.3 | 1.1
| `cna.c imm.cat` | 0.0 | 1.4 | 1.3 | 5.0 | 7.7
| | 0.1 | 0.3 | 0.3 | 0.3 | 1.0
| `harris-3.c rc11.cat` | 0.0 | 86 | 159 | 381 | 626
| | 1.9 | 1.8 | 6.1 | 5.7 | 15.5
| `harris-3.c imm.cat` | 0.0 | 58 | 68 | 395 | 521
| | 2.0 | 1.9 | 3.7 | 5.0 | 12.6
| `wsq.c rc11.cat` | 0.0 | 4.0 | 16.8 | 8.2 | 29.0
| | 0.2 | 0.2 | 1.0 | 0.6 | 2.0
| `wsq.c imm.cat` | 0.0 | 4.5 | 12.6 | 6.7 | 23.8
| | 0.2 | 0.3 | 0.7 | 0.5 | 1.7


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -151,26 +155,26 @@ private void run() {
         // ---- Initialize active set propagation queue ----
         final Map<Relation, List<EventGraph>> propagationQueue = new HashMap<>();
         axioms.forEach(a -> {
-            final EventGraph relevant = filterUnknowns(axiom2RelevantSets.get(a), a.getRelation());
+            final EventGraph relevant = filterUnknowns(new IndexedEventGraph(axiom2RelevantSets.get(a)), a.getRelation());
```
hernanponcedeleon: Why now you need to create a copy of the graph (via `new`) but before this was not the case?

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -365,29 +379,29 @@ public Map<Relation, EventGraph> visitComposition(Composition comp) {
...
+                final Set<Event> e1may1 = k1.getMaySet().getRange(e1);
+                for (Event e3 : news.getRange(e1)) {
+                    final Set<Event> e2Set = new IndexedSet<>((IndexedSet<Event>) may2In.getRange(e3));
+                    e2Set.retainAll(e1may1);
+                    set1.addRange(e1, e2Set);
+                    set2In.addRange(e3, e2Set);
                 }
-            });
+            }
+            final MutableEventGraph set2 = set2In.inverse();
```
hernanponcedeleon: I find the naming convention `In` weird.  At some point I thought it was about inverse, but then you use it in part that has nothing to do with inverse, and there are other inverse-related ones that do not use `In`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -437,27 +451,27 @@ public Map<Relation, EventGraph> visitProduct(CartesianProduct product) {
...
-                        }
-                        if (!k0.getMustSet().contains(e, e2)) {
-                            factors.add(e, e2);
-                        }
-                    }
+            final EventGraph k0MayIn = k0.getMaySet().inverse();
+            for (Event e1 : news.getDomain()) {
+                final Set<Event> e1k0May = k0.getMaySet().getRange(e1);
+                for (Event e3 : news.getRange(e1)) {
+                    final IndexedSet<Event> e2Set = exec.eventDomain().newSet(k0MayIn.getRange(e3));
```
hernanponcedeleon: why e2 as var name?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/Dependency.java
```diff
@@ -158,7 +158,7 @@ private static State process(Event reader, Set<Writer> state, Register register,
         List<RegWriter> candidates = state.stream()
                 .filter(e -> e.register.equals(register))
                 .map(e -> e.event)
-                .filter(e -> reader == null || !exec.areMutuallyExclusive(reader, e))
+                .filter(e -> reader == null || e == null || !exec.areMutuallyExclusive(reader, e))
```
hernanponcedeleon, line 161: All these null check are ugly. Why can we get `null` here now?

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
```diff
@@ -5,23 +5,36 @@
...
+    IndexedDomain<Event> eventDomain();
+    Set<Event> implyingEvents(Event implied);
+    Set<Event> excludingEvents(Event event);
 
     static ExecutionAnalysis fromConfig(Program program, ProgressModel.Hierarchy progressModel, Context context, Configuration config)
             throws InvalidConfigurationException {
         final BranchEquivalence eq = context.requires(BranchEquivalence.class);
-        return new DefaultExecutionAnalysis(program, eq, progressModel);
+        final var exec = new DefaultExecutionAnalysis(program, eq, progressModel);
+        exec.run();
```
hernanponcedeleon: Why do we now need to run it but it was not the case before?

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedSet.java
```diff
@@ -0,0 +1,442 @@
+package com.dat3m.dartagnan.utils.collections;
+
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/// Compact subset of a collection using a bitvector.
+public final class IndexedSet<E> extends AbstractSet<E> {
+    // assert Arrays.stream(domain).allMatch(o -> o instanceof E)
```
hernanponcedeleon: Remove all commented assertions

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedSet.java
```diff
@@ -0,0 +1,442 @@
...
+            }
+            if (!value) {
+                return false;
+            }
+        }
+        ensureMemberArray();
+        final long mask = 1L << (index % 64);
+        final boolean change = value == ((member[index / 64] & mask) == 0L);
+        member[index / 64] ^= change ? mask : 0L;
+        //shrinkIfPossible(); // This would be too costly.
```
hernanponcedeleon: Remove (same in other parts of the code)

hernanponcedeleon:
I get the following error
```
> java -jar dartagnan/target/dartagnan.jar cat/imm.cat benchmarks/locks/ttas.c --wmm.analysis.relationAnalysis=lazy
Test: benchmarks/locks/ttas.c
Result: ERROR
Reason: ClassCastException
Details:
        class com.dat3m.dartagnan.wmm.utils.graph.immutable.ImmutableMapEventGraph cannot be cast to class com.dat3m.dartagnan.wmm.utils.graph.mutable.IndexedEventGraph (com.dat3m.dartagnan.wmm.utils.graph.immutable.ImmutableMapEventGraph and com.dat3m.dartagnan.wmm.utils.graph.mutable.IndexedEventGraph are in unnamed module of loader 'app')
```

hernanponcedeleon (commented):
First pass up to `IndexedSet.java`

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
xeren: This line fits into the `inverse` scheme: `set2` is the inverse of its inverse.  But primarily, I named it to refer to the 'incoming' edges as in `getInMap()`.  (`set1` and `set2` could as well be named `set1Out` and `set2Out`.)  But  `set2Inverse` (and `may2Inverse`) may be better names.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
xeren: Because it better describes the position of the event between the iterated event pair.  `e2set` computes the set `{ x | e1 -may(r0)> x -may(r0)> e3 }`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
xeren: Seems to be a relict from before I changed the type of `axiom2RelevantSets`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
xeren: This is an eager memoization.  It avoids managing partial results and benefits from structure properties.  E.g. external events are only mutually exclusive if the `ThreadCreate` of one's thread is mutually exclusive to the other event.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/Dependency.java
xeren: We got `null` before.  Appearing in `Writer.event`, it states that `register` may be uninitialized.  `exec.areMutuallyExclusive(reader, e)` just silently returned `false` before, because it was using `HashMap`, which allowed null keys and missed the check.  `IndexedSet` does no longer allow null keys and throws `NullPointerException`, as should be expected of `exec`.
This seems to be the only place in the project where `null` is passed.

xeren:
> I get the following error
> 
> ```
> > java -jar dartagnan/target/dartagnan.jar cat/imm.cat benchmarks/locks/ttas.c --wmm.analysis.relationAnalysis=lazy
> Test: benchmarks/locks/ttas.c
> Result: ERROR
> Reason: ClassCastException
> Details:
>         class com.dat3m.dartagnan.wmm.utils.graph.immutable.ImmutableMapEventGraph cannot be cast to class com.dat3m.dartagnan.wmm.utils.graph.mutable.IndexedEventGraph (com.dat3m.dartagnan.wmm.utils.graph.immutable.ImmutableMapEventGraph and com.dat3m.dartagnan.wmm.utils.graph.mutable.IndexedEventGraph are in unnamed module of loader 'app')
> ```

Fixed that now.

`LazyRelationAnalysis` does not yet benefit from the compact sets.  Once it does, it may still out-perform `NativeRelationAnalysis` in some large benchmarks.  Otherwise, we could remove it.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
ThomasHaas: I think your answer is very disconnected to what Hernán asked.
Let me translate what you said:
- The previous `ExecutionAnalysis` did not do any precomputations during construction.
- The new `ExecutionAnalysis` does some precomputations to more quickly answer queries.
- The new run method contains all the precomputations.

Now, the question is why `run` is not just called inside the constructor. I guess this is subjective preference, but I would also just call `run` from within the constructor.

xeren: Yes, as @ThomasHaas wrote.
`ExecutionAnalysis` had no `run` method before, because its results were computed lazily.  Its result can be represented by two boolean matrices.  The way `exec` is used in the RA and ASA is to find implying events of an intermediate event in a composition and to filter mutually-exclusive event pairs from all graphs.  This goes really fast if the results are available as `IndexedSet`.
The costs of creating of these sets, however, grows squaredly with the program size.  This is because they had to be complete.  So, I had `ExecutionAnalysis` keep them for reuse.
Profiling the `challenging` benchmarks showed me that 15 sec were still spent creating these sets.  With the eager computation in `run()`, I got this down to 2 sec.  This is why there is this method now.

> Now, the question is why `run` is not just called inside the constructor. I guess this is subjective preference, but I would also just call `run` from within the constructor.

No objections here.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedDomain.java
```diff
@@ -0,0 +1,101 @@
...
+public final class IndexedDomain<E> {
+
+    private static final float HASH_TABLE_FILL_FACTOR = 1.5f;
+
+    final Object[] elements;
+    final int[] index;
+    final Map<IndexedDomain<?>, Boolean> compatibilityMap = new HashMap<>();
+
+    public IndexedDomain(Collection<E> elements) {
+        this.elements = elements.toArray();
```
ThomasHaas: I would use `E[] elements` and change the constructor to use `this.elements = elements.toArray((E[]) new Object[0]);` together with a `@SuppressWarnings("unchecked")` attribute.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedSet.java
```diff
@@ -0,0 +1,462 @@
...
+// - In its <i>array form</i>,
+//   it contains a bit array where the bit at position `domain.indexOf(e)` is set iff `e` is contained.
+// - It can only maintain its <i>loneMember form</i>,
+//   if it contains up to one element, in which case it stores its index.
+// Instances can be in <i>array form</i>, even if they contain less than two elements.
+// This happens e.g. after `addAll(Set.of(e1))` and `addAll(Set.of(e1,e2)); remove(e1)`.
+// This is because the check would be too expensive.
+public final class IndexedSet<E> extends AbstractSet<E> {
+    private final IndexedDomain<E> domain;
+    private final Object[] elements;
```
ThomasHaas: With the change to `IndexedDomain.elements`, you can trivially change this here to `E[]` and then get rid of all the casts in the code.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -272,36 +275,47 @@ public EventGraph visitAcyclicity(Acyclicity axiom) {
         // if the clauses {@code exec(x) implies before(x,y)} and {@code exec(z) implies before(y,z)} exist.
         // NOTE: Assumes that the must-set of rel+ is acyclic.
         private static EventGraph transitivelyDerivableMustEdges(ExecutionAnalysis exec, RelationAnalysis.Knowledge k) {
-            MutableEventGraph result = new MapEventGraph();
-            Map<Event, Set<Event>> map = new HashMap<>();
-            Map<Event, Set<Event>> mapInverse = new HashMap<>();
+            final MutableEventGraph result = new IndexedEventGraph(exec.eventDomain());
+            final IndexedDomain<Event> eventDomain = k.getMaySet() instanceof IndexedEventGraph g ? g.eventDomain()
+                    : exec.eventDomain();
```
hernanponcedeleon: I assume this is an optimization to use a smaller domain when possible. Can `g.eventDomain()` be smaller than `exec.eventDomain()`?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/utils/graph/mutable/IndexedEventGraph.java
```diff
@@ -0,0 +1,301 @@
...
+
+    public IndexedEventGraph(IndexedDomain<Event> d, EventGraph g) {
+        this(d, 0);
+        addAll(g);
+    }
+
+    private IndexedEventGraph(IndexedDomain<Event> d, int ignore) {
+        eventDomain = d;
+        emptySet = d.emptySet();
+        map = new IndexedSet[d.size()];
```
ThomasHaas: You can do something like `(IndexedSet<Event>[]) Array.newInstance(IndexedSet.class, d.size());` to create the array of type `(IndexedSet<Event>[])`. Then you can change the type of `map` to be more precise.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
```diff
@@ -31,23 +44,106 @@ static ExecutionAnalysis fromConfig(Program program, ProgressModel.Hierarchy pro
...
+    public Set<Event> implyingEvents(Event implied) {
+        return eventsByImpliedEvent.get(implied);
+    }
+
+    @Override
+    public Set<Event> excludingEvents(Event excluded) {
+        return eventsByExcludingEvent.get(excluded);
+    }
+
+    void run() {
```
hernanponcedeleon:
```suggestion
    private void run() {
```

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedDomain.java
```diff
@@ -0,0 +1,101 @@
+package com.dat3m.dartagnan.utils.collections;
+
+import java.util.*;
+
+/// Used for compact sets, see `IndexedSet`.
+// This implementation contains an immutable hash table to look up an index by element.
+public final class IndexedDomain<E> {
```
ThomasHaas, line 7: I'm surprised that you put a general-purpose hash table implementation here.
For our use case, we only ever instantiate this with `Event` and only ever after the program has been fixed. In that case, a simple array `int[] id2Index` based on`Event.getGlobalId()` could have replaced the hash table even more efficiently.

Anyway, I'm not complaining about the more general implementation, you can keep it :)

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
```diff
@@ -31,23 +44,106 @@ static ExecutionAnalysis fromConfig(Program program, ProgressModel.Hierarchy pro
...
+        computeMutuallyExclusiveEvents();
+    }
+
+    private void computeImplyingEvents() {
+        final Map<Thread, IndexedSet<Event>> threadEvents = new HashMap<>();
+        for (Thread thread : program.getThreads()) {
+            final var events = eventDomain.newSet();
+            events.addAll(thread.getEvents());
+            threadEvents.put(thread, events);
+        }
```
hernanponcedeleon: Can there really be a missmatch between `eventDomain` and all the events in the program?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -69,6 +76,14 @@ protected NativeRelationAnalysis(VerificationTask t, Context context, Configurat
         definitions = context.requires(ReachingDefinitionsAnalysis.class);
         alias = context.requires(AliasAnalysis.class);
         wmmAnalysis = context.requires(WmmAnalysis.class);
+        visibility = VisibilityAnalysis.newInstance(task.getMemoryModel());
+        allEvents = exec.eventDomain().newSet();
+        allVisibleEvents = new IndexedDomain<>(task.getProgram().getThreadEventsWithAllTags(VISIBLE)).newSet();
+        EMPTY = new Delta(new IndexedEventGraph(exec.eventDomain()), new IndexedEventGraph(exec.eventDomain()));
+        mutex = new IndexedEventGraph(exec.eventDomain());
+        if (!allEvents.domain().isCompatibleWith(allVisibleEvents.domain())) {
+            logger.warn("Incompatible IndexDomain.  Analysis performance may suffer.");
```
hernanponcedeleon:
```suggestion
            logger.warn("Incompatible IndexDomain. Analysis performance may suffer.");
```

hernanponcedeleon: Why can this affect performance? do you need to resize some of the domains?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -724,8 +732,9 @@ public MutableKnowledge visitLXSXPairs(LXSXPairs lxsx) {
                     final int start = iterate(end - 1, i -> i >= 0, i -> i - 1)
                             .filter(i -> exec.isImplied(store, events.get(i)))
                             .findFirst().orElse(0);
+                    final Set<Event> storeMutex = execExcluding(store);
                     final List<Event> candidates = events.subList(start, end).stream()
-                            .filter(e -> !exec.areMutuallyExclusive(e, store))
+                            .filter(e -> !storeMutex.contains(e))
```
hernanponcedeleon: Why this change? Are just repeating here how `areMutuallyExclusive` is implemented? (same in several other places)

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -1551,15 +1525,15 @@ public Delta visitDifference(Difference diff) {
...
         }
 
         @Override
         public Delta visitComposition(Composition comp) {
             final Relation r1 = comp.getLeftOperand();
             final Relation r2 = comp.getRightOperand();
-            MutableEventGraph maySet = new MapEventGraph();
-            MutableEventGraph mustSet = new MapEventGraph();
+            MutableEventGraph maySet = newGraphWithDomain(getKnowledge(comp.getDefinedRelation()).getMaySet());
+            MutableEventGraph mustSet = newGraphWithDomain(getKnowledge(comp.getDefinedRelation()).getMustSet());
```
hernanponcedeleon: Extract `getKnowledge(comp.getDefinedRelation())` to a variable

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -1571,18 +1545,22 @@ public Delta visitComposition(Composition comp) {
             return new Delta(maySet, mustSet);
         }
 
-        private void computeComposition(MutableEventGraph result, EventGraph left, EventGraph right, final boolean isMay) {
+        private void computeComposition(MutableEventGraph result, EventGraph left, EventGraph right, boolean isMay) {
+            final IndexedDomain<Event> eventDomain = ((IndexedEventGraph) result).eventDomain();
```
hernanponcedeleon: If the casting is safe, why no changing the parameter type?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
```diff
@@ -1702,4 +1676,36 @@ public MutableEventGraph getMustSet() {
...
+    }
+
+    private IndexedEventGraph newGraph(Relation relation) {
+        final IndexedSet<Event> events = visibility.mayHaveInvisibleEvents(relation) ? allEvents : allVisibleEvents;
+        return new IndexedEventGraph(events.domain());
+    }
+
+    private MutableEventGraph newGraphWithDomain(EventGraph other) {
+        return new IndexedEventGraph(((IndexedEventGraph) other).eventDomain());
+    }
```
hernanponcedeleon: If the cast is safe, why not changing the paremeter type?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/VisibilityAnalysis.java
```diff
@@ -0,0 +1,144 @@
...
+        private Result re(boolean invisibleDomain, boolean invisibleRange) {
+            return invisibleDomain ? invisibleRange ? Result.YY : Result.YN : invisibleRange ? Result.NY : Result.NN;
+        }
+    }
+
+    private enum Result { NN, NY, YN, YY }
+
+    private boolean isInvisibleTag(String tag) {
+        return switch (tag) {
+            case VISIBLE, MEMORY, READ, WRITE, FENCE, INIT -> false;
```
hernanponcedeleon: The following GPU events should probably also be considered not invidible
```
        public static final String CBAR = "CBAR";
        public static final String AVDEVICE = "AVDEVICE";
        public static final String VISDEVICE = "VISDEVICE";
```

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
ThomasHaas: I think `exec.eventDomain()` contains all events, ordered so that that all visible events come first and then the invisible one's.
`g.eventDomain()` can be the subset of visible events for many relations (only exceptions are idd, addrDirect, etc.).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
ThomasHaas: I think compatibility says that both sets order the elements in the same way and that the same bit positions describe the same elements.
The orderedness alone can make checks like `containsAll` faster (a single iteration through the collection works). Together with the bitset representation, you can even compare whole chunks with a single operation.
I think when the sets are incompatible, they do not need to get resized or anything, but more inefficient algorithms need to be used for comparisons.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/VisibilityAnalysis.java
```diff
@@ -0,0 +1,144 @@
...
+        @Override
+        public Result visitConstraint(Constraint constraint) { return Result.NN; }
+        @Override
+        public Result visitInternalDataDependency(DirectDataDependency idd) { return Result.YY; }
+        @Override
+        public Result visitAddressDependency(DirectAddressDependency addr) { return Result.YN; }
+        @Override
+        public Result visitControlDependency(DirectControlDependency ctrl) { return Result.YY; }
+        @Override
+        public Result visitFree(Free free) { return Result.YY; }
```
ThomasHaas: The `Free` definition only connects visible events (implicitly enforced by RA's may sets).

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationEventDomains.java
```diff
@@ -0,0 +1,144 @@
...
+        @Override
+        public Result visitAddressDependency(DirectAddressDependency addr) { return Result.YN; }
+        @Override
+        public Result visitControlDependency(DirectControlDependency ctrl) { return Result.YY; }
+        @Override
+        public Result visitFree(Free free) { return Result.YY; }
+        @Override
+        public Result visitProgramOrder(ProgramOrder po) {
+            // This could be more accurate.
+            return re(!(po.getFilter() instanceof TagFilter) || isInvisibleTag(po.getFilter().toString()));
```
ThomasHaas, line 89: I think we should just get rid of the filter for po. We only ever instantiate it with `Tag.VISIBLE` and I'm not sure if we ever want to do it differently.

If we decide to get rid of it, we could do that in a separate, quick PR

hernanponcedeleon:
You can also remove [this](https://github.com/hernanponcedeleon/Dat3M/blob/d8de5e8ef6aa4a697ff2bbe88fa8416dfcd4ac7c/svcomp/svcomp.properties#L4) as part of this PR

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedDomain.java
xeren: I have other element classes in mind, like `EventData`/`EventModel` and `InclusionBasedPointerAnalysis.Variable`.  Something like `GloballyIdentifiable` could handle the dependency injection.
```java
public interface GloballyIdentifiable {
    public int getGlobalId();
}
```

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
xeren: Exactly.  This is why I added `VisibilityAnalysis`.  There is currently only a two-class partitioning.  It should be possible to add more classes, like `W \ IW` -> `M \ IW` -> `_ \ IW` -> `_` -> `_ | RegReader | RegWriter` -> all.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
xeren: This uses a pre-fetched `storeMutex`.  I added this early in this MR, when `ExecutionAnalysis` was still lazy and `storeMutex` was created inside `NativeRelationAnalysis`.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
```diff
@@ -272,36 +275,47 @@ public EventGraph visitAcyclicity(Acyclicity axiom) {
         // if the clauses {@code exec(x) implies before(x,y)} and {@code exec(z) implies before(y,z)} exist.
         // NOTE: Assumes that the must-set of rel+ is acyclic.
         private static EventGraph transitivelyDerivableMustEdges(ExecutionAnalysis exec, RelationAnalysis.Knowledge k) {
```
ThomasHaas: Hernán noticed that this method takes a while: it can be 20x slower than the whole RA???
I don't know why exactly this happens, but we should redo the implementation anyhow.
It is correct (and possibly even more precise) to just return `must+;must+`, i.e., compute a single transitive closure followed by a single composition of that closure with itself.
Since RA already has methods to do so, I tried copying them over and using them with slight modifications (the original methods were instance methods, but I need static one's here) like this:

```Java
final EventGraph must =  k.getMustSet();
final var transClosure = computeTransitiveClosure(must, newGraph(must), false, exec);
final MutableEventGraph res = newGraphWithDomain(must);
computeComposition(res, transClosure, transClosure, false, exec);
return res;
```
As expected, this runs over 20x faster and gave the same result.
I don't know how to best go about this. Maybe make RA provide public methods to do basic event graph operations, possibly as static utility methods? I think static methods would be great as I can imagine them being useful in more cases.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/VisibilityAnalysis.java
xeren: This is for control barriers that have their `F` tag removed, right?

hernanponcedeleon: Yes (relaxed barriers are not considered fences). From [here](https://github.com/KhronosGroup/Vulkan-MemoryModel/blob/9f3fdaaf6bf93fc95f38cae9726b6161e036b81c/alloy/spirv.als#L125) I think it is clear that visible events are for vulkan.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
xeren: Yes.  I inserted this check when I moved the creation of `allEvents.domain()` into `ExecutionAnalysis` while `allVisibleEvents.domain()` remained here in NRA.  A better approach would be to handle both in the same class.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
xeren: No, I do not expect it here.  It is possible to exclude annotations and labels from the index, though.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
xeren: I am thinking about `ra.computeTransitiveClosure(must).getMustSet()` and `ra.computeComposition(transClosure, transClosure).getMustSet()`.
The computation would then happen inside `getMustSet()`.

ThomasHaas: Can you clean up the comment on this function to something like:
```
// The transitiviely derivable must-edges are computed via (must(k)^+;must(k)^+)
```
All other parts of the comment are irrelevant. In particular, I think the comment about acyclicity should get moved: the assumption of acyclicity is only crucial on the caller's side to compute a correct transitive reduction.

dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
```diff
@@ -31,23 +40,101 @@ static ExecutionAnalysis fromConfig(Program program, ProgressModel.Hierarchy pro
...
+        this.program = program;
         this.eq = eq;
         this.progressModel = progressModel;
 
         this.lowestIdThread = program.getThreads().stream().min(Comparator.comparingInt(Thread::getId)).get();
-    }
 
-    private boolean isSameThread(Event a, Event b) {
-        return a.getThread() == b.getThread();
+        computeResults(domainRepository.getDomain(EventDomainRepository.DomainBound.ALL));
```
ThomasHaas: I would pass `program` to `computeResults` instead of making it a member because after the constuctor has run, the `program` member variable is essentially dead.

hernanponcedeleon:
```
dartagnan cat/svcomp.cat --method=eager ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c --wmm.analysis.relationAnalysis=lazy
Test: ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c
Result: ERROR
Reason: IllegalArgumentException
Details:
        Missing domain for intersection.
```

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
hernanponcedeleon: You still have the casting here

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/SetUtil.java
```diff
@@ -14,6 +14,20 @@ public static <T> Set<T> identityHashSet(int capacity) {
         return Collections.newSetFromMap(new IdentityHashMap<>(capacity));
     }
 
+    public static <T> Set<T> intersection(Set<T> left, Set<T> right) {
+        if (left instanceof IndexedSet<T> || right instanceof IndexedSet<T>) {
+            return IndexedSet.intersection(left, right);
+        }
+        final Collection<T> filter = left.size() < right.size() ? right : left;
+        final Set<T> intersection = new HashSet<>();
+        for (T element : filter == right ? left : right) {
```
hernanponcedeleon: It would probably be more readable having
```
final Collection<T> iterator = left.size() < right.size() ? left : right;
for (T element : iterator) {
```

hernanponcedeleon:
I haven't yet checked in detail what you changed in the LAZY RA but there is a regression
```
> dartagnan -DlogLevel=info cat/svcomp.cat --method=eager ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c --wmm.analysis.relationAnalysis=lazy
[02.07.2026] 15:10:57 [INFO] EnvironmentInfo.logEnvironmentInfo - Git branch: development
[02.07.2026] 15:10:57 [INFO] EnvironmentInfo.logEnvironmentInfo - Git commit ID: a7e3e4843359dde3a0e29500a821030e2433e316
...
[02.07.2026] 15:11:11 [INFO] RelationAnalysis.fromConfig - Finished regular analysis in 12.988 secs
```
vs
```
> dartagnan -DlogLevel=info cat/svcomp.cat --method=eager ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c --wmm.analysis.relationAnalysis=lazy
[02.07.2026] 15:14:26 [INFO] GitInfo.logGitInfo - Git branch: indexedset
[02.07.2026] 15:14:26 [INFO] GitInfo.logGitInfo - Git commit ID: 9551006b0e8ea81be5b1cb809eccb654e1ec62d5
...
[02.07.2026] 15:15:07 [INFO] RelationAnalysis.fromConfig - Finished regular analysis in 39.260 secs
```

EDIT: actually, I do not think is a good idea to touch the lazy analysis. The performance of the native one now is so good that probably the main use case for lazy is to check both analyses compute the same results. For that, the current implementation of lazy should be good enough

xeren:
> ```
> dartagnan cat/svcomp.cat --method=eager ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c --wmm.analysis.relationAnalysis=lazy
> Test: ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c
> Result: ERROR
> Reason: IllegalArgumentException
> Details:
>         Missing domain for intersection.
> ```

Caused by `ActiveSetAnalysis`, where the new sets were expected.  I added some operations to `SetUtil`, where `IndexedSet` is preferred, but other sets still work.

> I haven't yet checked in detail what you changed in the LAZY RA but there is a regression
> 
> ```
> > dartagnan -DlogLevel=info cat/svcomp.cat --method=eager ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c --wmm.analysis.relationAnalysis=lazy
> [02.07.2026] 15:10:57 [INFO] EnvironmentInfo.logEnvironmentInfo - Git branch: development
> [02.07.2026] 15:10:57 [INFO] EnvironmentInfo.logEnvironmentInfo - Git commit ID: a7e3e4843359dde3a0e29500a821030e2433e316
> ...
> [02.07.2026] 15:11:11 [INFO] RelationAnalysis.fromConfig - Finished regular analysis in 12.988 secs
> ```
> 
> vs
> 
> ```
> > dartagnan -DlogLevel=info cat/svcomp.cat --method=eager ../sv-benchmarks/c/pthread-complex/elimination_backoff_stack-race.c --wmm.analysis.relationAnalysis=lazy
> [02.07.2026] 15:14:26 [INFO] GitInfo.logGitInfo - Git branch: indexedset
> [02.07.2026] 15:14:26 [INFO] GitInfo.logGitInfo - Git commit ID: 9551006b0e8ea81be5b1cb809eccb654e1ec62d5
> ...
> [02.07.2026] 15:15:07 [INFO] RelationAnalysis.fromConfig - Finished regular analysis in 39.260 secs
> ```
> 
> EDIT: actually, I do not think is a good idea to touch the lazy analysis. The performance of the native one now is so good that probably the main use case for lazy is to check both analyses compute the same results. For that, the current implementation of lazy should be good enough

I reverted the latest changes.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
xeren: Reverted the change.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
xeren: Removed the cast.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ActiveSetAnalysis.java
xeren: The method serves a non-trivial optimization.  I think the information is helpful and should either be where it is or at the call site.

The note about the acyclicity assumption could as well be located inside the function itself. If `relation` is cyclic, (which is possible for weird consistency models like `acyclic po | po^-1`) the method would cull some cycles entirely out of the encoding.

ThomasHaas: The method does not cull anything, it only computes some relation. It is only because the caller uses the method to compute a transitive reduction that he has to ensure acyclicity of the must-set. That's why I suggested moving the comment to the callsite.

If `transitivelyDerivableMustEdges` just took the must-set as input instead of the whole relation knowledge, the caller that tries to compute a transitive reduction of `r` could actually pass in any acyclic subset of `must(r)` and it would be correct. It's just that `must(r)` is usually acyclic (which is the assumption) and so it can be passed as is.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/RelationEventDomains.java
```diff
@@ -0,0 +1,156 @@
...
+import com.dat3m.dartagnan.wmm.Wmm;
+import com.dat3m.dartagnan.wmm.definition.*;
+import com.dat3m.dartagnan.wmm.utils.graph.mutable.IndexedEventGraph;
+
+import java.util.*;
+
+import static com.dat3m.dartagnan.program.event.Tag.*;
+import static java.util.Collections.disjoint;
+
+/// Estimates boundaries for the domains and ranges of relations.
```
hernanponcedeleon:
```suggestion
// Estimates boundaries for the domains and ranges of relations.
```

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Tearing.java
```diff
@@ -57,6 +58,7 @@ public void run(Program program) {
         AliasAnalysis alias;
         try {
             final Context analysisContext = Context.create();
+            analysisContext.register(EventDomainRepository.class, EventDomainRepository.forProgram(program));
```
hernanponcedeleon, line 61: Why do you need to register this if never used by the class?

dartagnan/src/main/java/com/dat3m/dartagnan/utils/collections/IndexedSet.java
```diff
@@ -0,0 +1,474 @@
+package com.dat3m.dartagnan.utils.collections;
+
+import com.google.common.base.Preconditions;
+
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+
+/// Compact subset of a collection using a bitvector.
```
hernanponcedeleon:
```suggestion
// Compact subset of a collection using a bitvector.
```

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -178,6 +174,7 @@ public static void preprocessMemoryModel(VerificationTask task, Configuration co
 
     public static void performStaticProgramAnalyses(VerificationTask task, Context analysisContext, Configuration config) throws InvalidConfigurationException {
         final Program program = task.getProgram();
+        analysisContext.register(EventDomainRepository.class, EventDomainRepository.forProgram(program));
```
hernanponcedeleon, line 177: Why do you need to register if not used (same for `RelationEventDomains` below)?

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
```diff
@@ -105,6 +106,7 @@ private void reachingDefinitionMustOverride(ReachingDefinitionsAnalysis.Method m
         MemoryAllocation.newInstance().run(program);
         Configuration config = Configuration.builder().setOption(REACHING_DEFINITIONS_METHOD, method.name()).build();
         Context context = Context.create();
+        context.register(EventDomainRepository.class, EventDomainRepository.forProgram(program));
```
hernanponcedeleon, line 109: Why the register?

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Tearing.java
ThomasHaas: It is used by the other analyses he instantiates afterwards, otherwise they will throw an exception due to missing requirements.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
ThomasHaas: Same as above

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
xeren: It's currently required by `ExecutionAnalysis` and `RelationEventDomains`.  With its low costs and the properties of `IndexedDomain.isCompatibleWith`, you could argue that all users can as well initialise their own instances.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
xeren: Same reason as in `ModelChecker`: required by `ExecutionAnalysis`.

dartagnan/src/main/java/com/dat3m/dartagnan/program/processing/Tearing.java
xeren: Same reason as in `ModelChecker`: Required by `ExecutionAnalysis`.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/analysis/NativeRelationAnalysis.java
xeren: I removed this warning when I created `EventDomainRepository`, which guarantees the compatibility.

xeren:
I do not plan any feature additions to this MR.  Something like `IndexedMap`, more fine-grained event domains, or infix-based compatibility can be added later, if the need arises.

hernanponcedeleon:
These are the latest results on SVCOMP benchmarks

| Category | Branch | Total CPU Time (s) | Total Solved Tasks | Avg Time per Task (s) | Performance Improvement (%) |
| :--- | :--- | :---: | :---: | :---: | :---: |
| **unreach-call** | MR | 15700 | 470 | 33.40 | 13.78% |
| **unreach-call** | Development | 17900 | 462 | 38.74 | - |
| **no-data-race** | MR | 19300 | 699 | 27.61 | 15.80% |
| **no-data-race** | Development | 22300 | 680 | 32.79 | - |
| **valid-memsafety** | MR | 16700 | 480 | 34.79 | 14.04% |
| **valid-memsafety** | Development | 19100 | 472 | 40.47 | - |
| **no-overflow** | MR | 14900 | 431 | 34.57 | 15.77% |
| **no-overflow** | Development | 17400 | 424 | 41.04 | - |

I am fine with the current version of the code, so unless @ThomasHaas has further comments, I think we can merge

ThomasHaas (approved):
I haven't checked the code in detail, but I think it is fine overall and the tests show it is correct as well. We can merge.

PR #1045, opened 2026-06-18
Fix alias analysis

xeren:
A wrong comparison in `ModifierTrait.MdLinear.shrinkToBounds(Md, int)` introduced in #1008 leads to false negative `mayAlias` results.  This PR fixes this.

**Explanation:**
A task is affected, if the program features at least one access with a dynamic address like `r0 + a + b * r1`, where
- `r0` must be a register that can only point into exactly one single object `o`,
- `a` and `b` are literal integers (if omitted, `a=0` and `b=1`),
- `r1` is a non-literal, i.e. leading to a non-functional modifier,
- `o.size` is known and `0 <= a - b` and `o.size < a + b` hold.  Thus, `a` and `a - b` might be valid addresses within bounds, but `a + b` is not.

Affected accesses receive the `&o + a` address set and miss at least the `&o + a - b` address.  Down-stream, the SMT queries could become falsely unsatisfiable.  `RefinementSolver` could return `Inconclusive`.

### DIFF
--- a/benchmarks/miscellaneous/alias-shrinkToBounds.c
+++ b/benchmarks/miscellaneous/alias-shrinkToBounds.c
@@ -1,0 +1,50 @@
+/*
+  Exposes a bug in Dartagnan's Alias Analysis, where address sets were incorrectly shrunk based on array size.
+  Expected verdict: FAIL.
+*/
+#include<pthread.h>
+#include<assert.h>
+#include<dat3m.h>
+
+#define NUM_LOCKS 2
+
+struct cache {
+  // Unused storage to increase the static offset for the affected address sets.
+  // For elements (given sizeof(int) == 4 and sizeof(pthread_mutex_t) == 8, the eager solver returned PASS.
+  // The lazy solver non-deterministically returned FAIL or UNKNOWN(inconclusive).
+  int garbage[17];
+  pthread_mutex_t mutex[NUM_LOCKS];
+} c;
+
+volatile int x;
+
+// Returns just `x`, but the Alias Analysis lost precision.
+// Replacing at least one `id(x)` with `x` yielded FAIL again.
+int id(int x) {
+  int i = __VERIFIER_nondet_int();
+  __VERIFIER_assume(i == x);
+  return i;
+}
+
+// The address sets of these lock events were falsely estimated by the Alias Analysis.
+void increment(int i) {
+  pthread_mutex_lock(&c.mutex[i]);
+  x++;
+  pthread_mutex_unlock(&c.mutex[i]);
+}
+
+void *t_fun(void *arg) {
+  increment(id(1));
+  return NULL;
+}
+
+int main () {
+  for (int i = 0; i < NUM_LOCKS; i++)
+    pthread_mutex_init(&c.mutex[i], NULL);
+  pthread_t t1;
+  pthread_create(&t1, NULL, t_fun, NULL);
+  increment(id(0));
+  pthread_join(t1, NULL);
+  assert(x == 2);
+  return 0;
+}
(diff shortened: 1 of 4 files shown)


ThomasHaas:
Here is an example that is wrong on `development` and that might fit your description.
I suppose this is due to the AA bug. It is certainly a very strange bug to reproduce.

```C
#include<pthread.h>
#include<assert.h>
#include<dat3m.h>

#define NUM_LOCKS 2
#define NUM_GARBAGE 17 // PASS
//#define NUM_GARBAGE 16 // FAIL
#define LOCK_INDEX 0

struct cache {
  int           garbage[NUM_GARBAGE];
  pthread_mutex_t mutex[NUM_LOCKS];
} c;

volatile int x;

void *increment(int i) {
  pthread_mutex_lock(&c.mutex[i]);
  x++;
  pthread_mutex_unlock(&c.mutex[i]);
}

void *t_fun(void *arg) {
  int i = __VERIFIER_nondet_int(); // Making this constant yields FAIL again
  __VERIFIER_assume(i == LOCK_INDEX);
  increment(i + 1);
  return NULL;
}

void init() {
  for (int i = 0; i < NUM_LOCKS; i++)
    pthread_mutex_init(&c.mutex[i], NULL);
 }

int main () {
  init();

  pthread_t t1;
  pthread_create(&t1, NULL, t_fun, NULL);

  int j = __VERIFIER_nondet_int(); // Making this constant does NOT yield FAIL
  __VERIFIER_assume(j == LOCK_INDEX);
  increment(j);

  pthread_join(t1, NULL);

  assert (x == 2); // Should fail: both threads use a different lock

  return 0;
}

```

hernanponcedeleon:
@ThomasHaas were you using eager to get PASS? I get all sort of unexpected results (suggesting the bug) ...
```
> java -jar dartagnan/target/dartagnan.jar cat/sc.cat test.c
Test: test.c
Result: FAIL
Reason: Program specification violation found
Details:
        T1:t_fun -> increment @test.c#26 -> @test.c#20  (E119): Unlocking an already unlocked mutex
Time: 0.740 secs

> java -jar dartagnan/target/dartagnan.jar cat/sc.cat test.c --solver=yices2
[19.06.2026] 21:19:40 [WARN] RefinementSolver.runInternal - Refinement procedure was inconclusive. Trying to find reason of inconclusiveness.
[19.06.2026] 21:19:40 [WARN] RefinementSolver.analyzeInconclusiveness - Found unexpected aliasing between:
        T0:main -> init @test.c#36 -> Iter#2 @test.c#31 -> @test.c#32   (E10)
AND
        T1:t_fun -> increment @test.c#26 -> Iter#1 @test.c#18 -> @test.c#18     (E101)
Possible out-of-bounds access in source code or error in alias analysis.
Test: test.c
Result: ERROR
Reason: RuntimeException
Details:
        Terminated verification due to inconclusiveness (bug?).

> java -jar dartagnan/target/dartagnan.jar cat/sc.cat test.c --method=eager
Test: test.c
Result: PASS
Time: 0.717 secs
```

@xeren can you add the test as a regression?

ThomasHaas:
Oh, I only checked eager. Since the program has no UB anymore, I thought that lazy will likely work.
But good to know that lazy is wildly broken here.

xeren:
> Here is an example that is wrong on development and that might fit your description.
> I suppose this is due to the AA bug. It is certainly a very strange bug to reproduce.

In your example, the AA from `development` falsely shrinks the address sets of all `pthread_mutex_lock` and `pthread_mutex_unlock` events to `&c.mutex + 0`.  The AA concludes `mustAlias` for all of them.
The eager encoding eliminates the address checks but keeps the constraint on `co`.  Under a memory model with Atomicity, the SMT encoding only explores executions where the critical sections do not interfere and finds no violations.

I discovered the bug on `chase-lev-fail.ll`, which features something like this, as well.

ThomasHaas:
Do you know if your fix comes with any regressions in terms of precision? Or does the fix specifically target an edge-case that rarely occurred in practice?

xeren:
There is a precision regression in rare cases:
By replacing `i + 1` with `i` in your example, the expected verdict becomes PASS.   On `development`, the AA shrinks the address sets due to the bug and returns `mustAlias`, which is correct for a different reason.  After this fix, it can only return `mayAlias` for these event pairs.

hernanponcedeleon (approved):

ThomasHaas:
@xeren Is the fix here unique to #1008 ?
Cause I think the AA before #1008 was more precise on `cachedq.c`. This means the bug was already present before #1008 or there is a regression (the regression on `cachedq.c` is quite large!).

PR #1046, opened 2026-06-20
Fix wrong optimization of signed comparison between zero extended operands

ThomasHaas:
Fixes:

```C
#include <assert.h>
#include <stdlib.h>

unsigned int n;

int main() {

  n = 3435973837;

//   if(n >= (4294967296 / 5)) { // PASS
  if(!(n < (4294967296 / 5))) { // FAIL
    abort();
  }

  assert(0);

  return 0;
}
```

### DIFF
--- a/benchmarks/miscellaneous/zext.c
+++ b/benchmarks/miscellaneous/zext.c
@@ -1,0 +1,9 @@
+#include <assert.h>
+#include <stdlib.h>
+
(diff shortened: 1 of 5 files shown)


hernanponcedeleon:
@ThomasHaas can you please add the test (without the commented line) as a regression?

hernanponcedeleon (approved):

PR #1047, opened 2026-06-21
Optimize encoding time using caches

ThomasHaas:
This PR adds the method `EdgeEncoder.withCache` which can transform an `EdgeEncoder` into a caching variant of it.
The caching variants are used in the encoding of transitive closures, compositions, and SAT acyclicity constraints.

For transitive closure and compositions we now generate the or clauses in a single n-ary `bmgr.or` call instead of many binary calls. This saves quite a bit of time as well.

Together, both changes can more than half the encoding time.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
@@ -35,11 +35,10 @@ import org.sosy_lab.java_smt.api.NumeralFormula.IntegerFormula;
 
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


hernanponcedeleon:
Which is the benchmark where you said you saw a huge speed up?

ThomasHaas:
Among ours, I used `challenging/cna.c` with `imm.cat` and `B=3` (or 4?).
Apart from that, all the SVCOMP one's that you mentioned. Generally, you should observe the effect on every larger benchmark + memory model that uses transitive closures.

hernanponcedeleon (approved):

PR #1048, opened 2026-06-24
SVCOMP races in cat

hernanponcedeleon:
Replaces the SVCOMP data race encoding by a cat definition.

While there is some performance overhead (in a run with a 1 min timeout the overhead was between 3x and 4x), in a full run with 15 min, the results are good. We even get better results than the official SVCOMP 2026 results (this is not apples-to-apples since there have been other changes since the version we used for the last competition, including the native compilation and I am running on a different server, but still): 
|  | SVCOMP 2026 | MR |
| :--- | :---: | ---: |
| true | 505 | 508 |
| false | 197 | 215 |
| points | 1207 | 1231 |


cat/svcomp.cat
```diff
@@ -4,5 +4,13 @@ SVCOMP
 let com = (rf | fr | co)
 
 (* Sequential consistency *)
-let hb = po | (com & int) | (rmw^-1? ; (com & ext))
-acyclic hb as sc
+acyclic po | (com & int) | (rmw^-1? ; (com & ext))
+
+(* Data Races *)
+let Atomic = RMW | A
+let sw = [Atomic]; rmw^-1? ;(com & ext); [Atomic]
```
ThomasHaas: Maybe we should change our atomic blocks to tag events inside as atomic A (in addition to RMW).
Then we can simplify this here, and it also makes more sense on the C language level.

cat/svcomp.cat
```diff
@@ -4,5 +4,13 @@ SVCOMP
...
+let hb = (po | sw)+
+let conflict = (ext & ((W * _) | (_ * W)) & loc) \ ((IW * _) | (_ * IW))
+let dr = conflict \ (hb | hb^-1 | (Atomic * Atomic))
```
ThomasHaas: I would put the exclusion of `(Atomic * Atomic)` into the `conflict` relation.

I always have in mind that a race candidate (a conflict) consists of two accesses of different threads to the same location, (at least) one being a write and one being non-atomic.
The race candidate is a proper race if it is unordered in `hb`.

Translating this into CAT would remove the atomic pairs already from `conflict`. But maybe that is just my personal preference :P

cat/svcomp.cat
```diff
@@ -4,5 +4,12 @@ SVCOMP
 let com = (rf | fr | co)
 
 (* Sequential consistency *)
-let hb = po | (com & int) | (rmw^-1? ; (com & ext))
-acyclic hb as sc
+acyclic po | (com & int) | (rmw^-1? ; (com & ext))
```
ThomasHaas: I dislike that this definition is inlined now.
I think we should just define two hb's but with different names, either
(1) `hb-sc` (for consistency) and `hb` (for races), or
(2) `hb` (for consistency) and `hb-race` (for races)
or something similar.

That's my final nitpick. We can merge afterwards :)

hernanponcedeleon: Actually, that was my initial implementation, but I really disliked it and that is why I ended up inlining things.

I might not be completely againts `hb-consistency` and `hb-race` which makes it more explicit why we have two `hb` definitions

ThomasHaas: I saw that you had two hb's before but with the same name. I agree that reusing the exact same name is not good here and that's why we should add suffixes.

ThomasHaas (approved):
LGTM.

PR #1050, opened 2026-06-26
Log OS and clang info

hernanponcedeleon:


dartagnan/src/main/java/com/dat3m/dartagnan/utils/SysInfo.java
```diff
@@ -0,0 +1,39 @@
+package com.dat3m.dartagnan.utils;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+public class SysInfo {
```
ThomasHaas: Maybe we can have an `EnvironmentInfo` which is a combination of this one and `GitInfo`.
At least I can imagine that we will eventually find something new to log about the execution environment of Dartagnan and we would end up creating a third class of this kind.

ThomasHaas (approved):
LGTM

PR #1051, opened 2026-06-28
Document required/recommended versions for clang and graalvm

hernanponcedeleon:
We recently found out that due to a [change in LLVM IR](https://releases.llvm.org/10.0.0/docs/ReleaseNotes.html#id3), our parser does not properly work with older versions of clang (e.g., we report wrong results for races in [this](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/blob/main/c/goblint-regression/04-mutex_04-munge_nr.c) program).

Also, we noticed due to a [difference garbage collector configuration up to (and including) 21.3](https://www.graalvm.org/jdk24/reference-manual/native-image/optimizations-and-performance/MemoryManagement/), we can have large overhead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,11 @@
 
 Requirements
 ======
-* [Maven](https://maven.apache.org/) 3.8 or above
-* [Java](https://openjdk.java.net/projects/jdk/17/) 17 or above
-* [GraalVM](https://www.graalvm.org/) (only if you run in native mode; recommended!) 22 or above for optimal performance
-* [Clang](https://clang.llvm.org) (only to verify C programs) 10 above
-* [Graphviz](https://graphviz.org) (only if option `--witness=png` is used)
+* [Maven](https://maven.apache.org/) version 3.8 or above
+* [Java](https://openjdk.java.net/projects/jdk/17/) version 17 or above
+* [GraalVM](https://www.graalvm.org/) version 22 or above for optimal performance (only if you run in native mode; recommended!)
+* [Clang](https://clang.llvm.org) version 10 or above (only to verify C programs)
+* [Graphviz](https://graphviz.org)
 
 Installation
 ======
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,8 +11,8 @@ Requirements
 ======
 * [Maven](https://maven.apache.org/) 3.8 or above
 * [Java](https://openjdk.java.net/projects/jdk/17/) 17 or above
-* [GraalVM](https://www.graalvm.org/) (only if you run in native mode; recommended!)
-* [Clang](https://clang.llvm.org) (only to verify C programs)
+* [GraalVM](https://www.graalvm.org/) (only if you run in native mode; recommended!) 22 or above for optimal performance
+* [Clang](https://clang.llvm.org) (only to verify C programs) 10 above
```
ThomasHaas: "10 or above".

Do you want to keep the comment `(only to verify C programs)` inbetween Clang the the version number?
It feels a bit off to me.

hernanponcedeleon: To be honest, I kind of dislike the whole `only if ...` in the documentation. Maybe we should completely get rid of those

ThomasHaas: In what way you want to get rid of those? Recommend them unconditionally or put them elsewhere? I could see adding an `Optional` section, but there you will still need to put a comment for each optional feature.
I could also see making the requirement for clang unconditional because 99.9% of people will use Dartagnan to verify C code (btw. technically we need the provided LLVM code to be in shape of version 10.0.0, irrespective of whether clang is installed or not). However, I would not make GraalVM unnecessarily a hard requirement.

hernanponcedeleon: The one that always "bothered me" was graphviz. I thought generating witnesses and running on C code would cover the majority of the users, but I forgot about graalvm and I agree we might not want to have this as a hard requirement. I guess I will keep the `(only if ...)` but move them to the end of each sentence

ThomasHaas (approved):
LGTM.

PR #1052, opened 2026-07-04
Cleanup pom

hernanponcedeleon:


xeren (dismissed):
Apparently, `commons-cli` stopped being used in `SVCompRunner.java` since #195.  That same MR also removed the usage of `commons-cli` in `Dartagnan.java`, so it could be possible to also remove the dependency from `dartagnan` (in another MR).

`pl.project13.maven` remains a direct dependency of `dartagnan`.

LGTM

hernanponcedeleon:
Good catch Rene!

I found a few more old dependencies, so I got rid of them too.

ThomasHaas:
Do we really not use Guava anywhere? That surprises me

EDIT: We surely use it. Why can we remove it from dartagnan's pom? Is it transitively included now?

hernanponcedeleon:
Oops, I accidentally edit your comment rather than answer it.

> EDIT: We surely use it. Why can we remove it from dartagnan's pom? Is it transitively included now?

I do not see it used anywhere

ThomasHaas:
Functions like `Lists.transform` and `Lists.reverse` are from google's library, both of which are used in, e.g., the `ProgramEncoder`.

> Oops, I accidentally edit your comment rather than answer it

Happened to me too. But it is possible to see the edit history, so you can recover the old message (I did already).

hernanponcedeleon:
> Functions like `Lists.transform` and `Lists.reverse` are from google's library, both of which are used in, e.g., the `ProgramEncoder`.

Those seem to be from `import com.google.common.collect.Lists;` rather than `guava`

ThomasHaas:
`google.common` is Guava, no? We certainly have no other google library included in our pom.
Btw. all the `Preconditions` are also from Guava.

hernanponcedeleon:
You were right, guava was being transitively added
```
[INFO] com.dat3m:dartagnan:jar:4.5.0-alpha
[INFO] \- org.sosy-lab:java-smt:jar:5.0.1:compile
[INFO]    \- com.google.guava:guava:jar:33.2.1-jre:compile
```
Fixed

ThomasHaas (approved):
LGTM

PR #1054, opened 2026-07-08
Remove obsolete class `Filter`

xeren:
`Filter` and its implementing classes used to describe unary predicates in a memory model.  It was later used to organise event caches in `Program` via `getThreadEvents(Filter)`.  Both functionalities have been replaced since.  In this MR, I refactor the last remaining users of this class.

I also remove `FenceGraph`, since this former user of `Filter` is itself not used.  IIRC fence relations are no longer built-in, but defined as derived in `cat/stdlib.cat`.


ThomasHaas (commented):
LGTM. @hernanponcedeleon feel free to merge :)

PR #1058, opened 2026-07-17
Slightly more precise and much easier optimization for acyclicity

ThomasHaas:
This is a simplification of #1015 with a slightly better precision.
@hernanponcedeleon you merged #1015 a little too early :P

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
@@ -14,10 +14,7 @@ import com.dat3m.dartagnan.smt.EncodingUtils;
 import com.dat3m.dartagnan.smt.FormulaManagerExt;
 import com.dat3m.dartagnan.wmm.*;
 import com.dat3m.dartagnan.wmm.analysis.RelationAnalysis;
-import com.dat3m.dartagnan.wmm.axiom.Acyclicity;
-import com.dat3m.dartagnan.wmm.axiom.Axiom;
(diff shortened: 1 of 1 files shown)


hernanponcedeleon:
> @hernanponcedeleon you merged #1015 a little too early :P

This was possibly the (merged) PR that remained open the longest :P

Do you have any example where the current code is more precise?

ThomasHaas:
> > @hernanponcedeleon you merged #1015 a little too early :P
> 
> This was possibly the (merged) PR that remained open the longest :P

Yeah, cause the evaluations on SVCOMP were not good and cause I was busy with other stuff :P.

> Do you have any example where the current code is more precise?

In terms of real benchmarks, probably under all weaker memory models (e.g. aarch64.cat).
In terms of artificial examples, I guess something like this:
```
a;
x;
if (*)
    z; // maybe a fence in TSO?
y;  
```
s.t. you have a must-edge between every pair (from top to bottom) except for `(a, y)`.

The old code did not make the `(z,y)` ordering unconditional, because it thought that this could then be composed with `(a,z)` to give `(a, y)` which is not a must-edge and so appears to be wrong. But `(a,y)` is a transitive must-edge via `(a, x);(x, y)` (skipping `z` altogether) and so everything is fine: `(a, y)` is already statically ordered even if it is no direct must-edge.

The new code just computes the transitive closure to do the reasoning which simplifies it a lot (the old code approximated this same reasoning somehow). Overall, the precision is not so much higher, especially for SC. But under ARM, I saw some bigger improvements. I have not really measured how it affects performance.

hernanponcedeleon (approved):

PR #1059, opened 2026-07-17
Deterministic encoding

ThomasHaas:
Our encoding was not deterministic. To search for issues, I repeatedly dumped the encoding of the same benchmark and diff'd it to see differences. I then removed the sources of differences, mostly by using `LinkedHashSet` in place of `HashSet`.
I don't see any differences anymore, so the encoding is now deterministic as far as I can see.

What I have tested/fixed:
- Encoding order of relations. To fix this, I needed to make `DependencyGraph` compute deterministic topological orderings (among other things).
- Encoding order of SAT encoding for acyclicity
- Encoding of non-termination
- Encoding of program data/control-flow. In this case, `getRegisterReads` and `ReachingDefinitionsAnalysis.getUsedRegisters` could return register accesses in non-deterministic order.

The only thing I have not tested are recursive memory models and CAT_SPEC encodings. But I think the fix in `DependencyGraph` should work for them as well.

**Impact on runtime:** the SMT solving times seem to be much more stable now (they still have some fluctuation though). However, it is unclear if they are on average better or worse (maybe the fixed order is a bad order?).
There might be a bit of extra overhead in processing time (and memory usage) for using `LinkedHashSet` now. But I think that should be negligible.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
@@ -90,9 +90,11 @@ public final class EncodingContext {
                 .map(n -> t.getMemoryModel().getRelation(n).getDefinition())
                 .toList();
         final Iterable<? extends Constraint> toEncode = Iterables.concat(c, anarchicConstraints);
-        constraintsToEncode = new LinkedHashSet<>(
-                DependencyGraph.from(toEncode, EncodingContext::computeConstraintDependencies).getNodeContents()
-        );
+        var depGraph = DependencyGraph.from(toEncode, EncodingContext::computeConstraintDependencies);
+        // NOTE: This guarantees a deterministic ordering of the constraints to be encoded
+        constraintsToEncode = t.getMemoryModel().getConstraints().stream()
+                .filter(constr -> depGraph.get(constr) != null)
+                .toList();
     }
 
     public static EncodingContext of(VerificationTask task, Context analysisContext, FormulaManager formulaManager) throws InvalidConfigurationException {
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
@@ -93,7 +93,7 @@ public final class EncodingContext {
         var depGraph = DependencyGraph.from(toEncode, EncodingContext::computeConstraintDependencies);
         // NOTE: This guarantees a deterministic ordering of the constraints to be encoded
         constraintsToEncode = t.getMemoryModel().getConstraints().stream()
-                .filter(constr -> depGraph.get(constr) != null)
+                .filter(depGraph::contains)
                 .toList();
     }
 
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/NonTerminationEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/NonTerminationEncoder.java
@@ -118,10 +118,10 @@ public class NonTerminationEncoder {
     private final VerificationTask task;
 
     private final List<Loop> allLoops = new ArrayList<>();
-    private final Map<Event, NonterminationCase> nonterm2Case = new HashMap<>();
-    private final Map<LoopAnalysis.LoopIterationInfo, Iteration> iterInfo2Iter = new HashMap<>();
+    private final Map<Event, NonterminationCase> nonterm2Case = new LinkedHashMap<>();
+    private final Map<LoopAnalysis.LoopIterationInfo, Iteration> iterInfo2Iter = new LinkedHashMap<>();
     // Maps an event to all the loop iterations it is contained in (sorted from innermost to outermost loop)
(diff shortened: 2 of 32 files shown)


ThomasHaas:
Wow, the change of `DependencyGraph.nodeMap` to a `LinkedHashSet` breaks Dartagnan.
I don't think the Tarjan algorithm should be affected. But I think the `BranchReordering` class might somehow depend on the iteration order over the computed dependency graph which would be a bug in that pass.
The processed program is certainly broken.
It surprises me very much: if `BranchReordering` is indeed broken, why was the non-determinism always nice to us and gave a working ordering?

ThomasHaas:
Ok, this is a very nasty one. It's actually no bug, but depending on how `BranchReordering` orders the branches, `NaiveLoopBoundAnnotation` may fail to detect the loop bounds. This is because the loop header may get moved:

```
B; // Independent block of code
ctr = 0;
loop:
  ctr++;
  if (ctr <= K) goto loop;
```
vs.
```
ctr = 0;  // Now separated from loop entry point -> LoopBoundAnnotation fails
B; // Independent block of code
loop: 
  ctr++;
  if (ctr <= K) goto loop;
```

In an attempt to avoid this problem, I have made `BranchReordering` compute its own topological ordering instead of relying on the one computed by `DependencyGraph`. The ordering tries to keep branches close together by their id. In general, I'm not sure if we can easily guarantee that code does not get moved around in a bad way.

EDIT: It still fails. I think it is going to be a nightmare to fix this.

ThomasHaas:
Ok, so I fixed the issue like this:
- `DependencyGraph` and `BranchReordering` remain mostly unchanged
- `EncodingContext.constraintsToEncode` does not rely on the order provided by `DependencyGraph` anymore and instead follows the order of `Wmm.getConstraints()` (which is deterministic)
- I switched `Wmm.getConstraints()` order to first list the relation definitions and then the axioms. This is more natural and combined with the previous point, means that we encode relations/axioms in the order in which they are listed in the `Wmm` which should largely align with the order in which they are declared in the CAT file (modulo the fact that axioms are always put last).

However, the fact that we have passes that are sensitive to the ordering provided by `BranchReordering` + the fact that the current ordering is the "good one" is a bit sketchy.
That being said, I think `BranchReordering` + `HashSet-based DependencyGraph` is deterministic because `BranchReordering` uses a deterministic hash-value computation (which also happens to give a good iteration order in the end).
`EncodingContext.constraintsToEncode` + `HashSet-based DependencyGraph` was not deterministic because the involved `Constraints` have a non-deterministic hash (based on object identity). That's why the second point of the fixes above is needed.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
```diff
@@ -90,9 +90,11 @@ private EncodingContext(VerificationTask t, Context a, FormulaManager m, Collect
                 .map(n -> t.getMemoryModel().getRelation(n).getDefinition())
                 .toList();
         final Iterable<? extends Constraint> toEncode = Iterables.concat(c, anarchicConstraints);
-        constraintsToEncode = new LinkedHashSet<>(
-                DependencyGraph.from(toEncode, EncodingContext::computeConstraintDependencies).getNodeContents()
-        );
+        var depGraph = DependencyGraph.from(toEncode, EncodingContext::computeConstraintDependencies);
+        // NOTE: This guarantees a deterministic ordering of the constraints to be encoded
+        constraintsToEncode = t.getMemoryModel().getConstraints().stream()
+                .filter(constr -> depGraph.get(constr) != null)
```
hernanponcedeleon: Is it really possible that we put a `null` node in the graph? If so, it might be a bit better to has a `DependencyGraph.hasNode()` implemented as the null check

hernanponcedeleon:
> However, the fact that we have passes that are sensitive to the ordering provided by `BranchReordering` + the fact that the current ordering is the "good one" is a bit sketchy.

Other than `NaiveLoopBoundAnnotation`, do we have any other pass that is sensitive to the ordering? Can you add a note to this (these) passes that make it clear once that gets improved, we can make `DependencyGraph` and `BranchReordering` _"more deterministic"_ as in your initial commit.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/EncodingContext.java
ThomasHaas: The null is possible because the dependecy graph only contains the subset of the constraints to be encoded.
I can add a helper method to check for nodes.

ThomasHaas:
> > However, the fact that we have passes that are sensitive to the ordering provided by `BranchReordering` + the fact that the current ordering is the "good one" is a bit sketchy.
> 
> Other than `NaiveLoopBoundAnnotation`, do we have any other pass that is sensitive to the ordering? Can you add a note to this (these) passes that make it clear once that gets improved, we can make `DependencyGraph` and `BranchReordering` _"more deterministic"_ as in your initial commit.

I don't think there are other passes sensitive to the order. `BranchReordering` is deterministic either way due to the deterministic hash codes (I removed the option for non-deterministic hash codes).

hernanponcedeleon (approved):

Issue #1039, opened 2026-06-05
Two possible GPU-related correctness inconsistencies


8ws26y8gnh-cloud:
Hi, I ran into two GPU-related behaviors that look suspicious to me. They may have different root causes, but I am grouping them here because both concern GPU frontend/model handling. I may be missing some intended preprocessing or configuration step, so I would appreciate a sanity check.

## 1. Possible missing Vulkan tags after SPIR-V frontend compilation

Several OpenCL lock examples from Dartagnan's own benchmark directory, compiled to SPIR-V via `clspv`, report data races under `vulkan.cat`, although these examples are expected to pass. The behavior also differs from the older ASPLOS'24 artifact, where the same style of benchmark passes.

### Reproduction sketch

I started from the OpenCL lock benchmarks in:

```text
benchmarks/opencl/benchmarks/
```

For example:

```text
benchmarks/opencl/benchmarks/caslock.cl
benchmarks/opencl/benchmarks/ticketlock.cl
benchmarks/opencl/benchmarks/ttaslock.cl
```

I used `clspv` to compile an OpenCL lock benchmark to SPIR-V:

```bash
clspv caslock.cl \
  --cl-std=CL2.0 \
  --inline-entry-points \
  --spv-version=1.6 \
  -o caslock.spv

spirv-dis caslock.spv > caslock.spvasm
```

The default `clspv` output contains:

```text
OpMemoryModel Logical GLSL450
```

As far as I can tell, Dartagnan's Vulkan SPIR-V frontend rejects this form. For the Vulkan run, I therefore adjusted the disassembled `.spvasm` to use the Vulkan memory model declaration/capability:

```text
OpCapability VulkanMemoryModel
OpExtension "SPV_KHR_vulkan_memory_model"
OpMemoryModel Logical Vulkan
```

I am not sure whether this manual conversion is a valid or recommended way to prepare these `clspv` outputs for Dartagnan's Vulkan frontend. If this preprocessing step is not valid, I would appreciate being pointed to the intended way to compile or prepare these benchmarks.

I also added the metadata annotations expected by Dartagnan's SPIR-V inputs, such as `@Config` and `@Input` lines, following the format described in the artifact documentation / existing SPIR-V examples. For example:

```text
; @Config: 2, 1, 3
; @Input: %17 = {{0}}
; @Input: %18 = {{0}}
; @Input: %19 = {{0, 0, 0, 0, 0, 0}}
```

[...]

PR #1061, opened 2026-07-20
Set eager and idl2sat as default

hernanponcedeleon:
I did some benchmarking to evaluate if it is still the case that the lazy encoding performs better than the eager one and to compare the IDL vs SAT encodings for acyclicity. I filtered the results where times were below 1s.

For lazy vs eager, the results are not super clear (except for power, where lazy is much better due to the recursive definition), but considering that currently we return wrong results with eager for ptx tests using proxy (#1039) and that one can expect more deterministic solving times using eager (at least once we merge #1059), I think it makes more sense to use that method as the default.

image
image
image
image

For IDL vs SAT, the results are clear.

image
image
image
image

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Method.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/configuration/Method.java
@@ -24,7 +24,7 @@ public enum Method implements OptionInterface {
     }
 
     public static Method getDefault() {
-        return LAZY;
+        return EAGER;
     }
 
     // Used to decide the order shown by the selector in the UI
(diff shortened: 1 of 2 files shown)


ThomasHaas:
Do you want to keep the unit tests as they are or also use different settings for them?

hernanponcedeleon:
Performance is not a problem ATM in the CI, so I would keep them as they are. 

Also, at some point we explicitly set some options to the unit tests exactly not to change their behavior if we ever changed the default options.

ThomasHaas (approved):
LGTM

PR #1062, opened 2026-07-22
Fix cat.include option

hernanponcedeleon:
While testing #1038, I noticed that the `cat.include` option is not working.

Apparently the sosy-lab library has some restrictions on how to handle paths in the the options ([there are ways of dealing with full paths, but not with directory paths AFAICT](https://sosy-lab.github.io/java-common-lib/api/org/sosy_lab/common/configuration/FileOption.Type.html)). 

Using strings is consistent with the other option we have related to files/paths (i.e., `bound.save` and `bound.load`).


ThomasHaas (approved):
LGTM

PR #469, opened 2023-06-22
Rework

hernanponcedeleon:
This PR simply merges the current status of `development_rework` (minus the code in the prototype folder and the LLVM grammar and parser) into `development`. The code has already been reviewed (see #444, #445, #447, #448, #450, #452, #453, #455, #456, #457, #460. #462, #464).

PR #1063, opened 2026-07-22
[DRAFT] Rework verification task/goal/result

ThomasHaas:
This PR attempts to generalize the idea of `VerificationTask` to allow for different types of tasks such as state-space enumeration (this PR does not implement state-space enumeration!).

The changes can roughly be summarized like this:
- `VerificationTask` contains a `TaskGoal` that captures what needs to be done (currently there is only `TaskGoal.Verify`).
- For every `TaskGoal.X`, there is a corresponding `TaskResult.X`, e.g., `TaskResult.Verify`. This result is produced by `TaskSolver`. 
- The previous `Result` enum has been renamed to `ResultStatus` which is part of every `TaskResult`. The `ResultStatus` has been changed to distinguish `BOUNDED` and `UNKNOWN`. As a result, output looks like this now:
```
[22.07.2026] 15:55:42 [INFO] AssumeSolver.runInternal - Verification finished with result BOUNDED
Test: /Users/thomashaas/IdeaProjects/Dat3M/benchmarks/challenging/cna.c
Result: BOUNDED
Reason: Not fully unrolled loops
Details:
	T1:thread_n -> cna_unlock @cna.c#53 -> @cna.h#102	(E223)
Time: 3.909 secs
```
This likely requires updating of external scripts that look for `UNKNOWN` in the output. But I guess we have to bite this bullet eventually :P.

**Possible alternatives:** Rather than having a single `VerificationTask` class with different `TaskGoals`, we could have different `VerificationTask` classes (e.g., `VerificationTask` and `EnumerationTask`. They could all either share a common base class `Task` which has the `Program/Wmm/Config/etc.`, or they all contain a `TaskContext` that contains `Program/Wmm/Config/etc.`.
Certainly, the current naming scheme of `VerificationTask` as universal base class together with non-verification goals like `TaskGoal.Enumeration` is odd.
I am inclined to test out one of these solutions.


ThomasHaas:
I tried the later solution now where `VerificationTask` has become a general `Task` and its subclasses represent the actual task to solve. `TaskGoal` is gone.
I think I like this solution a lot more.

EDIT: Btw. the changes in this PR are  just "first steps" towards a more general way of defining and solving new types of tasks. And some of the "next steps" I have to figure out once I try to add the `Enumeration/Exploration` task and unify some of the backend. In that sense, this is WIP.
However, if I do all changes in a single PR, I think it will get too big. Alternatively, instead of merging onto development, I could introduce an intermediary branch that I can target with these PRs for piece-wise review.
@hernanponcedeleon any preferences on how to deal with this?

hernanponcedeleon:
I definitely do not want a huge PR, so please split them into "steps".

If you are not yet sure about some of the next steps (which means some of the earlier decisions might change), maybe it is better to have an intermediate branch, target all "step X" branches there and then have a final (but easier) review from the large branch to development. IIRC this is how we did it for the IR / type system rework in 2023.

ThomasHaas:
I can use an intermediary branch, but most changes in the PRs are about renaming/moving stuff which makes them "large". Ignoring those changes, I think they are overall quite simple.
Now that i think about it, the part about "unifying backends" can probably be done separately anyhow; so the only thing left is adding the enumeration task for Telechat + whatever changes it brings with it.

ThomasHaas:
I created another branch called `exploration` that is targeted by this PR. 
However, I think now the CI won't run, right?

hernanponcedeleon:
> However, I think now the CI won't run, right?

Something [like this](https://github.com/hernanponcedeleon/Dat3M/pull/469/changes/0d6dc081e2cb6f5d67fb8741011f54ac0c0a243c) should work

ThomasHaas:
I'm not even sure if I need the intermediate branch anymore. I think I will do a whole rework of the input/output pipeline for verification tasks here and then have a second PR for adding the enumeration/exploration mode. I think the second PR will not really undo the ideas of the first PR, but only (maybe) generalize some classes.

EDIT: Btw. I will wait for #1066 getting merged before I keep working on this one.

ThomasHaas:
See #1074 for a new version

PR #1064, opened 2026-07-23
Vulkan: Partial co and align with Alloy

natgavrilenko:


dartagnan/src/test/resources/VULKAN-expected.csv
```diff
@@ -103,7 +103,7 @@ litmus/VULKAN/Manual/MP-mesa-fence-loop.litmus,0
 litmus/VULKAN/Manual/MP-mesa-optimized.litmus,1
 litmus/VULKAN/Manual/MP-avvis.litmus,0
 litmus/VULKAN/Manual/MP-no-avvis.litmus,1
-litmus/VULKAN/Manual/CoWW-RR.litmus,1
+litmus/VULKAN/Manual/CoWW-RR.litmus,0
```
natgavrilenko, line 106: This is a racy test, where expected result changed due to the updated fr definition. Does not impact non-racy tests.

cat/vulkan.cat
```diff
@@ -144,16 +144,25 @@ let locord = loc &
...
 (***********************)
 
+// Mutually ordered atomic writes must be ordered by asmo in either direction
+// Translated from Alloy rai[asmo] = (W -> W) & mutordatom
+// The opposite direction, empty asmo \ ((W * W) & mutordatom),
+// follows from the asmo definition let asmo = co & mutordatom.
+empty ((W * W) & mutordatom) \ asmo \ asmo^-1 as consistent-asmo
+
+// Enforcing strict partial order for asmo
+empty (asmo; asmo) \ asmo
```
ThomasHaas: This constraint enforces only transitivity and has nothing to do with strictness; I think the comment should change.

Btw. is this constraint not automatically satisfied since `co` and `mutordatom` are both transitive? I don't mind the redundancy though.

cat/vulkan.cat
```diff
@@ -144,16 +144,25 @@ let locord = loc &
...
+
+// Enforcing co from initial writes to other same-location stores
+empty (loc & (IW * (W \ IW))) \ co
+
```
ThomasHaas, line 165: I feel the fact that we do not already enforce initial writes to be before other writes for partial co's seems to be an oversight. I think we should fix this internally (in a separate PR though).

cat/vulkan.cat
ThomasHaas: The no-chains variant also needs updating

natgavrilenko: fixed

hernanponcedeleon: > Btw. is this constraint not automatically satisfied since `co` and `mutordatom` are both transitive? I don't mind the redundancy though.

We do not get this for partial `co`, do we?

ThomasHaas: Sure we do: "partial co" == "partially ordered co" and partial orders are transitive. 
But I need to revise my statement slightly: `mutordatom` is actually not transitive but it is "sufficiently transitive" so that `co & mutordatom` keeps its transitivity .

hernanponcedeleon (approved):

PR #1066, opened 2026-07-24
Simplify Parsers and use Path instead of File

ThomasHaas:
- I have replaced most usages of `File` in the Dartagnan core part by `Path`. The remaining usages of `File` are all in the unit tests and the UI code.
- The `ProgramParser` and the backend `Compilation` classes have been simplified a bit. The latter now always constructs a temporary .c file in a temp directory when compiling raw input. In particular, the UI does not create a `dat3m.c` in the directory of a loaded .c file anymore (this was ugly!).

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/smt/ProverWithTracker.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/smt/ProverWithTracker.java
@@ -4,69 +4,69 @@ import com.google.common.collect.ImmutableMap;
 import org.sosy_lab.java_smt.api.*;
 import org.sosy_lab.java_smt.api.SolverContext.ProverOptions;
 
-import java.io.File;
-import java.io.FileWriter;
 import java.io.IOException;
-import java.io.PrintWriter;
-import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.nio.file.*;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
+
+import static java.nio.file.StandardOpenOption.*;
 
 public class ProverWithTracker implements ProverEnvironment {
 
     private final FormulaManager fmgr;
     private final ProverEnvironment prover;
-    private final String fileName;
+    private final Path filePath;
     private final Set<String> declarations;
 
-    public ProverWithTracker(SolverContext ctx, String fileName, ProverOptions... options) {
+    public ProverWithTracker(SolverContext ctx, Path filePath, ProverOptions... options) {
         this.fmgr = ctx.getFormulaManager();
         this.prover = ctx.newProverEnvironment(options);
-        this.fileName = fileName;
+        this.filePath = filePath;
         this.declarations = new HashSet<>();
         init();
     }
 
     // An empty filename means there is no need to dump the encoding
     private boolean dump() {
-        return !fileName.isEmpty();
+        return filePath != null;
     }
 
     private void init() {
-        if(dump()) {
-            try {
-                Files.deleteIfExists(Paths.get(fileName));
-            } catch (IOException e) {
-                e.printStackTrace();
-            }
-            StringBuilder description = new StringBuilder();
-            LocalDate currentDate = LocalDate.now();
-            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
-            description.append("Generated on: ").append(currentDate.format(formatter)).append("\n");
-            description.append("Generator: Dartagnan\n");
-            description.append("Application: Bounded model checking for weak memory models\n");
-            description.append("""
-                    Publications:\s
-                    - Hernán Ponce de León, Florian Furbach, Keijo Heljanko, \
-                    Roland Meyer: Dartagnan: Bounded Model Checking for Weak Memory Models \
-                    (Competition Contribution). TACAS (2) 2020: 378-382
-                    - Thomas Haas, Roland Meyer, Hernán Ponce de León: \
-                    CAAT: consistency as a theory. Proc. ACM Program. Lang. 6(OOPSLA2): 114-144 (2022)"""
-            );
-            write("(set-info :smt-lib-version 2.6)\n");
-            write("(set-logic ALL)\n");
-            write("(set-info :category \"industrial\")\n");
-            write("(set-info :source |\n" + description + "\n|)\n");
-            write("(set-info :license \"https://creativecommons.org/licenses/by/4.0/\")\n");
+        if (!dump()) {
+            return;
         }
+
+        try {
+            Files.deleteIfExists(filePath);
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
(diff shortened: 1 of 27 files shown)


ThomasHaas:
Now I really replaced all usages of `File` outside of:
- the unit tests
- two usages in the UI where the file selector UI element naturally returns a `File` object.

@hernan-poncedeleon you might want to double check that witness generation and loading/storing to bounds files keeps working correctly. I have not checked this yet.

ThomasHaas:
I will not update the unit tests to also use `Path` in this PR. So I think the PR is ready for review.

hernanponcedeleon:
> So I think the PR is ready for review.

I am busy until Tuesday, so I cannot promise to find time to review before then, but I will do my best.

dartagnan/src/main/java/com/dat3m/dartagnan/smt/ProverWithTracker.java
```diff
@@ -4,69 +4,68 @@
...
+        write("(set-logic ALL)\n");
+        write("(set-info :category \"industrial\")\n");
+        write("(set-info :source |\n" + description + "\n|)\n");
+        write("(set-info :license \"https://creativecommons.org/licenses/by/4.0/\")\n");
     }
 
     @Override
     public void close() {
         if(dump()) {
-            removeDuplicatedDeclarations(fileName);
```
hernanponcedeleon, line 69: Why is this one gone?

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
```diff
@@ -32,26 +32,18 @@ public static Provider<Wmm> createWmmFromArch(Supplier<Arch> archSupplier) {
...
-
-    public static Provider<Wmm> createWmmFromPath(Supplier<String> pathSupplier) {
-        return createWmmFromFile(() -> new File(pathSupplier.get()));
+    public static Provider<Wmm> createWmmFromPath(Supplier<Path> pathSupplier) {
+        return Provider.fromSupplier(() -> new ParserCat().parse(pathSupplier.get()));
     }
 
     public static Provider<Wmm> createWmmFromName(Supplier<String> nameSupplier) {
-        return createWmmFromPath(() -> getRootPath("cat/" + nameSupplier.get() + ".cat"));
+        return createWmmFromPath(() -> Path.of(getRootPath("cat/" + nameSupplier.get() + ".cat")));
```
hernanponcedeleon, line 40: Isn't the `cat/` problematic on e.g., Windows?

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/WmmFromArchitectureProvider.java
```diff
@@ -26,14 +26,14 @@ public class WmmFromArchitectureProvider extends AbstractProvider<Wmm> {
...
-        ARCH_WMM_MAP.put(Arch.VULKAN, () -> new ParserCat().parse(new File(getRootPath("cat/vulkan.cat"))));
-        ARCH_WMM_MAP.put(Arch.OPENCL, () -> new ParserCat().parse(new File(getRootPath("cat/opencl.cat"))));
+        ARCH_WMM_MAP.put(Arch.TSO, () -> new ParserCat().parse(Path.of(getRootPath("cat/tso.cat"))));
+        ARCH_WMM_MAP.put(Arch.ARM8, () -> new ParserCat().parse(Path.of(getRootPath("cat/aarch64.cat"))));
+        ARCH_WMM_MAP.put(Arch.POWER, () -> new ParserCat().parse(Path.of(getRootPath("cat/power.cat"))));
+        ARCH_WMM_MAP.put(Arch.RISCV, () -> new ParserCat().parse(Path.of(getRootPath("cat/riscv.cat"))));
+        ARCH_WMM_MAP.put(Arch.LKMM, () -> new ParserCat().parse(Path.of(getRootPath("cat/linux-kernel.cat"))));
+        ARCH_WMM_MAP.put(Arch.IMM, () -> new ParserCat().parse(Path.of(getRootPath("cat/imm.cat"))));
+        ARCH_WMM_MAP.put(Arch.VULKAN, () -> new ParserCat().parse(Path.of(getRootPath("cat/vulkan.cat"))));
+        ARCH_WMM_MAP.put(Arch.OPENCL, () -> new ParserCat().parse(Path.of(getRootPath("cat/opencl.cat"))));
```
hernanponcedeleon, line 36: Again, shouldn't we use something like `Path path = Path.of(getRootPath(""), "cat", "tso.cat");`

hernanponcedeleon (commented):
I haven't tested the UI, but bound and witness files work as expected

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
ThomasHaas: `getRootPath` normalizes the path already in a platform-agnostic manner.
Also I think you can use `/` to specify paths on windows without problems (`/` is a valid path separator on essentially all operating systems). However, you cannot assume that `toString` will return the path with `/` separators and so you have to be careful there (that's why I needed to add a `containsSubpath` method).

dartagnan/src/main/java/com/dat3m/dartagnan/smt/ProverWithTracker.java
ThomasHaas: The line of code does nothing and was stupid for 3 reasons:
- It removed duplicate declarations in the file name?! Was it supposed to access the file content?
- The (deduplicated) return value is unused.
- The side-effect of the method call is irrelevant as the prover gets closed anyhow.

hernanponcedeleon: Ok, it seems this was a remanent of the initial implementation https://github.com/hernanponcedeleon/Dat3M/blob/61381f3210eced4d1cf4084a68a44c7348447cd6/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ProverWithTracker.java#L48

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
hernanponcedeleon: I think `ResourceHelper` can be improved by returning a `Path` rather than a String for most methods. However, most (if not all) of those are used by the unit tests, so we can do this in another PR.

hernanponcedeleon (approved):

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
ThomasHaas: Definitely. But those changes affect like 100 unit test files (mostly the SPIRV tests I think) and I didn't want to go through them all. Especially with active PRs like #1067 modifying unit test code and hence resulting in conflicts.

PR #1067, opened 2026-07-27
Vulkan: Reformat old model

natgavrilenko:
- Extract the common part of chains and no-chains models
- Use no-chains model by default, as it is more relaxed
- Minor cleanup of the common model file (no semantic changes)


cat/vulkan-common.cat
```diff
@@ -0,0 +1,180 @@
...
+let SEMSC12 = SEMSC1 & SEMSC2
+let SEMSC13 = SEMSC1 & SEMSC3
+let SEMSC23 = SEMSC2 & SEMSC3
+let SEMSC012 = SEMSC01 & SEMSC2
+let SEMSC013 = SEMSC01 & SEMSC3
+let SEMSC023 = SEMSC02 & SEMSC3
+let SEMSC123 = SEMSC12 & SEMSC3
+let SEMSC0123 = SEMSC012 & SEMSC3
+
+let AVSG = AV | SEMAV
```
hernanponcedeleon:
```suggestion
let AVSG = (AV | SEMAV) & (DV | QF | WG | SG)
```
This make it more consistent with the next few lines and how we are doing it for `inscope` definition

cat/vulkan-common.cat
```diff
@@ -0,0 +1,180 @@
...
+let AVWG = (AV | SEMAV) & (DV | QF | WG)
+let AVQF = (AV | SEMAV) & (DV | QF)
+let AVSHADER = (AV | SEMAV) & DV
+
+let VISSG = VIS | SEMVIS
```
hernanponcedeleon:
```suggestion
let VISSG = (VIS | SEMVIS) & (DV | QF | WG | SG)
```

cat/vulkan-common.cat
```diff
@@ -0,0 +1,180 @@
...
+let VISWG = (VIS | SEMVIS) & (DV | QF | WG)
+let VISQF = (VIS | SEMVIS) & (DV | QF)
+let VISSHADER = (VIS | SEMVIS) & DV
+
+(* Relations *)
+
+let inscope = (DV * DV) |
+    (sqf & ((DV | QF) * (DV | QF))) |
+    (swg & ((DV | QF | WG) * (DV | QF | WG))) |
+    (ssg & ((DV | QF | WG | SG) * (DV | QF | WG | SG)))
```
hernanponcedeleon, line 79: Would it make sense to distinguish between the "precise" and "at least" scope ... the same idea is used in [vmm](https://github.com/open-s4c/vatomic/blob/9cb5a3853947bf146136a712e81e5c39f5e5732a/vmm.cat#L91).

This would make this and the `AVX` and `VISX` definition a bit more readable IMO

cat/vulkan-common.cat
```diff
@@ -0,0 +1,180 @@
...
+    ([REL & ATOM]; rs; (rf & mutordatom); [ACQ & ATOM]) |
+    ([REL & F]; posemtosc; [ATOM & W]; hypors; (rf & mutordatom); [ACQ & ATOM]) |
+    ([REL & ATOM]; rs; (rf & mutordatom); [ATOM & R]; posctosem; [ACQ & F]) |
+    ([REL & F]; posemtosc; [ATOM & W]; hypors; (rf & mutordatom); [ATOM & R]; posctosem; [ACQ & F]) |
+    ([REL & F]; po?; [CBAR]; ((syncbar & inscope) \ id); [CBAR]; po?; [ACQ & F]))
+
+let ithb(SC, SEMSC) = (ssw |
+    ([SEMSC]; sw; [SEMSC]) |
+    ([SC | SEMSC]; po; [REL & SEMSC]) |
+    ([ACQ & SEMSC]; po; [SC | SEMSC]))+
```
hernanponcedeleon, line 110: I would use a more generic name for the input parameters (I think `SC` can be easily confused by a specific `SCX`), but I don't have a strong opinion here.

cat/vulkan-common.cat
natgavrilenko: There are already many tags in this model, so I wouldn't introduce more. But I can replace individual scope tags with inclusive scopes, as we do in the new model.

Note that also AV and VIS tags are redundant, because a non-empty scope for a write implies AV and a non-empty scope for a read implies VIS. 

I tried to keep a reasonable balance between readability and matching the alloy model, but I can prioritaze readablity if you prefer. I don't really have a strong opinion here.

cat/vulkan-common.cat
natgavrilenko: I would keep as it is: SC can be subsctitutes with any concrete SC0, SC1, SC2, SC01, etc and the same for SEMSC

PR #1068, opened 2026-08-04
Refactor output generation

ThomasHaas:
- Replaced `TaskResultAnalyzer` with `OutputGenerator`
- `OutputGenerator` automatically generates witnesses if possible (the witness configuration are part of that class now)
- `OutputGenerator` generates `Outputs` (replacement of `ResultSummary`) which contain an exit code and an arbitrary summary string. The summary string is more general than the previous `ResultSummary`. I'm not even sure if the exit code should be part of the output here or if we should generate plain strings.
- Removed `OutputLogger`. It didn't provide much more functionality than a `List<Output>` which we use now.

Possible TODOs:
- I'm wondering if I should put an additional `Map<String, Object>` into `Output` to allow for arbitrary extra information to be attached. For example, if a witness gets generated, we could add a mapping `"witness" -> witnessPath`. I'm also thinking about adding the witness file path to the summary of the output. After all, it is part of the output.


hernanponcedeleon:
Witness generation seems to be broken. The following does not create the png file.
```
CFLAGS="-DACQ2RX" dartagnan cat/vmm.cat benchmarks/locks/ticketlock.c --witness=png
```

ThomasHaas:
You ran the native build, right? I forgot to update the configs for that one. The Java version works fine for me

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -69,11 +63,11 @@ public static void main(String[] args) throws Exception {
         final Path catFile = getCatFileFromArgs(args);
         final List<Path> progFiles = getProgramFilesFromArgs(args);
         final boolean isBatchMode = progFiles.size() > 1;
+        final OutputGenerator outputGenerator = OutputGenerator.create(config);
+        outputGenerator.setBatchMode(isBatchMode);
```
hernanponcedeleon: Why not passing the batch mode to the constructor and make the `private boolean isBatchMode` final? I don't think we really have any use scenario where we would want to change this after creation

hernanponcedeleon (commented):
Since you are changing the generated output, I think the following change would also make sense as part of this PR.

Right now, when we run in batch mode, we log several entries like this
```
[05.08.2026] 20:52:06 [INFO] Dartagnan.lambda$getProgramFilesFromArgs$2 - Program(s) path: litmus/VULKAN/KhronosGroup
```
I think it would make more sense to move the following log call from `getProgramFilesFromArgs`
```
logger.info("Program(s) path: {}", path.normalize());
```
to the `for (Path progFile : progFiles)` loop, such that one can relate any logged statistics with the correct test

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: I will change it. I'm not yet sure if batch mode is really relevant though. It really is only used to give different names to witnesses when explicit witness names are provided. I think nobody will use this and we might as well delete the batch mode in `OutputGenerator`.

ThomasHaas:
Btw, I can also change the `UNKNOWN` result to `BOUNDED` or `BOUNDED_SAFE/FAIL` in this PR if you want.

ThomasHaas:
> Since you are changing the generated output, I think the following change would also make sense as part of this PR.
> 
> Right now, when we run in batch mode, we log several entries like this
> 
> ```
> [05.08.2026] 20:52:06 [INFO] Dartagnan.lambda$getProgramFilesFromArgs$2 - Program(s) path: litmus/VULKAN/KhronosGroup
> ```
> 
> I think it would make more sense to move the following log call from `getProgramFilesFromArgs`
> 
> ```
> logger.info("Program(s) path: {}", path.normalize());
> ```
> 
> to the `for (Path progFile : progFiles)` loop, such that one can relate any logged statistics with the correct test

It does change the semantics though. The current version can print the whole directory rather than each test in isolation (as your example shows).
We could, however, keep the printing of directories and move the printing of single files into the loop.

hernanponcedeleon:
> Btw, I can also change the `UNKNOWN` result to `BOUNDED` or `BOUNDED_SAFE/FAIL` in this PR if you want.

I am not sure about this change. On the one side, I acknowledge that maybe the current meaning of `UNKNOWN` is not crystal clear and something like `BOUNDED_PASS` might be more self-explanatory. However, this is not a "small" change (e.g., it is not just string in the detailed explanation, it is actually changing the result of the verification).

> It does change the semantics though.

I know it does. The point is: all data in one particular single-test verification run is specifically related to the single program, and not to the path

ThomasHaas:
> > Btw, I can also change the `UNKNOWN` result to `BOUNDED` or `BOUNDED_SAFE/FAIL` in this PR if you want.
> 
> I am not sure about this change. On the one side, I acknowledge that maybe the current meaning of `UNKNOWN` is not crystal clear and something like `BOUNDED_PASS` might be more self-explanatory. However, this is not a "small" change (e.g., it is not just string in the detailed explanation, it is actually changing the result of the verification).

It is not much to change, but it is true that there is a semantic change for the RefinementSolver which can produce a proper UNKNOWN (apart from that usage of UNKNOWN, I think all other one's are actually `BOUNDED_PASS` and thus a simple renaming will capture them).

> 
> > It does change the semantics though.
> 
> I know it does. The point is: all data in one particular single-test verification run is specifically related to the single program, and not to the path

So I assume you are fine with the change and don't want the hybrid approach that prints both.

dartagnan/src/main/java/com/dat3m/dartagnan/OutputGenerator.java
```diff
@@ -16,62 +16,108 @@
...
+        return Optional.ofNullable(witnessFile);
+    }
+
+    // ====================================================================================
+
+    public Output getOutputFromException(Exception exception, String programPath) {
         final String message = exception.getMessage() != null ? exception.getMessage() : "Unknown error occurred";
         final String details = "\t" + message;
-
+
```
hernanponcedeleon: Remove

dartagnan/src/main/java/com/dat3m/dartagnan/OutputGenerator.java
```diff
@@ -16,62 +16,108 @@
...
-
+        
         if (exception instanceof InterruptedException) {
             final ExitCode exitCode =
                     message.contains("Timeout") ? TIMEOUT_ELAPSED
                     : message.contains("canceled") ? CANCELED
                     : UNKNOWN_ERROR;
-            return new ResultSummary(programPath, "", INTERRUPTED, "", "", details, 0, exitCode);
+            return new Output(exitCode, toSummary(programPath, "", INTERRUPTED,
+                    "", "", details, 0, null));
```
hernanponcedeleon, line 116: The class member `witnessFile` is still null here, right? Should we use that instead of an explicit `null` (I had to go to the `toSummary` to check what actually this parameter was)?

Same for the else case below.

EDIT: nvm, since this method can be called once for each program in batch mode, I guess the witness from the previous iteration might still be there rather than `null`.

hernanponcedeleon:
Either than the minor format thing, once the conflict is fixed, I think we can merge

ThomasHaas:
I added back uncaught exceptions from #1020 but instead registered a handler for uncaught exceptions.
That handler prints the error to system.err and always generates the `UNKNOWN_ERROR` exit code.
This simplifies the code again and is also more robust.

hernanponcedeleon:
There is inconsistency on how we report missing cat file vs missing program file
```
> dartagnan cat/svcomp2.cat ../sv-benchmarks/c/pthread/lazy01.c
Test: ../sv-benchmarks/c/pthread/lazy01.c
Result: ERROR
Reason: NoSuchFileException
Details:
        cat/svcomp2.cat
```
vs
```
> dartagnan cat/svcomp.cat ../sv-benchmarks/c/pthread/lazy0.c
Path to input program(s) not given or format not recognized
```

ThomasHaas:
I changed the missing cat file output now. But note that if you really want to, you can always find some discrepancies.
For example, if you execute Dartagnan in batch mode and during solving you delete the .cat file, you will also get
```
Test: ...
Result: ERROR
Reason: NoSuchFileException
Details:
        cat/svcomp2.cat
```
instead of (the new)
```
CAT file cat/svcomp2.cat does not exist
```

Technically, you don't even need batch mode if you have a good timing and delete the cat file after it has been found but before it has been parsed :).

ThomasHaas:
Btw. if you prefer to always have an output of (at least) this shape for errors:
```
Result: ERROR
Reason: NoSuchFileException
Details:
        cat/svcomp2.cat
```
I can change it to this format. Note that there is no test case mentioned here (importantly, nothing beyond the exception itself is mentioned).
Once we do solving for a particular program, we can prepend the program to the error message.

ThomasHaas:
I made `OutputLogger.getOutputFromException` a static method with optional program parameter.
This allows the usage of the method everywhere in the code base without issues. Now all exceptions generate output of shape
```
[Test : ...]
Result: ERROR
Reason: NoSuchFileException
Details:
        cat/svcomp2.cat
```

hernanponcedeleon (approved):

PR #1072, opened 2026-08-12
Make register initialization work with spinloop detection

hernanponcedeleon:
@ThomasHaas I did some small renaming on top of your code. If you are fine with those changes, we can merge.


ThomasHaas:
Just keep in mind that externals tools like Telechat need to use the new option name now.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
> Just keep in mind that externals tools like Telechat need to use the new option name now.

Yes. That option is only used by telechat and I will tell Luke to update it before starting a new testing campaing

PR #1073, opened 2026-08-13
Fix some options

hernanponcedeleon:
Some options were not working in native mode, and also nor shown in the help (even for the JVM version). I also make options related to initialization (memory, registers, allocations) more consistent.


ThomasHaas (approved):
LGTM

PR #1074, opened 2026-08-15
Generalize VerificationTask and related classes

ThomasHaas:
This PR supersedes #1063.


hernanponcedeleon:
I'm traveling ATM and just with my phone. I'll take a look at the beginning of September.

ThomasHaas:
Fine with me

dartagnan/src/test/java/com/dat3m/dartagnan/llvm/AbstractCTest.java
```diff
@@ -2,12 +2,13 @@
...
 import com.dat3m.dartagnan.program.Program;
-import com.dat3m.dartagnan.utils.Result;
+import com.dat3m.dartagnan.verification.ResultStatus;
 import com.dat3m.dartagnan.utils.rules.Provider;
 import com.dat3m.dartagnan.utils.rules.Providers;
 import com.dat3m.dartagnan.utils.rules.RequestShutdownOnError;
-import com.dat3m.dartagnan.verification.TaskSolver;
 import com.dat3m.dartagnan.verification.VerificationTask;
+import com.dat3m.dartagnan.verification.VerificationTaskSolver;
+import com.dat3m.dartagnan.verification.Task;
```
xeren: Unused import

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
```diff
@@ -49,8 +50,8 @@ public static Provider<Program> createProgramFromPath(Supplier<Path> pathSupplie
     // =========================== Task related providers ==============================
 
     public static Provider<VerificationTask> createTask(Supplier<Program> programSupplier, Supplier<Wmm> wmmSupplier, Supplier<EnumSet<Property>> propertySupplier,
-                                                        Supplier<ProgressModel.Hierarchy> progressModelSupplier, Supplier<Configuration> config) {
-        return Provider.fromSupplier(() -> VerificationTask.builder().
+                                            Supplier<ProgressModel.Hierarchy> progressModelSupplier, Supplier<Configuration> config) {
+        return Provider.fromSupplier(() -> Task.builder().
```
xeren: Spacing

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -97,7 +98,10 @@ public void setShutdownManager(ShutdownManager shutdownManager) {
     }
 
     public boolean hasModel() {
-        final Property.Type propType = Property.getCombinedType(context.getTask().getProperty(), context.getTask());
+        if (!(context.getTask() instanceof VerificationTask veriTask)) {
+            return false;
+        }
+        final Property.Type propType = Property.getCombinedType(veriTask.getProperties(), veriTask);
```
xeren, line 104: `ModelChecker` could still require `VerificationTask` specifically.  Then `AssumeSolver` and `RefinementSolver` would not need to check this at runtime.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTaskSolver.java
```diff
@@ -0,0 +1,104 @@
...
+import org.sosy_lab.common.configuration.TimeSpanOption;
+import org.sosy_lab.java_smt.api.SolverException;
+
+import java.util.concurrent.TimeUnit;
+
+import static com.dat3m.dartagnan.configuration.OptionNames.METHOD;
+import static com.dat3m.dartagnan.configuration.OptionNames.TIMEOUT;
+
+@Options
+public final class VerificationTaskSolver extends TaskSolverBase<VerificationTaskSolver, VerificationTask, VerificationResult> implements AutoCloseable {
```
xeren, line 16: Unused imports.

xeren (commented):
### Additions
- `sealed interface Task permits VerificationTask`
- `TaskSolverBase implements TaskSolver`, `TaskResult`
- `VerificationResult implements`
- `ExecutionModelManager.fromIREvaluator(IREvaluator)`

### Movements
- `verification.VerificationTask` -> `verification.Task`
- `VerificationTask.getProperty()` -> `VerificationTask.getProperties()`
- `utils.Result` -> `verification.ResultStatus`
- `TaskSolver` -> `VerificationTaskSolver`
- `solver.getResult()` -> `solver.getResult().getStatus()`
- `VerificationTask.VerificationTaskBuilder` -> `Task.TaskBuilder`
- `TaskSolver.method` -> `VerificationTaskSolver.method`
- `TaskSolver.{timeout,shutdownManager,runtime}` -> `TaskSolverBase.{...}`
- `TaskSolver.model` -> `VerificationResult.model`

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
ThomasHaas: Sure, but it does depend a little bit on how I implement the exploration mode for Telechat: 
- as a separate solver class still inheriting from `ModelChecker` (this is the current "hacky" implementation on another branch)
- as a separate solver class not inheriting from `ModelChecker`
- as just another mode of `AssumeSolver` and possibly `RefinementSolver`.

So whatever choice I make now is likely to get changed in the following PR. But if you prefer, I can force `AssumeSolver` and `RefinementSolver` to require `VerificationTask` at compile time.

dartagnan/src/test/java/com/dat3m/dartagnan/utils/rules/Providers.java
ThomasHaas: I'm not sure about what you mean.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
ThomasHaas: I made `AssumeSolver` and `RefinementSolver` accept only `VerificationTask`. `ModelChecker` still accepts any task.

hernanponcedeleon:
@ThomasHaas can you please check I did not mess anything up when fixing the conflicts? I will try to review tomorrow

ThomasHaas:
The changes look fine

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
xeren: Ok, that works for now.  I'll probably be commenting on the hackiness in that other PR.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
```diff
@@ -15,8 +17,10 @@
 import org.sosy_lab.java_smt.api.SolverContext;
 import org.sosy_lab.java_smt.api.SolverException;
 
-import static com.dat3m.dartagnan.utils.Result.FAIL;
-import static com.dat3m.dartagnan.utils.Result.PASS;
+import java.util.EnumSet;
```
xeren, line 20: Unused, as is `Preconditions`.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/RefinementSolver.java
```diff
@@ -36,6 +37,7 @@
 import com.dat3m.dartagnan.wmm.axiom.Emptiness;
 import com.dat3m.dartagnan.wmm.definition.*;
 import com.dat3m.dartagnan.wmm.utils.Dimension;
+import com.google.common.base.Preconditions;
```
xeren, line 40: unused

dartagnan/src/test/java/com/dat3m/dartagnan/litmus/AbstractLitmusTest.java
```diff
@@ -6,12 +6,13 @@
...
 import com.dat3m.dartagnan.utils.ResourceHelper;
-import com.dat3m.dartagnan.utils.Result;
+import com.dat3m.dartagnan.verification.ResultStatus;
 import com.dat3m.dartagnan.utils.rules.Provider;
 import com.dat3m.dartagnan.utils.rules.Providers;
 import com.dat3m.dartagnan.utils.rules.RequestShutdownOnError;
-import com.dat3m.dartagnan.verification.TaskSolver;
 import com.dat3m.dartagnan.verification.VerificationTask;
+import com.dat3m.dartagnan.verification.VerificationTaskSolver;
+import com.dat3m.dartagnan.verification.Task;
```
xeren, line 15: unused

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
```diff
@@ -31,7 +35,7 @@ public static AssumeSolver create(VerificationTask task) throws InvalidConfigura
         return new AssumeSolver(task);
     }
 
-    protected Context preprocessAndAnalyse(VerificationTask task) throws InvalidConfigurationException {
+    protected Context preprocessAndAnalyse(Task task) throws InvalidConfigurationException {
```
hernanponcedeleon, line 38: Since this method is part of the `AssumeSolver` which requires not a general `Task`, but explictly a `VerificationTask`, wouldn't it make sense to also use `VerificationTask` here?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
hernanponcedeleon, line 104: What is the intention here? At first glance I would assume that VerificationTask is anything that gets encoded as an SMT problem and can thus have a model. But then exploration (at least as implemented for telechat) will also be encoded and SMT, but probably will get a different Task

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
```diff
@@ -165,24 +169,24 @@ public void close() {
...
         ProcessingManager.fromConfig(config).run(program);
     }
 
-    public static void preprocessMemoryModel(VerificationTask task, Configuration config) throws InvalidConfigurationException{
+    public static void preprocessMemoryModel(Task task, Configuration config) throws InvalidConfigurationException{
         final Wmm memoryModel = task.getMemoryModel();
 
         // We remove flagged axioms if we do not check for them.
-        if (!task.getProperty().contains(Property.CAT_SPEC)) {
+        if (task instanceof VerificationTask veriTask && !veriTask.getProperties().contains(Property.CAT_SPEC)) {
```
hernanponcedeleon, line 181: Why is this only meaninful for VerificationTask but not for others?

dartagnan/src/main/java/com/dat3m/dartagnan/verification/Task.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public TaskBuilder withBound(int k) {
+            checkArgument(k > 0 , "Unrolling bound must be positive.");
+            this.config.setOption(BOUND, Integer.toString(k));
+            return this;
+        }
+
+        public TaskBuilder withProgressModel(ProgressModel.Hierarchy progressModel) {
+            this.progressModel = progressModel;
```
hernanponcedeleon, line 72: we can add `checkNotNull`

dartagnan/src/main/java/com/dat3m/dartagnan/verification/Task.java
```diff
@@ -0,0 +1,100 @@
...
+            return this;
+        }
+
+        public TaskBuilder withSolverTimeout(int t) {
+            this.config.setOption(TIMEOUT, Integer.toString(t));
```
hernanponcedeleon, line 77: We can add `checkArgument(t > 0)`

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
```diff
@@ -679,7 +679,7 @@ public void aliasAnalysisDeterminism() throws Exception {
         final Configuration config = Configuration.builder()
                 .setOption(ENABLE_EXTENDED_RELATION_ANALYSIS, "false")
                 .build();
-        final VerificationTask task = VerificationTask.builder()
+        final Task task = Task.builder()
```
hernanponcedeleon, line 682: Why are you using Task here while in most of the test you use VerificationTask.

EDIT: Actually, not sure if Task or VerificationTask is used more, but I do not see a claer pattern why sometimes once is used, while in other palces the other one is used.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTaskSolver.java
```diff
@@ -0,0 +1,93 @@
...
+
+    // ===================================== Misc =====================================
+
+    @Override
+    protected VerificationTaskSolver getThis() {
+        return this;
+    }
+
+    @Override
+    public void close() {
```
hernanponcedeleon, line 81: Codex reports a potential problem here
```
  - VerificationTaskSolver.close() closes the model inside its VerificationResult, but VerificationResult remains accessible and still reports hasModel() == true. A second close() also closes the same model again. Before this refactor,
    closing TaskSolver cleared the stored model. The result should either own the model and become closeable/invalidated, or the solver should clear/invalidate it. 
```

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
ThomasHaas: `VerificationTask` refers to the task of "verifying a property" independent from how you solve it (BMC / SMC whatever). Such tasks may have witnesses in the form of models (usually counterexamples for the property).
Other tasks might not have such witnesses like e.g., `EnumerationTask` (or `ExplorationTask`) which enumerates all models. That's why I need the check in general.

Btw. the function existed like that already, modulo the check for verification tasks. One can argue that maybe the method should not reverse-engineer from the property and the outcome whether a model exists or not. Maybe the AssumeSolver/RefinementSolver should just set a "hasModel" variable or uncondtionally generate the model if it has found one. But that is a change orthogonal to this PR.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
ThomasHaas: I don't know if the optimization is sound for other tasks (that don't exist yet). I only know that if I verify properties and the propery I verify is not CAT_SPEC, I can delete flagged axioms. It will probably also sound for `EnumerationTask` in the future.

Btw. the `instanceof` check is also used to cast to `VerificationTask` because the general `Task` object has no `getProperties()` method. The new code is equivalent to the old code, just with the fact that the old code always had a `VerificationTask` at hand and so didn't need to do the cast.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/Task.java
ThomasHaas: I think we allow `0` to disable timeouts. The check `>= 0` could be made, but it is also part of the JavaSMT option annotation which can have integer bound constraints. The difference, of course, is that JavaSMT will throw an `InvalidConfigurationException` at injection time whereas the manually added check gives a `IllegalArgumentException` at construction time.

dartagnan/src/test/java/com/dat3m/dartagnan/others/miscellaneous/AnalysisTest.java
ThomasHaas: I think I used `Task` if the actual task was irrelevant. All of the analysis tests never actually solve a task, they just run some analyses. Most other unit tests actually generate and solve verification tasks, not any other task.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/VerificationTaskSolver.java
ThomasHaas: I forgot to null the result here.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/AssumeSolver.java
ThomasHaas: But it doesn't need any extra information from the `VerificationTask`. I guess the question goes into same direction as René's. It all depends on how we implement and solve other tasks.
For example, I could easily change `AssumeSolver` to solve both verification and enumeration tasks because they differ by not much and both use the same processing/analyses. Then `Task` here is the right choice.
I could also make a new solver entirely like it is implemented in the `telechat` branch.
In general though, both assume solving and refinement solving can solve both verification tasks and enumeration tasks (the method used and the task to be solved are orthogonal parameters).

Btw. IIRC I replaced ALL usages of the old `VerificationTask` with the more general `Task` and everywhere where the property was relevant/needed, I changed it back to `VerificationTask`. This should roughly explain the usages.

dartagnan/src/main/java/com/dat3m/dartagnan/verification/solving/ModelChecker.java
hernanponcedeleon: If an enumeration procedure just enumerate states or an exploration procedure just lists executions, I get it ... but what if I want to enumerate all states and show which of these violate a property? Would this be a VerificationTask?

ThomasHaas: No, I don't think those cases would classify as `VerificationTask`. `VerificationTask` is essentially a final class representing the task of verifying properties of a single program under a single memory model (exactly what it represents in `development` right now). Any other task would be reflected by a different variant of `Task` and I think in your example, we could have an `EnumerationTask` with an optional condition related to properties. But the result of such a task is not a single execution but either a list of executions or a list of states. Besides different models, the result might also not use `FAIL/PASS/UNKNOWN` as status because there is not necessarily a notion of "wrong" (e.g., if you do pure final state enumeration).
All this would be reflected in a corresponding `EnumerationTaskResult` that does not contain a single model but a whole list of models or final states.

Genereally, the idea is to have for any possible task(*) we want to solve three essential classes: `XTask`, `XTaskResult` and `XTaskSolver`, where the solver takes an `XTask` as input and generates an `XTaskResult` as output.
If you have multiple variants of the same task (similar to configurations), we can probably pack them into a single common task class with optional parameters.
The `XTaskSolver` can then use, based on configurations, different backends to do the actual solving, be it BMC, SMC, or anything else. Ideally, this class should also control the processing pipeline I think,
Backend classes like `ModelChecker` and `AssumeSolver` should be implementation details, and ideally wouldn't even need to know (too much) about what task they are actually solving. Maybe they can be used in different ways (e.g. in a loop for state enumeration) to solve different tasks.

(*) We still assume that all tasks involve a single program, a single memory model and a single progress model, i.e., I don't have generalized the idea of "tasks" to something like state-inclusion between different programs or equivalence under different memory models. This would require quite a bit more effort.

hernanponcedeleon (approved):

PR #1075, opened 2026-08-19
Fix branch equivalence

xeren:
Fixes a bug where uneliminated dead code leads to asymmetric results in `BranchEquivalence.areMutuallyExclusive(Event,Event)`.

Let `a` and `b` be unreachable events and `c` be reachable.  
```java
be.areMutuallyExclusive(a, a) // false -> true
be.areMutuallyExclusive(a, b) // false -> true
be.areMutuallyExclusive(a, c) // false -> true
be.areMutuallyExclusive(c, a) // true  -> true
```

Solves a problem in #1060.  A failing test showed different results between `LazyRelationAnalysis` and `NativeRelationAnalysis` (without XRA) on some programs.  The only affected relation was `loc`.  While LRA relied on the symmetry assumption of `loc`, NRA did not.  So at least one of the subroutines `AliasAnalysis.mayAlias` or `ExcutionAnalysis.areMutualExclusive` had to be asymmetric for some event pairs.  That MR exposed this bug because it reorders the branches and alters the results of LRA.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BranchEquivalence.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BranchEquivalence.java
@@ -229,6 +229,7 @@ public class BranchEquivalence extends AbstractEquivalence<Event> {
                 final Branch creatorBranch = decomposition.event2BranchMap.get(start.getCreator());
                 threadInitialBranch.mustPred.add(creatorBranch);
                 threadInitialBranch.exclusiveBranches.addAll(creatorBranch.exclusiveBranches);
+                creatorBranch.exclusiveBranches.forEach(b -> b.exclusiveBranches.add(threadInitialBranch));
                 creatorBranch.mustSucc.add(threadInitialBranch);
             } else {
                 unconditionalThreadInitialBranches.add(threadInitialBranch);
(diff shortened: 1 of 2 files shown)


dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/BranchEquivalence.java
```diff
@@ -339,6 +339,7 @@ private void createBranchClasses(BranchDecomposition decomposition) {
                     .get();
 
             unreachableClass.impliedClasses.addAll(this.getAllTypedEqClasses());
+            unreachableClass.exclusiveClasses.addAll(this.getAllTypedEqClasses());
```
ThomasHaas, line 343: You should probably remove `unreachableClass` afterwards. In practice it doesn't matter, but it feels awkward to say that a class is mutually exclusive with itself.

ThomasHaas (dismissed):
LGTM. You just need to rebase the branch

hernanponcedeleon:
> LGTM. You just need to rebase the branch

Have you tested this indeed solved the problem from #1060 ?

ThomasHaas:
I haven't. But this PR is reasonable either way. I think we can just merge and rebase the other PR to see if the problems are fixed.

hernanponcedeleon:
I merged both branch and tested locally and `RelationAnalysisTest` still fails.

xeren:
> I merged both branch and tested locally and `RelationAnalysisTest` still fails.

There was another line missing to pass `RelationAnalysisTest` on the other branch.  I must have skipped it during rebasing.

PR #1076, opened 2026-08-19
Fix indexed event graph

xeren:
Fixes a bug in `IndexedEventGraph.getRange(Event)` where the wrong `IndexedDomain` was used to check the domain-sided event.  This leads to graphs with different domains appearing smaller than they actually are.  (

Let `dom` and `rng` be of type `IndexedDomain<Event>`.  Let `a` and `b` be events.
```java
assume(dom.fullSet().contains(a));
assume(rng.fullSet().contains(b));
assume(!rng.fullSet().contains(a));
var g = new IndexedEventGraph(dom, rng);
g.add(a, b);
g.contains(a, b); // before: true, now: true
g.size(); // before: 1, now: 1
g.getRange(); // before: [b], now [b]
g.getRange(a) // before: [], now: [b]
g.toString(); // before: "[]", now: "[(a,b)]"
```

This bug affects `NativeRelationAnalysis` on relations with different domains, like `__iddTrans ; __addrDirect`.  All of them are only joined over visible events (in the smaller domain).  So the missing edges are inactive.  No verification result is affected.


ThomasHaas (approved):
LGTM

Issue #1078, opened 2026-08-21
[Bug] ARMv8 CAS instruction (AMO) should not produce ctrl-deps


ThomasHaas:
We produce different results than herd7 on several of the following tests:
https://github.com/herd/herdtools7/tree/master/catalogue/aarch64-cas/tests

It seems that a CAS should not produce a ctrl-dep unlike LX/SX loops do.
The fix for this is quite easy because I already introduced a `NO_CARRY_DEPS` tag in some earlier PR, which can be used to tag instructions so that they do not carry dependencies anymore (not fully implemented yet, but easy to do).

An open question is whether we should also remove ctrl-deps from AMO CASes of other architectures like RISCV or even C11.
Notice that for C11 there is probably no practical case where it matters because almost all usages of CAS are followed by an explicit check of its return value which again provides a control dependency.

EDIT: I think ctrl-deps are not the only bug, even fixing those, we have 4 or 5 tests with different results.

hernanponcedeleon:
> An open question is whether we should also remove ctrl-deps from AMO CASes of other architectures like RISCV or even C11. Notice that for C11 there is probably no practical case where it matters because almost all usages of CAS are followed by an explicit check of its return value which again provides a control dependency.

We should port some of the test where we give wrong results because of the extra Ctrl dep to c-litmus or riscv-litmus and see what herd7 generates. Since in both cases herd7 generates a single event (assuming for riscv this is amo-cas or whatever it is called), they probably do not generate a Ctrl dep.

ThomasHaas:
Some of the different results were due to uninitialized registers. Fixing those, I think there are two remaining tests which we forbid (even with removed ctrl-dep) but herd7 permits:
- MP+rel+CAS-ok-MRs-addr.litmus
- MP+rel+CAS-ok-RnRs-addr.litmus

hernanponcedeleon:
One of them I would not be surprised if it is related to some special behavior of the zero register.

ThomasHaas:
Actually, the test is quite wild. It seems that a succeeding CAS might not update the register containing the expected value with the loaded value since they are equal anyway. But then there is no data dependency from the load of a successful CAS to subsequent events, only from a failing one :O. Since the equality is checked without producing a ctrl-dep, there is neither a ctrl-dep: there are no dependencies at all! [EDIT: However, the CAS tunnels input dependencies through, see below for more details]

ThomasHaas:
[...]

PR #1079, opened 2026-08-21
[DRAFT] Fix CAS dep arm8

ThomasHaas:
See #1078 for details.

In short:
- CAS does not produce ctrl-deps unlike LX/SX.
  EDIT: I think the CAS should still produce a ctrl-dep to the CAS store but not beyond. This would be similar to language-level ctrl-deps. I guess we need better support from structured control-flow in the core (`IfAsJump` feels kinda hacky ATM).
- CAS does not (necessarily) produce a data dep on successful CAS. This seems surprising, but the idea is that a successful CAS can funnel through the expected value rather than the freshly loaded value (since they are equal anyhow). I assume that in this case, if the expected value is loaded from memory, the CAS may extend the dependency from that load instead of producing a new dependency!
EDIT: This behavior is actually non-deterministic and depends on whether the hardware can predict the success of the CAS. A faithful implementation would need to guess here (the current one always assumes correct prediction).

TODO: 
- Add all tests mentioned in #1078
- Fix the problems mentioned in the EDITs above

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/encoding/WmmEncoder.java
@@ -336,6 +336,7 @@ public class WmmEncoder {
 
         // ASSUMPTION: The active set of a static relation is a subset of the most precise may-set.
         //             This holds true as long as our RA computes the most precise may-set for static relations.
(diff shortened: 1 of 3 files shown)


ThomasHaas:
I think this behavior is extremely subtle and might be a good candidate for finding compiler bugs :)

ThomasHaas:
I will close this PR for now due to too many uncertainties in the handling of CASs and their dependencies.

PR #1080, opened 2026-09-02
Some improvements to the UI

hernanponcedeleon:


hernanponcedeleon:
Since we can now cancel the run (and can see time progress), maybe we should even get rid of the solver timeout pane.

ThomasHaas:
I haven't checked the changes but if the cancel button works fine, then we can get rid of the timeout pane.

ThomasHaas (dismissed):
LGTM

PR #1082, opened 2026-09-04
Flexible compilation

hernanponcedeleon:
This PR replaces the fixed C/Clang compilation with configurable compilation pipelines.

Compilation pipelines are defined in `compilation.yml` and can invoke arbitrary external tools to translate source files into a format natively parsed by Dartagnan. Pipeline configuration is validated when loaded, and the CLI reports the versions of the tools used by configured pipelines (those that are installed in the user environment).

`ProgramParser` and the UI now derive their supported source extensions from the configured pipelines. The UI compiles editor contents through the same pipeline mechanism, while preserving local include resolution for imported files and removing generated outputs afterwards.

The README and tests were updated to document and cover the new workflow.


ui/src/main/java/com/dat3m/ui/Dat3M.java
```diff
@@ -138,9 +148,8 @@ private void runTest() {
         final Editor programEditor = editorsPane.getEditor(EditorCode.PROGRAM);
         final String sourceCode = programEditor.getEditorPane().getText();
         final String format = programEditor.getLoadedFormat().isEmpty()
-                ? ProgramParser.EXTENSION_C            // We default to "c" code, if we do not know
+                ? ProgramParser.EXTENSION_LITMUS       // We default to litmus code, if we do not know
```
ThomasHaas: Why this change? Honestly, we should probably just put a "language/extension/format" dropdown where the user can select like C/Litmus/SPIRV (whatever formats we support). This would get rid of all the guesswork for when the user manually types in code or copy&pastes it.
And then if a user a imports a file, we can automatically set the option according to the file extension.

README.md
```diff
@@ -72,7 +75,7 @@ To run in JVM mode:
 ```
 java -jar $DAT3M_HOME/dartagnan/target/dartagnan.jar <CAT file> [--target=<arch>] <program file> [options]
 ```
-For programs written in `.c`, value `<arch>` specifies the programming language or architectures to which the program will be compiled. For programs written in `.litmus` format, if the `--target` option is not given, Dartagnan will automatically extract the `<arch>` from the litmus test header. `<arch>` must be one of the following: 
+For source programs processed by a compilation pipeline, value `<arch>` specifies the programming language or architecture used for verification. For programs written in `.litmus` format, if the `--target` option is not given, Dartagnan will automatically extract the `<arch>` from the litmus test header. `<arch>` must be one of the following:
```
ThomasHaas: The wording sounds like `arch` is related to the compilation pipeline introduced in this PR, but it is not at all. It is related to our internal compilation. To be honest, I think we should use more words/details to describe this parameter.

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/utils/PipelinesTest.java
```diff
@@ -0,0 +1,340 @@
...
+        Files.writeString(yaml, configuration);
+        return yaml;
+    }
+
+    private static void assertConfigurationError(Path yaml, IOException exception, String message) {
+        assertEquals("Invalid compilation pipeline configuration '%s': %s".formatted(yaml, message), exception.getMessage());
+    }
+
+    private record InvalidConfiguration(String yaml, String expectedMessage) {
+    }
```
ThomasHaas: As far as I see, this record is used in exactly one method. In that case, a neat little trick is to put the record declaration inside that method (you can define classes inside methods!). I often forget about this feature myself, but I think it is quite valueable.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -39,8 +40,13 @@ public class Dartagnan extends BaseOptions {
...
     private static final Logger logger = LoggerFactory.getLogger(Dartagnan.class);
 
-    private Dartagnan(Configuration config) throws InvalidConfigurationException {
+    private final Pipelines pipelines;
+    private final ProgramParser programParser;
+
+    private Dartagnan(Configuration config) throws InvalidConfigurationException, IOException {
         config.recursiveInject(this);
+        pipelines = Pipelines.load(getCompilationPipelinePath());
+        programParser = new ProgramParser(pipelines);
```
ThomasHaas: I'm not a big fan of this extra state here. I understand that you do it so that `getProgramFiles` can access the `programParser`. But I would just let those methods be static and pass in either the `ProgramParser` as a whole or just a `Predicate<Path>` with `programParser::isSupportedFile`. In the latter case one could, in principle, pass something like `p -> ProgramParser.isNativelySupported || pipelines.isSupported` and avoid the need to instantiate a single `ProgramParser` outside of the verification loop though I don't mind reusing the same instance over and over again.

Generally, I would rather go towards making `Dartagnan` a fully static class and just have `BaseOptions` as a standalone class in which we can inject the config. Then the main method would do this
```
final BaseOptions o  = new BaseOptions(config);
// instead of
// final Dartagnan o = new Dartagnan(config);
```
I wouldn't do this change in this PR, but I'm just saying that the PR shouldn't try to go the opposite direction of that goal :)

dartagnan/src/main/java/com/dat3m/dartagnan/utils/EnvironmentInfo.java
```diff
@@ -28,11 +30,16 @@ public static void initEnvironmentInfo () {
...
+    public static void logEnvironmentInfo(Set<String> tools) {
         logger.info("Git branch: {}", properties.getProperty("git.branch", "unknown"));
         logger.info("Git commit ID: {}", properties.getProperty("git.commit.id", "unknown"));
         logger.info("OS info: {}", getOSInfo());
-        logger.info("Clang version: {}", getClangInfo());
+        for (String tool : tools) {
+            final String version = getToolVersion(tool);
+            if (!version.equals("unknown")) {
+                logger.info("{} version: {}", tool, version);
+            }
```
ThomasHaas: I think it would be more elegant to have `getToolVersion` return an `Optional<String>`, seeing that you treat "unknown" exactly like an absent value.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ProgramParser.java
```diff
@@ -27,43 +34,60 @@ public class ProgramParser {
...
-    public static boolean isSupportedFile(Path filePath) {
-        return SUPPORTED_EXTENSIONS.contains(getFileExtension(filePath));
+    public ProgramParser() throws IOException {
+        this(Pipelines.load(getCompilationPipelinePath()));
+    }
+
+    public ProgramParser(Pipelines pipelines) throws IOException {
+        this.pipelines = pipelines;
+        supportedExtensions = Stream.concat(NATIVE_EXTENSIONS.stream(), pipelines.getSupportedExtensions().stream())
+                .collect(Collectors.toUnmodifiableSet());
```
ThomasHaas: Use `ImmutableSet` instead, also as type for the the class member `supportedExtensions` and the return type of `getSupportedExtensions`.

README.md
hernanponcedeleon: Better?

ThomasHaas:
I just need to check the UI changes. Everything else looks fine now.

hernanponcedeleon:
I just noticed that the SVCOMP script will not work cause it still uses CFLAGS. I will mark this PR as draft until I get access to a linux machine to test SVCOMP support there.

hernanponcedeleon:
I tried manually a few benchmarks and the setup is working correctly (the generated llvm has the correct instrumentation). There are a few svcomp violations I cannot find but I don't think those have anything to do with this PR.

ThomasHaas:
I tested the UI and got this error message when doing c code with .litmus extension:
```
Unknown litmus format#include
```
I have not checked where the message gets generated, but there is a missing space :).

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Pipelines.java
```diff
@@ -0,0 +1,294 @@
...
+            final String rawData = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8)
+                    .replace("$DAT3M_HOME", getHomeDirectory().toString())
+                    .replace("$DAT3M_OUTPUT", getOutputDirectory().toString());
+            final Object yamlData = new Yaml().load(rawData);
+            if (!(yamlData instanceof Map<?, ?> yamlMapping)) {
+                throw new IOException("Compilation pipeline configuration must be a YAML mapping");
+            }
+            final Map<String, Object> configuration = new LinkedHashMap<>();
+            for (Map.Entry<?, ?> entry : yamlMapping.entrySet()) {
+                if (entry.getKey() instanceof String key && !key.startsWith("x-")) {
```
xeren, line 69: Why skip `"x-"` prefixes?  Is this SnakeYAML-specific or a commentation alternative in configurations?

hernanponcedeleon: The `x-` prefix is a common YAML convention for auxiliary/template data. Those keys are not part of the `Pipelines` record, so passing them to Jackson would fail because unknown properties are deliberately rejected.

hernanponcedeleon:
> I tested the UI and got this error message when doing c code with .litmus extension:
> 
> ```
> Unknown litmus format#include
> ```
> 
> I have not checked where the message gets generated, but there is a missing space :).

Fixed

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLitmus.java
```diff
@@ -0,0 +1,29 @@
...
+            try {
+                // A failed parser may have consumed input before reporting its syntax error.
+                charStream.seek(0);
+                return parser.parse(charStream);
+            } catch (ParsingException ignored) {
+                // Try the next litmus grammar.
+            }
+        }
+        throw new ParsingException("No litmus parser recognizes the input.");
+    }
```
ThomasHaas, line 38: This looks a bit scary to me. Why the change? Also, this will hide all errors even if the litmus parser is the correct one.

hernanponcedeleon: It was an attempt to fix the ui problem you reported, but you are right that it would hide real parsing problems.

I ammended the last commit to keep the `LitmusParser` class (which simplifies some code), but I moved here the previous logic to select the correct litmus parser based on the first line.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLitmusAArch64.java
```diff
@@ -7,7 +7,7 @@
 import com.dat3m.dartagnan.program.Program;
 import org.antlr.v4.runtime.*;
 
-class ParserLitmusAArch64 implements ParserInterface {
+class ParserLitmusAArch64 extends ParserLitmus {
```
ThomasHaas: The inheritance here does not give you anything, no? I think you should revert it to `ParserInterface`.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/ParserLitmus.java
```diff
@@ -0,0 +1,55 @@
...
+    private static ParserInterface getParser(String type) {
+        return switch (type.toUpperCase()) {
+            case TYPE_AARCH64 -> new ParserLitmusAArch64();
+            case TYPE_PPC -> new ParserLitmusPPC();
+            case TYPE_X86 -> new ParserLitmusX86();
+            case TYPE_RISCV -> new ParserLitmusRISCV();
+            case TYPE_PTX -> new ParserLitmusPTX();
+            case TYPE_VULKAN -> new ParserLitmusVulkan();
+            case TYPE_C, TYPE_OPENCL -> new ParserLitmusC();
+            default -> throw new ParsingException("No litmus parser recognizes the input.");
```
ThomasHaas: The error message should be different again :)
The original was mostly fine, just the spacing was wrong.

hernanponcedeleon: Should we report the offending symbols? I.e., for the scenario you described, should the error popup be
```
Unknown litmus format #include
```
or just
```
Unknown litmus format
```
I find the former more confusing than helpful.

ThomasHaas: I think the former might be better like this:
```
Could not recognize litmus format from header: '#include'
```
Then it is somewhat clear how Dartagnan tries to recognize the format.
We could even list the supported formats `... Expected: X86, ...` if there are not too many.

README.md
```diff
@@ -66,13 +69,13 @@ There are three possible results for the verification:
...
-java -jar $DAT3M_HOME/dartagnan/target/dartagnan.jar <CAT file> [--target=<arch>] <program file> [options]
+java -jar $DAT3M_HOME/dartagnan/target/dartagnan.jar <CAT file> [--target=<target>] <program file> [options]
 ```
-For programs written in `.c`, value `<arch>` specifies the programming language or architectures to which the program will be compiled. For programs written in `.litmus` format, if the `--target` option is not given, Dartagnan will automatically extract the `<arch>` from the litmus test header. `<arch>` must be one of the following: 
+The `--target=<target>` option tells Dartagnan which programming-language or hardware-architecture semantics to use when verifying the program. It is independent of the programming language of the `<program file>`. A compilation pipeline performs the frontend step by translating a source file into a natively supported input format, while the target performs the backend step by selecting the semantics used during verification. For non-Litmus programs, the target must be given explicitly. For `.litmus` programs, Dartagnan extracts the target from the test header when `--target` is omitted. `<target>` must be one of the following:
```
ThomasHaas: I think we can word this better.
> It is independent of the programming language of the `<program file>`

Technically yes, but practical no. E.g., for SPIRV code, you can only really specify Vulkan/OpenCL.
I think we should be more explicit about the correlations between CAT models and the target. A possible suggestion:

> Dartagnan can verify programs relative to any given CAT memory model. However, different CAT memory models are defined over different instruction sets. Dartagnan can internally translate the program to the correct instruction set by specifying `--target=<target>`. For example, for LLVM programs (.ll) generated from C code, specifying `--target=C11` will keep the code as is (as expected by, e.g., `rc11.cat`) whereas specifying `--target=arm8` will cause Dartagnan to lower the C/LLVM code to ARM8 assembly (as expected by `aarch64.cat`) using standard compiler mappings. For non-Litmus programs, the target must be given explicitly. For `.litmus` programs, Dartagnan extracts the target from the test header when `--target` is omitted. `<target>` must be one of the following:
...

ThomasHaas (approved):
LGTM

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/utils/Pipelines.java
hernanponcedeleon: Does this answer your question? If so, I will merge

PR #1083, opened 2026-09-05
Add script and CI job to compare performance of PRs

hernanponcedeleon:
As an example, I ran the script locally comparing `development` with #1060 and this is what it generated (we might want to let the real CI run for that PR once the action is in place).

### Memory model: power

| Benchmark | Base branch | PR branch | Change |
| :--- | :--- | :--- | :--- |
| benchmarks/locks/cna.c | 68.667 ± 2.517 s | 66.667 ± 0.577 s | ✅ 2.9% |
| benchmarks/lfds/dglm.c | 6.199 ± 0.092 s | 5.906 ± 0.008 s | ✅ 4.7% |
| benchmarks/challenging/harris-3.c | 128.333 ± 11.240 s | 150.667 ± 13.429 s | ❌ 17.4% |

### Memory model: vmm

| Benchmark | Base branch | PR branch | Change |
| :--- | :--- | :--- | :--- |
| benchmarks/locks/cna.c | 8.353 ± 0.591 s | 8.977 ± 0.596 s | ❌ 7.5% |
| benchmarks/lfds/safe_stack.c | 8.117 ± 0.302 s | 6.197 ± 0.922 s | ✅ 23.7% |
| benchmarks/challenging/cna.c | 30.961 ± 0.965 s | 30.857 ± 1.985 s | ✅ 0.3% |
| benchmarks/challenging/harris-3.c | 5.372 ± 0.351 s | 5.521 ± 0.276 s | ❌ 2.8% |
| benchmarks/challenging/wsq.c | 133.667 ± 14.154 s | 145.000 ± 17.349 s | ❌ 8.5% |

### Memory model: aarch64

| Benchmark | Base branch | PR branch | Change |
| :--- | :--- | :--- | :--- |
| benchmarks/challenging/wsq.c | 10.177 ± 0.347 s | 11.175 ± 1.488 s | ❌ 9.8% |

*51 benchmark(s) omitted because both averages were below 5 seconds.*

### DIFF
--- a/.github/performance-benchmarks.yml
+++ b/.github/performance-benchmarks.yml
@@ -1,5 +1,5 @@
 # Each entry is one deliberately selected performance measurement. Paths are relative to the repository root.
-# Configurations may add command-line options, for example: options: ["--bound=2"].
+# Options on an entry apply to every configuration; configurations can add their own options.
 benchmarks:
   - program: benchmarks/locks/cna.c
     runs: 3
@@ -9,11 +9,60 @@ benchmarks:
       - cat: cat/vmm.cat
         target: c11
 
+  - program: benchmarks/locks/linuxrwlock.c
+    runs: 3
+    options: ["--bound=2"]
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+      - cat: cat/power.cat
+        target: power
+
+  - program: benchmarks/locks/mutex_musl.c
+    runs: 3
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+        options: ["--bound=5"]
+      - cat: cat/power.cat
+        target: power
+        options: ["--bound=4"]
+      - cat: cat/vmm.cat
+        target: c11
+        options: ["--bound=3"]
+
   - program: benchmarks/lfds/dglm.c
+    runs: 3
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+        options: ["--bound=3"]
+      - cat: cat/power.cat
+        target: power
+      - cat: cat/vmm.cat
+        target: c11
+        options: ["--bound=2"]
+
+  - program: benchmarks/lfds/ms.c
+    runs: 3
+    options: ["--bound=2"]
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+      - cat: cat/power.cat
+        target: power
+      - cat: cat/vmm.cat
+        target: c11
+
+  - program: benchmarks/lfds/treiber.c
     runs: 3
     configurations:
       - cat: cat/power.cat
         target: power
+        options: ["--bound=4"]
+      - cat: cat/vmm.cat
+        target: c11
+        options: ["--bound=3"]
 
   - program: benchmarks/lfds/safe_stack.c
     runs: 3
@@ -24,6 +73,9 @@ benchmarks:
   - program: benchmarks/challenging/cna.c
     runs: 3
     configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+        options: ["--bound=2"]
       - cat: cat/vmm.cat
         target: c11
 
@@ -32,5 +84,3 @@ benchmarks:
     configurations:
       - cat: cat/aarch64.cat
         target: arm8
-      - cat: cat/vmm.cat
-        target: c11
--- a/.github/performance-benchmarks.yml
+++ b/.github/performance-benchmarks.yml
@@ -1,37 +1,36 @@
-# Each entry selects matching source files below its directory. Paths are relative to the repository root.
-# A folder exclude skips files for every configuration; a configuration exclude skips files only for itself.
+# Each entry is one deliberately selected performance measurement. Paths are relative to the repository root.
+# Configurations may add command-line options, for example: options: ["--bound=2"].
 benchmarks:
-  - directory: benchmarks/locks
-    include: ["*.c"]
+  - program: benchmarks/locks/cna.c
     runs: 3
     configurations:
-      - cat: cat/aarch64.cat
-        target: arm8
       - cat: cat/power.cat
         target: power
       - cat: cat/vmm.cat
         target: c11
 
-  - directory: benchmarks/lfds
-    include: ["*.c"]
+  - program: benchmarks/lfds/dglm.c
     runs: 3
     configurations:
-      - cat: cat/aarch64.cat
-        target: arm8
       - cat: cat/power.cat
         target: power
-        exclude: ["safe_stack.c"]
+
+  - program: benchmarks/lfds/safe_stack.c
+    runs: 3
+    configurations:
       - cat: cat/vmm.cat
         target: c11
 
-  - directory: benchmarks/challenging
-    include: ["*.c"]
+  - program: benchmarks/challenging/cna.c
+    runs: 3
+    configurations:
+      - cat: cat/vmm.cat
+        target: c11
+
+  - program: benchmarks/challenging/wsq.c
     runs: 3
     configurations:
       - cat: cat/aarch64.cat
         target: arm8
-      - cat: cat/power.cat
-        target: power
-        exclude: ["cna.c", "wsq.c"]
       - cat: cat/vmm.cat
         target: c11
--- a/.github/performance-benchmarks.yml
+++ b/.github/performance-benchmarks.yml
@@ -4,8 +4,6 @@ benchmarks:
   - program: benchmarks/locks/cna.c
     runs: 3
     configurations:
-      - cat: cat/power.cat
-        target: power
       - cat: cat/vmm.cat
         target: c11
 
--- a/.github/performance-benchmarks.yml
+++ b/.github/performance-benchmarks.yml
@@ -1,0 +1,37 @@
+# Each entry selects matching source files below its directory. Paths are relative to the repository root.
+# A folder exclude skips files for every configuration; a configuration exclude skips files only for itself.
+benchmarks:
+  - directory: benchmarks/locks
+    include: ["*.c"]
+    runs: 3
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+      - cat: cat/power.cat
+        target: power
+      - cat: cat/vmm.cat
+        target: c11
+
+  - directory: benchmarks/lfds
+    include: ["*.c"]
+    runs: 3
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+      - cat: cat/power.cat
+        target: power
+        exclude: ["safe_stack.c"]
+      - cat: cat/vmm.cat
+        target: c11
+
+  - directory: benchmarks/challenging
+    include: ["*.c"]
+    runs: 3
+    configurations:
+      - cat: cat/aarch64.cat
+        target: arm8
+      - cat: cat/power.cat
+        target: power
+        exclude: ["cna.c", "wsq.c"]
+      - cat: cat/vmm.cat
+        target: c11
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -64,7 +64,7 @@ jobs:
             --base-checkout "$GITHUB_WORKSPACE/base" \
             --head-checkout "$GITHUB_WORKSPACE/head" \
             --timeout 180 \
-            --jobs 1 \
+            --jobs 3 \
             --min-average-seconds 5 \
             --markdown performance/report.md \
             --json performance/results.json
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -64,7 +64,7 @@ jobs:
             --base-checkout "$GITHUB_WORKSPACE/base" \
             --head-checkout "$GITHUB_WORKSPACE/head" \
             --timeout 300 \
-            --jobs 3 \
+            --jobs 1 \
             --min-average-seconds 5 \
             --markdown performance/report.md \
             --json performance/results.json
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -14,8 +14,18 @@ concurrency:
 
 jobs:
   compare:
-    name: Check for performance regressions
-    runs-on: ubuntu-latest
+    name: Check for performance regressions (${{ matrix.platform }})
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: Linux x64
+            runner: ubuntu-latest
+            artifact: linux-x64
+          - platform: macOS ARM64
+            runner: macos-14
+            artifact: macos-arm64
     steps:
       - name: Check out base revision
         uses: actions/checkout@v4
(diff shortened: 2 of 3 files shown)


ThomasHaas:
What bounds are used for the tests? Also, I don't think we should run `harris`. I don't think Dartagnan can do anything with that benchmark (I assume you used `--bound=1`?)

hernanponcedeleon:
The initial version of the code supported fine grained bound selection, but it was really hard to find specific bounds for each benchmark + cat combination what would not blow up the CI time. Also, it made the script and the yaml file more complex without any visible benefit. 

With the default bound, we still get 9 benchmarks taking at least more than 5 seconds (i.e., some threshold were hopefully time differences is not just noise), thus I went with this simpler solution. The table above shows that Harris runs to completion for all memory models (with `--bound=1`).

ThomasHaas:
I'm not even sure if `harris` with `--bound=1` can reach the assertion, The benchmark is so far beyond what Dartagnan can handle that it is quite pointless I think. Maybe we should also move `wsq.c` to the regular lfds tests (and maybe exclude it under power like safe_stack). There is also the question of whether you should run on the `benchmarks` folder at all or directly on the already compiled benchmarks in the `resources` directory (like the unit tests do).
I could even imagine having a dedicated folder for performance tests with more specific bound configurations. Running it on all benchmarks when 90%+ get filtered out due to low relevance is certainly not an optimal choice.

Btw. you check all properties right? That would at least explain why `wsq` is so much slower under `vmm.cat`, though it still surprises me quite a bit.

hernanponcedeleon:
> I'm not even sure if `harris` with `--bound=1` can reach the assertion, The benchmark is so far beyond what Dartagnan can handle that it is quite pointless I think. Maybe we should also move `wsq.c` to the regular lfds tests (and maybe exclude it under power like safe_stack). 

Moving `wsq` would still not allow us to completely get rid of `challenging` from the `yaml` file since we have `challenging/cna`. This even suggests that we might need a more precise definition of what `challenging` means.

> There is also the question of whether you should run on the `benchmarks` folder at all or directly on the already compiled benchmarks in the `resources` directory (like the unit tests do).

I used the c files on purpose since this gives us more testing that the CLI works as expected. The times used in the table are those we report internally and thus compilation time is not counted.

> I could even imagine having a dedicated folder for performance tests with more specific bound configurations. 

What would be the point of having other folder with repeated files? We could keep the files where they are and add the fine-grained configurations, but this is quite a lot of work (not only initially, but also for maintainance) wihtout any clear advantage AFAICS.

> Running it on all benchmarks when 90%+ get filtered out due to low relevance is certainly not an optimal choice.

Most of the filtered benchmarsk run in `0.0X` (we use native) so I prefer to waste some resources but keep the configuration simple. I can make the script a bit smarter and only run a benchmark once (instead of 5 times) if the first run is below the filtering timeout.

> 
> Btw. you check all properties right? That would at least explain why `wsq` is so much slower under `vmm.cat`, though it still surprises me quite a bit.

Yes. I chose the memory models such that we have an easy one (aarch64), one with recursion (power), and one with a definition of races (vmm).

ThomasHaas:
> > I could even imagine having a dedicated folder for performance tests with more specific bound configurations.
> 
> What would be the point of having other folder with repeated files? We could keep the files where they are and add the fine-grained configurations, but this is quite a lot of work (not only initially, but also for maintainance) wihtout any clear advantage AFAICS.

It wouldn't need to contain the literal benchmarks but a file that points to the performance benchmarks with possible parameters like bounds. Given the list you show above, and excluding harris, there are currently only 7 tests (and less actual programs) all of which are cases with UNKNOWN/FAIL result (performance on PASS is not at all represented). Maintenance-wise, it doesn't look at all intimidating to me. I mean, it doesn't look any more complex than adding a unit test (or 7 for that matter). Together with the fact that we extremely rarely add new benchmarks (especially with >5s runtime for B=1), I think having more control over the (few) tests we have is more valuable than automatically picking up any random test we add.
I could totally see the value of automation if the performance tests would be related to our unit tests though (e.g., same tests with same configs).

That being said, the current state of the PR is already useful so I'm not objected to merging, I just think that it could be even more useful. The only "objective flaw" is that PASS benchmarks are not reflected I think.

hernanponcedeleon:
I did as you suggested and added a fine-grained configuration. I also used a confidense interval rather than a direct comparison between averages to account for the std deviation between runs and filter out noise.

I al still trying to bring down the CI time. Hopefully using parallel jobs will work.

ThomasHaas:
I like the changes. I think having a total sum over all times would also be nice.

Btw. have you tried running development against development? I think this would show how noisy Dartagnan is and how much noise the performance script actually compensates for.

hernanponcedeleon:
> Btw. have you tried running development against development? I think this would show how noisy Dartagnan is and how much noise the performance script actually compensates for.

I haven't explicitly, but this PR does not touch the code, so the results here give you the same data.

ThomasHaas:
> > Btw. have you tried running development against development? I think this would show how noisy Dartagnan is and how much noise the performance script actually compensates for.
> 
> I haven't explicitly, but this PR does not touch the code, so the results here give you the same data.

Oh, I was still under the impression that you compared against #1060 like in the initial table. So you are saying the newest table is effectively `development` vs. `development` and despite having done 3 runs per benchmark we still have cases with over 20% of difference? Hmm, that sucks a lot.

hernanponcedeleon:
Being precise, it is `development` vs `performance-ci`, but both generate the same binary. 

As the table shows, the is a lot of std deviation between the runs where the difference is large. This is why I added a more complex formula in the last column. This shows that despite de difference, there is no statistical difference.

ThomasHaas:
I guess you consider any interval containing `0` as showing no statistical difference? Then it makes sense. But seeing huge intervals like `-26.7% [-117.5%, +64.1%]` and heavily lobsided intervals like `➖ +19.1% [-4.7%, +43.0%]` still make me quite concerned. I feel like a hypothetical optimization that is strictly 10% faster everywhere will completely get caught up in the noise. If so, and if we only trust the confidence intervals, I'm wondering if we ever have find any "significant improvment".

ThomasHaas:
I see the updated table... `❌ -23.4% [-42.2%, -4.5%]` says that with a high confidence we have lost performance, right?
So the data is pretty much useless, no? Btw. I don't think I see such large deviations on my machine. Is the CI just that inconsistent?

hernanponcedeleon:
> I see the updated table... `❌ -23.4% [-42.2%, -4.5%]` says that with a high confidence we have lost performance, right? 

Yes, this is bad. Maybe it was not a good idea to run 3 jobs in parallel since this might be influencing. I will rollback to run all test sequenatilly, and hopefully results will looks better. If this is the case, I can split this into two jobs, once per branch to bring down the CI times again.

hernanponcedeleon:
Numbers look better without parallelism. Only two benchmarks have 2 digits diff, and all are marked as no statistical difference.

ThomasHaas:
Yeah, it looks a lot better

hernanponcedeleon:
It seems splitting the runs in two different VMs introduced systematic machine bias. I think I will go back to the version that runs everything on the same machine and we will have to leave with the fact that the CI will take a bit longer.

hernanponcedeleon:
I made a local run on my machine and I also get noise between runs of the same benchmarks with the same binary. I am not too sure there is much we can do about it (other than removing the few non-determinism that remains in the encoding generation after #1059; I think we should be able to fix the remaining non-determinism after #1075).

## Performance comparison

### Memory model: vmm

| Benchmark | Base branch | PR branch | Improvement (99% CI) | Result |
| :--- | :--- | :--- | :--- | :--- |
| `benchmarks/locks/cna.c` | 6.749 ± 0.367 s | 6.728 ± 0.308 s | ➖ -0.0% [-55.7%, +55.6%] | UNKNOWN |
| `benchmarks/locks/mutex_musl.c` | 14.727 ± 1.306 s | 13.977 ± 0.079 s | ➖ +4.6% [-42.8%, +52.0%] | UNKNOWN |
| `benchmarks/lfds/dglm.c` | 17.592 ± 0.197 s | 16.766 ± 0.292 s | ✅ +4.7% [+1.1%, +8.3%] | UNKNOWN |
| `benchmarks/lfds/ms.c` | 25.566 ± 0.541 s | 24.446 ± 0.315 s | ➖ +4.3% [-13.1%, +21.8%] | UNKNOWN |
| `benchmarks/lfds/treiber.c` | 7.201 ± 0.210 s | 7.193 ± 0.053 s | ➖ +0.1% [-17.1%, +17.2%] | UNKNOWN |
| `benchmarks/lfds/safe_stack.c` | 6.047 ± 0.131 s | 5.606 ± 0.202 s | ➖ +7.3% [-0.2%, +14.9%] | UNKNOWN |
| `benchmarks/challenging/cna.c` | 21.489 ± 0.407 s | 21.621 ± 0.466 s | ➖ -0.6% [-8.1%, +6.9%] | UNKNOWN |

### Memory model: aarch64

| Benchmark | Base branch | PR branch | Improvement (99% CI) | Result |
| :--- | :--- | :--- | :--- | :--- |
| `benchmarks/locks/linuxrwlock.c` | 5.939 ± 0.068 s | 5.759 ± 0.199 s | ➖ +3.0% [-12.8%, +18.9%] | UNKNOWN |
| `benchmarks/lfds/dglm.c` | 5.115 ± 0.081 s | 5.446 ± 0.040 s | ❌ -6.5% [-11.7%, -1.2%] | PASS |
| `benchmarks/lfds/ms.c` | 6.338 ± 0.058 s | 6.560 ± 0.048 s | ❌ -3.5% [-6.5%, -0.5%] | UNKNOWN |
| `benchmarks/challenging/cna.c` | 13.278 ± 1.562 s | 12.804 ± 0.153 s | ➖ +2.7% [-63.1%, +68.5%] | UNKNOWN |
| `benchmarks/challenging/wsq.c` | 7.627 ± 0.616 s | 7.791 ± 0.744 s | ➖ -2.3% [-49.6%, +45.0%] | UNKNOWN |

### Memory model: power

| Benchmark | Base branch | PR branch | Improvement (99% CI) | Result |
| :--- | :--- | :--- | :--- | :--- |
| `benchmarks/locks/linuxrwlock.c` | 13.880 ± 1.335 s | 14.000 ± 0.445 s | ➖ -1.3% [-43.4%, +40.7%] | UNKNOWN |
| `benchmarks/locks/mutex_musl.c` | 13.778 ± 0.163 s | 14.053 ± 0.150 s | ➖ -2.0% [-12.6%, +8.6%] | UNKNOWN |
| `benchmarks/lfds/ms.c` | 24.231 ± 1.155 s | 20.401 ± 1.015 s | ➖ +15.8% [-5.3%, +36.8%] | UNKNOWN |
| `benchmarks/lfds/treiber.c` | 8.904 ± 0.137 s | 11.022 ± 0.397 s | ❌ -23.8% [-38.3%, -9.3%] | UNKNOWN |

### Total

| Benchmarks | Base branch | PR branch | Improvement (99% CI) |
| :--- | :--- | :--- | :--- |
| All reported benchmarks | 198.462 ± 3.121 s | 194.174 ± 0.883 s | ➖ +2.1% [-5.9%, +10.2%] |

2 benchmarks were omitted because both averages were below 5 seconds.

To check if there is any use for the current script, I ask it to compare `development` with a dummy branch that changes the alias analysis to a more imprecise one (`FIELD_INSENSITIVE`). There, it is a bit more clear that we have a regression and it is not just noise.

## Performance comparison

### Memory model: vmm

| Benchmark | Base branch | PR branch | Improvement (99% CI) | Result |
| :--- | :--- | :--- | :--- | :--- |
| `benchmarks/locks/cna.c` | 6.095 ± 0.411 s | 6.921 ± 0.640 s | ➖ -13.5% [-109.3%, +82.2%] | UNKNOWN |
| `benchmarks/locks/mutex_musl.c` | 12.330 ± 0.310 s | 13.986 ± 0.561 s | ➖ -13.4% [-67.3%, +40.4%] | UNKNOWN |
| `benchmarks/lfds/dglm.c` | 17.152 ± 0.465 s | 16.964 ± 0.167 s | ➖ +1.1% [-15.6%, +17.8%] | UNKNOWN |
| `benchmarks/lfds/ms.c` | 24.032 ± 0.608 s | 25.149 ± 0.301 s | ➖ -4.6% [-22.1%, +12.8%] | UNKNOWN |
| `benchmarks/lfds/treiber.c` | 6.763 ± 0.080 s | 7.114 ± 0.345 s | ➖ -5.2% [-28.6%, +18.3%] | UNKNOWN |
| `benchmarks/lfds/safe_stack.c` | 5.652 ± 0.189 s | 6.022 ± 0.387 s | ➖ -6.6% [-33.5%, +20.4%] | UNKNOWN |
| `benchmarks/challenging/cna.c` | 22.673 ± 0.167 s | 22.863 ± 0.319 s | ➖ -0.8% [-8.9%, +7.2%] | UNKNOWN |

### Memory model: aarch64

| Benchmark | Base branch | PR branch | Improvement (99% CI) | Result |
| :--- | :--- | :--- | :--- | :--- |
| `benchmarks/locks/linuxrwlock.c` | 5.536 ± 0.089 s | 6.652 ± 0.123 s | ❌ -20.2% [-30.7%, -9.6%] | UNKNOWN |
| `benchmarks/lfds/dglm.c` | 5.507 ± 0.048 s | 6.119 ± 0.161 s | ❌ -11.1% [-23.7%, -1.5%] | PASS |
| `benchmarks/lfds/ms.c` | 6.061 ± 0.017 s | 6.978 ± 0.155 s | ❌ -15.1% [-24.5%, -5.8%] | UNKNOWN |
| `benchmarks/challenging/cna.c` | 12.619 ± 0.377 s | 14.111 ± 0.549 s | ❌ -11.8% [-39.5%, -2.1%] | UNKNOWN |
| `benchmarks/challenging/wsq.c` | 7.622 ± 0.418 s | 8.154 ± 0.787 s | ➖ -7.0% [-70.8%, +56.9%] | UNKNOWN |

### Memory model: power

| Benchmark | Base branch | PR branch | Improvement (99% CI) | Result |
| :--- | :--- | :--- | :--- | :--- |
| `benchmarks/locks/linuxrwlock.c` | 13.665 ± 0.125 s | 14.798 ± 0.134 s | ❌ -8.3% [-14.2%, -2.4%] | UNKNOWN |
| `benchmarks/locks/mutex_musl.c` | 13.864 ± 0.201 s | 14.923 ± 0.521 s | ➖ -7.6% [-43.7%, +28.4%] | UNKNOWN |
| `benchmarks/lfds/ms.c` | 21.883 ± 0.515 s | 22.132 ± 0.384 s | ➖ -1.1% [-20.2%, +18.0%] | UNKNOWN |
| `benchmarks/lfds/treiber.c` | 8.351 ± 0.041 s | 11.088 ± 0.130 s | ❌ -32.8% [-38.2%, -27.4%] | UNKNOWN |

### Total

| Benchmarks | Base branch | PR branch | Improvement (99% CI) |
| :--- | :--- | :--- | :--- |
| All reported benchmarks | 189.815 ± 2.109 s | 207.974 ± 2.377 s | ❌ -9.6% [-19.4%, -0.8%] |

2 benchmarks were omitted because both averages were below 5 seconds.

ThomasHaas:
I don't think there is any real non-determinism left. The dumped encoding seems to be always identical now and it should closely reflect the order in which we generate the encoding internally. That being said, I only checked the stability of the encoding on a couple of benchmarks under a few memory models. Maybe recursive models like Power are still non-deterministic?

You could try setting Z3's random seed explicitly, but I think it is already fixed to 0 by default

hernanponcedeleon:
Results look MUCH stable now
- latest run is [this](https://github.com/hernanponcedeleon/Dat3M/pull/1083#issuecomment-5555890721)
- data from the previous run is available [here](https://github.com/hernanponcedeleon/Dat3M/actions/runs/34160686872/artifacts/10033295520)

ThomasHaas:
Is it just due to the switch from Z3 to Yices2?

hernanponcedeleon:
TBH, I am not sure. Another potential factor is at which time of the fsy the job is run and how busy GitHub runners are. I would not be surprised if they are busier during US working hours. I'll rerun the job later today to test this hypothesis.

I'll also do a test commit changing the alias analysis to something more imprecise to test if we would catch that regression. Is there any other option we can use which we expect to reduce performance?

hernanponcedeleon:
> Is it just due to the switch from Z3 to Yices2?

I'll also change the job to additionally run on a macos/windows machine (which uses z3 as default) so we get data about both solvers.

ThomasHaas:
> I'll also do a test commit changing the alias analysis to something more imprecise to test if we would catch that regression. Is there any other option we can use which we expect to reduce performance?

- `encoding.wmm.idl2sat=true`
- `encoding.wmm.reduceAcyclicityRelevantSets=false` (this one should impact `wsq.c` heavily)
- `wmm.analysis.extendedRelationAnalysis=false`
- Maybe `program.processing.spinloops=false` which disables dynamic spinloop detection (the name of the option is odd tbh).
- Maybe `program.processing.constantPropagation=false` to disable `SCCP`?

hernanponcedeleon:
Two analysis tests [fail](https://github.com/hernanponcedeleon/Dat3M/actions/runs/34234133159/job/102090051770) with the field insisitive alias analysis. @xeren are those tests written under the assumption we will use `LINEAR_MD_FIELDS`? If this is the case, the test should specify which alias analsysi to use and not rely on the default one.

xeren:
Yes, both tests should specify either `LINEAR_FIELDS` or `LINEAR_MD_FIELDS`.

hernanponcedeleon:
The current status is as follows
- The table shows (almost) no red X / green marks when comparing the same binary. There have been at least 4 such tables since yesterday and I think only once I saw a green tick on an individual benchmark, but the bottom table (sum of times) showed no statistical diff (and in many cases, mostly in Linux, pretty similar times).
- Large regressions (like the imprecise AA) are very visible.
- "Small" regressions (like the one when XRA is set to false) results in few red crosses, but we see consistent 10-12% slow-down even if the confidence interval contains 0 (and thus the few red marks).
- I would expect "small" (< 20%) improvements similarly not to result on green marks, but hopefully a consistent sum improvement at the bottom of the table.

ThomasHaas (approved):
LGTM

PR #1085, opened 2026-09-07
Make yices2 the default solver for Linux

hernanponcedeleon:


ThomasHaas:
FYI, this will not change the solver we use in unit tests and hence the CI.
But I guess it will change the solver used by #1083, and I assume that is what you are going for?

hernanponcedeleon:
I know unit tests define their own solver (or at least most unit tests do).

The main use of this PR is not #1083 but rather the fact fact that Linux users were using z3 by default which is much slower than yices in general.

ThomasHaas (approved):
LGTM

PR #1086, opened 2026-09-08
Add EnumerationTask and related classes

ThomasHaas:
- `EnumerationTask` represents the task of enumerating final states. Currently, limited to litmus code.
- There is a new option `--mode={verification, enumeration}` to select the type of task to solve.

Possible TODOs: 
- Add mode option to UI
- Maybe merge `EnumerationSolver` with `AssumeSolver`: they share a lot of code.
- Maybe some package refactoring.
- Unit tests?
- Make enumeration bound configurable (currently fixed at 10000 states)
- Add bounds checks to report incomplete enumeration
- Rename `--mode=verification/enumeration` to the shorter `--mode=verify/enumerate`?


hernanponcedeleon:
> * `EnumerationTask` represents the task of enumerating final states. Currently, limited to litmus code.
> * There is a new option `--mode={verification, enumeration}` to select the type of task to solve.

I haven't test it, but seeing a new option and no changes to the relection file, it might be the case the options does not work in native. Also, if the option is in a new class, it will not be shown in the help. I will add (in another PR) some test so make sure we do not miss this situation anymore.

> 
> Possible TODOs:
> 
> * Unit tests?

Done.

Many of the tests will fail, but this does not necessarily mean the exploraiton is wrong. I generated the expected results using the cat files in herd's repo, thus there might be a missmatch with our cat models. But if this is the case, we should figure out how and why the models differ.

ThomasHaas:
From local testing, we still have 41 test failures. One of them times out due to OOTA.
Most of them have a spec and still list `locations` which causes a mismatch where Dartagnan enumerates less than intended. However, for the C11 litmus test `../litmus/C11/manual/imm-R2-alt.litmus`, we enumerate `13` instead of `12` states. This one might be the only genuine discrepancy between Dartagnan and herd7.

ThomasHaas:
I extracted a litmus test on which herd7 and Dartagnan disagree under `c11.cat`:
```
C imm-R2

(*
 * Bridging the Gap between Programming Languages and Hardware Weak Memory Models
 *)

{}

P0 (atomic_int* x, atomic_int* y) {
    atomic_store_explicit(y, 1, memory_order_relaxed);
}

P1 (atomic_int* x) {
    atomic_thread_fence(memory_order_release);
    int r0 = atomic_fetch_add_explicit(x, 1, memory_order_acquire);
    atomic_store_explicit(x, 3, memory_order_relaxed);
}

P2 (atomic_int* x, atomic_int* y) {
    int r0 = atomic_load_explicit(x, memory_order_acquire);
    int r1 = atomic_load_explicit(y, memory_order_relaxed);
}

exists (1:r0=1 /\ 2:r0=2 /\ 2:r1=0) // Reachable in Dartagnan, unreachable in Herd7
```

It looks like a release-acquire chain interrupted by an RMW. Looking at the `c11.cat` memory model, it does not seem to have a construct like `(rf;rmw)*` to extend release-acquire chains like other models do.

xeren:
Dartagnan's `c11.cat` does feature a term `R & W`, which might rather be `RMW`.  The above litmus test fails with this change.

ThomasHaas:
That's a good point. `R & W` is clearly wrong in our model.

EDIT: Replacing `R & W` with `RMW` fixed the problem without breaking anything else in the litmus tests.

ThomasHaas:
@hernanponcedeleon Except for the one timeout due to OOTA, I think all remaining failing tests are due to the explicitly mentioned `locations` inside the tests that cause herd7 to enumerate more.
I think we have the following options:
- Remove the `locations` bit from those tests and let herd7 generate new expected values
- Skip those tests only for enumeration but not for verification (requires a separate skip list)
- Add support for `locations` in litmus tests (probably the hardest solution to implement).

ThomasHaas:
The verdict on `linuxrwlock` under `c11.cat` has changed from `FAIL` to `UNKNOWN` which looks reasonable to me.

hernanponcedeleon:
> * Add support for `locations` in litmus tests (probably the hardest solution to implement).

Implementing this should not be hard, right? We just need to parse the locations/registers in this list, store them in the program as some "metadata", and compute the notion of state not only based on the variables of the spec, but rather the union of those with the ones computed here.

> The verdict on linuxrwlock under c11.cat has changed from FAIL to UNKNOWN which looks reasonable to me.

can you explain me why the "bug" is gone?

ThomasHaas:
> > * Add support for `locations` in litmus tests (probably the hardest solution to implement).
> 
> Implementing this should not be hard, right? We just need to parse the locations/registers in this list, store them in the program as some "metadata", and compute the notion of state not only based on the variables of the spec, but rather the union of those with the ones computed here.

Just as "metadata" is a bit problematic, but yeah, we need to store the list somewhere similar to how we store the spec.
However, we need to be careful that existing passes are updated, e.g., `Tearing` may also need to tear the stored expressions.
 
> > The verdict on linuxrwlock under c11.cat has changed from FAIL to UNKNOWN which looks reasonable to me.
> 
> can you explain me why the "bug" is gone?

The `c11.cat` model had a place where it used `R & W` to refer to RMWs in the context of release-sequences. Clearly, that relation is always empty in our modeling, which broke release sequences. I assume the `FADD` chains that could form in `linuxrwlock` were affected by that. Replacing `R & W` by `RMW` restores the release sequences as intended.

PR #1090, opened 2026-09-09
Add feature tags to memObj copy

tonghaining:
Fix for OpenCL MemoryTransformer: memObj copy should also carry Global/Local feature tags.

The local_barrier_flag test expectation is updated to be consistent with its annotation since avvis is not relevant to OpenCL.


hernanponcedeleon:
Sorry @tonghaining for all the dummy commits ... your PR detected a problem on a CI job we recently added and it took me some time to find a problem solution.

I will try to review ASAP.

hernanponcedeleon:
I think the changes can add more than just GLOBAL/LOCAL (e.g., also `C11.NON_ATOMIC_LOCATION`). If so, the title of the PR needs to be updated (something about fetaure tags should work).

Other than that, I am fine with the changes. @natgavrilenko let me know if you want to take a look, otherwise I will merge once the title is updated.

natgavrilenko (approved):

PR #1091, opened 2026-09-10
Fix performance comparison job to be also run on PRs from forks

hernanponcedeleon:
This PR addresses a limitation identified by #1090 (coming from a fork): performance benchmarks ran successfully, but the report comment was skipped because fork workflows cannot receive write permissions.

The performance comparison and report publication are now separate workflows. The unprivileged workflow runs the benchmarks and uploads the results, while a trusted workflow publishes the report to the originating PR.

### DIFF
--- a/.github/workflows/performance-comparison.yml
+++ b/.github/workflows/performance-comparison.yml
@@ -1,4 +1,4 @@
-name: Performance regression
+name: Generate performance report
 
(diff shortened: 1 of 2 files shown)


ThomasHaas (approved):
I blindly trust the changes :)

PR #1094, opened 2026-09-10
Fix performance comparison job to be also run on PRs from forks

hernanponcedeleon:
This is exaclty as #1091. I removed that one from the history because I thought it was wrong (i.e., no table was generated), but apparently those changes will only work once they land in the master branch.

We shouuld merge this one to development, then I will merge the whole development into master and then we can rebase all opened PRs.

### DIFF
--- a/.github/workflows/performance-comparison.yml
+++ b/.github/workflows/performance-comparison.yml
@@ -1,4 +1,4 @@
-name: Performance regression
+name: Generate performance report
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


ThomasHaas:
You want to merge development fully into master? That's basically the release of a new Dartagnan version, no?

EDIT: Also why does this require to be in `master` but the previous solution did not?

ThomasHaas:
Btw. none of the our active PRs touches performance aspects, so I don't think that we must urgently merge this PR and development into master.

hernanponcedeleon:
> You want to merge development fully into master? That's basically the release of a new Dartagnan version, no?

I do not need to make a release to merge to `master`. Normally I avoid merging to `master` too often to give us time to find regressions we merged into `development` before they can land in the latest version of the code people might use, but this is not a strong requirement. Many times in the past I merged to `main` (without a release) for the same reason that some CI configurations need to be in `master` to become effective.

> EDIT: Also why does this require to be in `master` but the previous solution did not?

Because I need a `workflow_run` to overcome the permission problem of PRs from forks and `workflow_run` only triggers when the yaml file is in the default branch of the repo (`master` in our case).

> Btw. none of the our active PRs touches performance aspects, so I don't think that we must urgently merge this PR and development into master.

No, but I do not see any reason why to delay it ... since having it in place will lets us know if the solution works.

ThomasHaas (approved):

PR #1095, opened 2026-09-10
Fix transitivity of ExecutionAnalysis.isImplied

xeren:
Fixes an imprecision in `ExecutionAnalysis` introduced in #1044:  The cached `implyingEvents` were not closed transitively and thus **the sets were substantially smaller than expected**.  Precisely, if `e2` was implied by a `ThreadCreate` or `ThreadStart` `e1`, then `implyingEvents(e2).addAll(implyingEvents(e1))` was missing.  There were `e0` with `isImplied(e0, e1) && isImplied(e1, e2) && ! isImplied(e0, e2)`.

This pull has the following concrete effects:
- Events now imply `Init` events again (see the new test `AnalysisTest.impliedInits()`)
- Events in a spawned thread can now imply many events in the creator thread again.

As a consequence, the precision of `RelationAnalysis` and `ActiveSetAnalysis` should increase, as well.  I expect a small positive effect on the eager solver's performance.

Thanks to @JnsGlnr for pointing this out.

### DIFF
--- a/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
+++ b/dartagnan/src/main/java/com/dat3m/dartagnan/program/analysis/ExecutionAnalysis.java
@@ -5,6 +5,7 @@ import com.dat3m.dartagnan.program.IRHelper;
 import com.dat3m.dartagnan.program.Program;
 import com.dat3m.dartagnan.program.Thread;
 import com.dat3m.dartagnan.program.event.Event;
+import com.dat3m.dartagnan.program.event.core.threading.ThreadCreate;
 import com.dat3m.dartagnan.program.event.core.threading.ThreadStart;
 import com.dat3m.dartagnan.utils.collections.IndexedDomain;
 import com.dat3m.dartagnan.utils.collections.IndexedSet;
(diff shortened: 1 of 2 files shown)


ThomasHaas:
The reason you recompute the implied events the way you do was for performance reasons, right? But the recomputation did not capture every implication that `BranchEquivalence` provided, correct?

I'm wondering how your changes relate to the TODO in `BranchEquivalence`:
```
    TODO: (5) Transitively close the merged implication graph,
              and saturate the mutual exclusion relation using the rule
                   A => B, B ~ C, C <= D   ---->   A ~ D  (where ~ denotes mutual exclusion)
```
Do you now also capture transitive implications that were missing in `BranchEquivalence`?

xeren:
> Do you now also capture transitive implications that were missing in `BranchEquivalence`?

No, the transitivity in EA still requires the transitivity in BE.  If it fails for the program below, i.e. `! be.isImplied(worker2.entry, worker1.entry.creator)`, then EA also misses that pair.
```c
void worker2() {
}
void worker1() {
    if (a)
        thread_create(worker2);
}
void main() {
    if (b)
        thread_create(worker1);
}
```

ThomasHaas:
We certainly get smaller active sets now and the CI shows no issues.

ThomasHaas (approved):
LGTM

PR #1097, opened 2026-09-11
[SPIRV] Track id to variable mapping to be able to use original variable name…

hernanponcedeleon:
This PR adds mappings between SPIRV ids and original variables names. This allows to write the program header (inputs/outputs/specification) in terms of sourcecode variables. One can still used the SPIRV id so there should not be regressions. 

I updated the headers in `benchmarks/opencl/benchmarks` and `benchmarks/opencl/patterns` so now those can be passed directly as input to dartagnan, e.g.,
```
dartagnan cat/vulkan.cat --target=vulkan benchmarks/opencl/benchmarks/caslock.cl           
Test: benchmarks/opencl/benchmarks/caslock.cl
Result: PASS
Time: 0.032 secs
```

Notice I removed many of the `// Currently, clspv inserts Coherent decorations only if ...` since this has been fixed in https://github.com/google/clspv/pull/1651.


ThomasHaas:
Oh, you found another usage for Metadata. Now I feel like we should create a top-level `metadata` package with all metadata classes, seeing that with this PR and #1104 we already have 3 different metadata packages that are quite deeply nested.
Alternatively, we could also put `Metadata` as inner classes of the class they should be associated with, e.g., `MemoryObject.VariableName`, `Program.SourceLocation`, and `Wmm.CutAnnotation`.

natgavrilenko (commented):
Your approach should work, but I think it creates a lot of tangled code that will be painful to maintain in the long term. Specifically, I don't like that VisitorSpirvVariableNames goes everywhere it wants, even though the same ops are also processed by the other visitors.

It might make sense to invest a bit more time into a cleaner implementation. For example:
- Instead of VisitorSpirvInput (and VisitorSpirvVariableNames), start directly with accepting svpInstructions
- In this case, we will not have initial values and concrete size for runtime arrays
- After parsing the whole shader and collecting the necessary information (including bindings, descriptor sets, opencl reflection etc), parse inputs and outputs
- Before building the threads, add a new pass that sets initial inputs values and converts "abstract" types such as RuntimeArray to concrete types such as an array with a fixed number of elements

PR #1098, opened 2026-09-12
Generalize metadata

ThomasHaas:
We used Metadata only for Events so far. I want to generalize Metadata also to other classes, in particular, relations and constraints of the memory model.
One application could be to mark relations that should be shown in a witness, for example, when the CAT file contains `show r` statements (I think we ignore these statements right now). In principle, we can also generate source location data for relations and constraints (which is probably not that useful though).
Another application which I have in mind is the ability to mark relations/constraints for cutting. In particular, we could lift the baseline/bias option for refinement to a simple `WmmProcessor` that adds constraints and marks them for cutting. Technically, we could use a bias like NO_OOTA even for eager solving.

EDIT: Btw. I moved `SourceLocation` to a more general package because it is usable beyond events. But since it is quite unlikely that we will use it for anything else, I can also move it back to reduce the number of changes slightly.


hernanponcedeleon:
> Another application which I have in mind is the ability to mark relations/constraints for cutting. In particular, we could lift the baseline/bias option for refinement to a simple `WmmProcessor` that adds constraints and marks them for cutting.

You mean, using something like `cut hb` directly on the cat file?

> Technically, we could use a bias like NO_OOTA even for eager solving.

I do not get this. Isn't cutting basically "eagerly encoding something"?

ThomasHaas:
> > Another application which I have in mind is the ability to mark relations/constraints for cutting. In particular, we could lift the baseline/bias option for refinement to a simple `WmmProcessor` that adds constraints and marks them for cutting.
> 
> You mean, using something like `cut hb` directly on the cat file?

Yes, but not only that. See below.

> > Technically, we could use a bias like NO_OOTA even for eager solving.
> 
> I do not get this. Isn't cutting basically "eagerly encoding something"?

I mean that you can specify a bias even for eager solving (no refinement): it simply adds another constraint to the memory model. Practically this might not be interesting (maybe NO_OOTA could be useful in some edge cases), but technically there is no reason that we can only do it for refinement except for that fact that those added constraints should also get cut. Because of that, refinement has to add the biases internally so that it can remember them for cutting.
If we can mark arbitrary relations/constraints for cutting, we can move all of this into a dedicated pass.

ThomasHaas:
Btw. I'm wondering if I should make the `MetadataCarrier` interface simply have a `MetadataMap getMetadataMap()` method and then default-implement all the other interface methods with that one. The current implementation uses an extra base class to provide the implementation without exposing the underlying `MetadataMap`. I'm not sure if it is worth the extra effort just to hide the underlying map...

hernanponcedeleon:
> I mean that you can specify a bias even for eager solving (no refinement): it simply adds another constraint to the memory model. Practically this might not be interesting (maybe NO_OOTA could be useful in some edge cases), but technically there is no reason that we can only do it for refinement except for that fact that those added constraints should also get cut. 

I still do not see the point of this. Wouldn't you get exaclty the same behavior if you just add the constraint (axiom) direcly on the cat file?

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
```diff
@@ -14,7 +15,7 @@
  * Given an execution, describes a set of events or a binary relation between events,
  * constrained by the associated {@link Definition}.
  */
-public final class Relation implements Dependent<Relation> {
+public final class Relation extends MetadataCarrierBase<Relation> implements Dependent<Relation> {
```
hernanponcedeleon, line 18: We do have passes that transform the memory model and its relations (e.g., `MergeEquivalentRelations`). Metadata should probably be conserved during these passes.

dartagnan/src/main/java/com/dat3m/dartagnan/utils/metadata/MetadataCarrierBase.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public boolean hasMetadata(Class<? extends Metadata> metadataClass) { return metadataMap.contains(metadataClass); }
+    @Override
+    public <TMeta extends Metadata> TMeta getMetadata(Class<TMeta> metadataClass) { return metadataMap.get(metadataClass); }
+    @Override
+    public <TMeta extends Metadata> TMeta setMetadata(TMeta metadata) { return metadataMap.put(metadata); }
+
+    @Override
+    public void copyAllMetadataFrom(T other) {
+        ((MetadataCarrierBase<?>) other).metadataMap.getAllMetadata().forEach(this.metadataMap::put);
```
hernanponcedeleon: Shouldn't this be `MetadataCarrier`, otherwise I think the cast might not be safe in general (even if might be in the current implementation).

ThomasHaas: `MetadataCarrier` does not give access to the `metadataMap`. The assumption is that you only copy metadata between same types (e.g. from relation to relation) so this will work. But I will put an `instanceof` check and throw an error in the case there is a mismatch.

ThomasHaas:
> I still do not see the point of this. Wouldn't you get exaclty the same behavior if you just add the constraint (axiom) direcly on the cat file?

The point is refactoring/modularization of the code that adds biases into a processing pass. If we have a way to specify cuts inside the CAT file, the whole pass can be mimicked by the user. Are you saying that we then should get rid of the pass entirely? That is totally fine with me.

hernanponcedeleon:
I see the value of being hable to replace the currently hardcoded bias computation by letting the user do this directly in the cat file. 

What I do no see is how would this helps the eager encoding since whatever that would mark as a bias is already being eagerly encoded anyway.

ThomasHaas:
Forget about eager, I shouldn't have mentioned it. My point is that the whole bias option is independent from refinement solving (modulo the cutting aspect of it) and that there are biases that do not already exist in the eager encoding (e.g. NO_OOTA in C11/LKMM). I'm not saying that it is practically useful to use biases with eager solving.

hernanponcedeleon:
Ok, that clarifies a bit. Modulo the 2 open comments above, the rest LGTM.

dartagnan/src/main/java/com/dat3m/dartagnan/wmm/Relation.java
ThomasHaas: I made `MergeEquivalentRelations` merge the metadata now but have not touched other passes. Generally though, it is not so clear what needs to be preserved and how. For example, if you have two equivalent definitions with different source locations, how do you merge their source locations? The cleanest solution would probably to preserve both source locations, though we currently cannot do it.

I think we need to change metadata merging and preservation on a case-by-case basis depending on the type of metadata. As of right now, there is no metadata usable for Wmm's, so it is hard to tell what to do exactly. 

Btw. I also moved `SourceLocation` back to the event package for now to reduce changes. We can move it to a more general location if we ever really need it for other things.

hernanponcedeleon: > I think we need to change metadata merging and preservation depending on a case-by-case basis depending on the type of metadata. As of right now, there is no metadata usable for Wmm's, so it is hard to tell what to do exactly.

Can you add a TODO or comment somewhere so we keep track of this potential problem in the future if we add more Metadata?

hernanponcedeleon (approved):

PR #1100, opened 2026-09-12
[SPIRV] Add support for OpCopyMemory

hernanponcedeleon:
Since `OpCopyMemory` makes use of memory operands, it plays a role from the point of view of the memory mode. Intuitively, I think it should not be any different than a program that explicity "chunks" the whole copy as a sequence of loads/stores as the implementation is doing, but TBH I have not thought very hard about possible problems from the memory model side (I mostly focused on validating the syntax restictions). @natgavrilenko let me know if you see any memory model problem.


dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -121,28 +164,40 @@ private List<Event> visitMemoryAccess(String id, Type type, Expression pointer,
...
+            return type;
+        }
+        if (type instanceof AggregateType aggregateType) {
+            return aggregateType.getFields().get(index).type();
         }
-        return Set.of();
+        if (type instanceof ArrayType arrayType) {
+            return arrayType.getElementType();
+        }
+        throw new IllegalArgumentException("Cannot index memory access type %s".formatted(type));
```
ThomasHaas, line 177: We have `ExpressionHelper.extractType` which does the same.
Maybe we should move it to `TypeFactory` though.

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,231 @@ public void testStoreWithIllegalTags() {
         }
     }
 
+    @Test
+    public void testCopyScalarMemory() {
```
natgavrilenko: To be consistent with naming, the test should be called testCopyMemoryScalar

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,231 @@ public void testStoreWithIllegalTags() {
...
+        builder.mockIntType("%int", 32);
+        builder.mockPtrType("%int_ptr", "%int", "Workgroup");
+        ScopedPointerVariable target = builder.mockVariable("%target", "%int_ptr");
+        ScopedPointerVariable source = builder.mockVariable("%source", "%int_ptr");
+
+        // when
+        parse(input);
+
+        // then
+        List<Event> events = builder.getCurrentFunction().getEvents();
```
natgavrilenko: Since you are using this in each test, maybe add a new private method
`private List<Event> getLastEvents(int count) { ... }` ?

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,231 @@ public void testStoreWithIllegalTags() {
...
+        // The mocked program may contain setup events, so access the events generated by OpCopyMemory from the end.
+        Load load = (Load) events.get(events.size() - 2);
+        Store store = (Store) events.get(events.size() - 1);
+        assertEquals(source, load.getAddress());
+        assertEquals(target, store.getAddress());
+        assertEquals(load.getResultRegister(), store.getMemValue());
+        assertTrue(load.hasTag(Tag.Spirv.SC_WORKGROUP));
```
natgavrilenko: I would just compare the whole tag sets - this is more precise and avoids long lists of assertTrue(hasTag) and assertFalse(hasTag):
```
assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.READ, Tag.Spirv.SC_WORKGROUP), load.getTags());
assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.WRITE, Tag.Spirv.SC_WORKGROUP), store.getTags());
```
The same for the other tests

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,231 @@ public void testStoreWithIllegalTags() {
...
+        Load load = (Load) events.get(events.size() - 2);
+        Store store = (Store) events.get(events.size() - 1);
+        assertTrue(load.hasTag(Tag.Spirv.SC_WORKGROUP));
+        assertTrue(store.hasTag(Tag.Spirv.SC_UNIFORM));
+    }
+
+    @Test
+    public void testCopyMemoryWithSharedMemoryOperands() {
+        // given
+        String input = "OpCopyMemory %target %source Volatile";
```
natgavrilenko: I would use NonPrivatePointer instead of Volatile, just to make the test more relevant to the memory model

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,231 @@ public void testStoreWithIllegalTags() {
...
+        List<Event> events = builder.getCurrentFunction().getEvents();
+        // The mocked program may contain setup events, so access the events generated by OpCopyMemory from the end.
+        Load load = (Load) events.get(events.size() - 2);
+        Store store = (Store) events.get(events.size() - 1);
+        assertTrue(load.hasTag(Tag.Spirv.MEM_VOLATILE));
+        assertTrue(store.hasTag(Tag.Spirv.MEM_VOLATILE));
+    }
+
+    @Test
+    public void testCopyMemoryWithSharedAvailabilityAndVisibility() {
```
natgavrilenko, line 257: Did you try to check all these combinations with spirv-val? Are there any disagreements about what we consider legal and illegal?

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,231 @@ public void testStoreWithIllegalTags() {
...
+        assertTrue(store.hasTag(Tag.Spirv.MEM_AVAILABLE));
+        assertTrue(store.hasTag(Tag.Spirv.DEVICE));
+        assertFalse(store.hasTag(Tag.Spirv.MEM_VISIBLE));
+        assertFalse(store.hasTag(Tag.Spirv.WORKGROUP));
+    }
+
+    @Test
+    public void testCopyMemoryWithSeparateMemoryOperands() {
+        // given
+        String input = "OpCopyMemory %target %source Nontemporal Volatile";
```
natgavrilenko: I didn't see much Nontemporal and Volatile in practice. Maybe one can be NonPrivatePointer and the other Aligned to make it more realistic. The same for the other Volatile.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/helpers/HelperTags.java
```diff
@@ -38,23 +38,31 @@ public static Set<String> parseMemorySemanticsTags(String id, Expression expr) {
         return tags;
     }
 
-    public static Set<String> parseMemoryOperandsTags(List<String> operands, Integer alignment,
-                                                      List<String> paramIds, List<Expression> paramsValues) {
+    public static MemoryOperandTags parseMemoryOperandsTags(List<String> operands, Integer alignment,
```
natgavrilenko: I think it would be cleaner to return the set of tags as before and leave splitting to the OpCopyMemory visitor. Here we just need to:
1) remove the error for `tagSet.contains(MEM_AVAILABLE) && tagSet.contains(MEM_VISIBLE)`
2) Validate the number of scopes depending on which of MakePointerAvailable and MakePointerVisible flags are set + add tests for the incorrect counts.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -86,8 +84,53 @@ public Event visitOpLoad(SpirvParser.OpLoadContext ctx) {
...
+        final Expression source = builder.getExpression(sourceId);
+        if (!(target.getType() instanceof ScopedPointerType targetType)
+                || !(source.getType() instanceof ScopedPointerType sourceType)
+                || !targetType.getPointedType().equals(sourceType.getPointedType())) {
+            throw new ParsingException("Mismatching pointer types in OpCopyMemory");
+        }
+
+        final Type type = targetType.getPointedType();
+        final String registerId = "copy_memory_" + copyMemoryCount++;
+        final List<Event> loads = visitMemoryAccess(registerId, type, source, (i, address) -> {
```
natgavrilenko: visitMemoryAccess looks very similar to the one in visitOpLoad. Maybe extract it into a separate function?

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
hernanponcedeleon: Yes, both dartagnan and spirv-val agree on all tests: they reject `testCopyMemoryRejectsVisibleTargetMask` and `testCopyMemoryRejectsAvailableSourceMask` and validate al others. 

As I wrote in #1111, I think we should get rid of all the rules in dartagnan and always run the validator whenever the input is a `*.spvasm` (implementing this is trivial after #1082). This would simplify our code and we will get always the most up to date rules (assuming we have an up to date spirv-val).

natgavrilenko: Did you also check multiple and single avvis operand cases, like  "NonPrivatePointer|MakePointerAvailable|MakePointerVisible %device %workgroup" and "NonPrivatePointer|MakePointerAvailable|MakePointerVisible %device"? Not sure if such exotic cases have been carefully tested in spirv-val.

Do you have a spirv file to tests this, so that I don't need to compose it manually?

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,214 @@ public void testStoreWithIllegalTags() {
...
+                Tag.Spirv.SC_WORKGROUP), load.getTags());
+        assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.WRITE, Tag.Spirv.MEM_NON_PRIVATE,
+                Tag.Spirv.SC_WORKGROUP), store.getTags());
+    }
+
+    @Test
+    public void testCopyMemoryWithSharedAvailabilityAndVisibility() {
+        // given
+        String input = "OpCopyMemory %target %source "
+                + "NonPrivatePointer|MakePointerAvailable|MakePointerVisible %workgroup %workgroup";
```
natgavrilenko: I would make on of the scopes %device to make sure that the scopes are assigned correctly

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,214 @@ public void testStoreWithIllegalTags() {
...
+        // then
+        List<Event> events = getLastEvents(2);
+        Load load = (Load) events.get(0);
+        Store store = (Store) events.get(1);
+        assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.READ, Tag.Spirv.MEM_NON_PRIVATE,
+                Tag.Spirv.MEM_VISIBLE, Tag.Spirv.WORKGROUP, Tag.Spirv.SC_WORKGROUP), load.getTags());
+        assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.WRITE, Tag.Spirv.MEM_NON_PRIVATE,
+                Tag.Spirv.MEM_AVAILABLE, Tag.Spirv.WORKGROUP, Tag.Spirv.SC_WORKGROUP), store.getTags());
+    }
+
```
natgavrilenko: There is one corner case that we didn't cover. What if MakePointerVisible is before MakePointerAvailable and the scopes are different for them? Is the first scope always for availability? Or are scopes listed in the order of the operand flags? Maybe we can tests with spirv-as and spirv-dis.

Example:
```
    @Test
    public void testCopyMemoryWithSharedAvailabilityAndVisibilityInverted() {
        // given
        String input = "OpCopyMemory %target %source "
                + "NonPrivatePointer|MakePointerVisible|MakePointerAvailable %device %workgroup";
        builder.mockIntType("%int", 32);
        builder.mockPtrType("%int_ptr", "%int", "Workgroup");
        builder.mockVariable("%target", "%int_ptr");
        builder.mockVariable("%source", "%int_ptr");
        builder.mockConstant("%device", "%int", 1);
        builder.mockConstant("%workgroup", "%int", 2);

        // when
        parse(input);

        // then
        List<Event> events = getLastEvents(2);
        Load load = (Load) events.get(0);
        Store store = (Store) events.get(1);
        assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.READ, Tag.Spirv.MEM_NON_PRIVATE,
                Tag.Spirv.MEM_VISIBLE, Tag.Spirv.DEVICE, Tag.Spirv.SC_WORKGROUP), load.getTags());
        assertEquals(Set.of(Tag.VISIBLE, Tag.MEMORY, Tag.WRITE, Tag.Spirv.MEM_NON_PRIVATE,
                Tag.Spirv.MEM_AVAILABLE, Tag.Spirv.WORKGROUP, Tag.Spirv.SC_WORKGROUP), store.getTags());
    }
```

Now this test fails, but I am not sure which one is the correct behavior. I can also raise this question in the SPIR-V meeting or at the Memory Model TSG.

dartagnan/src/test/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemoryTest.java
```diff
@@ -187,6 +187,214 @@ public void testStoreWithIllegalTags() {
...
+        builder.mockVariable("%source", "%int_ptr");
+        builder.mockConstant("%device", "%int", 1);
+
+        try {
+            // when
+            parse(input);
+            fail("Should throw exception");
+        } catch (ParsingException e) {
+            // then
+            assertEquals(String.format("OpCopyMemory cannot contain tag '%s'", Tag.Spirv.MEM_VISIBLE),
```
natgavrilenko: The error message is misleading: OpCopyMemory can contain MEM_VISIBLE but not for the target operand. The same for MEM_AVAILABLE.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/program/visitors/spirv/VisitorOpsMemory.java
```diff
@@ -64,30 +68,66 @@ public Event visitOpLoad(SpirvParser.OpLoadContext ctx) {
...
-        checkAndPropagateTags(events, tags, Tag.Spirv.MEM_AVAILABLE, ctx.pointer().getText(), "OpLoad");
+        Set<String> tags = parseAndValidateMemoryAccessTags(
+                ctx.memoryAccess(), Tag.Spirv.MEM_AVAILABLE, "OpLoad");
+        addMemoryEvents(events, tags, ctx.pointer().getText());
+        return null;
+    }
+
+    @Override
+    public Event visitOpCopyMemory(SpirvParser.OpCopyMemoryContext ctx) {
+        final String targetId = ctx.targetIdRef().getText();
```
natgavrilenko: Could you remove these 'final' to be consistent with the rest of the file and spirv visitors? Modern Java is smart enough understand that these variables are effectively final.

The same in the other new methods

PR #1101, opened 2026-09-13
[CI] Add job to check for unused imports and variables and missing Override

hernanponcedeleon:


ThomasHaas:
If you find a way to make IntelliJ remove unused imports without modifying them in any other way (collapsing/expanding), let me know :). In case that everyone uses IntelliJ, we could also just let it do its thing.

hernanponcedeleon:
Maybe we can try [Error Prone](https://github.com/google/error-prone). It can be integrated with maven, intellij (and other IDEs) and would even allow us to write our own rules if we want to enforce certain coding styles.

hernanponcedeleon:
> If you find a way to make IntelliJ remove unused imports without modifying them in any other way (collapsing/expanding), let me know :). In case that everyone uses IntelliJ, we could also just let it do its thing.

Running `mvn --batch-mode -Pstyle-fix clean test-compile` will do the job now

ThomasHaas:
Hmm, I guess this complains about all unused data, seeing that some `fromConfig` methods are gone and some parameters in the `RefinementSolver`. I'm not against it, but at least some choices were deliberate (at least back then when they were made).

hernanponcedeleon:
The job just complained about the unused `config` but it did not remove the `fromConfig`. This I did manually cause I did not see the point of having this method for a class where we do not have options.

ThomasHaas:
It was mainly to avoid situations where you add options, then update all callers from `newInstance` to `fromConfig` and when you later move/delete the options, you then change back all the callers from `fromConfig` to `newInstance`.
It's not a big deal to do so, but that was basically the intention. It also guaranteed that you can add options without changing any callers at all, since they are forced to pass a `Configuration` anyways. When all of this was added for the first time, I think every class had only a `fromConfig` method and `newInstance` didn't exist at all.

ThomasHaas:
Does `mvn --batch-mode -Pstyle-fix clean test-compile` only complain about issues or does it automatically fix some? If it is the former, then it's not better than the IDE's code analysis, no?

hernanponcedeleon:
There are two profiles: one for checking (this is what the CI runs) and one for fixing (this is responsibility of the developer; I don't think is a good idea let the CI change the code).

ThomasHaas (approved):
LGTM

PR #1104, opened 2026-09-14
Add cutting to CAT

ThomasHaas:
- `cut name` where "name" is a relation or axiom name can be used to force cutting for refinement solving. If the name is not recognized, we only give a warning (I could make this an exception).
- Removed `refinement.baseline` option.


ThomasHaas:
I enabled a certain cut by default which seems to have some negative impact. I guess this shows that the cutting actually works though :P.

cat/stdlib.cat
```diff
@@ -13,6 +13,6 @@ let fri = fr & int
 let fencerel(F) = po;[F];po
 
 // Enforce that reads from uninitialized memory are impossible if an init write exists.
-empty (([R] \ [range(rf)]);loc;[IW])
+empty ([R \ range(rf)]);loc;[IW] as force-init-read
```
hernanponcedeleon, line 16: Now that you removed cutting this, I guess there is no need to name the axiom (the meaning is clear from the comment above)

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
```diff
@@ -379,6 +380,31 @@ public Relation visitExprCartesian(ExprCartesianContext c) {
...
+
+        // Failed to annotate relation, try to annotate axiom.
+        wmm.getAxioms().stream()
+                .filter(a -> name.equals(a.getName()))
+                .findFirst()
+                .ifPresentOrElse(a -> a.setMetadata(CutAnnotation.get()),
+                        () -> logger.warn("Unable to cut '{}': name is not recognized.", name));
+
+        return null;
+    }
```
hernanponcedeleon, line 404: Cutting a relation would enforce eagerly encoding it (and its dependencies) while cutting an axiom would do this (for the constraint relation) + eagerly encoding the axiom per se, right?

hernanponcedeleon (commented):
> If the name is not recognized, we only give a warning (I could make this an exception).

I think the exception would be better.

`benchmarks/challenging/cna.c` has a bunch of references to the option you are removing.

`Baseline.java` can be removed.

dartagnan/src/main/java/com/dat3m/dartagnan/parsers/cat/VisitorCat.java
ThomasHaas: yes

cat/stdlib.cat
ThomasHaas: sure, but do you think it is bad to have the name?

hernanponcedeleon: I initially thought it wat a bit inconsistent, but at least the other axiom in this file has a name, so it is ok

hernanponcedeleon (approved):

PR #1102, opened 2026-09-13
[WIP][SVCOMP] Add support for sv-witness generation in 2.2. format

hernanponcedeleon:
## Summary

This PR adds support for generating violation witnesses in the SV-Witness 2.2 YAML format. It supports reachability, data-race, overflow, and memory-safety violations.

The generated witnesses describe a coarse concurrent execution containing the relevant thread ordering, assumptions, and violation targets.

## Validation

The 30 fastest successful SV-COMP 2026 `false(unreach-call)` benchmarks were tested (using the only two available validators for the new format) with a 30-second limit per tool.

| Outcome | Dartagnan | Theta | CPAchecker |
| :--- | :---: | :---: | :---: |
| Benchmarks considered | 30 | 30 | 30 |
| Witnesses generated / validator invoked | 29 | 29 | 29 |
| Validated | — | 10 | 9 |
| Rejected | — | 3 | 3 |
| Timeout | 1 | 2 | 13 |
| Error or unsupported feature | 0 | 14 | 4 |
| Skipped | — | 1 | 1 |

Most unsuccessful validations resulted from timeouts or unsupported features in the validators.

PR #1105, opened 2026-09-14
Add source metadata to program

hernanponcedeleon:
This contains the changes from #1102 that were unreleaded to the witness generation.


dartagnan/src/main/java/com/dat3m/dartagnan/program/Program.java
```diff
@@ -110,6 +113,11 @@ public void setName(String name) {
         this.name = name;
     }
 
+    public Path getInputPath() {
+        Preconditions.checkState(hasMetadata(Source.class), "Program has no source metadata");
+        return getMetadata(Source.class).path();
+    }
+
```
ThomasHaas, line 120: I'm not sure if we should make any metadata mandatory. We can keep it like this for now, but it does feel a bit strange to me.

hernanponcedeleon: But then we should adapt other classes (e.g., `OutputGenerator`) which right now require this metadata to exists.

ThomasHaas: Yes, classes would need adapt. We can also make the input path an explicit field of `Program` (like you had it in #1102) to signify that is mandatory. We can also have mandatory metadata, but so far we didn't have that.

I think we will need to discuss guidelines for what `Metadata` should be used for, if it must be optional or can be mandatory, if it is allowed to affect semantics, and so on. At least some basic principles to follow would be useful.

hernanponcedeleon: > I think we will need to discuss guidelines for what `Metadata` should be used for, if it must be optional or can be mandatory, if it is allowed to affect semantics, and so on. At least some basic principles to follow would be useful.

Can you draft a PR for this? That PR can also do the restructuring you suggested [here](https://github.com/hernanponcedeleon/Dat3M/pull/1097#issuecomment-5665970908) (I think having the metadata as an inner class would be the best since otherwise we might forget we have some specific metadata  for some class when doing changes to that class).

ThomasHaas: I can do the changes I mentioned in #1097. Do you want to use that PR as a discussion platform regarding `Metadata`? I thought about opening a discussion thread instead.

hernanponcedeleon: Let's have those changes and the documentation on guidelines as a single PR.

We can probably discuss offline about a first draft and during review we see if we need to adapt.

ThomasHaas: I'm still not so happy about the helper method here that throws an exception.
Can't we just get the metadata in `OutputGenerator` (the only user of that data, right? via `getMetadata`) and if it is absent default to the program's name instead? We could even go further and check if the program has a name and if not default to "unknown".

I would then also change the ` toSummary` method back to use a string instead of a `Path` so that we can pass the program's name and possibly dummy value.

EDIT: Then you also don't need to invent fake source information in the UI code I think. Setting the name of the program is enough to get the correct output.

ThomasHaas (approved):
LGTM

PR #1107, opened 2026-09-15
Metadata refactor & guidelines

ThomasHaas:
As discussed in #1105, I moved the basic metadata classes to a top-level package.
I made all Event-specific metadata inner classes of `Event`.
I moved `SourceLocation` to the top-level package, seeing that is could (and probably should) be reused for programs (#1105).


ThomasHaas:
We wanted to discuss some guidelines about the usage of metadata. I will bring up some points/questions (not exhaustive) that we should talk about with tentative answers.

(1) **Q:** Is metadata always optional or can it be mandatory?
      **A: Preferred optional?**
      **Current state:** metadata is generally optional. However, there is a `UnrollingBound` metadata used to remember unrolling bounds at bound events which is unconditionally used in Dartagnan's loop bound dumping (though it probably could be made optional).
Maybe metadata that we are guaranteed to internally generate can be made mandatory? 

(2) **Q:** Should metadata be able to modify the semantics of, e.g., events?
      **A: No.** Semantic-defining information should always be explicit.

(3) **Q:** Can the presence of metadata affect the tools outcome?
      **A: Yes?** While metadata should not directly affect the ground semantics (and hence the expected verdict), it could be allowed to affect precision of the tool. For example, type-based alias analysis (TBAA) using source language types can be implemented via Metadata (see #617). Similarly, one could imagine value range metadata attached to, e.g., loads.
If the metadata is sound it can change UNKNOWN to PASS, but if it is unsound, it could change FAIL to PASS.

(4) **Q:** How complex can metadata be? 
**A: Simple?** Metadata should likely consist of simple types (ints, strings, paths, etc.) and not refer to IR-specific types (events, functions, registers, etc.). Furthermore, metadata classes should be immutable.

(5) **Q:** Where should metadata classes be located?
**A: Depends.** Metadata that is widely usable (e.g., source locations) should be in the general `metadata` package, but metadata that is specific to one type of class (e.g., event-specific metadata) should be an inner class.
The decision may not be clear: #1104 adds `CutAnnotation` metadata used for both `Relation` and `Constraint` which have no common superclass (maybe they should?) for `CutAnnotation` to live in. Maybe it could be put into the `Wmm` class?
An alternative would be to put all metadata into the `metadata` package, possibly within a subpackage for division (`metadata.program`, `metadata.wmm`, etc.). There should only be one layer of subpackages (no deep nesting) and the subdivision should be coarse (e.g. `metadata.program` instead of `metadata.event`, `metadata.memory`, etc.).

hernanponcedeleon:
> I moved `SourceLocation` to the top-level package, seeing that is could (and probably should) be reused for programs (#1105).

I think `SourceLocation` is more specific (e.g., it contains a line number) than what we would need for programs. I think We could have `Source` as general metadata (maybe useful for witnesses too (?)) and maybe change the string in `SourceLocation.Generic` to `Source`.

ThomasHaas:
I think there are different possible ways to go about it. We can try with a basic `Source` metadata and have `SourceLocation` enrich `Source` with `Location` data.

hernanponcedeleon:
> We wanted to discuss some guidelines about the usage of metadata. I will bring up some points/questions (not exhaustive) that we should talk about with tentative answers.
> 
> (1) **Q:** Is metadata always optional or can it be mandatory? **A: Preferred optional?** **Current state:** metadata is generally optional. However, there is a `UnrollingBound` metadata used to remember unrolling bounds at bound events which is unconditionally used in Dartagnan's loop bound dumping (though it probably could be made optional). Maybe metadata that we are guaranteed to internally generate can be made mandatory?
> 

I think the "if its presence can be guaranteed by us, making it mandatory is ok" is a good trade-off.

> (2) **Q:** Should metadata be able to modify the semantics of, e.g., events? **A: No.** Semantic-defining information should always be explicit.

Agree.

> 
> (3) **Q:** Can the presence of metadata affect the tools outcome? **A: Yes?** While metadata should not directly affect the ground semantics (and hence the expected verdict), it could be allowed to affect precision of the tool. For example, type-based alias analysis (TBAA) using source language types can be implemented via Metadata (see #617). Similarly, one could imagine value range metadata attached to, e.g., loads. If the metadata is sound it can change UNKNOWN to PASS, but if it is unsound, it could change FAIL to PASS.
> 

I think the following would fit in this Q. Imagine we would implement support for the `locations` annotations in litmus needed in #1086 as metadata. On the one side, I can see this would make sense if you think of it as "lets add a zoom filter to no notion of state"; one the other side, it will definitely affect the results of the exploration mode.

> (4) **Q:** How complex can metadata be? **A: Simple?** Metadata should likely consist of simple types (ints, strings, paths, etc.) and not refer to IR-specific types (events, functions, registers, etc.). Furthermore, metadata classes should be immutable.

> I cannot envision any use case where we would require IR-specific things. I would go for "just simple" and we can revise if the day comes.

> 
> (5) **Q:** Where should metadata classes be located? **A: Depends.** Metadata that is widely usable (e.g., source locations) should be in the general `metadata` package, but metadata that is specific to one type of class (e.g., event-specific metadata) should be an inner class. The decision may not be clear: #1104 adds `CutAnnotation` metadata used for both `Relation` and `Constraint` which have no common superclass (maybe they should?) for `CutAnnotation` to live in. Maybe it could be put into the `Wmm` class? An alternative would be to put all metadata into the `metadata` package, possibly within a subpackage for division (`metadata.program`, `metadata.wmm`, etc.). There should only be one layer of subpackages (no deep nesting) and the subdivision should be coarse (e.g. `metadata.program` instead of `metadata.event`, `metadata.memory`, etc.).

As we already discussed offine, I am in favor of keeping it class-internal as we already did in #1104.

BTW, I think we should add this documentation as part of the `Metadata` interface (e.g., this would help agents to spot wrong usages of it).

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/SourceLocation.java
```diff
@@ -0,0 +1,30 @@
...
+        public String toString() {
+            return threadName + "#" + lineNumber;
+        }
+    }
+
+    record Generic(String source, int lineNumber) implements SourceLocation {
+        @Override
+        public String toString() {
+            // If source is a path, we take the last path element
+            // if source is not a path, the function will return it as is
```
hernanponcedeleon: You mean `Path.of(source).getFileName().toString() == source` if the path does not exists? If so I would explicitly say so to avoid confusion.

EDIT: it seems `Path.of("/").getFileName()` would return `null`

ThomasHaas: It's not about path existence but rather path shape: `not/a/path/to/a/file` would still yield `file`.
Technically, `Path.of()` can throw an exception if there are invalid symbols in the path though. I can also revert it back to the old which used explicit substring manipulation, but now that I think about it, the old code expected `/` as separator which may fail under Windows... I guess we should just use a proper `Path`.

> it seems Path.of("/").getFileName() would return null

Interesting, I expected it to return the empty string.

ThomasHaas:
> > (3) **Q:** Can the presence of metadata affect the tools outcome? **A: Yes?** While metadata should not directly affect the ground semantics (and hence the expected verdict), it could be allowed to affect precision of the tool. For example, type-based alias analysis (TBAA) using source language types can be implemented via Metadata (see #617). Similarly, one could imagine value range metadata attached to, e.g., loads. If the metadata is sound it can change UNKNOWN to PASS, but if it is unsound, it could change FAIL to PASS.
> 
> I think the following would fit in this Q. Imagine we would implement support for the `locations` annotations in litmus needed in #1086 as metadata. On the one side, I can see this would make sense if you think of it as "lets add a zoom filter to no notion of state"; one the other side, it will definitely affect the results of the exploration mode.
> 
> > (4) **Q:** How complex can metadata be? **A: Simple?** Metadata should likely consist of simple types (ints, strings, paths, etc.) and not refer to IR-specific types (events, functions, registers, etc.). Furthermore, metadata classes should be immutable.
> 
> > I cannot envision any use case where we would require IR-specific things. I would go for "just simple" and we can revise if the day comes.

You just mentioned one in point (3) though. The `locations` annotations refer to registers and memory locations in the IR and so if it was Metadata, it would refer to IR objects. You can, of course, try to use only names in the metadata and later connect them back to the IR to avoid direct references. But I don't think this circumvents the actual problem, namely, that anything that refers to IR (directly or indirectly) can get easily broken by processing passes.

Given these arguments, I think `locations` should not be metadata. That doesn't mean we have to make every non-metadata piece a field in `Program` though. I could imagine to have something like a modular `ProgramExtension` member which can be `LitmusExtension = {Specification, Locations, Filter}` for litmus code and `None` for plain LLVM/SPIRV. 

EDIT: If the location metadata would be a `LocationAnnotation` directly attached to the relevant registers and memory objects, I could imagine it as a sort of metadata (its a bit in the gray zone :)). However, I still would prefer the above sketched solution.

> BTW, I think we should add this documentation as part of the `Metadata` interface (e.g., this would help agents to spot wrong usages of it).

Sure, but first we need to have some consensus :). I didn't want to put my proposed rules into the text before anyone showed agreement :)

ThomasHaas:
```
[ERROR] com.dat3m.dartagnan.litmus.comparison.C11AndOpenCLRaceTest.testAssume[0: ../litmus/OPENCL/portedFromC11/manual/imm-E3.3.litmus]
java.lang.AssertionError: expected:<true> but was:<false>
```
This is a new one. Given how many CI runs we had without this error, debugging this one is going to be nightmare. I think RA and its reasoning over stratifications might be the most likely culprit here.

hernanponcedeleon:
> ```
> [ERROR] com.dat3m.dartagnan.litmus.comparison.C11AndOpenCLRaceTest.testAssume[0: ../litmus/OPENCL/portedFromC11/manual/imm-E3.3.litmus]
> java.lang.AssertionError: expected:<true> but was:<false>
> ```
> 
> This is a new one. Given how many CI runs we had without this error, debugging this one is going to be nightmare. I think RA and its reasoning over stratifications might be the most likely culprit here.

I asked an agent to reproduced and it could not. This plus the fact that I never seen that test fail (not only in this PR history, but any any previous run) suggests the chances to hit it again are extremely low.

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/SourceLocation.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    record Litmus(String threadName, int lineNumber) implements SourceLocation {
+        @Override
+        public String toString() {
+            return threadName + "#" + lineNumber;
+        }
+    }
+
+    record Generic(Path sourcePath, int lineNumber) implements SourceLocation {
```
hernanponcedeleon, line 26: Using `Path` here may cause problems. This metadata is populated using debug information from LLVM/SPIRV. If the file was generated by Windows, but we are running the code now from Linux, the incompatibility between systems will results in something like `"null#line"`

ThomasHaas:
I couldn't reproduce it either and it is orthogonal to this PR. However, if you wanna debug it, I don't think you should just run the code until the error occurs but instead check the may/must sets in every run. If the error is due to RA, then maybe there are fluctuations in RA that do not immediately affect the verdict but still show the presence of a bug. Apart from RA, I cannot imagine what else could be wrong.

hernanponcedeleon:
A few comments about the current state of the metadata following the proposed guidelines:
- `CustomPrinting` provides a `stringify` method that refers to our IR (thus violating (1)). Can we move the metadata to `AbstractEvent` (which makes use of `stringify` via `this`) and remove the `Event` parameter from the method?
- `MemoryOrder` is used by `VisitorC11`to decide whether an access is atomic, thus affecting semantics and violating (2).

I also think this pattern is currently allowed and can cause problems
```
setMetadata(new SourceLocation.Litmus(...));
setMetadata(new SourceLocation.Generic(...));
getMetadata(SourceLocation.class);
```
We should probably have some guidelines (and maybe even automatic checks that these are followed) about which kind of metadata can be combined.

About the guidelines: should we maybe split the matadata into different "sub-types"? Something like
- provenance: source locations, original names, original IDs. It may affect output but not accepted programs or verification results.
- optimization hints: TBAA, ranges. Absence must use a conservative fallback; consuming the hint must be sound.
- internally generated: unrolling/compilation IDs. The producer and consumer must be documented.

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/SourceLocation.java
ThomasHaas: I don't understand. The point of `Path` is that it is OS-agnostic, no? Why would we have issues there?
Is it because the LLVM visitor constructs the path via `directory + "/" + file` and so the separator is inconsistent (we should use `Path.of(directory, file)` though)? Even then, `/` is a valid separator on Windows I think.

hernanponcedeleon: Suppose you are given a *.ll file generated from Windows. The debug information contains `C:\projects\demo\test.c`. 

Dartagnan later parses that LLVM file on Linux:
```
Path sourcePath = Path.of("C:\\projects\\demo\\test.c");
sourcePath.getFileName();
```
Because `\` is not a path separator on Linux, the result is the entire string `C:\projects\demo\test.c` instead of `test.c`.

To be fair, I think we have the same problem on `development`. Also, since debug information is probably less useful when we are dealing with the compiled code rather than the source, this is not much of a real problem.

ThomasHaas:
> A few comments about the current state of the metadata following the proposed guidelines:
> 
> * `CustomPrinting` provides a `stringify` method that refers to our IR (thus violating (1)). Can we move the metadata to `AbstractEvent` (which makes use of `stringify` via `this`) and remove the `Event` parameter from the method?

The point of `(1)` was about state referring to IR because passes can invalidate IR breaking the metadata. Methods referring to IR are no problem and `CustomPrinting` is stateless.

> * `MemoryOrder` is used by `VisitorC11`to decide whether an access is atomic, thus affecting semantics and violating (2).

Yeah, that sucks a bit. For the most part, the metadata is generated within `VisitorC11` (as part of new event creation) and so its existence is guaranteed. But the cases of plain `Load/Store` events is different: those carry the metadata since the parser created them and there are plenty opportunities for the metadata to be dropped.
The best solution is for `VisitorC11` to not use the metadata IMO. `tagList` probably needs to check for every memory-order-related tag (RLX/REL/ACQ/SC) and attach atomicity tags based on that.
Alternatively, maybe the frontend should be responsible for marking events as atomic/nonatomic (in some sense, the parser's output is "incorrectly tagged" and the compilation pass fixes this in hindsight). After all, the frontend is also responsible for putting the correct memory order tags.

> I also think this pattern is currently allowed and can cause problems
> 
> ```
> setMetadata(new SourceLocation.Litmus(...));
> setMetadata(new SourceLocation.Generic(...));
> getMetadata(SourceLocation.class);
> ```
> 
> We should probably have some guidelines (and maybe even automatic checks that these are followed) about which kind of metadata can be combined.

I don't think this is a real issue to be honest. The frontend should be generating that metadata and it should know not to attach multiple types of source information to the same object. Trying to come up with a fix to a problem that will likely never arise is wasted effort IMO.

> About the guidelines: should we maybe split the matadata into different "sub-types"? Something like
> 
>     * provenance: source locations, original names, original IDs. It may affect output but not accepted programs or verification results.
> 
>     * optimization hints: TBAA, ranges. Absence must use a conservative fallback; consuming the hint must be sound.
> 
>     * internally generated: unrolling/compilation IDs. The producer and consumer must be documented.

I also thought about a more fine-granular distinction, but I'm not sure what distinctions are practically useful and if they are needed yet. I like the provenance category insofar as it captures the fact that the metadata should be preserved through modifications of the annotated object (in particular copies should keep that metadata).
In that sense, I also think `UnrollingId` and `CompilationId` are a kind of provenance (they are also optional btw.).
In contrast, other types of metadata may get invalidated by transformations. Currently, we blindly copy all metadata during processing but it has ugly consequences, for example, `CustomPrinting` of an ARM8 `CAS` is preserved during compilation and so the resulting subevents have a wrong printer attached (the printer is smart to realize this and fallback to the default printing).
As for optimization hints, I guess `CutAnnotation` falls into this category.
`CustomPrinting` and `MemoryOrder` do not quite fit anywhere though. Both were intended only for IR printing and should not affect Dartagnan's behaviour in any way (modulo the fact that `VisitorC11` abuses `MemoryOrder` right now). I'm not sure if "internally generated" is a reasonable category and what consequences it has. I think it is rather the case that there is "uncategorized metadata" (no matter where it is generated) which requires more accurate documentation precisely because it lacks a category.

Btw. now that I think about it, maybe we should move `UnrollingId` and `CompilationId` into the classes that generate them? They have a unique producer after all. Generally, if metadata has a unique producer or a unique consumer, we should maybe put it into that class. If neither is unique, we place it into the annotated class as we do right now?
By "unique" I don't mean that we currently happen to have exactly one class producing/consuming the metadata but rather that there is no expectation to have any other class in the future that also produce/consume it.

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/SourceLocation.java
ThomasHaas: Ok, from Windows to Linux there are problems (the other way around is fine). It does not result in `null#line` though but rather an overly verbose source location. The same problem does not exist on `development` because we explicitly use `directory + "/" + file` which will result in a path like `C:\\projects\\demo/test.c` with mixed separators. We then do the split on "/" to extract the file name.
I guess we need to add a utility that replaces "\\\\" by "/" in paths to normalize them.

ThomasHaas:
I think we should avoid making the framework too rigorous: the guidelines are "soft constraints" and not hard rules.

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/SourceLocation.java
ThomasHaas: Hmm, I think there is a different underlying problem. If the source path was specified as `C:\\projects\\demo\\test.c` and we verify this on Linux/MacOS, the source path we track should still be the windows path, right? I mean, are we allowed to represent the path differently at all? I wouldn't be so strict about it and allow the path to be tracked as `C:/projects/demo/test.c` instead, but it certainly is something to consider.

ThomasHaas:
Btw. we use several tags internally like `__NOOPT` and `__SPINLOOP`. I'm wondering if they could (and should?) be replaced by some form of metadata. I think some of those tags can actually influence the semantics. But maybe that just means that those should not be tags in the first place?

dartagnan/src/main/java/com/dat3m/dartagnan/program/event/Event.java
```diff
@@ -85,4 +87,27 @@ public interface Event extends Comparable<Event>, MetadataCarrier<Event> {
...
+    record CompilationId(int value) implements Metadata { }
+
+    @FunctionalInterface
+    interface CustomPrinting extends Metadata {
+        Optional<String> stringify(Event e);
+    }
+
+    // Marks a specific tag as "memory order". Mostly used for printing.
+    //TODO: Add a factory with caching, because we can share the same MemoryOrder instance between different events.
+    record MemoryOrder(String value) implements Metadata { }
```
hernanponcedeleon: > The best solution is for VisitorC11 to not use the metadata IMO.
Can you please add a TODO here?

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/Metadata.java
```diff
@@ -0,0 +1,31 @@
...
+
+    (5) Metadata classes should be inner classes of the class they intend to annotate.
+        If a Metadata class applies to multiple related classes, it can be placed in a related class.
+        If a Metadata class applies to unrelated classes, it should be placed in the metadata package.
+
+    (EXCEPTIONS)
+    For metadata that Dartagnan can guarantee to generate internally, e.g. as part of its processing pipeline,
+    the rules can be relaxed:
+    - Rule (4) does not strictly apply: we can have mandatory metadata, but optional metadata is still preferred.
+
```
hernanponcedeleon, line 30: We should add some guidelines about copying metadata. 

We should also add a general TODO about lifetime of the metadata and specific TODOs for those that are currently broken (e.g., `CustomPrinting` on an some ARM8 events)

dartagnan/src/main/java/com/dat3m/dartagnan/metadata/SourceLocation.java
hernanponcedeleon: > The frontend should be generating that metadata and it should know not to attach multiple types of source information to the same object.

Please document this here

hernanponcedeleon (approved):

PR #1108, opened 2026-09-15
Fix unordered floating-point comparison encoding

hernanponcedeleon:
### Description

Fixes the SMT encoding of unordered floating-point comparisons.

JavaSMT’s floating-point comparison operations are ordered and therefore evaluate to `false` when either operand is `NaN`. LLVM unordered comparisons instead evaluate to `true` when either operand is `NaN` (see [here](https://llvm.org/docs/LangRef.html#fcmp-instruction)). The previous encoding consequently produced incorrect results for `ueq`, `ult`, `ule`, `ugt`, and `uge`.

A new test was added to `benchmarks/miscellaneous/floats.c`. It needs to be compiled with `-O1` to generate the corresponding unordered LLVM `fcmp` instructions. The helper functions are marked `noinline` to ensure all predicates remain present and reachable in the generated LLVM IR.

---

### Reproducing from the C source

This has been tested with Clang 14 and Clang 23.

Save the following compilation pipeline as `test.yml`.
```yaml
workdir: "$DAT3M_OUTPUT"

pipelines:
  - pipeline: ".c"
    output: "{basename}.ll"
    commands:
      - name: "Compile test 17"
        tool: "clang"
        input: "{pipeline_input}"
        output: "{basename}.ll"
        args:
          - "-O1"
          - "-DTEST_ID=17"
          - "-I"
          - "$DAT3M_HOME/include"
          - "-S"
          - "-emit-llvm"
          - "-g"
          - "{cmd_input}"
          - "-o"
          - "{cmd_output}"
```

Run as
```bash
java -jar dartagnan/target/dartagnan.jar \
    --compilation.pipeline=test.yml \
    cat/imm.cat \
    --target=imm \
    --property=program_spec \
    benchmarks/miscellaneous/floats.c
```
The command reports `FAIL` on development and `PASS` with this fix.


dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -523,12 +523,12 @@ public TypedFormula<BooleanType, BooleanFormula> visitFloatCmpExpression(FloatCm
...
+                case UEQ -> fromOrdToUnord(l, r, fpmgr.equalWithFPSemantics(l, r));
                 case UNEQ -> bmgr.not(fpmgr.equalWithFPSemantics(l, r));
-                case ULT -> fpmgr.lessThan(l, r);
-                case ULTE -> fpmgr.lessOrEquals(l, r);
-                case UGT -> fpmgr.greaterThan(l, r);
-                case UGTE -> fpmgr.greaterOrEquals(l, r);
+                case ULT -> fromOrdToUnord(l, r, fpmgr.lessThan(l, r));
+                case ULTE -> fromOrdToUnord(l, r, fpmgr.lessOrEquals(l, r));
+                case UGT -> fromOrdToUnord(l, r, fpmgr.greaterThan(l, r));
+                case UGTE -> fromOrdToUnord(l, r, fpmgr.greaterOrEquals(l, r));
```
ThomasHaas: It is strange got have `fromOrdToUnord` and `fromUnordToOrd` applied to the same `fpmgr.cmpop` function, because it implies that `fpmgr.cmpop` is both ordered and unordered?
I think the functions should simply be `toOrd` and `toUnord`; they both handle NaN explicitly so that the behavior of `fpmgr.cmpop` on NaN is irrelevant.

Also, `ORD` is equivalent to `toOrd(l, r, bmgr.true())` and `UNO` is equivalent to `toUnord(l, r, bmgr.false())`.
Using this, we can avoid duplicating the NaN reasoning at two places.

dartagnan/src/main/java/com/dat3m/dartagnan/encoding/ExpressionEncoder.java
```diff
@@ -540,6 +540,12 @@ private BooleanFormula fromUnordToOrd(FloatingPointFormula l, FloatingPointFormu
             return fmgr.ifThenElse(bmgr.or(fpmgr.isNaN(l), fpmgr.isNaN(r)), bmgr.makeFalse(), cmp);
```
ThomasHaas: I would change this to `bmgr.and(bmgr.not(fpmgr.isNaN(l)), bmgr.not(fpmgr.isNaN(r)), cmp);` to better align with `toUnords` implementation (and also with the LLVM IR language ref).

ThomasHaas (approved):
LGTM

PR #1111, opened 2026-09-16
[SPIRV] Relax diff labels rule for conditional jumps prior to SPIRV 1.6

hernanponcedeleon:
The rule is only valid from SPIRV 1.6 onwards, thus we are currently rejecting perfectly valid kernels.


xeren (approved):
LGTM

hernanponcedeleon:
It seem `spirv-val` actually checks for this rule. E.g., the validation passes for this shader
```
  ; SPIR-V
  ; Version: 1.5
  ; Generator: Khronos SPIR-V Tools Assembler; 0
  ; Bound: 8
  ; Schema: 0
                 OpCapability Shader
                 OpMemoryModel Logical GLSL450
                 OpEntryPoint GLCompute %1 "main"
         %void = OpTypeVoid
         %bool = OpTypeBool
         %true = OpConstantTrue %bool
            %5 = OpTypeFunction %void
            %1 = OpFunction %void None %5
            %6 = OpLabel
                 OpBranchConditional %true %7 %7
            %7 = OpLabel
                 OpReturn
                 OpFunctionEnd

```
but fails for this one
```
  ; SPIR-V
  ; Version: 1.6
  ; Generator: Khronos SPIR-V Tools Assembler; 0
  ; Bound: 8
  ; Schema: 0
                 OpCapability Shader
                 OpMemoryModel Logical GLSL450
                 OpEntryPoint GLCompute %1 "main"
         %void = OpTypeVoid
         %bool = OpTypeBool
         %true = OpConstantTrue %bool
            %5 = OpTypeFunction %void
            %1 = OpFunction %void None %5
            %6 = OpLabel
                 OpBranchConditional %true %7 %7
            %7 = OpLabel
                 OpReturn
                 OpFunctionEnd

```
We should probably remove the check all-together and let `spirv-val` validate the kernels before we run them with dartagnan. I can add a new [compilation pipeline](https://github.com/hernanponcedeleon/Dat3M/blob/master/compilation.yml) rule which would do this for `spvasm` inputs.

ThomasHaas:
I wanted to comment about this yesterday but held back on it. I was always under the impression that Dartagnan might be doing too many basic validity checks that should never fail unless the SPIRV code was handwritten as the SPIRV-generating frontend should be doing those checks already. I mean, having those (redundant?) checks is not bad but it takes quite a bit of effort to get them all right for practically little benefit.
In the case of this PR, the checks for wrong labels/jumps is not even semantically relevant, i.e., it seems like an arbitrary syntactic restriction (there is probably a legitimate reason for the restriction though).

hernanponcedeleon:
>  was always under the impression that Dartagnan might be doing too many basic validity checks that should never fail unless the SPIRV code was handwritten as the SPIRV-generating frontend should be doing those checks already.

Writing SPIRV manually is a valid way of doing it (AFAIR Natalia has done this many times). Also, we should not rely on a generator having checks (I have seen generators getting it wrong). However, spirv-tools has a validator which should have most up-to-date rules in place, so the best way would be to use it.

natgavrilenko:
> > was always under the impression that Dartagnan might be doing too many basic validity checks that should never fail unless the SPIRV code was handwritten as the SPIRV-generating frontend should be doing those checks already.
> 
> Writing SPIRV manually is a valid way of doing it (AFAIR Natalia has done this many times). Also, we should not rely on a generator having checks (I have seen generators getting it wrong). However, spirv-tools has a validator which should have most up-to-date rules in place, so the best way would be to use it.

I don't think that removing validation is a good idea. First, spirv-val can have mistakes itself. Second, when we validate spirv instruction we also consider and test corner-case scenarios. But if just blindly trust SPIR-V, we can easily overlook those. We have a robust and well-tested parser, why to make it more sloppy?

PR #1112, opened 2026-09-17
Allow to pass compiler flags via CFLAGS

hernanponcedeleon:


ThomasHaas:
Isn't `CFLAGS` strictly associated with C compilers not general compilers? If so, I think using `CFLAGS` in all compilation pipelines is dangerous. What if the user has their CFLAGS set-up in the environment specifically for their C compiler (maybe `-Wall` is always set via a shell config) and then invokes Dartagnan on SPIRV code? The SPIRV compiler will probably fail due to the unrecognized options passed via CFLAGS, no?

Can't we just allow any `$OPTION` in the .yml as a standalone argument and let the user decide what he wants to call that option by modifying the .yml? We can of course put `$CFLAGS` by default for `.c` code but leave it out for other pipelines (unless they also have a commonly used name for the compiler flags option).

hernanponcedeleon:
In my mind I was always thinking of CFLAGS as "compiler flags", but you are right that CFLAGS/CXXFLAGS are standard flags. I create a dat3m specific one and refactor a bit the yml configuration.

Now these two work
```
DAT3M_COMPILER_OPTIONS="-DACQ2RX" java -jar dartagnan/target/dartagnan.jar cat/vmm.cat benchmarks/locks/ticketlock.c      
CFLAGS="-DACQ2RX" DAT3M_COMPILER_OPTIONS=$CFLAGS java -jar dartagnan/target/dartagnan.jar cat/vmm.cat benchmarks/locks/ticketlock.c   // CFLAG could be already set rather than passing by argument
```

ThomasHaas (approved):
LGTM

PR #1113, opened 2026-09-18
[RFC] Program extensions

ThomasHaas:
This is a quick implementation of an idea I had.

Some program formats like Litmus specify extended information such as filters and specs that go beyond the program itself.
To capture this, I added a `ProgramExtension` member to `Program` which can be instantiated to a `ProgramExtension.Litmus` object containing all the extra information. I also used this for SPIR-V which uses the same filter/spec structure as litmus.
As noted in #1086, the litmus format can also specify "locations" for the purpose of enumeration. Supporting this would be straightforward by adding a new member to `ProgramExtension.Litmus`.

I'm not sure if this is the best way to go about it and if there are any pitfalls, but the idea seems reasonable to me.


ThomasHaas:
If the only goal is to handle litmus `locations`, we can also use the cheap solution of just adding another member to `Program`.  The question is whether we want throw everything into the `Program` class. An imaginable advantage is that we could reuse this even for non-litmus programs. 

A different perspective would be to just consider all of this a "program spec" rather than a "program extension" and therefore use a less general `ProgramSpecification` class instead.

hernanponcedeleon:
What does the "extension" in `ProgramExtension` refers to? the format (*.litmus, *.c, ...) or something "extra" that we attach to the program.

I like the idea of having a way to attached extra stuff to a program. In this view, the extension would be similar to the metadata but we would allow it to affect semantics.

However, I see two problems with the current proposal (assuming the "view" of above):
- the filter and the specification are more related to the verification goal than to the program per se (i.e., they do not affect the semantics of the program).
- I would use more fine grained extensions (e.g., one per each item below; at least those that we want to implement this way).

I see all the following could be extensions in the view above (I am not saying we might really want to implemented them all as this)
1. locations
2. aliases as used in PTX/Vulkan
3. which thread synchronize via kernel launch (model via `ssw` annotation in Vulkan)
4. the thread grid
5. the forward progress guarantees
6. the entry point

2-5 kind of represent the environment where the program executes.

5,6 is probable something we would still want to control from options rather than "mark it" in the source code as we do the others.

ThomasHaas:
> What does the "extension" in `ProgramExtension` refers to? the format (*.litmus, *.c, ...) or something "extra" that we attach to the program.

The latter: "something extra".

> I like the idea of having a way to attached extra stuff to a program. In this view, the extension would be similar to the metadata but we would allow it to affect semantics.

Yes, that was the idea.

> However, I see two problems with the current proposal (assuming the "view" of above):
> 
> the filter and the specification are more related to the verification goal than to the program per se (i.e., they do not affect the semantics of the program).

I thought about this too. The issue is that all this verification goal information is inside the program's syntax and as such, the `ProgramParser` has to extract that information somehow. Note that this information is there even if it is irrelevant for our verification goal. 
I was also contemplating about restricting this feature only to spec information and call it `ProgramSpec` instead.
Either way, the first (and maybe only?) goal was to somehow capture the additional information/spec in the source language that goes beyond the program itself.

> I would use more fine grained extensions (e.g., one per each item below; at least those that we want to implement this way).
> I see all the following could be extensions in the view above (I am not saying we might really want to implemented them all as this)
> 
>     1. locations
> 
>     2. aliases as used in PTX/Vulkan
> 
>     3. which thread synchronize via kernel launch (model via `ssw` annotation in Vulkan)
> 
>     4. the thread grid
> 
>     5. the forward progress guarantees
> 
>     6. the entry point

I think if we can add features that make sense directly on our internal language, we probably do not need a generic extension mechanism.
For example, `forward progress guarantees` make sense on every program and "not having this extension" is equivalent to simply having fair progress. A similar point holds for entry points I think.
Even the thread grid makes sense generically, something I tried to capture a while ago in #882. Regular CPU programs can also be understood as arranged in a thread grid/hierarchy but that hierarchy has only a single thread layer below the root node.

The biggest outlier among those points is really the `locations` bit since it tells nothing about the program itself and only the spec, and only when doing enumeration. So I think there are two natural ways to go about it:
1. Change `ProgramExtension` to a generic `ProgramSpec` (in principle the specs do not need to relate to the source language). It has the ugly point to it that the spec may not actually be checked for, i.e., it can be ignored.
2. Rename `ProgramExtension` to something (I don't know what to call it yet) that represents extra data extracted from parsing the source program like embedded specs without conveying the idea that it is more than that. It has the benefit that the parser can just dump all filter/spec clauses/locations/whatever into this and it is not necessarily weird that parts of the information are ignored in Dartagnan (depending on the verification goal). This is basically what I implemented right now.

hernanponcedeleon:
My main issue with the current proposal is its lack of modularity. I think the PR should introduce a mechanism through which a program can carry multiple independent attachments.
  
The common interface could be called `SourceProgramAttachment`, `ProgramAttachment`, or `ParsedAttachment`; all of these seem clearer than `ProgramExtension` to me.
  
Initially, we would add two implementations:  
- `ProgramSpecification`, containing `SpecificationType specType` and `Expression spec`.
- `ExecutionFilter`, containing `Expression filter`.
  
Both attachments would be produced when parsing Litmus and SPIR-V programs.
  
Then, as part of #1086, we could add:
- `ProgramLocations`, containing the locations that should be considered during state enumeration.

We can then separately discuss whether any of the other information mentioned above (aliases, grid, etc) should also use this mechanism.

ThomasHaas:
> My main issue with the current proposal is its lack of modularity. I think the PR should introduce a mechanism through which a program can carry multiple independent attachments.

If you fully modularize it like this, I think it will boil down to exactly what we have right now: a single (optional) member per extension (e.g., spec and filter). The only difference is that we would wrap all the members in wrapper classes and maybe use a map to store all attachments.

> ProgramLocations, containing the locations that should be considered during state enumeration.

Then I would just add a `List<Expression> locations` directly to `Program` and avoid the wrapper classes / extension mechanism. In that sense, we can close the PR and just use the easiest solution.

ThomasHaas:
I will close this for now

PR #1089, opened 2026-09-09
Reorganize unit tests

xeren:
Changes multiple parts in the testing package of Dartagnan.
- All solver test suites provide both `testAssume` and `testRefinement`.  Those who had just one enabled before fail `org.junit.Assume.assumeTrue` to skip the other method.  (This can be undone.)
- Remove most usages of the `Provider` class, only leaving `Provider<ShutdownManager>`.
- *edit:* `ArrayValidTest` and `BranchTest` instances no longer share static `Wmm` instances.
- *edit:* `com.dat3m.dartagnan.litmus.{comparison|compilation}.*` suites now compare result status instead of model existence.
#### Changes to the test suite hierarchy
- `AbstractVerificationTaskSolverTest` (new)
  - `AbstractAsmTest` (new, 6 existing subclasses)
  - `AbstractCTest`
  - `AbstractLitmusTest`
    - *edit:* `ArrayValidTest`, `BranchTest` (+ existing subclasses)
  - `AbstractSpirvTest` (new)
    - `AbstractSpirvOpenclTest` (new, 11 existing subclasses)
    - `AbstractSpirvVulkanTest` (new, 12 existing subclasses)
  - *edit:* `AbstractComparisonTest`
  - *edit:* `AbstractCompilationTest`
#### Changes to the test utilities
- *edit:* Now located in package `com.dat3m.dartagnan.test`
- `Program TestHelper.parseProgram(Path)`
- `Wmm TestHelper.parseWmm(Path)`
- `Path ResourceHelper.getCatPath(Arch,String)`
#### Ideas for further changes
- `AbstractComparisonTest` and `AbstractCompilationTest` could extend `AbstractVerificationTaskSolverTest`.  The source task solver would provide an optional `expected` value for the target task solver.
*edit: implemented.*

PR #1114, opened 2026-09-19
Refactor BaseOptions (minor changes)

ThomasHaas:
I made `BaseOptions` a non-abstract class and moved it to the top-level outside of the utils package (it was no utility).
~Maybe we should also make it an inner class of `Dartagnan`, i.e., `Dartagnan.Options`.~

EDIT: I made `BaseOptions` an inner class now. SVCOMPRunner no longer references the base options (it used to inherit from `BaseOptions` but no inherited option was ever used)


dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
```diff
@@ -212,4 +207,73 @@ private static List<Path> getProgramFiles(Path path, Predicate<Path> isSupported
             return List.of();
         }
     }
+
+    // ========================================== Options ==========================================
+
+    @org.sosy_lab.common.configuration.Options
```
hernanponcedeleon: Use `import` rather than fully qualified name

hernanponcedeleon:
There are problems with reflection both in dartanan and svcomp modules. The CI did not catch them because the tests are not enabled in the current CI configuration (#1089 should fix this).

Running `mvn -pl dartagnan -Dtest=OptionRegistrationTest test` locally should catch the dartagnan one.

Once #1116 is merged, `mvn -pl svcomp -Dtest=OptionRegistrationTest test` will catch the other one.

dartagnan/src/main/java/com/dat3m/dartagnan/Dartagnan.java
ThomasHaas: I can't because the name clashes with the inner class which is also called `Options`. I don' think Java can create aliases, can it?

hernanponcedeleon: Actually, when I read the PR the changes, I was not too happy about  the inconsistency between `Options` and `SVCOMPOptions`. Now that I see this, I would rather use `DartagnanOptions` and avoid the problem with names clashing.

ThomasHaas: Fair enough. I first had `BaseOptions` but then I realized inheritance is pointless. Then I changed it to `DartagnanOptions/Dat3mOptions` but later I made it an inner class and since `Dartagnan.DartagnanOptions` looked stupid, I changed it to just `Options`. I forgot to do the same for `SVCOMPRunner` and also call it `Options` there. TBH, I would rather take the name clash with the annotation than adding the redundancy to the name.
On the other hand, it is unlikely now that external classes would want to reference the inner class, so we do not actually run into cases like `Dartagnan.DartagnanOptions`. In that sense, I guess it is fine to add the redunancy.

Now that I think about it, we should just call it `Config` like we do in many other classes. This is the most consistent name and has no name clash.

ThomasHaas: I changed the name to `DartagnanOptions` and `SVCOMPOptions` now, simply because the name `Config` lead to code like this:
```
Configuration config = ...;
Config o = ...
```
which looked really bad to me :).

dartagnan/src/main/resources/META-INF/native-image/com.dat3m.dartagnan/dartagnan/reflect-config.json
```diff
@@ -40,7 +40,7 @@
         "allDeclaredConstructors": true
     },
     {
-        "name":"com.dat3m.dartagnan.Dartagnan",
+        "name":"com.dat3m.dartagnan.Dartagnan.DartagnanOptions",
```
hernanponcedeleon:
```suggestion
        "name":"com.dat3m.dartagnan.Dartagnan$DartagnanOptions",
```

svcomp/src/main/resources/META-INF/native-image/com.dat3m.svcomp/svcomp/reflect-config.json
```diff
@@ -1,6 +1,6 @@
 [
     {
-        "name":"com.dat3m.svcomp.SVCOMPRunner",
+        "name":"com.dat3m.svcomp.SVCOMPRunner.SVCOMPOptions",
```
hernanponcedeleon:
```suggestion
        "name":"com.dat3m.svcomp.SVCOMPRunner$SVCOMPOptions",
```

hernanponcedeleon (commented):
Both tests still fail
```
  mvn -pl dartagnan -Dtest=OptionRegistrationTest test
  mvn -pl svcomp -Dtest=OptionRegistrationTest test
```

dartagnan/src/main/resources/META-INF/native-image/com.dat3m.dartagnan/dartagnan/reflect-config.json
hernanponcedeleon: can be removed

hernanponcedeleon (approved):

PR #1115, opened 2026-09-19
Add missing tests to CI job

hernanponcedeleon:
I expect this to catch problems in #1114


hernanponcedeleon:
I just realized this is also covered by #1089, thus closing

PR #1117, opened 2026-09-19
Implement pthread_once

hernanponcedeleon:
The intrinsic models each once-control object using three states:
- 0: routine uninitialized
- 1: routine running
- 2: routine complete

Its behavior is equivalent to:

```
  if (compare_exchange(once_control, 0, 1, acquire)) {
      init_routine();
      store(once_control, 2, release);
  } else {
      do {
          state = load(once_control, acquire);
      } while (state != 2);
  }

  return 0;
```

The compare-exchange ensures that exactly one thread runs init_routine. Other callers wait until the routine completes (as required by POSIX).

I additionally validated this whit 27 new SVCOMP benchmarks. We report correct results on 24, 2 do not compile on my macos, and the remaining one timeout because it has 100 threads.


ThomasHaas:
~Devirtualization runs before the intrinsics pass which I think breaks indirect function pointers passed to `pthread_once`. You might have to check that the function pointer is actually a `Function`.~

~I find this confusing though, because `pthread_create` with indirect function pointers should also fail but I thought that actually works (it certainly did at some point)?
Can we maybe run devirtualization after inlining intrinsics? Is there a reason we do not do so?~

EDIT: Nvm, most intrinsics are actually inlined before devirtualization. Only some are done later.

ThomasHaas (approved):
LGTM

hernanponcedeleon:
I think we actually have a problem with intrinsics and devirtualization, but it is more general than just `pthread_once` (or even`pthread_create`) and thus, we should solve it in a different PR. 

Early intrinsics called through function pointers are not recognized before devirtualization. After devirtualization resolves the call target, the early-inlining pass has already run, leaving the intrinsic call unprocessed and eventually reported as an undefined function. Running devirtualization before early intrinsic inlining fixes this. 

However, we cannot simply move the existing devirtualization pass because early intrinsic expansion may itself introduce indirect calls, such as callbacks passed to `pthread_once` or `pthread_create`. Therefore, devirtualization is required both before early intrinsic inlining to resolve calls to intrinsics and afterward to resolve calls introduced by their expansion.

ThomasHaas:
> I think we actually have a problem with intrinsics and devirtualization, but it is more general than just `pthread_once` (or even`pthread_create`) and thus, we should solve it in a different PR.
> 
> Early intrinsics called through function pointers are not recognized before devirtualization. After devirtualization resolves the call target, the early-inlining pass has already run, leaving the intrinsic call unprocessed and eventually reported as an undefined function. Running devirtualization before early intrinsic inlining fixes this.

This was partly deliberate since function pointers to "real intrinsics" are not valid because they are no real functions (e.g., functions pointers to LLVM intriniscs are not possible). The issue is that what we call "Intriniscs" right now are not really intriniscs for the most part: they are just internally-defined implementations of some libraries like pthread or stdlib.

I suggest to rework the handling of libraries by adding a `Library(Implementation)` interface with a `void link(Program)` function and implementing it in classes like `PthreadLibrary` and `CStdLibrary` that provide the implementation of said library.
Instead of replacing function calls like `Inlining` does, they should just add a function body to the function declaration (precisely what linking against the real library would have done). Maybe we can have a special mechanism to still allow resolving functions calls directly instead of adding a function body (sometimes defining a body is not possible).

Funnily enough, I played around with this exact idea a few minutes ago and already made a prototype implementation where I moved all `Pthread` functionality into a `PthreadLibrary` class.

> However, we cannot simply move the existing devirtualization pass because early intrinsic expansion may itself introduce indirect calls, such as callbacks passed to `pthread_once` or `pthread_create`. Therefore, devirtualization is required both before early intrinsic inlining to resolve calls to intrinsics and afterward to resolve calls introduced by their expansion.

Which is no problem I think. We can run it twice.

EDIT: Maybe we cannot run it twice actually. I didn't look at what the implementation does

